PR #1, opened 2020-01-10
Project directory restructure.

Britefury:
Added setup.py and requirements.txt


flax/__init__.py
```diff
@@ -15,5 +15,5 @@
 """Flax API."""
 
 # Allow `import flax`; `flax.nn.[...]`, and the same for `flax.optim.[...]`
-from . import nn
+from flax import nn
```
avital: Let's not make aesthetic changes in this CL. (If we think avoiding `from .` is desirable, we can discuss in a separate CL that only changes all of those occurrences.)

flax/optim.py
```diff
@@ -72,7 +72,7 @@ def loss_fn(model):
 from jax import lax
 import jax.numpy as jnp
 
-from .nn import base
+from flax.nn import base
```
avital: ditto

PR #9, opened 2020-01-24
Examples/vae

makora9143:
Hi! 

This project is very interesting to me.

I add an example of VAE with using Flax.

~I'm wondering how to implement some methods for models (e.g. sampling/generating of VAE) via Flax.~

~If you have any suggestions, please let me know.~

~Update: Since I found @nn.module_method decorator, I added it to define `generate` method, but it's not working due to loading params :(~


makora9143:
@googlebot I signed it!

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
+from absl import app
+from absl import flags
+
+import jax.numpy as jnp
+import numpy as np
+
+import jax
+from jax import nn as jnn
```
jheek: relu and sigmoid are available in nn already.

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+    return - 0.5 * jnp.sum(1 + logvar - jnp.power(mean, 2) - jnp.exp(logvar))
+
+
+@jax.vmap
+def binary_cross_entropy(probs, labels):
+    return - jnp.sum(labels * jnp.log(probs + EPS) + (1 - labels) * jnp.log(1 - probs + EPS))
+
+
+def compute_metrics(recon_x, x, mean, logvar):
+    BCE = binary_cross_entropy(recon_x, x)
```
jheek: please use lower case variable names

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+flags.DEFINE_integer(
+    'num_epochs', default=10,
+    help=('Number of training epochs')
+)
+
+
+class Encoder(nn.Module):
+    def apply(self, x):
+        x = nn.Dense(x, 400, name='enc_fc1')
```
jheek: the flax codebase has 2 spaces indentation instead of 4

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+from flax import nn
+from flax import optim
+
+import tensorflow_datasets as tfds
+
+from utils import save_image
+
+
+EPS = 1e-8
+KEY = random.PRNGKey(0)
```
jheek: this is a bit of a subtle issue but creating jax arrays before the flags are parsed results in incorrect behavior when JAX/XLA depends on flags. So you need to create the PRNGKey in the main function

examples/vae/README.md
```diff
@@ -0,0 +1,9 @@
+# Basic VAE Example
```
jheek: Do you have some example images or a test set ELBO that this model achieves?

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+        logvar_x = nn.Dense(x, 20, name='enc_fc22')
+        return mean_x, logvar_x
+
+
+class Decoder(nn.Module):
+    def apply(self, z):
+        z = nn.Dense(z, 400, name='dec_fc1')
+        z = jnn.relu(z)
+        z = nn.Dense(z, 784, name='dec_fc2')
+        z = jnn.sigmoid(z)
```
jheek: I think log_sigmoid is more stable?

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+@jax.vmap
+def kl_divergence(mean, logvar):
+    return - 0.5 * jnp.sum(1 + logvar - jnp.power(mean, 2) - jnp.exp(logvar))
+
+
+@jax.vmap
+def binary_cross_entropy(probs, labels):
+    return - jnp.sum(labels * jnp.log(probs + EPS) + (1 - labels) * jnp.log(1 - probs + EPS))
```
jheek: can we avoid the eps by using logits instead of probs?

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+class VAE(nn.Module):
+    def apply(self, x):
+        mean, logvar = Encoder(x, name='encoder')
+        z = reparameterize(mean, logvar)
+        recon_x = Decoder(z, name='decoder')
+        return recon_x, mean, logvar
+
+    @nn.module_method
+    def generate(self, z):
+        params = self.get_param("decoder")
```
jheek: " -> '

examples/vae/main.py
```diff
@@ -0,0 +1,147 @@
...
+        return Decoder.call(params, z)
```
jheek: This should also work:
```
@nn.module_method
def generate(self, z):
  return Decoder(z, name='decoder')
```

I think it is fine to really on the strings being the same in this case but you can also consider making the Decoder in a private function

```
def _create_decoder(self):
  return Decoder.shared(name='decoder')
```
this way we don't really on the names being the same in apply and generate.

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    Modify: torch.tensor -> jax.numpy.DeviceArray
+
+    If you want to know about this file in detail, please visit the original code:
+        https://github.com/pytorch/vision/blob/master/torchvision/utils.py
+"""
+import math
+import jax
+import jax.numpy as np
+
+irange = range
```
jheek: We try to not shadow globally defined names so I would prefer to change the argument named range to something else.

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    This code is created with reference to torchvision/utils.py.
+
+    Modify: torch.tensor -> jax.numpy.DeviceArray
+
+    If you want to know about this file in detail, please visit the original code:
+        https://github.com/pytorch/vision/blob/master/torchvision/utils.py
+"""
+import math
+import jax
+import jax.numpy as np
```
jheek: jax.numpy as jnp

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        normalize (bool, optional): If True, shift the image to the range (0, 1),
+            by the min and max values specified by :attr:`range`. Default: ``False``.
+        range (tuple, optional): tuple (min, max) where min and max are numbers,
+            then these numbers are used to normalize the image. By default, min and max
+            are computed from the tensor.
+        scale_each (bool, optional): If ``True``, scale each image in the batch of
+            images separately rather than the (min, max) over all images. Default: ``False``.
+        pad_value (float, optional): Value for the padded pixels. Default: ``0``.
+    """
+    if not (isinstance(ndarray, np.DeviceArray) or
```
jheek: I think it is better to check `isinstance(ndarray, jnp.ndarray)`

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+import jax.numpy as np
+
+irange = range
+
+
+def make_grid(ndarray, nrow=8, padding=2,
+              normalize=False, range=None, scale_each=False, pad_value=0.0):
+    """Make a grid of images.
+    Args:
+        ndarray (array_like or list): 4D mini-batch ndarray of shape (B x H x W x C)
```
jheek: ndarray -> images?

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        scale_each (bool, optional): If ``True``, scale each image in the batch of
+            images separately rather than the (min, max) over all images. Default: ``False``.
+        pad_value (float, optional): Value for the padded pixels. Default: ``0``.
+    """
+    if not (isinstance(ndarray, np.DeviceArray) or
+            (isinstance(ndarray, list) and all(isinstance(t, np.DeviceArray) for t in ndarray))):
+        raise TypeError('array_like or list of tensors expected, got {}'.format(type(ndarray)))
+
+    # if list of tensors, convert to a 4D mini-batch ndarray
+    if isinstance(ndarray, list):
```
jheek: It seems easier to replace the assertions and this list check by `ndarray = jnp.asarray(ndarray)`

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+        ndarray = np.stack(ndarray, 0)
+
+    if ndarray.ndim == 2:  # single image H x W
```
jheek: this function accepts many more input signatures than the docstring provides. I think it is better to just support B,H,W,C case

examples/vae/utils.py
```diff
@@ -0,0 +1,90 @@
...
+    for y in irange(ymaps):
+        for x in irange(xmaps):
+            if k >= nmaps:
+                break
+            grid = jax.ops.index_update(grid, jax.ops.index[y * height + padding: (y+1) * height, x * width + padding: (x+1) * width], ndarray[k])
+            k = k + 1
+    return grid
+
+
+def save_image(ndarray, fp, nrow=8, padding=2,
```
jheek: why not call save_image with the outputs of make_grid?

examples/vae/main.py
makora9143: Thank you for the comment.
I removed this line and use instead `nn.relu` and `nn.sigmoid`.

examples/vae/main.py
makora9143: Replace them as `bce_loss`.

examples/vae/main.py
makora9143: I changed the code to have 2 spaces.

makora9143:
Hi, @jheek 
Very thank you for your review!
I'll try to fix the code little by little.

examples/vae/main.py
makora9143: I understand.
I removed this line and put `KEY = random.PRNGKey(0)` into the first line of `main()` function.

examples/vae/main.py
makora9143: Fix.

examples/vae/main.py
makora9143: Thank you for the very useful suggestion!
With your suggestion, I cannot decide to remain which method for an example.
Hence, I created three types of `generate` method: `generate`, `generate_one_liner` and `generate_shared`.

Could you tell me which method is better for an example?

examples/vae/main.py
makora9143: I decided to remove this line to make `decoder` return `logits`.

examples/vae/main.py
makora9143: Since the `decoder` returns `logits`,
I changed this loss function to use `nn.log_sigmoid` so that I can remove `EPS`.

If you have other suggestions for this function, could you tell me?

examples/vae/utils.py
makora9143: Fix it.

examples/vae/utils.py
makora9143: Fix it.

examples/vae/utils.py
makora9143: Fix it

examples/vae/utils.py
makora9143: Fix it

examples/vae/utils.py
makora9143: I just removed the supports except for B,H,W,C case.

examples/vae/utils.py
makora9143: Fixed.

examples/vae/utils.py
makora9143: I merged `make_grid` and `save_image`!

examples/vae/README.md
makora9143: I added a generated image and ELBO of a test set (mean and std for 5 trials)

makora9143:
Finally, I finished the code fixing.
@jheek could you check my code?

Thank you for your support!

examples/vae/main.py
```diff
@@ -0,0 +1,155 @@
...
+    z = nn.relu(z)
+    z = nn.Dense(z, 784, name='dec_fc2')
+    return z
+
+
+class VAE(nn.Module):
+  def apply(self, x):
+    mean, logvar = Encoder(x, name='encoder')
+    z = reparameterize(mean, logvar)
+    # recon_x = Decoder(z, name='decoder')
```
jheek: can this line be removed?

examples/vae/main.py
```diff
@@ -0,0 +1,155 @@
...
+                                recon_xs[:8].reshape(-1, 28, 28, 1)])
+
+  generate_xs = model.generate(z)
+  generate_xs = generate_xs.reshape(-1, 28, 28, 1)
+
+  return compute_metrics(recon_xs, xs, mean, logvar), comparison, generate_xs
+
+
+def main(argv):
+  KEY = random.PRNGKey(0)
```
jheek: variables names should be lowercase

examples/vae/main.py
```diff
@@ -0,0 +1,155 @@
...
+  train_ds = tfds.load('mnist', split=tfds.Split.TRAIN)
+  train_ds = train_ds.cache().shuffle(1000).batch(FLAGS.batch_size)
+  test_ds = tfds.as_numpy(tfds.load('mnist', split=tfds.Split.TEST, batch_size=-1))
+
+  _, vae = VAE.create_by_shape(KEY, [((1, 784), jnp.float32)])
```
jheek: `create_by_shape` is deprecated use `_, params = VAE.init_by_shape(...)`
and then `vae = nn.Model(VAE, params)`

examples/vae/requirements.txt
```diff
@@ -0,0 +1,4 @@
+jax
+flask
```
jheek: is this supposed to be flax?

examples/vae/requirements.txt
```diff
@@ -0,0 +1,4 @@
...
+tensorflow
+Pillow
```
jheek: should it be lower case?

examples/vae/README.md
```diff
@@ -0,0 +1,17 @@
...
+```bash
+pip install -r requirements.txt
+python main.py
+```
+
+## Examples;
+
+If you run the code by above command, you can get some generated images:
+
+![generated_mnist](https://github.com/makora9143/flax/blob/examples/vae/examples/vae/example.png)
```
jheek: please use relative links see: https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-readmes#relative-links-and-image-paths-in-readme-files

makora9143:
Hi, @jheek!
Thank you for your code review and comments.
I've reflected on your comments on my code.
Could you check my code again?

jheek:
looks all good to me!

PR #12, opened 2020-01-31
Add dataclasses to requirements

bnwebcode:
Dataclasses added to distribution in Python 3.7 but README.md suggests
the aim is to support Python 3.5 or later. With this change flax
appears to run correctly on 3.6.9

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,3 +4,4 @@ jax
 # tensorflow is needed for tensorflow-datasets
 tensorflow
 tensorflow-datasets
+dataclasses


bnwebcode:
@googlebot I signed it!

noahtren:
Just ran into this problem with Python 3.6 and realized I had to download `dataclasses` directly. Thanks!

jheek:
You also need to add the dependency to setup.py to make sure it gets installed when doing a pip install.

marcvanzee:
@bnwebcode It seems 16 days ago since the last reply on this PR. It seem Jonathan asked you to add the dependency to setup.py, could you please do this so we can submit this change?

jheek:
requirements.txt has been removed and setup.py has been updated recently. This PR is no longer necessary.

PR #14, opened 2020-02-11
initialize polyak params

### LINKED ISSUE #13: Polyak Averaging Params_ema initiailzation
Hi, I was applying the changes in your HOWTO to add Polyak averaging and there seems to be some code missing, specifically params_ema is not initialized and so the line 
`optimizer, params_ema = train_step(optimizer, params_ema, batch)`
causes an UnboundLocalError (params_ema referenced before assignment).

joaogui1:
Fixes #13

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,7 +215,7 @@ Note that no special framework support was needed.
 
  @jax.jit
  def eval(model, eval_ds):
-@@ -59,12 +59,14 @@ def train():
+@@ -56,12 +59,14 @@
    optimizer = flax.optim.Momentum(
        learning_rate=0.1, beta=0.9).create(model)
 
--- a/README.md
+++ b/README.md
@@ -219,7 +219,7 @@ Note that no special framework support was needed.
    optimizer = flax.optim.Momentum(
        learning_rate=0.1, beta=0.9).create(model)
 
-+  params_ema = optimizer.target.params
++  params_ema = model.params
 +
    for epoch in range(10):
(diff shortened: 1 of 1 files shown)


avital:
There is no need to instantiate an optimizer. You can do `params_ema = model.params`

joaogui1:
The optimizer was already instantiated since we were going to train the model, but it's probably more intuitive your way

joaogui1:
Hey @avital do you think it's good to go now?

avital:
We don't need there to be an optimizer within the eval function at all...

joaogui1:
Ok, so on that.
I copied the mnist example, changed things so it was using polyak averaging, generated the diff and then changed the readme
On readme there's this line:
``` 
def eval(model, eval_ds):
@@ -59,9 +62,9 @@ def train():
```
Shouldn't there be the def eval?
Sorry for the confusion, I think I don't fully understand the formatting used by the HOWTOs and so I tried to change the least amount possible

joaogui1:
So I double checked and it was correct before, despite the weird syntax if you ran `patch -p0 -i patch.patch -o b/mnist-polyak.py` it would generate the correct patched file in b/mnist-polyak.py with the instantiation of the optimizer only on the training and not on eval. You can check it for yourself @avital.

avital:
@marcvanzee is in the process of changing the example diff in the README and putting up our HOWTOs in a more structured way, so I'm closing this for now. Thanks @joaogui1!

PR #15, opened 2020-02-16
Quick fix for build errors

joaogui1:
Right not the build is erroing in a few tests because of badly formatted imports, this PR fixes it. 
Specifically it:

1. Adds examples/utils to the main package, another solution would be to duplicate it into the tests directory, but I find the utils quite useful and maybe adding them to the main package could be good.

2. Changes the imports in the examples tests so they work with python

Feedback on these decisions is more than welcome


joaogui1:
@levskaya fixed it

PR #20, opened 2020-02-21
Fix rtd build

jheek:
ReadTheDocs will fail with an OOM because TensorFlow is too large a dependency.
I think we shouldn't force users to install this anyway because flax does not actually need TF.

This PR also fixes a small bug in the notebook and adds a missing line of configuration that RTD needs to build

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,4 @@
 numpy
 jaxlib
 jax
-# tensorflow is needed for tensorflow-datasets
-tensorflow
-tensorflow-datasets
 msgpack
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -1,7 +1,4 @@
 numpy
 jaxlib
 jax
-# tensorflow is needed for tensorflow-datasets
```
avital: If we remove this from Flax core requirements, I think we need to add them to a new requirements.txt and setup.py files (or maybe just one of them?) inside the examples/ directories

jheek: I just found that we are also using it in `checkpoint.py`. I will tweak this a bit more.
I'll add requirements.txt to flax/examples

avital:
I went ahead and merged this and deleted the remote branch (was that ok?)

PR #22, opened 2020-02-21
Proposal: Remove `module.create()` and `module.create_by_shape()`

avital:


MostafaDehghani:
This is done in #63.

MostafaDehghani:
This is done in #63.

PR #25, opened 2020-02-28
Ensure shared modules appear in params only once.

avital:
Concretely, create a shared submodule, then pass it in to a child module
and apply it there. Test that the parameters are only stored once, in
the frame where the shared module was created.

Co-authored-by: jheek <heekjonathan@gmail.com>

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -44,7 +44,32 @@ def _track_outputs(x):
 
 
 class _ModuleFrame:
-  """A ModuleFrame contains all the information needed to apply a Module."""
+  """A ModuleFrame the context needed to init or apply a Module.
+
+  In particular, `self.params` is a dictionary where parameters are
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@
 .DS_Store
 docs/_autosummary
 docs/_build
+.vscode
```
jheek: this should probably be removed

avital: Done

avital:
Sorry for the late delay. I rewrote the docstring per the feedback you shared with me off-thread, and added a docstring for ModuleFrame to help explain the machinery.

Let me know if this looks good.

avital:
I merged with a rebase. Per your suggestion @jheek let's start doing that rather than merge commits unless it's genuinely a long merge branch.

PR #27, opened 2020-02-28
Minor fixes to the flax_intro notebook

gnecula:
I have started to read the documentation more carefully, and I decided to go ahead and fix
the typos that I found. In places, I have added some new text, please check. 

I found the `State` section to be confusing and I don't think that I quite understood it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Now install `flax` from Github:
 ## GPU accelerated installation
 
 First install `jaxlib`; please follow the instructions in the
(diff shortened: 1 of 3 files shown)


jheek:
Docs and in particular how we explain state is still WIP. We have recently made some changes to simplify how state works and these should soon result in clearer documentation as well.

avital:
@jheek can we merge this in the meanwhile and start another GitHub issue for state documentation improvement?

PR #32, opened 2020-03-04
Add fast sampling to lm1b example, checkpointing, bugfixes.

levskaya:
Adds fast, cached decoding and simple temperature sampling for the language model example. Adds checkpointing, small bugfixes, and removes some unused code.


examples/lm1b/decode.py
```diff
@@ -33,53 +43,87 @@ def multinomial(rng, logits):
...
-  return generated
+  batch_size = prompt_inputs.shape[0]
+  max_decode_len = prompt_inputs.shape[1]
+  end_marker = jnp.array(eos_token)
+  temperature = jnp.array(temperature)
+
+  # Initialize sampling loop state.
+  rng0 = prng_key
+  i0 = jnp.array(0)
+  token0 = jnp.zeros((batch_size, 1)).astype(jnp.int32)
```
jheek: doesn't zeros have a dtype argument?

examples/lm1b/train.py
```diff
@@ -365,6 +452,23 @@ def main(argv):
...
         eval_summary_writer.flush()
 
+      # Fast inference of prompt extension using trained LM.
+      rng, subrng = jax.random.split(rng)
+      pred_rngs = random.split(subrng, jax.local_device_count())
+      prompt = jnp.array(encoder.encode(FLAGS.prompt))
+      prompt = jax_utils.replicate(prompt)
+      prompt = jnp.reshape(prompt, (prompt.shape[0], 1, prompt.shape[1]))
+      cache = jax_utils.replicate(
+          cache_def.initialize_cache((1, FLAGS.max_predict_token_length)))
```
jheek: initialize_cache creates zero constants. So it might be faster / more memory efficient when initlaize_check is inside predict_step this way they will not be materialized as device buffers

examples/lm1b/decode.py
```diff
@@ -33,53 +43,87 @@ def multinomial(rng, logits):
...
-
-
-def sample_sequence(model, length, context, num_samples=1, temperature=1.0):
-  """Samples from the given model.
+# TODO(levskaya): add top-k and/or top-p "nucleus" sampling.
+def temperature_sample(prompt_inputs,
+                       init_cache,
+                       tokens_to_logits,
+                       prng_key,
+                       temperature=0.6,
```
jheek: odd default?

examples/lm1b/decode.py
```diff
@@ -33,53 +43,87 @@ def multinomial(rng, logits):
   probs = jnp.cumsum(probs, axis=-1)
   a = jax.random.uniform(rng, logits.shape[:-1] + (1,))
   out = jnp.argmin(a > probs, axis=-1)
-  return out[..., None]
+  return out
```
jheek: multinomial seems useful to upstream to jax.random?

examples/lm1b/decode.py
levskaya: ;) will fix.

examples/lm1b/decode.py
levskaya: will fix.

examples/lm1b/decode.py
levskaya: probably!  but we'll do that separately then later remove this.

examples/lm1b/train.py
levskaya: We'd still need to provide the cache_def as a static arg though, no?  The amount of memory used here is tiny right now though.

examples/lm1b/train.py
```diff
@@ -25,65 +26,101 @@
...
-  _, model = model_def.create_by_shape(key, [(input_shape, jnp.float32)])
-  return model
+  with nn.attention.Cache().mutate() as cache_def:
+    _, model = model_def.create_by_shape(key,
+                                         [(input_shape, jnp.float32)],
+                                         cache=cache_def)
+  return model, cache_def
 
 
 def create_optimizer(model, learning_rate):
```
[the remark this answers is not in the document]
avital: I don't think we should remove the use of ReplicatedOptimizer here, let's keep each PR with a single goal...

examples/lm1b/train.py
```diff
@@ -293,30 +365,44 @@ def main(argv):
...
-  for step, batch in zip(range(num_train_steps), train_iter):
+  for step, batch in zip(range(start_step, num_train_steps), train_iter):
     batch = common_utils.shard(jax.tree_map(lambda x: x._numpy(), batch))  # pylint: disable=protected-access
     optimizer, metrics, dropout_rngs = p_train_step(
         optimizer, batch, dropout_rng=dropout_rngs)
     metrics_all.append(metrics)
 
-    if (step + 1) % eval_freq == 0:
+    # Save a Checkpoint
+    if step % FLAGS.checkpoint_freq == 0 and step > 0:
```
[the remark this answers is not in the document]
levskaya: sure!

examples/lm1b/train.py
```diff
@@ -293,30 +365,44 @@ def main(argv):
 
   rng = random.PRNGKey(random_seed)
```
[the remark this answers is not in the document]
levskaya: correct - because of the simple dynamic batching implementation this is forced to be single-host example, but this is good to add anyway as a clear example of that use.

examples/lm1b/decode.py
```diff
@@ -33,53 +43,87 @@ def multinomial(rng, logits):
...
+  max_decode_len = prompt_inputs.shape[1]
+  end_marker = jnp.array(eos_token)
+  temperature = jnp.array(temperature)
+
+  # Initialize sampling loop state.
+  rng0 = prng_key
+  i0 = jnp.array(0)
+  token0 = jnp.zeros((batch_size, 1)).astype(jnp.int32)
+  ended0 = jnp.zeros((batch_size, 1)).astype(jnp.bool_)
+  sequences0 = prompt_inputs
```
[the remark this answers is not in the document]
levskaya: Sure!

examples/lm1b/train.py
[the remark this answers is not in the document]
levskaya: We do need to fix this, but I'd also like to do that separately.

PR #34, opened 2020-03-05
Fix collection anchor

jheek:
This PR aims to fix some edge cases that fail because Collection relies on inspecting the module stack at arbitrary points (not the top of the stack). This used to be okay because the stack corresponded to the child parent structure of the frames. But with the recent changes to shared modules this duality no longer holds.

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -112,6 +112,14 @@ class _ModuleFrame:
       path += self.name
     return path
 
+  def is_child_of(self, frame):
+    if frame is self.parent:
+      return True
+    elif self.parent:
+      return self.parent.is_child_of(frame)
+    else:
+      return False
+
   def create_name(self):
     name = str(self._name_counter)
     self._name_counter += 1
@@ -883,7 +891,7 @@ class Collection:
     self.state = state
     # the anchor is used to determine the prefix of the collection.
     # this way we can create/nest collections inside modules.
-    self._anchor = len(_module_stack)
+    self._anchor = _module_stack[-1] if _module_stack else None
 
     self._mutable = False
     self._root = None
@@ -931,21 +939,23 @@ class Collection:
     Returns:
       The previous value stored in the collection or None.
     """
-    _top_frame('store')
+    frame = _top_frame('store')
     if not self._mutable:
       raise ValueError('Collection is not mutable. Use the `mutate` method to'
                        'create a mutable copy.')
-    assert len(_module_stack) > self._anchor
-
     # the root of a Collection is the first module scope that gets created
     # inside the mutate scope of the Collection. By allowing only one unique
     # root scope we guarantee that state is not accidentally shared
     # between different models. When a user specifies an explicit name we can
     # distinguish models and a collection can have multiple roots.
-    root = _module_stack[self._anchor]
+    assert frame.is_child_of(self._anchor)
+    root = frame
+    while root.parent is not self._anchor:
+      root = root.parent
+
     if self._root is None:
       self._root = root
-    elif self._root is not root:
+    elif self._root != root:
       if self._root.name is None or root.name is None:
         raise ValueError('When multiple top-level module calls use a Collection'
                          ' each top-level module should have a name.')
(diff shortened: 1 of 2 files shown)


avital:
Hi @jheek , gonna take a look now but in the meanwhile do you mind force pushing a new version where the commit message describes what motivated the change and in general terms what the change does?

tests/nn_test.py
```diff
@@ -383,6 +383,36 @@ def apply(self, x):
...
+  def test_collection_store_fails_if_out_of_scope(self):
+    @nn.module
+    def stateful_module(coll):
+      coll.store(1)
+
+    @nn.module
+    def test_inner(f):
+      with nn.Collection().mutate() as coll:
+        pattern = 'Trying to capture state outside the scope'
+        with self.assertRaisesRegex(ValueError, pattern):
```
avital: Same here, can we please put the assertion at the top level of the test?

flax/nn/base.py
```diff
@@ -931,29 +940,62 @@ def store(self, value):
...
     if self._root is None:
       self._root = root
-    elif self._root is not root:
+    elif self._root != root:
       if self._root.name is None or root.name is None:
+        # Example:
+        # with nn.Collection() as coll:
+        #   StatefulModule.call(params, coll)
+        #   StatefulModule.call(params2, coll)
         raise ValueError('When multiple top-level module calls use a Collection'
```
avital: Can you explain more about why this is a problem?  Is it because you want to make it explicit whether they share state or not? I spent a few minutes thinking about it and couldn't figure it out.

flax/nn/base.py
```diff
@@ -931,29 +940,62 @@ def store(self, value):
...
                        'create a mutable copy.')
-    assert len(_module_stack) > self._anchor
-
     # the root of a Collection is the first module scope that gets created
     # inside the mutate scope of the Collection. By allowing only one unique
     # root scope we guarantee that state is not accidentally shared
     # between different models. When a user specifies an explicit name we can
     # distinguish models and a collection can have multiple roots.
-    root = _module_stack[self._anchor]
+    if frame is self._anchor:
```
avital: is the use of `is` instead of `==` here important? if so that's a sharp edge that a lot of people won't know about.

flax/nn/base.py
```diff
@@ -931,29 +940,62 @@ def store(self, value):
...
+    have multiple unnamed roots.
+
+    Args:
+      - frame: the frame of which we want to know the root
+    Returns:
+      The root of the given frame.
+    """
+    assert frame.is_child_of(self._anchor)
+    root = frame
+    while root.parent is not self._anchor:
```
avital: same question: is `is not` important? would `!=` work?

flax/nn/base.py
```diff
@@ -112,6 +112,15 @@ def path(self):
       path += self.name
     return path
 
+  def is_child_of(self, frame):
```
avital: I think the right term is `is_descendent_of`. child typically describes first-order children (the converse of "parent")

flax/nn/base.py
```diff
@@ -931,29 +940,62 @@ def store(self, value):
...
     path = self._current_path()
     old_value = self.state.get(path, None)
     self.state[path] = value
     return old_value
 
+  def _find_root(self, frame):
+    """Find the root frame with respect to the anchor.
+
+    The root frame is defined as the first child of anchor 
+    that is a parent of frame.
```
avital:
```suggestion
    The root frame is defined as the child of anchor 
    that is an ancestor of frame.
```

flax/nn/base.py
```diff
@@ -931,29 +940,62 @@ def store(self, value):
...
+      #   coll.store(1)
+      raise ValueError('State should be stored from within a module.'
+                       ' Consider using the value directly instead of'
+                       ' storing it in a Collection.')
+    if not frame.is_child_of(self._anchor):
+      # edge case where the Collection cannot capture the scope of a shared Module
+      # See test_collection_store_fails_if_out_of_scope in nn_test.py
+      raise ValueError('Trying to capture state outside the scope of this Collection.'
+                       ' Most likely due to passing around a shared Module.')
+    root = self._find_root(frame)
```
avital: I don't quite follow the logic of using `_find_root` here.

flax/nn/base.py
```diff
@@ -931,29 +943,62 @@ def store(self, value):
...
     if self._root is None:
       self._root = root
-    elif self._root is not root:
+    elif self._root != root:
       if self._root.name is None or root.name is None:
+        # Example:
```
avital:
```suggestion
        # In the following examples, should the two calls to `StatefulModule` share state or not?
        # because it's ambiguous, we throw an error and require the user to explicitly separate state
        # by giving each instance a separate name, or to explicitly (XXXXX??) in order to explicitly share
        # state
```

jheek:
Fixed the 2 issues. I think it is ready to merge

avital:
Sorry I added a clarification to one of my comments -- I don't know how
users can share state but not parameters. (Or maybe that's something we
don't want to support right now?)

jheek:
I filled the placeholder

avital:
LGTM!

PR #38, opened 2020-03-06
add the implementation for a VAE with conv and conv transpose

jeroenvlek:
So I started working on this and only later found the already existing PR (#9 by @makora9143 which actually helped me find out how to apply a submodule through call(), so thanks for that!). That VAE example already looks excellent. 

This one adds KL-loss scaling + burn in and only learning the variance to avoid a latent space collapse. Furthermore, the encoder is based on Flax's convolutional MNIST example, and the conv transpose is done with an LHS dilution, following "A guide to convolution arithmetic for deeplearning" by Dumoulin and Visin, and it also contains serialization and pickling of models. (Would you want a ConvTranspose layer down the line, wrapping jax.lax.conv_transpose?)

I'm sure it can be polished, it might even be a bit much for one example. Let me know if you're interested, then I'm happy to adapt the example to make it suitable as an example.

### DIFF
--- a/examples/vae_conv/vae_conv.py
+++ b/examples/vae_conv/vae_conv.py
@@ -1,0 +1,219 @@
+import argparse
+import math
+import numpy
+import pickle
+
+import flax
+import flax.serialization
+import jax
+import jax.numpy as jnp
+import tensorflow_datasets as tfds
+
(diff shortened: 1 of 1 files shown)


jeroenvlek:
@googlebot I signed it!

PR #39, opened 2020-03-06
add the implementation for a VAE with conv and conv transpose

jeroenvlek:
So I started working on this and only later found the already existing PR (#9 by @makora9143 which actually helped me find out how to apply a submodule through call(), so thanks for that!). That VAE example already looks excellent.

This one adds KL-loss scaling + burn in and only learning the variance to avoid a latent space collapse. Furthermore, the encoder is based on Flax's convolutional MNIST example, and the conv transpose is done with an LHS dilution, following "A guide to convolution arithmetic for deeplearning" by Dumoulin and Visin, and it also contains serialization and pickling of models. (Would you want a ConvTranspose layer down the line, wrapping jax.lax.conv_transpose?)

I'm sure it can be polished, it might even be a bit much for one example. Let me know if you're interested, then I'm happy to adapt the example to make it suitable as an example.

### DIFF
--- a/examples/vae_conv/vae_conv.py
+++ b/examples/vae_conv/vae_conv.py
@@ -1,0 +1,219 @@
+import argparse
+import math
+import numpy
+import pickle
+
+import flax
+import flax.serialization
+import jax
+import jax.numpy as jnp
+import tensorflow_datasets as tfds
+
+import matplotlib.pyplot as plt
+
+IMG_LENGTH = 28
+LEARNING_RATE = 0.001
+
+KL_WARM_UP_STEPS = 1000
+KL_LOSS_SCALING = 10e-3
+
+KEY = jax.random.PRNGKey(0)
+
+@jax.vmap
+def draw_sample(z_mean, z_log_var):
+    epsilon = jax.random.normal(KEY, z_mean.shape)
+    sample = z_mean + (jax.numpy.exp(0.5 * z_log_var) * epsilon)
+    return sample
+
+
+class Encoder(flax.nn.Module):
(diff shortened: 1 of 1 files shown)


jheek:
As you noticed there is already a PR for a VAE example.
I think it would be good to give the other PR by @makora9143 some more time to land.
It seems like you have also thought about some details around stability (eg. avoiding collapse).
It would be great if we could unify all the insights into a single VAE example but maybe it is easier to refine the VAE after the other PR has merged. Let me know what you think...

jeroenvlek:
Yes, that makes sense. First come, first serve ;) 

Happy to integrate any of this work into that. Give this PR a kick when you need me, please.

jheek:
Hi Jeroen, 
PR #9 just got merged. Let me know if you would like to bring some of your ideas into the VAE example.

jeroenvlek:
> Hi Jeroen,
> PR #9 just got merged. Let me know if you would like to bring some of your ideas into the VAE example.

Sorry for the late reply. What are you looking for? Shall I copy this file to the same directory? Or actually modify the other example and add a few ideas in there?

jheek:
It would be great if you could improve the existing VAE example. Let me know if that works or not.

avital:
Hi @jeroenvlek -- we'd love to showcase more elaborate VAE examples as part of our example showcase. If you take a look at https://github.com/google/flax/blob/master/examples/README.md, you'll see that we now link to external repositories with Flax implementations. We highly encourage you to file a pull request adding a link to your repository!

I'll close this issue but feel free to comment if you have any questions.

PR #40, opened 2020-03-07
Modify MNIST code in docs to do dtype conversion for TPUs.

levskaya:
The actual MNIST example will work on TPU, but the documented MNIST in readme and the annotated mnist doesn't do dtype conversion away from uint8 needed for TPUs.


jheek:
This changes seem correct but I am surprised this didn't result in an error from XLA. Are you sure there isn't a silent bug in XLA TPU that should be reported too?

PR #41, opened 2020-03-08
Subsume '_' into the prefix kwarg of {save,restore}_checkpoint.

ronw:
Making restore_checkpoint more permissive makes it possible to load a checkpoint at a step earlier than the latest by specifying a different prefix.

Updates save_checkpoint to be consistent.

### DIFF
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -67,7 +67,7 @@ def natural_sort(file_list, signed=True):
 def save_checkpoint(ckpt_dir,
                     target,
                     step,
-                    prefix='checkpoint',
+                    prefix='checkpoint_',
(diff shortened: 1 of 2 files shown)


levskaya:
Hi Ron - sorry I can't edit this PR otherwise I'd just do it - but would you mind modifying the unit tests this changes at [checkpoints_test.py](https://github.com/google/flax/blob/829cc4c260a94ac0161015900c78c2e82de05757/tests/checkpoints_test.py#L61) - if you search/replace all the kwargs there with `prefix='test'` to `prefix='test_'` I think the unit tests should pass.

ronw:
Done.  Sorry, I'm learning my way around the github workflow.

PR #42, opened 2020-03-09
Support replica groups for distributed batchnorm

jekbradbury:
Requires https://github.com/google/jax/pull/2382. Needed for large scale (i.e., small per-device batch size) training of ResNets, where the ideal number of examples to normalize over seems to be about 128 (and normalizing over the whole pmap is both unnecessarily slow and gives worse results).

### DIFF
--- a/flax/nn/normalization.py
+++ b/flax/nn/normalization.py
@@ -43,7 +43,8 @@ class BatchNorm(base.Module):
             scale=True,
             bias_init=initializers.zeros,
             scale_init=initializers.ones,
-            axis_name=None):
+            axis_name=None,
+            replica_groups=None):
     """Normalizes the input using batch statistics.
 
     Args:
@@ -65,6 +66,8 @@ class BatchNorm(base.Module):
       scale_init: initializer for scale, by default, one.
       axis_name: the axis name used to combine batch statistics from multiple
         devices. See `jax.pmap` for a description of axis names (default: None).
+      replica_groups: the custom replica groups used to combine batch statistics
+        over smaller subsets of the overall pmap axis (default: None).
 
     Returns:
       Normalized inputs (this same shape as inputs).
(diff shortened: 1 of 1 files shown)


jheek:
The changes look good to me.
Do you know if this is possible to test this on travis CPU?

flax/nn/normalization.py
```diff
@@ -65,6 +66,8 @@ def apply(self,
       scale_init: initializer for scale, by default, one.
       axis_name: the axis name used to combine batch statistics from multiple
         devices. See `jax.pmap` for a description of axis names (default: None).
+      replica_groups: the custom replica groups used to combine batch statistics
```
avital: Maybe this is known to everyone but me, but maybe we can document better what "replica groups" are?

jheek: I think it should refer to lax.psum/pmean for more details

avital: I looked up the documentation and source code of `lax.psum` but I can't find the term "replica groups" there.

Apologies as maybe I am "new here" but give that JAX lacks internal docstrings, I think we need to own explaining this in the Flax API, unless we can point people to other JAX references (which would be better!)

jekbradbury: The `replica_groups` kwarg is added in https://github.com/google/jax/pull/2382. We might end up using a different name, though. (The name and concept are definitely not known to everyone, but the idea is something many people want to express: doing batch normalization over more than just the examples on each accelerator alone, but less than the entire global batch). TF code often uses a `distributed_group_size` keyword argument and then converts that into XLA replica groups later.

avital: Got it. Perhaps a link to that pull request in the docstring is a simple solution?

jheek:
Actually I have one more question about this: What is the benefit of using `replica_groups` over a nested `pmap`? You could do something like `pmap(pmap(train_step, 'bn_group'), 'batch')` and use `BatchNorm(axis_name='bn_group')` to get a similar result, correct?

jekbradbury:
Yes—the difference is only that dist BN is a library feature that ideally wouldn’t affect the top-level training loop. But I’ll try seeing if I can do what I need with nested pmap; looks like that might be easier than wiring through replica group support in a way we’re fully happy with.

avital:
Hi @jekbradbury -- what's the latest on this? Have you been able to use nested `pmap`s or is this PR still necessary for the training that you're doing?

avital:
Looks like this PR may be stale. I'll close it for now, but feel free to re-open with additional context as appropriate.

avital:
Looks like this PR may be stale. I'll close it for now, but feel free to re-open with additional context as appropriate.

PR #47, opened 2020-03-09
Updated README.md

marcvanzee:
Makes some changes to the FLAX README:

* show ensembling diff instead of Polyak since it is more well-known
* show annotated link examples higher up at the top
* add link to readthedocs are the top


README.md
```diff
@@ -185,52 +187,122 @@ def train():
...
-Here are a few examples.
-
-
+Here is an example.
 
-### Polyak averaging
 
-This diff shows how to modify the MNIST example above to evaluate with
-an exponential moving average of parameters over the course of training.
+### Ensemble learning
```
jheek: ensemble training with pmap seems a bit niche. Why do we want this to be the in the top level README?
I don't see why someone needs to see an example of how howto's work. I think it makes more sense to give a link to the howto's or a showcase of 5 HOWTO's that we use to sell the concept

avital: People seemed to think Polyak averaging was even more niche, I think. We just want to show one HOWTO to explain what it is, and then point to the list of HOWTOs in the docs. This seems like the first HOWTO that decently short and people might care about. But we can change this going forward.

README.md
```diff
@@ -185,52 +187,122 @@ def train():
...
++  optimizer_def = optim.Momentum(
++      learning_rate=FLAGS.learning_rate, beta=FLAGS.momentum)
++  optimizer = optimizer_def.create(create_model(rng))
+   return optimizer
+ 
+@@ -110,7 +113,7 @@ def compute_metrics(logits, labels):
+   return metrics
+ 
+-@jax.jit
++@functools.partial(jax.pmap)
```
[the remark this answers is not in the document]
avital: (Ah yes I forgot to notice that here; Please apply all of the suggestions I made in https://github.com/google/flax/pull/44 to this as well -- and in the future we should replace this with something that automatically generates the README from the diffs in the HOWTOs)

README.md
marcvanzee: I agree this view is not very nice, and I would be happy to replace it with a link to a diff view of the actual HOWTO. However, we currently don't have a howto branch yet, so I suggest we leave this here and replace it once we have the actual branch (which should be in a few days). WDYT?

README.md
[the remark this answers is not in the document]
marcvanzee: Ok, the pain-points of having non-dynamic HOWTO diffs are showing up already :-).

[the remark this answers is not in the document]
marcvanzee: I've applied the changes.

README.md
```diff
@@ -232,11 +232,11 @@ index e93321f..7ac4fe2 100644
    return compute_metrics(logits, batch['label'])
  
 -def train_epoch(optimizer, train_ds, batch_size, epoch, rng):
-+def replicate(tree_obj, num_replicas):
-+  return jax.tree_map(lambda x: onp.array([x] * num_replicas), tree_obj)
++def replicate(tree_obj):
++  return jax.tree_map(lambda x: onp.array([x] * jax.device_count()), tree_obj)
```
jheek: This looks like a less efficient version of jax_utils.replicate

README.md
```diff
@@ -281,14 +281,13 @@ index e93321f..7ac4fe2 100644
  
    batch_size = FLAGS.batch_size
    num_epochs = FLAGS.num_epochs
-+  num_models = jax.device_count()
  
 -  model = create_model(rng)
 -  optimizer = create_optimizer(model, FLAGS.learning_rate, FLAGS.momentum)
-+  optimizers = create_optimizers(random.split(rng, num_models))
++  optimizers = create_optimizers(random.split(rng))
```
jheek: how many splits are supposed to be created here? I would guess num_devices?

README.md
marcvanzee: What do you mean exactly? How would you like me to change it?

README.md
marcvanzee: Yes good point, I updated it.

README.md
marcvanzee: I've now used jax_utils.replicate, and tested the code using a google-internal colab.

jheek:
Im still not really sold on the value of having the command line style diff in the README. I find it quite difficult to parse given the lack of visual clues. Also it currently states that the ensemble HOWTO is an example but there is no follow up on seeing the other HOWTOs. Would it make sense to park showcasing HOWTOs until they are actually visible to users?

README.md
```diff
@@ -185,52 +187,113 @@ def train():
...
+ def eval_step(model, batch):
+   logits = model(batch['image'])
+   return compute_metrics(logits, batch['label']) 
+ 
+-def train_epoch(optimizer, train_ds, batch_size, epoch, rng):
++def train_epoch(optimizers, train_ds, batch_size, epoch, rng):
+   """Train for a single epoch."""
+   train_ds_size = len(train_ds['image'])
+   steps_per_epoch = train_ds_size // batch_size
+@@ -139,25 +141,27 @@ def train_epoch(optimizer, train_ds, batch_size, epoch, rng):
```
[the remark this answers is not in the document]
avital: This is how `git diff` writes its outputs -- I believe it uses some heuristic to general pseudo-human-intelligible markers for surrounding context.

avital:
> Would it make sense to park showcasing HOWTOs until they are actually visible to users?
I am OK with removing the HOWTOs from the main README for now...

marcvanzee:
I removed the HOWTO from the README.

PR #48, opened 2020-03-09
add linterrc

jheek:
This adds the pylintrc that is a best effort implementation of the Google Styleguide.


avital:
Can you also modify the Travis configuration so that it runs this on every PR?

avital:
I am going to merge this, because it's an improvement. The linter doesn't yet pass, we'll deal with that in a subsequent PR.

PR #49, opened 2020-03-09
Stateful check

jheek:
Introduces a check that raises an error when mixing stateful or stochastic with jax transformations.

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -948,6 +948,10 @@ class Collection:
                        ' create a mutable copy.')
     # use the Jax TraceMaster to determine if a Collection is modified from
     # inside a nested jax transformation.
+    # In this case, we throw an error because transforming a stateful function
+    # is ill-defined (eg. what does vmap of BatchNorm do?).
+    # TODO(jheek): Add doc guide on combining jax transforms and state.
+    # TODO(jheek): Should some transformations be excempt from this error?
     master = utils._tracer_of_value(value)
     value_level = master.level if master else float('-inf')
     state_level = self._master.level if self._master else float('-inf')
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -895,6 +895,7 @@ class Collection:
     self._anchor = _module_stack[-1] if _module_stack else None
 
     self._mutable = False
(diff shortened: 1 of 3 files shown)


flax/nn/base.py
```diff
@@ -943,7 +945,13 @@ def store(self, value):
...
     if not self._mutable:
       raise ValueError('Collection is not mutable. Use the `mutate` method to'
-                       'create a mutable copy.')
+                       ' create a mutable copy.')
+    master = utils._tracer_of_value(value)
+    value_level = master.level if master else -1000
+    state_level = self._master.level if self._master else -1000
+    if value_level > state_level:
+      raise ValueError('Stateful operations are not allowed when the Collection'
+                       ' is created outside of the current Jax transformation')
```
avital: It would be great if this error message guided users to a doc link that shows one or two examples of "how to interleave JAX transformations with Flax modules".

We don't need to do that in this pull request, but if not, I think we should track this in a new issue.

flax/nn/base.py
```diff
@@ -943,7 +945,13 @@ def store(self, value):
     frame = _top_frame('store')
     if not self._mutable:
       raise ValueError('Collection is not mutable. Use the `mutate` method to'
-                       'create a mutable copy.')
+                       ' create a mutable copy.')
+    master = utils._tracer_of_value(value)
```
avital: I'd like to see a comment before this block. I can't follow what Tracers do when reading the JAX code that lacks comments.

Is there a 2-3 sentence description of what JAX tracers hold and how this test works?

Or are they documented anywhere that I'm missing? (I also searched in readthedocs)

flax/nn/base.py
```diff
@@ -943,7 +945,13 @@ def store(self, value):
...
+    value_level = master.level if master else -1000
```
avital: The magic constant bugs me but maybe it's fine, consider a short comment, e.g. "If you're inside a 1000-layer JAX transformation, all bets are off"

flax/nn/base.py
jheek: The doc is in the making. But I staged it out to a seperate PR indeed

PR #50, opened 2020-03-10
Setup a bug report template

AlexeyG:


jheek:
@googlebot Think again!

PR #53, opened 2020-03-10
Add Testimonials

marcvanzee:
All authors have approved their text.


marcvanzee:
As discussed offline, let's hold off on pushing this for now.

PR #55, opened 2020-03-10
Remove replicate and optimize calls from examples

jheek:


avital:
I don't know if you're looking for a review yet but this looks great! Can you (either here or in a follow-up) also add deprecation warnings for `optimizer.optimize()` and `optimizer.compute_gradient()`?

jheek:
> I don't know if you're looking for a review yet but this looks great! Can you (either here or in a follow-up) also add deprecation warnings for `optimizer.optimize()` and `optimizer.compute_gradient()`?

optimize() is already deprecated. `compute_gradient` still needs to be deprecated in a seperate PR indeed

PR #57, opened 2020-03-10
Add the Flax FAQs and the annotated MNIST example notebook.

MostafaDehghani:
Add the Flax FAQs and the annotated MNIST example notebook that can be run from a google colab.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,0 +1,251 @@
+# The Flax FAQ
+
+This FAQ is compiled from various questions asked on github issues, mailing list, chatrooms, and from personal conversations. 
+
+---
+**Question: How to initialize a Flax model?**
+
+**Context:** Modules seem to have just one function `apply` that takes as 
+arguments both the input and additional parameters (e.g. size of some layer). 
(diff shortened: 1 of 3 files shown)


docs/faq.md
```diff
@@ -0,0 +1,251 @@
...
+This FAQ is compiled from various questions asked on github issues, mailing list, chatrooms, and from personal conversations. 
+
+---
+**Question: How to initialize a Flax model?**
+
+**Context:** Modules seem to have just one function `apply` that takes as 
+arguments both the input and additional parameters (e.g. size of some layer). 
+How to initialize and use a model? The MNIST example has a CNN with no 
+additional parameters, so initialization happens by only calling 
+`create_by_shape`, but there seem to be multiple ways to create/initialize a
```
avital: We'll have to fix this reference to `create` and `create_by_shape` but we can put that in as a TODO for the moment and fix it afterwards.

avital:
Looks good, thanks! We'll have to update the MNIST example and the FAQ to no longer talk about create but we can do that in a follow-up commit

PR #59, opened 2020-03-11
Moves HOWTOs to readthedocs

marcvanzee:
For now this is manually, we could iterate on this on how to automatically add HOWTO links, but for now I've simply moved it to readthedocs.


docs/howtos.md
```diff
@@ -0,0 +1,74 @@
...
+## Adding a new HOWTO
+
+In order to create a new HOWTO, you can simply make the desired changes local
+changes and run:
+
+```
+./howtos/script/pack_howto_diff.sh
+```
+
+This will pack the changes made on `howto-<name>` into a diff file and stage it
```
jheek: How is name determined here?

marcvanzee: Sorry this comment was outdated so it doesn't make much sense now. I've rephrased it, the user can specify the name of the howto when executing pack_howto_diff.

PR #65, opened 2020-03-13
move scan_in_dim to jax_utils

jheek:


avital:
This isn't a breaking change because of
https://github.com/google/flax/pull/65/files#diff-0c838f30c9b987fac713b378080a6fc7R163

But I would want to change that line. Instead of just aliasing it, wrap it
such that it prints a deprecation warning, like here:
* https://github.com/google/flax/blob/prerelease/flax/optim.py#L275
* https://github.com/google/flax/blob/prerelease/flax/optim.py#L288

jheek:
added deprecation warning

PR #66, opened 2020-03-13
Adds Polyak averaging HOWTO

marcvanzee:
With Polyak Averaging, the training loop keeps track of additional parameters, which are the exponential moving average of the parameters over the course of training. Then, making predictions with these EMA parameters typically leads to better predictions.

### DIFF
--- a/docs/howtos.md
+++ b/docs/howtos.md
@@ -15,6 +15,7 @@ documentation through readthedocs.
 Currently the following HOWTOs are available:
 
 * [Ensembling](https://github.com/google-research/flax/compare/prerelease..howto-ensembling?diff=split)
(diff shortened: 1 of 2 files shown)


marcvanzee:
The HOWTO tooling worked out nicely! One thing that is a bit tricky is if you want to add a HOWTO, and also some other changes (such as the documentation I added). In that case you clearly don't want to run `pack_howtos.sh`after you made all the changes, since the documentation change will also be part of the diff then. This is a mistake I made, its quite easy to resolve, but I don't see an easy way to avoid people from making the mistake (except for letting them only change .py files, but that feels a bit strange as well).

avital:
Should we restrict the howto diff to only looks at files within examples/?
That seems reasonable, simple and will resolve that concern. What do you
think?

Issue #72, opened 2020-03-16
A flax module without trainable parameter changes the rng


MostafaDehghani:
In a flax module, I turned a function that applies a bunch of numpy operations to a flax (sub-)module that has no trainable parameter, e.g.:
```
def add(x, y):
  return x + y
```
to
```
@nn.module
def add(x, y):
  return x + y
```
I noticed that I get different numbers at the end. Seems this is because the submodule changes the random number generators by splitting the rng of the parent module, while this behaviour is not expected in this case.
Maybe it makes more sense to condition splitting the rng  on the existence of trainable parameters in the submodule?

jheek:
hash based rngs fix this issue and a unit test was added in #77 that tests this particular edge case

PR #77, opened 2020-03-16
add hash based paramter rngs

jheek:
hash based rng splitting results in more stable PRNG keys that are the same when using the same name for modules/parameters.

This also fixes issue #72

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -236,6 +236,7 @@ class _ModuleMeta(abc.ABCMeta):
 
 
 def _fold_in_str(rng, data):
+  """Fold a string into a jax.random.PRNGKey using its SHA-1 hash."""
   m = hashlib.sha1()
   m.update(data.encode('utf-8'))
   d = m.digest()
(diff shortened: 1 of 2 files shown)


avital:
Can we add a unit test that verifies this behavior?

jheek:
added unit test

avital:
Thanks for adding the test, and I don't think we need to change anything now but I don't think this test is well written.

As it stands, the test essentially just rewrites what the code does and makes sure it... does what it says it does.

But the test should instead test for the user-facing behavior that matters.

Concretely, the test could have been something like this:

Module 1:
```
x = nn.Dense(x, name='foo', ...)
x = nn.Dense(x, name='bar', ...)
```

Module 2:
```
x = nn.Dense(x, name='foo', ...)
x = nn.Dense(x, name='baz', ...)
x = nn.Dense(x, name='bar', ...)
```

Now test that in both Module 1 and Module 2, when initializing with the same PRNG, 'foo' and 'bar' have the same values.

Then do the same for parameters and not submodules.

PR #82, opened 2020-03-17
update setup for pypi wheel

jheek:
- add minimal versions
- remove outdated requirements.txt

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,7 +18,7 @@ import os
 from setuptools import find_packages
 from setuptools import setup
 
-version = "0.0.1-alpha"
+version = "0.1.0rc1"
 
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -27,9 +27,9 @@
   README = ""
 
 install_requires = [
-    "numpy",
-    "jaxlib",
-    "jax",
+    "numpy>=1.12",
```
AlexeyG: Why this specific version of numpy?

jheek: For now using the same constraint as Jax. This way we make the implicit constraint explicit so that we know what version we can depend on.

AlexeyG: Makes sense. Thank you for the explanation!

PR #85, opened 2020-03-18
Remove use of nn.stochastic from seq2seq example

jheek:


examples/seq2seq/train.py
```diff
@@ -169,12 +168,12 @@ def apply(self,
...
 
 
-def create_model():
+def create_model(rng):
   """Creates a seq2seq model."""
   vocab_size = CTABLE.vocab_size()
   _, initial_params = Seq2seq.init_by_shape(
-      nn.make_rng(), [((1, get_max_input_len(), vocab_size), jnp.float32),
-                      ((1, get_max_output_len(), vocab_size), jnp.float32)])
+      rng, [((1, get_max_input_len(), vocab_size), jnp.float32),
```
marcvanzee: Any reason why you don't just write "jax.random.PRNGKey(0)" here directly, and avoid passing it as an argument to create_model?

jheek: I find it convenient to keep the global random seed at the top level so that it can be split up from there if necessary or perhaps be initialized using a flag like `rng = random.PRNGKey(FLAGS.seed)`.

marcvanzee: Ok, makes sense.

PR #87, opened 2020-03-18
Fix missing imports in example

Skyy93:
Got an issue because flax. in  model = nn.Model(... was missing
and import of tensorflow for casting was also missing

### DIFF
--- a/docs/annotated_mnist.md
+++ b/docs/annotated_mnist.md
@@ -3,6 +3,7 @@
 ```py
 import jax
 import flax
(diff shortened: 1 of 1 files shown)


Skyy93:
@googlebot I signed it!

avital:
Thanks! (It's a bit unfortunate that we `import tf` here just for casting in `tfds`. We should re-visit this... The annotated example is structured a little differently than examples/mnist/train.py...)

But this fixes the example so that's good.

PR #88, opened 2020-03-18
Avoid trying to load temporary checkpoint in `training.checkpoints.restore_checkpoint`.

adarob:


adarob:
An alternative approach would be to make the temporary checkpoint not match `glob_path`. We could name it something like `tmp.{prefix}`.

PR #89, opened 2020-03-18
Save unreplicated optimizer in lm1b example.

adarob:


levskaya:
Whaaat.  I thought I'd fixed this... I guess I did so in the full translation wmt example and somehow just forgot to fix it on lm1b.   Thanks so much for catching this, no desire to save stupidly redundant checkpoints!

PR #90, opened 2020-03-19
Make decoding jittable.

adarob:
Also returns actual logits instead of log probabilities.

### DIFF
--- a/examples/seq2seq/train.py
+++ b/examples/seq2seq/train.py
@@ -157,7 +157,7 @@ class Seq2seq(nn.Module):
     # No teacher forcing, feeding actual output back into the decoder.
     next_inputs = jnp.zeros((batch_size, vocab_size))
     output = []
-    for i in range(max_output_len):
+    for _ in range(max_output_len):
       next_inputs = next_inputs[:, np.newaxis]
       carry, decoder_outputs = decoder(carry, next_inputs, vocab_size)
       decoder_outputs = decoder_outputs.squeeze()
@@ -242,7 +242,6 @@ def train_step(optimizer, batch):
   metrics = compute_metrics(logits, batch['answer'])
   return optimizer, metrics
 
-
 def log_decode(question, inferred, golden):
   """Log the given question, inferred query, and correct query."""
   # Remove last token from inferred string and first token from golden.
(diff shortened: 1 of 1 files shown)


adarob:
Note that I have also implemented support for the following features, which I'd be happy to add:
  * Stacked LSTM
  * Bidirectional encoder
  * Variational loss (Recurrent VAE)
  * Scheduled sampling
  * Sampling w/ temperature

avital:
@adarob that's super awesome! can we make those other modifications into [HOWTOs](https://flax.readthedocs.io/en/latest/howtos.html)? Happy to set up a call with you to work through the details though that doc should do a pretty decent job explaining the system.

examples/seq2seq/train.py
```diff
@@ -156,17 +155,19 @@ def apply(self,
...
+    output = []
+    for _ in range(max_output_len):
+      next_inputs = next_inputs[:, np.newaxis]
+      carry, decoder_outputs = decoder(carry, next_inputs)
+      decoder_outputs = decoder_outputs.squeeze()
+      output.append(decoder_outputs)
+      # Select the argmax as the next input.
+      next_inputs = jnp.equal(
+          decoder_outputs,
+          jnp.max(decoder_outputs, axis=-1)[:, None]
```
marcvanzee: Just for my understanding: axis=1 would also work here, right? What the intuition behind choosing -1?

examples/seq2seq/train.py
```diff
@@ -156,17 +155,19 @@ def apply(self,
...
-    return output
+    next_inputs = jnp.zeros((batch_size, vocab_size))
+    output = []
+    for _ in range(max_output_len):
+      next_inputs = next_inputs[:, np.newaxis]
+      carry, decoder_outputs = decoder(carry, next_inputs)
+      decoder_outputs = decoder_outputs.squeeze()
+      output.append(decoder_outputs)
+      # Select the argmax as the next input.
+      next_inputs = jnp.equal(
```
marcvanzee: Nice improvement! :)

examples/seq2seq/train.py
adarob: That's correct. I typically like to use axis=-1 in these cases since that is _almost always_ the logit/label dimension whereas sometimes dimension 1 is batch, sometimes it is sequence position, etc. Of course, we should try to be consistent with this.

examples/seq2seq/train.py
```diff
@@ -251,14 +252,20 @@ def log_decode(question, inferred, golden):
   logging.info('DECODE: %s = %s %s', question, inferred, suffix)
 
 
+@jax.jit
+def decode(model, inputs):
+  """Decode inputs."""
+  decoder_inputs = encode_onehot(jnp.array(['='])).squeeze()
```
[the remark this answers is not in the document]
adarob: Done.

PR #93, opened 2020-03-20
add flatten dict utilities

jheek:


flax/optim.py
```diff
@@ -488,25 +488,21 @@ def _check_inputs(inputs):
...
-    if not isinstance(x, dict):
-      # x is a leaf
-      if self._filter_fn(path, x):
-        yield x
-    else:
-      for key, value in _sorted_items(x):
-        yield from self._iterate(value, '{}/{}'.format(path, key))
+    flat_dict = traverse_util.flatten_dict(x)
+    for key, value in _sorted_items(flat_dict):
+      path = '/' + '/'.join(key)
```
avital: Probably not worth changing here but wouldn't it be nicer to have the flattened keys be actual arrays of keys, e.g. `['foo', 'bar']` instead of `"/foo/bar"`?

flax/optim.py
```diff
@@ -488,25 +488,21 @@ def _check_inputs(inputs):
           'ModelParamTraversal can only traverse a flax Model instance.')
 
   def _iterate(self, x, path=''):
```
avital: I think the `path` argument is no longer necessary.

flax/traverse_util.py
```diff
@@ -49,6 +49,71 @@
...
+  """Flatten a nested dictionary.
+
+  The nested keys are flattened to a tuple.
+
+  Example::
+
+    xs = {'foo': 1, 'bar': {'a': 2, 'b': {}}}
+    flat_xs = flatten_dict(xs)
+    print(flat_xs)
+    # {
```
avital: Thank you for this informative docstring!

flax/traverse_util.py
```diff
@@ -49,6 +49,71 @@
...
+  """
+  def _flatten(xs, prefix):
+    if not isinstance(xs, dict):
+      return {prefix: xs}
+    result = {}
+    for key, value in xs.items():
+      path = prefix + (key,)
+      result.update(_flatten(value, path))
+    return result
+  return _flatten(xs, ())
```
avital: I think it'd be good to assert that `xs` is a dict here -- otherwise you get `flatten_dict(3) == {(): 3}` which seems surprising?

tests/traverse_util_test.py
```diff
@@ -145,6 +145,25 @@ def test_traversal_set(self):
     with self.assertRaises(ValueError):
       traversal.set([3, 4, 5], x)  # too many values
 
+  def test_flatten_dict(self):
+    xs = {'foo': 1, 'bar': {'a': 2, 'b': {}}}
```
avital: Given that flatten discards empty dicts (which seems correct to me, like `'b'` here), I think we should add a sentence to that effect in the flatten docstring.

flax/optim.py
jheek: `flatten_dict` uses a tuple of strings `('foo', 'bar')` because arrays are not hashable but yes we could consider using the same syntax for the traversals. The downside is that pattern matching on a tuple of strings is a little harder. I will not do it in this PR anyway because it would be a breaking change

flax/traverse_util.py
jheek: agreed

jheek:
@avital I fixed the issues and added a comment about `('foo', bar')` vs `/foo/bar`

flax/optim.py
avital: Ah yes, tuple is better than array.

avital:
LGTM

PR #97, opened 2020-03-22
Improve HOWTOs

avital:
1. Render HOWTOs as nicely formatted inline diffs in docs.
   These diffs are syntax highlighted for both Python and diffs

2. Simplify the multi-device parallel training HOWTO

3. Re-pack the ensembling diff. (Once I applied it and re-packed it
   it got a lot shorter… I wonder if this is something we need to do
   for every commit?)

4. Make the HOWTO docs page directly show HOWTOs, and separate the
   mechanics of the HOWTO system into a separate “HOWTOs HOWTO” page.

5. Modify the pack_howto_diffs.sh script to only pack changes in
   examples/. This significantly simplified the process of iterating
   on HOWTO diffs.

### DIFF
--- a/docs/.gitignore
+++ b/docs/.gitignore
@@ -1,0 +1,1 @@
+_formatted_howtos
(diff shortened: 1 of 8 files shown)


avital:
Please use this PR instead: https://github.com/google/flax/pull/100

PR #100, opened 2020-03-23
Improve HOWTOs

avital:
1. Render HOWTOs as nicely formatted inline diffs in docs.
   These diffs are syntax highlighted for both Python and diffs

2. Simplify the multi-device parallel training HOWTO

3. Re-pack the ensembling diff. (Once I applied it and re-packed it
   it got a lot shorter… I wonder if this is something we need to do
   for every commit?)

4. Make the HOWTO docs page directly show HOWTOs, and separate the
   mechanics of the HOWTO system into a separate “HOWTOs HOWTO” page.

5. Modify the pack_howto_diffs.sh script to only pack changes in
   examples/. This significantly simplified the process of iterating
   on HOWTO diffs.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -17,4 +17,5 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	python3 format_howtos.py
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/docs/format_howtos.py
+++ b/docs/format_howtos.py
@@ -10,6 +10,7 @@ import pygments
 import pygments.formatters
 from pygments.lexers import PythonLexer
 
+import re
 import os
 
 def main():
@@ -29,11 +30,17 @@ def format_howto(input_file, output_file):
   # index 51d2fde..a9d7dcb 100644
   diff = diff[2:]
 
-  # Remove more than one empty line in a row
+  # Remove double newlines of diff context from `diff` (which is a
+  # list of lines).
+
+  # Diff lines start with a charater in {'+', '-', ' '} designating
+  # insert, remove or context.  Create a regexp that matches empty
+  # lines that are either of these three types.
+  empty_line_regexp = re.compile('[+\\- ]\n')
   diff = [diff[lineno] for lineno in range(len(diff))
       if lineno == 0 or not (
-        diff[lineno].rstrip(' \n') == '' and
-        diff[lineno-1].rstrip(' \n') == ''
+        empty_line_regexp.match(diff[lineno]) and
+        empty_line_regexp.match(diff[lineno-1])
        )]
 
   # Don't do any special formatting.
--- a/docs/format_howtos.py
+++ b/docs/format_howtos.py
@@ -42,7 +42,7 @@ def format_howto(input_file, output_file):
       return source
 
   # Run `diff` through the normal pygments Python syntax
-  # highlighter.  Get back an array of HTML lines
+  # highlighter. Get back an array of HTML lines.
   colored_diff = (
     pygments.highlight('\n'.join(diff), PythonLexer(), RawHtmlFormatter())
   ).splitlines()
--- a/docs/format_howtos.py
+++ b/docs/format_howtos.py
@@ -36,7 +36,7 @@ def format_howto(input_file, output_file):
         diff[lineno-1].rstrip(' \n') == ''
        )]
 
-  # Don't do any special formatting
+  # Don't do any special formatting.
   class RawHtmlFormatter(pygments.formatters.HtmlFormatter):
     def wrap(self, source, outfile):
       return source
--- a/docs/format_howtos.py
+++ b/docs/format_howtos.py
@@ -1,0 +1,78 @@
+"""
+Read all of the HOWTO .diff files and convert them into .html files
+that are both Python syntax highlighted /and/ diff syntax highlighted.
+
+Then these can be included directly in readthedocs as inline HTML
+files.
+"""
+
+import pygments
+import pygments.formatters
+from pygments.lexers import PythonLexer
+
+import os
+
+def main():
+    for diff_filename in os.listdir(os.path.join('..', 'howtos', 'diffs')):
+        format_howto(os.path.join('..', 'howtos', 'diffs', diff_filename),
+                     os.path.join('_formatted_howtos', diff_filename + '.html'))
+
+def format_howto(input_file, output_file):
+    # Load one of our HOWTO diff files.
+    with open(input_file) as f:
+        diff = f.readlines()
+
+    # Ignore the first two line, which look like:
+    #
+    # diff --git a/examples/mnist/train.py b/examples/mnist/train.py
+    # index 51d2fde..a9d7dcb 100644
+    diff = diff[2:]
+
+    # Remove more than one empty line in a row
+    diff = [diff[lineno] for lineno in range(len(diff))
+            if lineno == 0 or not (
+                diff[lineno].rstrip(' \n') == '' and
+                diff[lineno-1].rstrip(' \n') == ''
+           )]
+
+    # Don't do any special formatting
+    class RawHtmlFormatter(pygments.formatters.HtmlFormatter):
+        def wrap(self, source, outfile):
+            return source
+
+    # Run `diff` through the normal pygments Python syntax
+    # highlighter.  Get back an array of HTML lines
+    colored_diff = (
+        pygments.highlight('\n'.join(diff), PythonLexer(), RawHtmlFormatter())
+    ).splitlines()
+
+    # Write a newly formatted diff-and-Python syntax highlighted
+    # output HTML file, that can be inline included into any .rst
+    # file.
+    with open(output_file, "w") as out_file:
+        # Add the relevant DIVs that Sphinx adds around code blocks
+        print('<div class="highlight-default notranslate"><div class="highlight">',
+              file=out_file)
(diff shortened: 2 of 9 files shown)


docs/Makefile
```diff
@@ -17,4 +17,5 @@ help:
 # Catch-all target: route all unknown targets to Sphinx using the new
 # "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
 %: Makefile
+	python3 format_howtos.py
```
jheek: I don't think travis uses the make file currently it probably should know that it actually does something more than calling sphinx

docs/format_howtos.py
```diff
@@ -0,0 +1,78 @@
+"""
```
jheek: How difficult is it to incorporate this into sphinx? Ideally we would load this code as a sphinx plugin and use something like `howto:: [name]` in the rst files.

docs/format_howtos.py
```diff
@@ -0,0 +1,78 @@
...
+    # Run `diff` through the normal pygments Python syntax
+    # highlighter.  Get back an array of HTML lines
+    colored_diff = (
+        pygments.highlight('\n'.join(diff), PythonLexer(), RawHtmlFormatter())
+    ).splitlines()
+
+    # Write a newly formatted diff-and-Python syntax highlighted
+    # output HTML file, that can be inline included into any .rst
+    # file.
+    with open(output_file, "w") as out_file:
```
jheek: nit: " -> '

docs/format_howtos.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+import pygments
+import pygments.formatters
+from pygments.lexers import PythonLexer
+
+import os
+
+def main():
+    for diff_filename in os.listdir(os.path.join('..', 'howtos', 'diffs')):
```
jheek: nit: this file has 4 spaces instead of 2

docs/format_howtos.py
```diff
@@ -0,0 +1,78 @@
...
+        format_howto(os.path.join('..', 'howtos', 'diffs', diff_filename),
+                     os.path.join('_formatted_howtos', diff_filename + '.html'))
+
+def format_howto(input_file, output_file):
+    # Load one of our HOWTO diff files.
+    with open(input_file) as f:
+        diff = f.readlines()
```
jheek: I guess we load the raw .diff file here directly? Would git produce different diffs due to its merging strategy?
Especially the line numbers will quickly bit rot when other parts of the code are updated without causing conflicts.

docs/Makefile
avital: OK I'll look at the travis script as well.

docs/format_howtos.py
avital: I was thinking about that as well. Honestly I don't know, but the approach in this CL has the added benefit that it's easy to debug/test the format_howto script. If it were entirely in Sphinx I could see that being harder. This version is more like "unix standalong executables that combine via pipes".

So, given that this approach works, doesn't have obvious flaws and allows for testing/debugging, I'd like to err on the side of "let's keep it the way it is".

docs/format_howtos.py
avital: Acknowledged.

docs/format_howtos.py
avital: Acknowledged.

docs/format_howtos.py
avital: Yes the line number will rot, but git diffs keep metadata about which commit they were forked from which (barring history rewriting on prerelease which we shouldn't do) allows them to be applied later as well. (This is how Git 3 way merge works)

But we don't show the line numbers directly to anyone.

What is true is that at times when the base examples change, then the diff here will be longer than it should be (imagine we brought in part of a HOWTO into the base example). At that point you would want to re-apply and re-pack all of the HOWTOs to make the cleaner.

We could consider doing that as a hook.

avital:
I pushed a new commit (hopefully) resolving your concerns, please take another look

docs/format_howtos.py
```diff
@@ -0,0 +1,79 @@
...
+    diff = f.readlines()
+
+  # Ignore the first two line, which look like:
+  #
+  # diff --git a/examples/mnist/train.py b/examples/mnist/train.py
+  # index 51d2fde..a9d7dcb 100644
+  diff = diff[2:]
+
+  # Remove more than one empty line in a row
+  diff = [diff[lineno] for lineno in range(len(diff))
```
marcvanzee: Are you trying to remove double newlines? It is not entirely clear to me from the comment. Perhaps it is easier to understand if you use a regex? https://stackoverflow.com/a/28902081

docs/format_howtos.py
```diff
@@ -0,0 +1,79 @@
...
+        diff[lineno-1].rstrip(' \n') == ''
+       )]
+
+  # Don't do any special formatting
+  class RawHtmlFormatter(pygments.formatters.HtmlFormatter):
+    def wrap(self, source, outfile):
+      return source
+
+  # Run `diff` through the normal pygments Python syntax
+  # highlighter.  Get back an array of HTML lines
```
marcvanzee:
```suggestion
  # highlighter. Get back an array of HTML lines.
```

docs/format_howtos.py
```diff
@@ -0,0 +1,79 @@
...
+  diff = diff[2:]
+
+  # Remove more than one empty line in a row
+  diff = [diff[lineno] for lineno in range(len(diff))
+      if lineno == 0 or not (
+        diff[lineno].rstrip(' \n') == '' and
+        diff[lineno-1].rstrip(' \n') == ''
+       )]
+
+  # Don't do any special formatting
```
marcvanzee:
```suggestion
  # Don't do any special formatting.
```

docs/format_howtos.py
```diff
@@ -0,0 +1,79 @@
...
+  # highlighter.  Get back an array of HTML lines
+  colored_diff = (
+    pygments.highlight('\n'.join(diff), PythonLexer(), RawHtmlFormatter())
+  ).splitlines()
+
+  # Write a newly formatted diff-and-Python syntax highlighted
+  # output HTML file, that can be inline included into any .rst
+  # file.
+  with open(output_file, 'w') as out_file:
+    # Add the relevant DIVs that Sphinx adds around code blocks
```
marcvanzee:
```suggestion
    # Add the relevant DIVs that Sphinx adds around code blocks.
```

docs/howtos-howto.md
```diff
@@ -64,14 +50,16 @@ The workflow script for the Github action can be found at
 
 ## Resolving a conflict with an existing HOWTO
 
-> :warning: TODO: This is still in progress. Currently conflicts are resolved by
-            checking whether the Github action fails, and if so, manually making
-            the required fixes.
+TODO: This is still in progress. Currently conflicts are resolved by
+checking whether the Github action fails, and if so, manually making
+the required fixes.
+</div>
```
marcvanzee: Remove this?

docs/howtos.rst
```diff
@@ -0,0 +1,53 @@
+Flax HOWTOs
+===========
+
+Flax aims to be a thin library of composable primitives. You compose
+these primites yourself into a training loop that you write and is fully under
+your control. You have full freedom to modify the behavior of your training loop,
+and you should generally not need special library support to implement
+the modifications you want.
+
+To help you get started, we show some sample diffs, which
```
marcvanzee: This newline seems wrong, will it render correctly?

docs/howtos.rst
```diff
@@ -0,0 +1,53 @@
...
+we call "HOWTOs". These HOWTOs show common modifications to training loops. For instance,
+the HOWTO for ensembling learning demonstrates what changes should be made to
+the standard MNIST example in order to train an ensemble of models on
+multiple devices.
+
+Note that these HOWTOs do not require special library support, they just
+demonstate how assembling the JAX and Flax primitives in different ways
+allow you to achieve various training loop modifications.
```
marcvanzee:
```suggestion
allows you to achieve various training loop modifications.
```

Also, "achieve various training loop modifications" sounds a bit strange. I don't think you "achieve" a modification, you "make" it.

docs/format_howtos.py
avital: Well the file is already split into lines, so you can't use a regexp. (And the lines aren't actually empty, they contain the first space that designates whether it's an addition or removal of a line, or just context).

I'll modify the comment.

docs/howtos-howto.md
avital: Oh? What fixed this? Do we have this test set up on Travis?

docs/howtos.rst
avital: I'm not sure what newline you mean. Is it that the line is shorter than 80c? Single newlines don't have any meaning in restructuredText. But two newlines in a row are a paragraph delineator?

docs/format_howtos.py
avital: I pushed another change rewriting the logic a bit and adding a clarifying comment. Please take another look.

docs/howtos.rst
avital: You're right, I fixed that.

docs/format_howtos.py
marcvanzee: Thanks for the clarifying comment!

docs/howtos-howto.md
marcvanzee: No I mean removing the "</div>", this is markdown, right?

docs/howtos.rst
marcvanzee: Indeed I meant the newline on the line that is shorter than 80c. SG.

docs/howtos-howto.md
avital: Oh oops, yes! Done.

PR #104, opened 2020-03-23
[draft] improve module method api

jheek:
This PR adds the following features:
- `__init__` is called on Module construction and can declare keyword arguments. these keyword arguments are not passed to apply or other module methods.
- module methods can be called like normal methods on the instance (eg. `self.encode`)

auto encoder example:
```
class AutoEncoder(nn.Module):

  def __init__(self, encoder_features, decoder_features):
    self.encoder = nn.Dense.shared(features=encoder_features, name='encoder')
    self.decoder = nn.Dense.shared(features=decoder_features, name='decoder')

  def apply(self, x):
    z = self.encode(x)
    return self.decode(z)

  @nn.module_method
  def encode(self, x):
    return self.encoder(x)

  @nn.module_method
  def decode(self, x):
    return self.decoder(x)
```

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -139,25 +139,23 @@ def module_method(fn):
 
     class AutoEncoder(nn.Module):
 
-      def apply(self, x, **hparams):
-        encoder, decoder = self._create_modules(**hparams)
+      def __init__(self, encoder_features, decoder_features):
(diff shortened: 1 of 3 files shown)


avital:
@jheek do we still plan on considering this? If not, should we close the pull request?

PR #109, opened 2020-03-24
Update nlp_seq example README.md

AlexeyG:


examples/nlp_seq/README.md
```diff
@@ -19,13 +19,13 @@ From|ADP the|DT AP|PROPN comes|VBZ this|DT story|NN :|:
 ### Supported setups
 The model should run with other configurations and hardware, but explicitly tested on the following.
 
-| Hardware |  Batch size  | Learning rate | Steps/Second | Accuracy  | TensorBoard |
+| Hardware |  Batch size  | Learning rate | Training time | Accuracy  | TensorBoard.dev |
 |:---:|:---:|:---:|:---:|:---:|:---:|
-| TITAN V  | 64  |  0.05 | 3.3  | 71.73% | comming soon |
+| Nvidia V100 (16GB) | 64  |  0.05 |  1d 11h 11m  | 72.20% | [2020-03-22](https://tensorboard.dev/experiment/IyswaKpRQbOpk0AfLwBu3A/) |
```
jheek: 1.5 days of training? I'm not super familiar with this stuff but that seems a little slow?

jheek: Actually looking at the tensorboard.dev it looks like its training for way too long and overfitting?

AlexeyG: To me the model indeed looks like its overfitting, but I kept it running for the entire 500k steps specified in the config. Most of that was not very useful, since the best results were obtained much earlier.

I believe that bohnetbd@ is working towards adding early stopping to the example. Once it lands, I expect it would address both your comments.

bohnetbd: Yes, it is overfitting. I improved it already with better regularisation (dropout variant) and I have early stopping in now. 

Secondly, yes it takes too long. 75k steps instead of 500k are fine, I looked also on way larger example (3x times data), still 75k steps are fine. So the time would be down to about 1/6. Larger batches might speed it up too.

AlexeyG: @jheek can we merge the README as is, or do you prefer that it first reflects a run with fewer steps, or @bohnetbd 's improvements?

jheek: The tensorboard in particular is a bit confusing now. Can we not just change the default config to 75k steps?

AlexeyG: Done.

AlexeyG:
Something went wrong during the rebase - had to re-add the changes in a new commit and will need the LGTM again. Sorry :/

PR #115, opened 2020-03-25
Include output from format_howtos.py in git

avital:
This should let readthedocs correctly render HOWTOs. RTD doesn't seem
to have a way to configure a command to run before docs build.

Perhaps indeed it would have been better to hook directly into
Sphinx's extensions but I didn't spend the time to figure that out,
and right now that doesn't seem like a super high priority thing
to change. This solution seems Good Enough.


avital:
I am going to merge this proactively because it's necessary for v1 of "rendered HOWTOs" which don't currently appear in RTD. We can revisit.

PR #116, opened 2020-03-25
Bring back the correct ensembling HOWTO

avital:
(I accidentally had pushed a copy of the multi-device training
HOWTO on top of it)


avital:
I will also proactively merge this as what our docs currently show it wrong. Happy to discuss more.

(It actually does suck to see the HTML in the commits now... [see #115])

PR #118, opened 2020-03-26
Build HOWTO HTML snippets during docs build time.

avital:
This undoes PR #115 while making builds work on RTD.


docs/conf.py
```diff
@@ -32,6 +32,12 @@
 sys.path.insert(0, os.path.abspath('..'))
 
 
+# Hacky but decent pre-process step: Format HOWTOs into color-coded
+# and syntax highlighted HTML files
```
marcvanzee:
```suggestion
# and syntax highlighted HTML files.
```

PR #119, opened 2020-03-26
Tiny tweaks to comments in lm1b example.

malmaud:
CC @levskaya , trying to build some momentum with an easy PR :)


levskaya:
Thanks!

PR #125, opened 2020-03-27
add stochastic check

jheek:


avital:
LGTM for the change but please add a commit message that explains what this change does.

jheek:
amended  commit with description

adarob:
This still lacks documentation, and I'm not sure what to do to make my test in #111 pass.

adarob:
Does this mean that nn.stochastic doesn't pass through any functions or just jitted functions? It appears to be the former. Is that intentional?

adarob:
Figured it out -- I didn't realize the `nn.stochastic` scope should also not cross the `lax.scan` boundaries, so this is WAI. However, I think we should still add documentation to `stochastic.py` to make this clear.

Issue #123, opened 2020-03-27
`Module.partial().__qualname__` is not copied


Conchylicultor:
I don't know if this is expected but `Module.partial` do not seem to forward `__qualname__`.

To reproduce:
```python
class MyModule(flax.nn.Module):
  def apply():
    pass

print(MyModule.partial().__name__)  # MyModule
print(MyModule.partial().__qualname__)  # Module.partial.<locals>.PartialModule
```

jheek:
Good point I will make a PR for this

PR #127, opened 2020-03-27
Add test requirements

danielsuo:
Hi from Google Princeton! Just started to look at `flax`; thanks for working on this! Tiny, tiny contribution, but it did take a moment for me to figure out how to run tests.

Additions:
1. **Added test requirements to `setup.py`**: there should also probably be some documentation on how to set up a development environment, but I didn't add because I'm not sure where you'd want to add (e.g., `dev`, `contributing.md`, somewhere else)
2. **Added `bash` to `tests/run_all_tests.sh`**: very minor detail, but I use `zsh` and the test script fails for me without this line. Shouldn't break anything existing.

I did notice that running `pytest tests` via `pytest-xdist` caused a lot of failures related to `cuda`, but haven't looked into them yet. Running `pytest` without the `-n 4` flag works fine.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,8 @@ install_requires = [
 ]
 
 tests_require = [
+    "pytest",
+    "tensorflow_datasets",
 ]
 
 setup(
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ install_requires = [
 
 tests_require = [
     "pytest",
+    "pytest-xdist",
     "tensorflow_datasets",
 ]
 
--- a/tests/run_all_tests.sh
+++ b/tests/run_all_tests.sh
@@ -1,3 +1,5 @@
+#!/bin/bash
+
 pytest -n 4 tests -W ignore
 
 # we apply pytest within each example to avoid pytest's annoying test-filename collision.


avital:
Thanks! LGTM. Can you please rebase your branch and then we can merge it? (we're trying to avoid commits that merge mater into feature branches)

danielsuo:
Done!

avital:
Ah, this is annoying -- I just merged another change into master so we need to update this branch again. I wonder what other projects that use GitHub do. Is there history really littered with "merge master in foo-branch" commits? 

Anyhow, I'll just let it have the merge commit for now until we figure out what the right flow is. I wouldn't want to keep you waiting.

danielsuo:
Oh no! I appreciate that, but this was hardly anything urgent.

I believe there's a little drop down arrow next to `Merge pull request` that lets you select from a few choices, including rebase.

image

avital:
I think that dropdown is different -- that says "don't introduce a 'merged PR #123' commit on master". I actually do want that commit, it helps people browsing history see the discussion around commits.

The problem is that when your branch is not up-to-date, there's a button that says "update branch" (I think that's what it's called) which merges master back in to your feature branch (not a rebase). I guess they did it this way because it's really harmful if GitHub would re-write history for a user's feature branch. Could lead to lots of unexpected issues, so this is the safer approach... But still.

danielsuo:
Oh! I totally misunderstood your question. Yeah, rewriting history automatically would be bad (and, I think, an unavoidable problem in a realistic contributor's workflow).

PR #128, opened 2020-03-27
Fix incorrect variable in apply_howto_diffs.sh

avital:


avital:
(I believe this is the cause of the breaking GitHub action: https://github.com/google/flax/actions?query=workflow%3A%22Apply+HOWTO+diffs+to+branches%22)

PR #131, opened 2020-03-27
Document module.call and module.apply

avital:


flax/nn/base.py
```diff
@@ -279,6 +279,22 @@ def __new__(cls, *args, name=None, **kwargs):
...
+    """Override to implement how a module is initialized and called.
+
+    Within `apply`, you can generally do two things:
+
+    * Call `self.param(name, shape, initializer)` to declare
+      parameters for this module.
+
+    * Call `MySubModule(...)` to instantiate and apply submodules.
+
+    `apply` isn't expected to be a pure function, due to the inlined
```
jheek: It is unclear to me what this paragraph tries to explain.
It contains a few clues but I wouldn't expect users to understand when to actually use init and call based on this.
There's a lot of vague terms here that are hard to understand like "init mode", "access the logic", "purely functional paths", "inlined parameter decelerations".

If I understand correctly you want to explain that init and call can be used inside apply if you want to explicitly pass the parameters to a module. I'm not sure if that pattern belongs in the docstring of apply. I would recommend putting the docs in the lowest level that encapsulates the major subjects. So in this case we are talking about the interaction between apply, init, and call so I think that belongs in the docstring of Module.

flax/nn/base.py
```diff
@@ -279,6 +279,22 @@ def __new__(cls, *args, name=None, **kwargs):
 
   @abc.abstractmethod
   def apply(self, *args, **kwargs):
+    """Override to implement how a module is initialized and called.
+
+    Within `apply`, you can generally do two things:
+
+    * Call `self.param(name, shape, initializer)` to declare
```
jheek: Perhaps the first point here should be that you can do arbitrary numerical computation using Jax

avital:
I agree with @jheek 's point that much of this is better explained as a docstring on Module. But the current Module abstraction leads to a somewhat confusing mental model, that we're trying to address (for example, one possible direction is #208  but there are others being considered).

I think I'd rather address this issues as part of a larger restructuring.

PR #137, opened 2020-03-28
Avoid singleton dimensions in batchnorm params

jekbradbury:
This improves performance on TPU (because the layout JAX forces for e.g. an `f32[1,1,1,256]` array makes for slower fusion code than an `f32[256]`). It also avoids an XLA bug on GPU (fix forthcoming) that caused iteration-order-dependent fusion failures (and thus significantly worse performance) for some models.

### DIFF
--- a/flax/nn/normalization.py
+++ b/flax/nn/normalization.py
@@ -72,6 +72,7 @@ class BatchNorm(base.Module):
     x = jnp.asarray(x, jnp.float32)
     axis = axis if isinstance(axis, tuple) else (axis,)
     axis = _absolute_dims(x.ndim, axis)
+    feature_shape = tuple(d if i in axis else 1 for i, d in enumerate(x.shape))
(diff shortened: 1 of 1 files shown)


flax/nn/normalization.py
```diff
@@ -89,27 +89,28 @@ def apply(self,
...
-        mean2 = lax.psum(mean2, axis_name=axis_name) / axis_size
+        mean2 = lax.pmean(mean2, axis_name=axis_name)
       var = mean2 - lax.square(mean)
 
       if ra_mean and not self.is_initializing():
         ra_mean.value = momentum * ra_mean.value + (1 - momentum) * mean
         ra_var.value = momentum * ra_var.value + (1 - momentum) * var
 
-    y = x - mean
+    y = x - mean.reshape(feature_shape)
```
levskaya: shouldn't this be `reduced_feature_shape`?  this is what's causing test breakage I think.

jekbradbury: Forgot to copy over a line; should be working now.

PR #141, opened 2020-03-30
Add ConvTranspose Module

j-towns:
Also rename the conv kwargs

  `lhs_dilation` -> `input_dilation`
  `rhs_dilation` -> `kernel_dilation`

### DIFF
--- a/flax/nn/linear.py
+++ b/flax/nn/linear.py
@@ -181,8 +181,8 @@ class Conv(base.Module):
             kernel_size,
             strides=None,
             padding='SAME',
-            lhs_dilation=None,
-            rhs_dilation=None,
+            input_dilation=None,
+            kernel_dilation=None,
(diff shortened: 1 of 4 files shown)


flax/nn/linear.py
```diff
@@ -181,8 +181,8 @@ def apply(self,
             kernel_size,
             strides=None,
             padding='SAME',
-            lhs_dilation=None,
-            rhs_dilation=None,
+            input_dilation=None,
```
jheek: I like this but it is a breaking change. We should probably send an email to flax-announce@ to make people aware

avital: (And add the inaugural entry in CHANGELOG :) )

j-towns:
Don't merge this yet cos I need to re-update the changelog.

PR #143, opened 2020-03-30
Fix vae

jheek:


examples/vae/main.py
```diff
@@ -142,30 +139,37 @@ def eval(model, eval_ds, z):
...
 
   optimizer = optim.Adam(learning_rate=FLAGS.learning_rate).create(vae)
 
+  rng, z_key, eval_rng = random.split(rng, 3)
+  z = random.normal(z_key, (64, 20))
+
   for epoch in range(FLAGS.num_epochs):
     for batch in tfds.as_numpy(train_ds):
+      rng, key = random.split(rng)
       batch['image'] = batch['image'].reshape(-1, 784) / 255.0
```
AlexeyG: Why are images flattened and  then unflattened again in `eval`? If flattening serves a purpose, consider moving the first flattening op into the IO pipeline

examples/vae/main.py
```diff
@@ -142,30 +139,37 @@ def eval(model, eval_ds, z):
...
   vae = nn.Model(VAE, params)
 
   optimizer = optim.Adam(learning_rate=FLAGS.learning_rate).create(vae)
 
+  rng, z_key, eval_rng = random.split(rng, 3)
+  z = random.normal(z_key, (64, 20))
+
   for epoch in range(FLAGS.num_epochs):
     for batch in tfds.as_numpy(train_ds):
+      rng, key = random.split(rng)
```
AlexeyG: What's the reason for manually handling key splitting instead of using the stochastic context? Using the latter would showcase more of Flax's features.

examples/vae/main.py
AlexeyG: After dividing by 255. the data becomes "continuous". You need to dynamically binarise it- both for train and eval. Many VAEs report results on dynamically binarised MNIST and on statically binarised MNIST (some fixed dataset, which I believe isn't a part of TFDS).

examples/vae/main.py
```diff
@@ -142,30 +139,37 @@ def eval(model, eval_ds, z):
...
+  rng = random.PRNGKey(0)
+  rng, key = random.split(rng)
+
   train_ds = tfds.load('mnist', split=tfds.Split.TRAIN)
   train_ds = train_ds.cache().shuffle(1000).batch(FLAGS.batch_size)
   test_ds = tfds.as_numpy(tfds.load('mnist', split=tfds.Split.TEST, batch_size=-1))
 
-  _, params = VAE.init_by_shape(key, [((1, 784), jnp.float32)])
+  _, params = VAE.init_by_shape(
+      key, [((1, 784), jnp.float32)], z_rng=random.PRNGKey(0))
```
AlexeyG: It doesn't make a difference for this specific model, but I think it'd be cleaner to specify the actual batch dimension during model init (instead of dim 1 as it is now).

examples/vae/README.md
```diff
@@ -12,6 +12,10 @@ python main.py
```
AlexeyG: Git doesn't let me comment on parts that are not considered different. But could you please remove a semicolon after `## Examples` ?

examples/vae/main.py
```diff
@@ -142,30 +139,37 @@ def eval(model, eval_ds, z):
...
   vae = nn.Model(VAE, params)
 
   optimizer = optim.Adam(learning_rate=FLAGS.learning_rate).create(vae)
 
+  rng, z_key, eval_rng = random.split(rng, 3)
+  z = random.normal(z_key, (64, 20))
```
AlexeyG: 64 -> FLAGS.batch_size ?

AlexeyG: Can we also make 20 (latent dimensionality) a parameter? I don't expect it being used a lot, but giving it a name would improve readability IMO.

examples/vae/main.py
```diff
@@ -127,10 +124,10 @@ def loss_fn(model):
...
 
 @jax.jit
-def eval(model, eval_ds, z):
+def eval(model, eval_ds, z, z_rng):
   xs = eval_ds['image'] / 255.0
   xs = xs.reshape(-1, 784)
-  recon_xs, mean, logvar = model(xs)
+  recon_xs, mean, logvar = model(xs, z_rng)
 
   comparison = jnp.concatenate([xs[:8].reshape(-1, 28, 28, 1),
```
AlexeyG: Return all samples+reconstructions and subsample+concat outside of eval before saving the image?

examples/vae/main.py
AlexeyG: Scrap half of the previous comment - TFDS *does* have the binarized mnist dataset. It's called `binarized_mnist` :)

jheek: What is your recommendation here? To eval on both?

AlexeyG: Just stick to the `binarized_mnist` in training and eval. This should be equivalent to the "static MNIST" setup from VAE literature.

examples/vae/main.py
jheek: latents are a flag now 64 is because the visualization is 8x8

examples/vae/main.py
jheek: There are a lot of reconstructions (10,000) so I think pre-processing the data for the comparison image is not a bad idea

examples/vae/main.py
jheek: done

examples/vae/main.py
jheek: done

examples/vae/main.py
jheek: I like to use the simple and explicit Jax APIs as much as possible. `nn.stochastic` was introduced to avoid having to pass around rngs through a complex model. But within a single function I think manually splitting rngs is still better.

examples/vae/main.py
AlexeyG: I didn't realise this. I thought it was going to be just the regular batch size.

examples/vae/main.py
jheek: flattening is now part of the input pipeline

examples/vae/main.py
```diff
@@ -127,10 +124,10 @@ def loss_fn(model):
 
 
 @jax.jit
-def eval(model, eval_ds, z):
+def eval(model, eval_ds, z, z_rng):
   xs = eval_ds['image'] / 255.0
```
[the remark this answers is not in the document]
jheek: also part of the input pipeline now

examples/vae/main.py
```diff
@@ -32,18 +32,18 @@
 FLAGS = flags.FLAGS
 
 flags.DEFINE_float(
-  'learning_rate', default=1e-3,
-  help=('The leanring rate for the Adam optimizer')
+    'learning_rate', default=1e-3,
+    help=('The leanring rate for the Adam optimizer')
```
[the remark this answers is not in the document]
jheek: done

examples/vae/main.py
```diff
@@ -107,14 +100,18 @@ def binary_cross_entropy_with_logits(logits, labels):
 def compute_metrics(recon_x, x, mean, logvar):
   bce_loss = binary_cross_entropy_with_logits(recon_x, x)
   kld_loss = kl_divergence(mean, logvar)
-  return {'bce': jnp.mean(bce_loss), 'kld': jnp.mean(kld_loss), 'loss': jnp.mean(bce_loss + kld_loss)}
+  return {
+      'bce': jnp.mean(bce_loss),
+      'kld': jnp.mean(kld_loss),
+      'loss': jnp.mean(bce_loss + kld_loss)
```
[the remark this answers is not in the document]
jheek: done

PR #144, opened 2020-03-30
Regression testing for Flax examples

AlexeyG:
# Prototype regression testing / benchmark framework for Flax examples

This provides an API/framework for users to write regression tests for their examples. Two goals were kept in mind when designing this:
* Examples code should not require any modification (or only require minimal modification) in order to be ready for regression testing.
* The code expected to be written by the user of the framework should be simple.

## Using TensorBoard logs to extract regression metrics
Most examples already log the metrics of interest (e.g. time per epoch, training accuracy, perplexity, etc) to TensorBoard (TB) during training. These logs thus contain all the information required for detecting a regression. Extracting this information from the TB logs thus immediately makes most examples ready for regression testing without any modifications.

This holds true for all but the simplest examples (e.g. MNIST) that do not rely on TB. For these examples we can either add TB logging (recommended solution), or do something weird like capturing `stderr` output and parsing it. The latter is cumbersome, but allows for keeping the simplest examples simple.

Currently, the framework only includes a thin wrapper for reading TB scaler summaries.

## API
The user API was kept minimal and similar to what unit tests would use. Specifically, it supports the following
* Extracting metrics from TB summaries
* Asserting that metric value are within pre-defined bounds
* Reporting metrics

### Reporting
We also allow reporting "extras" - any additional textual information. Currently this is used for adding information about failed assertions and test descriptions.

Reporting is done via the `report_metric('name', value)` and `report_metrics({'name': value})` methods for the metrics; and `report_extra('name', 'value')` and `report_extras({'name': 'value'})` methods for the extras. There is also a separate `report_wall_time(value)` method consumed by the CI framework we plant to interface with (more on that later).

### Asserting
We allow using the full spectrum of the `unittest` / `absltest` `self.assert*` methods for verifying benchmark metrics / correctness. This achieves two goals. First, this provides the familiar unit testing API to the end users; and second, this makes reporting metrics and checking their values completely independent, thus allowing for simple APIs for both.

The cost of this simplicity is that we need to override the default behaviour of `self.assert*` methods. Whereas normally a failed assertion would immediately cause the test to halt, we now detect failures as before, but *defer* raising them until the end of the test. This guarantees that the benchmark always produces output/report, even if the metric values were outside of the permitted value ranges. Furthermore, overriding assertion methods allows for automatically detecting and reporting benchmark failures.

### Example benchmark
Together, this allows for concise benchmark code. See example below for CIFAR10.

```python
  @flagsaver
  def test_1x_v100(self):
    """Run Wide ResNet CIFAR10 on 1x V100 GPUs for 2 epochs."""
    FLAGS.num_epochs = 2
    FLAGS.arch = 'wrn26_10'
    FLAGS.model_dir = model_dir

    start_time = time.time()
    train.main([])
    benchmark_time = time.time() - start_time
    summaries = self.read_summaries(model_dir)

    # Summaries contain all the information necessary for the regression
    # metrics.
    wall_time, _, eval_error_rate = zip(*summaries['eval_error_rate'])
    wall_time = np.array(wall_time)
    sec_per_epoch = np.mean(wall_time[1:] - wall_time[:-1])
    end_error_rate = eval_error_rate[-1]

    # Assertions are deferred until the test finishes, so the metrics are
    # always reported and benchmark success is determined based on *all*
    # assertions.
    self.assertBetween(sec_per_epoch, 80., 84.)
    self.assertBetween(end_error_rate, 0.30, 0.36)

    # Use the reporting API to report single or multiple metrics/extras.
    self.report_wall_time(benchmark_time)
    self.report_metrics({'sec_per_epoch': sec_per_epoch,
                         'error_rate': end_error_rate})
```

## Interfacing with monitoring / CI frameworks
The current plan is to interface with an external monitoring / CI framework by exporting benchmark results as a simple JSON file with the following format
```
{
      "name": <class.testMethod>
      "succeeded": true / false
      "wall_time": float (containing wall-time for the benchmark)
      "metrics": {
        "string" -> float map of other performance metrics
      }
      "extras": {
        "string" -> "string" map containing anything else of interest
      }
}
```

The support for this is implemented behind the scenes. So long as end uses use the `self.report_*` methods in their benchmark code, the framework will take care of aggregating the metrics, extras, figuring out the name of the test and the file, etc. As a bonus, if any of the assertions fail, the framework will output their error messages under `extras`.

JSON files (one per test method) *should* automatically written to disk after each test if the `benchmark_output_dir` is specified as an absl flag. I haven't actually tested that this works since I couldn't figure out how to get the pytest running to set flags.

## What's included in this PR?
This PR implements the framework described above, and adds (for now) toy benchmarks for ImageNet (2 epochs on 8xV100), CIFAR10 (2 epochs on 1xV100) and MNIST (full training on CPU). The MNIST example was also modified to produce TB summaries.
this


examples/imagenet/train_benchmark.py
```diff
@@ -0,0 +1,62 @@
...
+  """Benchmarks for the ImageNet Flax example."""
+
+  @flagsaver
+  def test_8x_v100_half_precision(self):
+    """Run ImageNet on 8x V100 GPUs in half precision for 2 epochs."""
+    model_dir = tempfile.mkdtemp()
+    FLAGS.batch_size = 2048
+    FLAGS.half_precision = True
+    FLAGS.loss_scaling = 256.
+    FLAGS.num_epochs = 2
```
MostafaDehghani, line 33: With this way of benchmarking, I assumed that we go for full training, don't we?

examples/imagenet/train_benchmark.py
```diff
@@ -0,0 +1,62 @@
...
+    wall_time, _, eval_accuracy = zip(*summaries['eval_accuracy'])
+    wall_time = np.array(wall_time)
+    sec_per_epoch = np.mean(wall_time[1:] - wall_time[:-1])
+    end_accuracy = eval_accuracy[-1]
+
+    # Assertions are deffered until the test finishes, so the metrics are
+    # always reported and benchmark success is determined based on *all*
+    # assertions.
+    self.assertBetween(sec_per_epoch, 210, 240)
+    self.assertBetween(end_accuracy, 0.06, 0.09)
```
MostafaDehghani, line 52: I wonder how informative this can be. I mean two epochs of training on imagenet is supposed to get us to this accuracy, but not sure if the accuracy of a code with a silent bug after two epochs is always going to be less than 0.06.

flax/testing/benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+    if not tag in data_by_key:
+      data_by_key[tag] = []
+    data_by_key[tag].append((wall_time, step, value))
+  return data_by_key
+
+
+class Benchmark(absltest.TestCase):
+  """Benchmark class for Flax examples.
+
+  This class overrides changes the behaviour of `self.assert*` methods to
```
MostafaDehghani, line 88: Remove either "overrides" or "changes".

flax/testing/benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+  deferred instead of immediate failure. This allows for using absltest
```
MostafaDehghani, line 89: 'to be deferred'?

flax/testing/benchmark.py
```diff
@@ -0,0 +1,241 @@
...
+               'extras': self._reported_extras}
+    if self._reported_wall_time is not None:
+      results['wall_time'] = self._reported_wall_time
+    if not succeeded:
+      msg = '\n'.join([str(fail) for fail in self._outstanding_fails])
+      results['extras']['failed_assertions'] = msg
+
+    results_str = json.dumps(results)
+    logging.info(results_str)
+    print(results_str)
```
MostafaDehghani, line 233: Is this left intentionally?

examples/imagenet/train_benchmark.py
AlexeyG: Yes, eventually. But not while setting up the system / debugging.

examples/imagenet/train_benchmark.py
AlexeyG: I think this refers back to the previous comment. These are "toy" regression tests. Having them run fast is convenient for the time being, as we're setting up the CI / monitoring that will run them.

flax/testing/benchmark.py
AlexeyG: Done.

flax/testing/benchmark.py
AlexeyG: Semi-intentionally - I could not figure out how to get pytest to print info-level logs. Removed the print statement now.

AlexeyG:
Thanks for the review, Mostafa!

The class support training full example for just a few epochs or running full training. The attached benchmarks for imagenet and cifar10 are "toy" benchmarks. They run much faster than what complete training would take and the idea is to use them while we figure out the intergration with the systems that will consume benchmark reports (JSON files produced for each tests). Once that's running smooth, we can switch to complete training - this should be just a matter of updating 3 lines or so for each of the examples.

flax/testing/benchmark.py
AlexeyG: Resolved.

examples/cifar10/train_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+
+  @flagsaver
+  def test_1x_v100(self):
+    """Run Wide ResNet CIFAR10 on 1x V100 GPUs for 2 epochs."""
+    model_dir = tempfile.mkdtemp()
+    FLAGS.num_epochs = 2
+    FLAGS.arch = 'wrn26_10'
+    FLAGS.model_dir = model_dir
+
+    start_time = time.time()
```
avital, line 34: We could consider making this work with the following abstraction:

```py
with self.benchmark() as duration, summaries:
  train.main([])
```

(`duration` is what was called `benchmark_time` in your code)

examples/cifar10/train_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    train.main([])
+    benchmark_time = time.time() - start_time
+    summaries = self.read_summaries(model_dir)
+
+    # Summaries contain all the information necessary for the regression
+    # metrics.
+    wall_time, _, eval_error_rate = zip(*summaries['eval_error_rate'])
+    wall_time = np.array(wall_time)
+    sec_per_epoch = np.mean(wall_time[1:] - wall_time[:-1])
```
avital, line 43: Is `sec_per_epoch` basically the same as `benchmark_time / len(wall_time)`?

flax/testing/benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+    Based on tensorflow/python/platform/benchmark.py
+
+    Args:
+      prefix: str. Prefix that the caller method must have.
+
+    Returns:
+      Resolved test name as `ClassName.test_name`.
+    """
+    # Find the caller method (outermost Benchmark class).
+    stack = inspect.stack()
```
avital, line 173: Interesting approach, it's great if it works!

flax/testing/benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+
+    results_str = json.dumps(results)
+    logging.info(results_str)
+
+    # Maybe save results as a file for pickup by CI / monitornig frameworks.
+    benchmark_output_dir = FLAGS.benchmark_output_dir
+    if benchmark_output_dir:
+      os.makedirs(benchmark_output_dir, exist_ok=True)
+      filename = os.path.join(benchmark_output_dir, name + '.json')
+      with open(filename, 'w') as fout:
```
avital, line 238: I think we'll need to use`tf.gfile` (or if there's something equivalent that doesn't depend on TF) to allow writing files to network file systems as well such as Google Cloud Storage (`gs://`) or others.

But we can cross that bridge when we get there.

examples/cifar10/train_benchmark.py
AlexeyG: I like how this makes the code more concise, however, it makes a few assumptions:
1) That `train.py` always accepts `model_dir` as a flag; and
2) That we always rely on summaries to get the necessary metrics.

* Assumption 2) is currently true, however, I can imagine that for some (really small) examples we may decide to instead rely on parsing their logging output to get the (very simple) metrics.
* Assumption 1) is also currently true for models that output TensorBoard summaries, but again, unclear whether this is how we'll always do things.
* I do see one non-disputable benefit of the construct you proposed - it makes it easy to get the duration (i.e. is a timer scope).


I think we can either:
1) Make a timer-only scope
2) Go ahead with making the two assumptions above, but we should keep in mind that not all examples fall under this abstraction.

Curious to hear your thoughts.

examples/cifar10/train_benchmark.py
AlexeyG: I think so, almost. However, the first epoch has a few one-off things happening in it, such as data caching or XLA compilation. Not taking it into account allows for a better estimate of the actual compute efficiency.

flax/testing/benchmark.py
AlexeyG: It seems to, at least in the examples I've tried. Personally, I wasn't too happy with the solution, but couldn't think of anything nicer.

flax/testing/benchmark.py
AlexeyG: Yes. This crossed my mind. I tried quickly searching for a gfile implementation that is not bound to tf, and I couldn't find one. I wanted to avoid creating a dependency on TensorFlow just to be able to use gfile. But not sure what other alternatives we have.

examples/cifar10/train_benchmark.py
avital: Got it, thanks.

A thought (please don't block this work on it): If we also want to measure start-up/teardown time (which I guess is also important incase, say JIT compilation time becomes slower in JAX), we could compute:

benchmark_time - (wall_time[-1] - wall_time[0])

Which is maybe off by some constant but should be (may be?) consistent across runs.

This is probably not worth doing... Just thought I'd share.

examples/cifar10/train_benchmark.py
avital: OK yes I now agree with you, let's not prematurely make it harder to use the framework more flexibly (like your point about parsing output)

The timer scope could be cute but it's not really that important -- one extra line vs two.

flax/testing/benchmark.py
avital: @mohitreddy1996, I'm curious -- do you know of something open source like tf.io.gfile (that supports network file systems natively) that's not bound to TF?

mohitreddy1996: Hi Avital and Alexey, 

After filing [Issue 210](https://github.com/google/flax/issues/210), I started investigating if tf.gfile could be replaced. I found the following alternatives, may be we could discuss more on how they could be useful or used as alternatives:

- [Dask](https://docs.dask.org/en/latest/remote-data-services.html#remote-data) has an IO library which allows reading/writing to local, network FS and cloud object stores (gcs, s3). Using this would mean creating a dependency on Dask (which I am not sure is a good idea as we are trying to get rid of an external dependency already - tf). 
- [fsspec](https://filesystem-spec.readthedocs.io/en/latest/intro.html) supports local, network FS, cloud object stores and even reading from Github. Although the [built-in implementations](https://filesystem-spec.readthedocs.io/en/latest/intro.html) doesn't list GCS, but looking at the [implementation](https://github.com/intake/filesystem_spec/blob/de54f06c08989e902e1703621ed99b15feac4085/fsspec/registry.py#L31), I believe it should be possible.
- [GCSFS](https://gcsfs.readthedocs.io/en/latest/): Part of Dask, but can be used as a standalone module. I haven't seen any documentation which says 'local' FS is supported (which means we will need to implement additional utility module to handle cases), but I haven't tried it out yet. Also the usage is not really extendable i.e doesn't support other cloud storages etc.  
- [Pyfilesystems](https://docs.pyfilesystem.org/en/latest/index.html) - Has a support for GCS from a [third party implementation](https://www.pyfilesystem.org/page/index-of-filesystems/). Also fsspec [claims to be better](https://filesystem-spec.readthedocs.io/en/latest/intro.html#not-pyfilesystems) than this.

Other FS interfaces rejected:
- [Apache Arrow's FS Interface](https://arrow.apache.org/docs/python/filesystems.html) - dependency on Apache Arrow also doesn't support GCS.

examples/cifar10/train_benchmark.py
mohitreddy1996: I think avital@ has a fair point of considering measurements of setup and teardown differently since it captures important details like JAX compilation and caching. I can actually see how that information could be very helpful in the future. 

May be this could be a feature which one of us could add in the future.

examples/cifar10/train_benchmark.py
```diff
@@ -0,0 +1,61 @@
...
+    end_error_rate = eval_error_rate[-1]
+
+    # Assertions are deffered until the test finishes, so the metrics are
```
mohitreddy1996, line 46: nit: deferred

examples/mnist/train_benchmark.py
```diff
@@ -0,0 +1,58 @@
...
+    summaries = self.read_summaries(model_dir)
+
+    # Summaries contain all the information necessary for the regression
+    # metrics.
+    wall_time, _, eval_accuracy = zip(*summaries['eval_accuracy'])
+    wall_time = np.array(wall_time)
+    sec_per_epoch = np.mean(wall_time[1:] - wall_time[:-1])
+    end_eval_accuracy = eval_accuracy[-1]
+
+    # Assertions are deffered until the test finishes, so the metrics are
```
mohitreddy1996, line 44: nit: deferred

flax/testing/benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+
+def _make_events_generator(path):
+  """Makes a generator yielding TensorBoard events from files in `path`."""
+  return directory_watcher.DirectoryWatcher(
+      path,
+      event_file_loader.EventFileLoader).Load()
+
+
+def _process_event(event):
+  """Parse TensorBoard scalars into a (tag, wall_time, step, scalar) tuple."""
```
mohitreddy1996, line 57: Curious if description should also be included here?

flax/testing/benchmark.py
AlexeyG: Hi Mohit,

Thank you for looking into this - this is super useful; and these look like viable leads. However, just keep things easy and pragmatic, perhaps we should keep using `os` and `open` for the moment. And once we find that this is insufficient, we should probably just switch to `tf.gfile` (thus  introducing the dependency), and later figure out how to avoid the dependency.

I am also now thinking that perhaps depending on TensorFlow isn't too bad. We already have to depend on it through `tensorflow_datasets` (which is typically used in the examples), so in most cases depending on `tf.gfile` will not introduce a new dependency.

examples/cifar10/train_benchmark.py
AlexeyG: Fixed

examples/mnist/train_benchmark.py
AlexeyG: Fixed.

flax/testing/benchmark.py
mohitreddy1996: Yup that sounds good as well :)

flax/testing/benchmark.py
AlexeyG: What do you mean that each event tuple should include a description?

mohitreddy1996: My bad, I think its being done correctly. I thought that the description of the metrics reported by a test case as an 'extra', if we should include it here. I didn't realize it was being considered different from the metrics. 

Took a closer look at benchmark.py, sorry about that. Thanks!

examples/cifar10/train_benchmark.py
AlexeyG: Agreed, we should track it - it's free extra information.

PR #145, opened 2020-03-30
run_all_tests.sh should die on failure.

levskaya:
Since factoring out the test script, we have silent failures happening now because several test commands are run - only the last one failing would be counted as a failure before.  This is the fix. (and there's an existing batchnorm test failure exposed now - that's not fixed here.)

### DIFF
--- a/tests/nn_test.py
+++ b/tests/nn_test.py
@@ -542,9 +542,9 @@ class NormalizationTest(absltest.TestCase):
       y = model(x)
     ema = state['/']
     onp.testing.assert_allclose(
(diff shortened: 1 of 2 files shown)


avital:
We can't merge this until we fix the batchnorm test failure. (Which I can't get to in the next few hours)

levskaya:
Set the full `set -euxo pipefail` incantation and fixed the trivial batchnorm test failure (due to james' BN layout improvement).

levskaya:
I forgot that we can't use `set -e` because pytest emits errcode 5 when it doesn't find tests, so I wrote a bash error trap that behaves like `set -e` (immediate exit) for all other error codes.

PR #147, opened 2020-03-30
Closes #129: reorganize optim into directory structure

### LINKED ISSUE #129: Reorganizing optim into directory structure?
Hello again! At the Princeton office, we work on, among other things, optimization algorithms for deep learning. We're interested in using `flax` and wanted to add some other well-known algorithms. Would you guys be open to reorganizing `optim.py` into a directory a la [pytorch](https://github.com/pytorch/pytorch/tree/master/torch/optim)? Happy to submit a PR if so!

Usually, this _would_ accompany a PR, but being new around here, wanted to understand how (if at all) you wanted to reorganize.

One possibility: All subclasses of `OptimizerDef` (except `MultiOptimizer`, which appears to have a circular dependency with `OptimizerDef`) live in their own files (e.g., `Momentum`, `GradientDescent`)

danielsuo:
Here's a first cut at reorganizing `flax.optim`. Some notes below. Happy to make whatever changes needed!

- As discussed in #129, I moved most classes that subclass `OptimizerDef` into their own files
  - I'm not really sure how we want to deal with "meta" `OptimizerDef`s e.g., `WeightNorm` and `MultiOptimizer`, so put the former in its own file and left the latter in `base.py` due to its circular dependency with `OptimizerDef`
  - I named the file for `GradientDescent` as `sgd.py`, but could rename to `gradient_descent.py` for consistency
  - I left out `ReplicatedOptimizer` since it's deprecated, but can move to its own file
- I tried to mirror formatting / structure of existing files, but might be helpful to have some guidelines on this in `contributing.md` (e.g., use this formatter with this config file; `pylintrc` is a start)
- I added an `__all__` to `optim/__init__.py`, but saw that it wasn't the custom in other `__init__.py` files
- We may eventually want to separate out the tests into different files as well, but for now, left as is.


levskaya:
The split as is looks good - ReplicatedOptimizer should stay in the "base" definitions and won't be around long, I think it's fine to use `__all__` in this way here, we can see what others think.  
Weird, I thought we had a pylintrc, but we'll double-check that.

danielsuo:
There is indeed a pylintrc! Just meant that it didn’t appear to be uniquely defining formatting. I didn’t want to use my formatted and spam with lots of white space changes :)

PR #149, opened 2020-03-31
Add SST-2 NLP example.

bastings:
This adds an NLP example: text classification on the SST-2 data set.
The model is a simple LSTM with MLP classifier on top, trained with binary cross entropy loss.


examples/sst2/input_pipeline.py
```diff
@@ -0,0 +1,175 @@
...
+from typing import Dict, Sequence, Text
+
+from absl import logging
+
+import numpy as np
+import tensorflow.compat.v2 as tf
+import tensorflow_datasets as tfds
+
+def build_vocab(datasets: Sequence[tf.data.Dataset],
+                special_tokens: Sequence[Text] = (b'<pad>', b'<unk>', b'<s>', b'</s>'),
```
bohnetbd: I was wondering why to use bytes and not python strings?

examples/sst2/model.py
```diff
@@ -0,0 +1,169 @@
...
+  model_def = TextClassifier.partial(train=False, **model_kwargs)
+  _, initial_params = model_def.init_by_shape(
+      jax.random.PRNGKey(seed),
+      [((batch_size, max_len), jnp.int32),
+       ((batch_size,), jnp.int32)])
+  model = nn.Model(model_def, initial_params)
+  return model
+
+
+def word_dropout(inputs: jnp.ndarray, rate: float = None, unk_idx: int = None):
```
bohnetbd: Very nice functionality! I was wondering about the interface if this could be more similar to other modules, providing train, and not having default None values for rate and unk_idx. Maybe, just adapting train. Up to you of cause.

examples/sst2/model.py
```diff
@@ -0,0 +1,169 @@
...
+  def apply(self,
+            inputs: jnp.ndarray,
+            hidden_size: int = None,
+            output_size: int = None,
+            output_bias: bool = False,
+            dropout: float = None,
+            train: bool = None):
+    # inputs.shape = <float32>[batch_size, seq_length, hidden_size]
+    hidden = nn.Dense(inputs, hidden_size, name='hidden')
+    hidden = nn.tanh(hidden)
```
bohnetbd: No action needed but relu or elu lead often to better results and more numerical stability - no nan.

examples/sst2/input_pipeline.py
bastings: TFDS is returning text fields as bytes so I sticked to that for performance. Sounds like a good idea to make this consistent where possible.

examples/sst2/train.py
```diff
@@ -0,0 +1,352 @@
...
+    eval_ds: A dataset to be used for the evaluation. Typically valid or test.
+
+  Returns:
+    A dict with the evaluation results.
+  """
+  count = 0
+  total_loss = 0.
+  total_correct = 0
+
+  for ex in tfds.as_numpy(eval_ds):
```
bohnetbd: Would not pass google python review :) No abbreviations that are not standard.

bastings: Fixed.

examples/sst2/model.py
bastings: I added "deterministic" like dropout has, so it's now a more standard API.

examples/sst2/model.py
bastings: I'll experiment with it and update if it indeed works better.

avital:
Could you please also add a link from the README to this example?

PR #151, opened 2020-03-31
remove jaxlib dependency

jheek:
jaxlib should not be installed automatically because users have a choice between
cpu-only/gpu and different cuda versions.

Jax also doesn't include jaxlib as a dependency in setup.py


levskaya:
eh - I see this branch needs updated.

PR #153, opened 2020-03-31
Fix lint errors and missing rngs in Seq2Seq example.

adarob:


levskaya:
Travis CI gives a green checkmark for the final version of this code but github for some reason is confused and doesn't see that.  Manually merging after verifying it looks good.

PR #155, opened 2020-03-31
Use '_rng' for init rng arg to allow apply to have an 'rng' arg.

levskaya:
Some users would like to use 'rng' as an argument name in their module apply functions, but the internal initialization time rng argument was named that, blocking that usage, this fixes that issue.

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -376,7 +376,7 @@ class Module(metaclass=_ModuleMeta):
     return PartialModule
 
   @classmethod
-  def create(cls, rng, *args, name=None, **kwargs):
+  def create(cls, _rng, *args, name=None, **kwargs):
     """Create a module instance by evaluating the model.
 
     DEPRECATION WARNING:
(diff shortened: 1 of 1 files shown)


avital:
Are we sure this is a good change? the `rng` argument to `init`/`init_by_shape` is meaningful (passed through to all of the initializers). What would a user expect if they do something like `module.init(jax.random.PRNGKey(0), rng=jax.random.PRNGKey(1))`?

(Maybe there's an obvious answer but I'm not sure...) Isn't it better to disallow using a `rng` kwarg altogether? (not sure how to do that off the top of my head)

avital:
Actually, I'm curious about the usecase -- why would a module take in another PRNG during init that's not the same as the `rng` argument?

adarob:
It's not for the init, it's for training/inference. See https://github.com/google/flax/pull/111/commits/ebc27036d8750f8ab5380a9d0b704cdfca1685f1

levskaya:
@jheek there will be in python 3.8!  https://www.python.org/dev/peps/pep-0570/

PR #163, opened 2020-04-01
Make SST-2 readme render correctly

AlexeyG:


bastings:
This is a duplicate of #164 
I added a bit more fixing there, so please close this one. Thanks for the quick action though @AlexeyG  :)

PR #164, opened 2020-04-01
Fix setups table in README

bastings:


bastings:
@marcvanzee quick fix for the table in the README

PR #170, opened 2020-04-02
Use github actions for continuous integration

jheek:


jheek:
When GH actions is enabled on the fork it will just work. Arguably, it is a litte easier because you don't have to install the Travis App.

PR #172, opened 2020-04-02
Add scheduled sampling howto for seq2seq example.

adarob:


jheek:
you probable want to add your howto to the docs in `docs/howtos.rst`.

adarob:
@avital @marcvanzee

howtos/diffs/scheduled-sampling.diff
```diff
@@ -0,0 +1,125 @@
...
+-def train_step(optimizer, batch, rng):
++def train_step(optimizer, batch, sample_probability, rng):
+   """Train one step."""
+   labels = batch['answer'][:, 1:]  # remove '=' start token
+ 
+   def loss_fn(model):
+     """Compute cross-entropy loss."""
+-    with nn.stochastic(rng):
+-      logits, _ = model(batch['query'], batch['answer'])
++    logits, _ = model(
```
avital, line 85: Is `with nn.stochastic(rng)` no longer necessary? That would surprise me...

avital: (Or if so, it should be removed from the base example. I suppose we could do that in a separate commit)

adarob: Fixed and tested.

avital:
Super cool!

Issue #157, opened 2020-04-01
Fully deprecate `optimizer.compute_gradient`


avital:
We're replacing the use of `optimizer.compute_gradient` with calls to JAX's `jax.grad` or `jax.value_and_grad`.

- [ ] The docstring for `compute_gradient` should specify that it's deprecated and print a warning, like for `optimizer.optimize()`.
- [ ] The ImageNet example shouldn't use` compute_gradient`

joaogui1:
Guess we can close this @avital

avital:
I agree :) Thanks!

PR #173, opened 2020-04-03
Fix fully deprecate `optimizer.compute_gradient` #157

shiyi001:


shiyi001:
@googlebot I signed it!

flax/optim/base.py
```diff
@@ -245,13 +245,22 @@ def apply_gradient(self, grads, **hyper_param_overrides):
   def compute_gradient(self, loss_fn):
     """Computes gradient of loss_fn.
 
+    DEPRECATION WARNING:
+    compute_gradient() is deprecated.
+    Use jax.grad() or jax.value_and_grad() and apply_gradient() instead.
```
avital, line 250: The "and apply_gradient" part isn't correct.

flax/optim/base.py
```diff
@@ -245,13 +245,22 @@ def apply_gradient(self, grads, **hyper_param_overrides):
...
     Args:
       loss_fn: a function that receives the target and returns a loss or a
         tuple of the loss and auxiliary outputs.
     Returns:
       A tuple consisting of the loss, auxiliary outputs if any,
         and a list of gradient.
     """
+    warnings.warn('compute_gradient() will be removed soon.'
+                  ' Use jax.grad() or jax.value_and_grad()'
+                  ' and apply_gradient() instead.',
```
avital, line 261: ditto

danielsuo:
Two things:

1. Maybe we merge this effort with #207? Looks like that one's further along
2. Agree that automatically testing the examples would be great. That being said...I'm downloading imagenet for the first time and...I can see why imagenet isn't automatically tested! @avital: did you have something else in mind for testing some subset of the imagenet example?

avital:
> 1. Maybe we merge this effort with #207? Looks like that one's further along
I agree. This PR looks stale (and incorrect as it currently stands). Let's close and push on #207 instead.

> 2. Agree that automatically testing the examples would be great. That being said...I'm downloading imagenet for the first time and...I can see why imagenet isn't automatically tested! @avital: did you have something else in mind for testing some subset of the imagenet example?
All TFDS datasets come with a way to generate fake data of the same shape, dtype (and maybe data distibution? not sure about that one) so that would allow us to test one step of our model with the same data preprocessing, but without the dataset. We could also use smaller hyperparameters (e.g. batch size) to make sure they run on CPU.

avital:
> 1. Maybe we merge this effort with #207? Looks like that one's further along

I agree. This PR looks stale (and incorrect as it currently stands). Let's close and push on #207 instead.

> 2. Agree that automatically testing the examples would be great. That being said...I'm 
downloading imagenet for the first time and...I can see why imagenet isn't automatically tested! @avital: did you have something else in mind for testing some subset of the imagenet example?

All TFDS datasets come with a way to generate fake data of the same shape, dtype (and maybe data distibution? not sure about that one) so that would allow us to test one step of our model with the same data preprocessing, but without the dataset. We could also use smaller hyperparameters (e.g. batch size) to make sure they run on CPU.

PR #174, opened 2020-04-04
Fix moco_train example.

us:
I was getting this error.
```
TypeError: 'ShapeDtypeStruct' object is not iterable
```


levskaya:
Just a quick question, in [model_resnet.py#L77](https://github.com/google/flax/blob/master/examples/wip/moco/model_resnet.py#L77) you see that these models are returning a pair of arrays.  Did you maybe alter the model definition used here?  I wasn't able to reproduce the original error when I tried it from code at the head of master.

us:
Hah okey @levskaya, sorry I missed that.

PR #178, opened 2020-04-07
make stateful check work on pytrees

jheek:
The state check should look inside pytrees to find tracers when checking for illegal combinations of state and jax transforms

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -958,7 +958,7 @@ class Collection:
     self._anchor = _module_stack[-1] if _module_stack else None
 
     self._mutable = False
-    self._master = None
+    self._master_level = None
     self._root = None
 
   def as_dict(self):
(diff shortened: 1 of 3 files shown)


flax/nn/base.py
```diff
@@ -1010,10 +1010,9 @@ def store(self, value):
     # is ill-defined (eg. what does vmap of BatchNorm do?).
     # TODO(jheek): Add doc guide on combining jax transforms and state.
     # TODO(jheek): Should some transformations be excempt from this error?
-    master = utils._tracer_of_value(value)
-    value_level = master.level if master else float('-inf')
-    state_level = self._master.level if self._master else float('-inf')
-    if value_level > state_level:
+    value_level = utils._level_of_value(value)
+    print('state_level, value_level', self._master_level, value_level, value)
```
levskaya, line 1014: Do we really want to always print these values?

levskaya: Yo - just checking in again about this PR ;)

jheek: Good catch I left some debug code in my PR

PR #184, opened 2020-04-08
Add init_by_shape method usage example.

us:
I added the module usage example to make the doc more understandable.(This commit just a starter commit, I will continue after I understand where to add on correct way)

Will docstring be used in the future ? So should I add the examples there or to a different place?

### DIFF
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -482,6 +482,10 @@ class Module(metaclass=_ModuleMeta):
       **kwargs: keyword arguments passed to the module's apply function
     Returns:
       A pair consisting of the model output and the initialized parameters
+    Example:
+      input_shape = (batch_size, image_size, image_size, 3)
+      model_output, initial_params = model.init_by_shape(jax.random.PRNGKey(0),
+                                                         input_specs=[(input_shape, jnp.float32)])
     """
     stochastic_rng = None
     try:
--- a/flax/nn/base.py
+++ b/flax/nn/base.py
@@ -483,9 +483,11 @@ class Module(metaclass=_ModuleMeta):
     Returns:
       A pair consisting of the model output and the initialized parameters
     Example:
+      ```
       input_shape = (batch_size, image_size, image_size, 3)
       model_output, initial_params = model.init_by_shape(jax.random.PRNGKey(0),
-                                                         input_specs=[(input_shape, jnp.float32)])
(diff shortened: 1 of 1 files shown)


flax/nn/base.py
```diff
@@ -398,6 +398,10 @@ def create(cls, _rng, *args, name=None, **kwargs):
       **kwargs: keyword arguments passed to the module's apply function
     Returns:
       A pair consisting of the model output and an instance of Model
+    Example:
+      input_shape = (batch_size, image_size, image_size, 3)
+      model_output, initial_params = model.init_by_shape(jax.random.PRNGKey(0),
```
MostafaDehghani, line 403: `init_by_shape` is the counterpart of `create_by_shape`, and using `init` is the alternative of using `create`. 
But regardless of that,  both `create` and `create_by_shape` are going to be removed soon and since they are deprecated, we left no pointer to these two in the examples or documentation to fade them out.

us: Thank you for review! 
but sorry, I write the doc to wrong place :/ I corrected now. Do you have any templates for writing usage examples, etc..

us:
@MostafaDehghani should I continue like that type 

```
Args: ........
Returns:.......
Examples:..........
```
 or do you recommend different style for usage example?

MostafaDehghani:
Thank you @us.
Your proposed format is pretty good (and I think it's a [standard](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)).

The only thing that can make it a bit nicer is to wrap the code in the `Example` section by backticks. 

e.g. In your commit:
````
...
Returns:
      A pair consisting of the model output and the initialized parameters	      A pair consisting of the model output and the initialized parameters
    Example:
     ```
      input_shape = (batch_size, image_size, image_size, 3)
      model_output, initial_params = model.init_by_shape(jax.random.PRNGKey(0),
                                                         input_specs=[(input_shape, jnp.float32)])
     ```
````

And, yes! You are more than welcome to send PRs that add `Example` section to docstrings! :)

levskaya:
Thanks - also when you edit it make sure it fits in the 80char limit to match our style.

us:
Okey @levskaya thank you I will now !

us:
@MostafaDehghani it's didn't look right. How can I fix that. Any resource about that style of doc?
Screen Shot 2020-04-15 at 9 18 20 PM

Issue #138, opened 2020-03-28
Review and update all TFDS pipelines to use modern splitting API.


levskaya:
As noted in #136 the lm1b example we're still using the deprecated TFDS splitting API, we should review and update all of our pipelines to use the modern version.  This is the tracking issue for those fixes.

AlexeyG:
Marking this as PRs welcome, however, we will need to verify that with the new API we get the same results. So the PR should include TensorBoard logs on 8x V100 (see `README.md` of `examples/lm1b` for expected results).

avital:
@andsteing Do you know if we've already made this update in the Linen examples?

levskaya:
Closing this issue as none of the existing linen examples still use the old tfds subsplit API.

Issue #136, opened 2020-03-27
TransformerLM lm1b model accept float and out-of vocabulary token as input


Conchylicultor:
I was experimenting with the `TransformerLM` from the lm1b example (https://github.com/google/flax/blob/master/examples/lm1b/models.py).

I noticed that `TransformerLM` do not raise any error when:
* The input is float (e.g. `np.float32`)
* The input ids are outside the range `[0, vocab_size]`

Both of those seems confusing to me. Is it the expected behavior ?

Example:

```python
from flax.examples.lm1b import models as lm1b_models

model_cls = lm1b_models.TransformerLM.partial(vocab_size=32)

y, params = model_cls.init_by_shape(
  jax.random.PRNGKey(0),
  [((1, 3), jnp.float32)],  # < Tracing with float don't raise error
)
model = flax.nn.Model(model_cls, params)

model(jnp.array([[0.3, 0.5, 2.9]]))  # < Float don't raise error
model(jnp.array([[100, 101, 102]]))  # < Out of vocabulary token ids don't raise error
```

-----
As a side note, I noticed that the input pipeline is using the deprecated TFDS sub-split API: `tfds.Split.TRAIN.subsplit` which is incompatible with the last versions of TFDS (`split='train[90%:]'` should be used instead). 

(minor cleanup) I also think 
```python
dataset_builder = tfds.builder(dataset_name, data_dir=data_dir)
...
train = tfds.load(dataset_name, data_dir=data_dir, split='train')
valid = tfds.load(dataset_name, data_dir=data_dir, split='test')
```
could be rewritten as:
```python
builder = tfds.builder(dataset_name, data_dir=data_dir)
builder.download_and_prepare()  # No-op if data already exists
...
train = builder.as_dataset(split='train')
valid = builder.as_dataset(split='test')
```
This would avoid reloading three time the dataset, which may save a few seconds startup when loading from a remote file systems.

levskaya:
Yeah, currently the model forcibly casts vocab data as `int32` type.  We _could_ be more strict and issue a warning instead (there _is_ a type check in the embedding layer).

Out of vocab is trickier - we can't raise exceptions easily (or at all) on accelerators, so out-of-bounds access silently clamps.  There is discussion in JAX to instead emit NaNs for out-of-bounds access which would error-out harder.
eg.
```python
jnp.take(jnp.array([[1.0,0.0,0.0],[0.0,1.0,0.0],[0.0,0.0,1.0]]), 
[...]

PR #185, opened 2020-04-08
Update examples TFDS input_pipeline

us:
Related issues: #136, #138

### DIFF
--- a/examples/lm1b/input_pipeline.py
+++ b/examples/lm1b/input_pipeline.py
@@ -56,20 +56,20 @@ def train_and_eval_dataset(dataset_name,
   splits = dataset_builder.info.splits
   if tfds.Split.TRAIN not in splits:
     raise ValueError("To train we require a train split in the dataset.")
-  train_ri = tfds.core.ReadInstruction('train')
+  train_split = 'train'
   if eval_holdout_size > 0:
     holdout_percentage = int(eval_holdout_size * 100.0)
     train_percentage = 100 - holdout_percentage
-    train_ri = tfds.core.ReadInstruction('train',  to=train_percentage, unit='%')
-    eval_ri = tfds.core.ReadInstruction('train', from_=train_percentage, to=100, unit='%')
+    train_split = 'train[:{}%]'.format(train_percentage)
+    eval_split = 'train[{}%:]'.format(train_percentage)
   else:
     if tfds.Split.VALIDATION not in splits and "test" not in splits:
       raise ValueError("We require a validation or test split in the dataset.")
-    eval_ri = tfds.core.ReadInstruction('validation')
+    eval_split = 'validation'
     if tfds.Split.VALIDATION not in splits:
-      eval_ri = tfds.core.ReadInstruction('test')
-  train = dataset_builder.as_dataset(split=train_ri, shuffle_files=train_shuffle_files)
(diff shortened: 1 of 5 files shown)


AlexeyG:
The ReadInstruction API seems unnecessarily complicated to me. Could you instead rewrite it using a string definition for the split?

E.g. see https://www.tensorflow.org/datasets/api_docs/python/tfds/core/ReadInstruction
or the proposal in #136 .

us:
@AlexeyG Thank you for reviewing! 

I fixed this please check.

examples/lm1b/input_pipeline.py
```diff
@@ -51,32 +51,25 @@ def train_and_eval_dataset(dataset_name,
...
   if tfds.Split.TRAIN not in splits:
     raise ValueError("To train we require a train split in the dataset.")
-  train_split = tfds.Split.TRAIN
+  train_split = 'train'
   if eval_holdout_size > 0:
     holdout_percentage = int(eval_holdout_size * 100.0)
     train_percentage = 100 - holdout_percentage
-    train_split = tfds.Split.TRAIN.subsplit(tfds.percent[:train_percentage])
-    eval_split = tfds.Split.TRAIN.subsplit(tfds.percent[train_percentage:])
+    train_split = 'train[:{}]'.format(train_percentage)
```
AlexeyG, line 63: I believe that this is not the correct split definition. It will take `train_percentage` examples instead of `train_percentage` percentage of the entire split. Please use the specification in the links I provided earlier.

Can you also provide TensorBoard logs confirming that the metrics after this change are in line with what `README.md` reports?

us:
```
train_split = tfds.percent[:train_percentage]
train = dataset_builder.as_dataset(split=train_split, shuffle_files=train_shuffle_files)
``` 
like that?

AlexeyG: Sorry, should have been more clear. I meant that you're missing the `%`character. So
```
train_split = 'train[:{}%]'.format(train_percentage)
```

us: Ohh I missed it. Thank you! 
 I reseted the wrong commit and forced to push. I think it's solved.

us: @AlexeyG I haven't got a powerful computer so I can't work with a machine like in the report. Also a small computer for training `cifar10`.

AlexeyG:
Thank you for making the change @us 

I tried running the lm1b example, and it crashes. It seems to be due to a bunch of small bugs that crawled into the lm1b example. Once #198  lands, please rebase your branch to master and I'll give the lm1b run another go.

I see that you're adding more commits that are updating the example IO pipelines. I'm inclined to err on the side of caution and request that we see new TensorBoard logs for each one of them. This is also something that #138 describes for lm1b.

I propose that you:
1) Split this into multiple PRs - one per example - this should be much easier/faster to check in than if we wait for all experiments to be re-run. 
2) Focus on smaller models like MNIST or VAE, which can train on CPU.
3) Verify that examples run and submit them with TensorBoard logs (e.g. a link to TensorBoard.dev)

We can then think of moving to larger examples, or may you want to wait until we finish implementing our regression testing framework (e.g. #144 or something similar) that should make example testing much easier.

Let me know what you think!

us:
Okey I closed this pr. I'll send on different prs. Thank you!

PR #186, opened 2020-04-08
Add an Example: pix2pix

us:
TODO: 
- [x] : Input pipeline 
- [x] : Generator and Discriminator Model 
- [ ] : Calculate gradient 
- [x] : Jupyter notebook 

I use [TF pix2pix tutorial](https://www.tensorflow.org/tutorials/generative/pix2pix#training) resource as a guide and I am stuck when calculating the gradient on `train_step` function. Can you help me? 
Thank you!

### DIFF
--- a/examples/pix2pix/train.py
+++ b/examples/pix2pix/train.py
@@ -104,10 +104,17 @@ def train():
   key = jax.random.PRNGKey(0)
   generator_model, generator_state = create_model(key, BATCH_SIZE, IMG_HEIGHT,
                                                   Generator)
-  discriminatogr_model, discriminator_state = create_model(key, BATCH_SIZE,
+  discriminator_model, discriminator_state = create_model(key, BATCH_SIZE,
                                                            IMG_HEIGHT,
                                                            Discriminator)
 
   generator_optimizer = create_optimizer(generator_model, 2e-4, 0.5)
   discriminator_optimizer = create_optimizer(discriminator_model, 2e-4, 0.5)
-  p_train_step = jax.pmap(functools.partial(train_step), axis_name='batch')
+  # p_train_step = jax.pmap(functools.partial(train_step), axis_name='batch')
+  epochs = 100
+  for epoch in range(epochs):
+    print("Epoch: ", epoch)
+    #Train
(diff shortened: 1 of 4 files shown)


AlexeyG:
Hi @us 

I am not familiar with the pix2pix model, so not sure if I can be of much help he, but I can try. Can you be more specific with what exactly you're stuck with?

Are you envisioning this as an example that would become a part of the canonical Flax examples? If so, please check our guidelines regarding adding new examples in `examples/README.md`.

us:
Thank you! I read that and opened the issue #193. 

I need to get grad the loss function but it's the correct way of that? I duplicated code. 
Loss function of pix2pix needs to two model and when getting the get grad of loss_fn, is it correct way? I am so confused.
How can I get grads of two models? please check [this](https://www.tensorflow.org/tutorials/generative/pix2pix#training) for understanding loss func and grad of pix2pix.  
I hope you understand this.

examples/pix2pix/train.py
```diff
@@ -0,0 +1,120 @@
...
+    return gen_total_loss, disc_loss
+
+  step = generator_opt.state.step
+  gen_grad_fn = jax.value_and_grad(loss_fn, has_aux=True)
+  disc_grad_fn = jax.value_and_grad(loss_fn, has_aux=True)
+  _, gen_grad = gen_grad_fn(generator_opt.target, discriminator_opt.target)
+  (gen_total_loss, disc_loss), disc_grad = disc_grad_fn(generator_opt.target,
+                                                        discriminator_opt.target)
+
+  new_gen_opt = generator_opt.apply_gradient(gen_grad)
```
avital, line 95: This won't work as-is. You should have a single optimizer (because the optimizer keeps track of the model parameters). As you have it here, you're really not even optimizing the same model with these two calls to `apply_gradient`.

avital:
Given no response in the past 26 days, I'll close this now. If you get a fully working version of pix2pix, we'd be happy to link to it form https://github.com/google/flax/blob/master/examples/README.md, per our new policy of preferring linking to external repos rather than maintaining them as part of Flax core.

PR #195, opened 2020-04-10
Rename `model_def` to `module` in all instances.

avital:
This was confusing people, and I'm not sure what the original rationale
was (maybe Module used to be named ModelDef?). All instances of `model_def`s
were really just modules that optionally went through calls to `partial`.

One part of the guided tour seems to imply that the intent was that a
"model definition" is a module combined with hyperparameters but that
seems like a pretty minor distinction that's not worth the mental
overhead cost of our users trying to understand what the difference
between model, model def and module are. Fewer mental concepts is
a good direction.


avital:
cc @j-towns

PR #198, opened 2020-04-10
Fix lm1b example: checkpoint loading was not done correctly; the refe…

AlexeyG:
* Checkpoint loading was not done correctly: if there's no checkpoint the optimizer would have been left already replicated.
* The referred to the missing Flax `psum` implementation.


AlexeyG:
It now fails after the first train and eval, so there's at least one more bug. I'll keep digging.

AlexeyG:
Th last commit should do it. I checked that it runs and that samples make sense (are eligible English).

PR #199, opened 2020-04-11
Add github status reporting to github Build action.

levskaya:
This is needed for copybara integration, which assumes the presence of traditional "travis-ci style" commit status metadata.


levskaya:
Note: this extra commit status only works for copybara pushes to the main repo where this extra flag is needed for integrating github actions results.  (PRs won't update this commit status since they don't have the github permissions to do so, but PRs do _not_ need to set a commit status.)

PR #201, opened 2020-04-13
Add Adagrad and RMSProp optimizers

danielsuo:
Hopefully fairly straightforward. If anyone has comments on the computation or feel there should be additional parameters, let me know!

### DIFF
--- a/flax/optim/adagrad.py
+++ b/flax/optim/adagrad.py
@@ -24,6 +24,7 @@ class _AdagradHyperParams:
   """Adagrad hyper parameters"""
 
   learning_rate: float
+  eps: float
 
 
 @struct.dataclass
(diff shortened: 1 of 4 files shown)


flax/optim/adagrad.py
```diff
@@ -0,0 +1,59 @@
...
+  def init_param_state(self, param):
+    """Initialize parameter state"""
+    return _AdagradParamState(jnp.zeros_like(param))
+
+  def apply_param_gradient(self, step, hyper_params, param, state, grad):
+    """Apply per-parameter gradients"""
+
+    assert hyper_params.learning_rate is not None, 'no learning rate provided.'
+    new_G = state.G + jnp.square(grad)
+    new_param = param - hyper_params.learning_rate * grad / jnp.sqrt(new_G)
```
levskaya, line 56: It might be a good idea to also add an epsilon hyperparameter to stabilize the denominator here?  I know for adagrad a zero initial 'G' is a rare edge case, but I imagine it could happen...  if that seems dumb no worries.

danielsuo: Good catch! Added, thanks!

PR #206, opened 2020-04-16
Make the tests run by transforming warnings into errors

mohitreddy1996:
Introduce pytest.ini and introduce configuration to filter and transform warnings into errors.

Few exceptions are documented in the same file. We filter out the specific warning using regex and its type instead of ignoring a generic type of warning which might lead to us missing out on something newly introduced in the flax codebase. Notably causing due to:
- Importing tensorflow.compat.v2.io import gfile in training/checkpoints.py
- XLA level Warning.
- RuntimeWarning due to scipy compiled against an older numpy than installed.

@avital PTAL.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,10 @@
+[pytest]
+# RuntimeWarning is due to scipy (or another package) that was compiled against an older numpy than is installed.
+# UserWarning is due to statement: tensorflow.compat.v2.io import gfile
+# DeprecationWarning is due to statement: tensorflow.compat.v2.io import gfile
+filterwarnings =
+    error
+    ignore:numpy.ufunc size changed.*:RuntimeWarning
+    ignore:No GPU/TPU found, falling back to CPU.*:UserWarning
+    ignore:can't resolve package from.*:ImportWarning
+    ignore:the imp module is deprecated.*:DeprecationWarning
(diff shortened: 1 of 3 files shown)


pytest.ini
```diff
@@ -0,0 +1,10 @@
+[pytest]
+# RuntimeWarning is due to scipy (or another package) that was compiled against an older numpy than is installed.
+# UserWarning is due to statement: tensorflow.compat.v2.io import gfile
+# DeprecationWarning is due to statement: tensorflow.compat.v2.io import gfile
```
avital, line 4: Should we be actually importing gfile in a differnet way?

avital:
Thanks, this is a clear improvement! Should we file separate GitHub tickets to investigate why the other deprecation warnings happen and see if we can eliminate them altogether?

pytest.ini
avital: (to avoid this warning)

mohitreddy1996: Good suggestion. I didn't try it, but let me give it a shot. Thanks!

mohitreddy1996:
> Thanks, this is a clear improvement! Should we file separate GitHub tickets to investigate why the other deprecation warnings happen and see if we can eliminate them altogether?

Yes, I will go ahead and create them.

avital:
Thanks @mohitreddy1996!

Issue #176, opened 2020-04-06
Add tests to all examples that one step of training works without crashing.


avital:

MostafaDehghani:
There is already 

- [ ] mnist
- [ ] Module.apply
- [ ] Module.call
- [ ] Module should get a docstring
- [ ] nn.stateful
- [ ] nn.stochastic
- [ ] Module.get_param (explain when it should be used)

levskaya:
I think the challenge here is running w. minimized-model hparams such that our CI doesn't slow to a terrible crawl, and in that case we're not really testing the training loop, though we would at least be exercising the code paths.

MostafaDehghani:
@AlexeyG has sent a PR with a nice proposal for regression testing for Flax examples (#144). Not sure if we really want to train all our examples to the convergence (although it would be nice), but having that, we probably don't need to have tests with a single step of training for flax examples?

avital:
We won't be able to run full end-to-end tests for all examples on every
commit. And it's valuable to run tests even before you submit a pull
request, as part of your experimentation flow. Moreover, we would want to
run this test on all HOWTOs as well, which won't be covered by the
end-to-end integration and performance testing.

Thus, it's still very valuable to have a small unit test per example (and
is good practice regardless)

danielsuo:
Does it make sense to have different test suites run at different times? E.g.,:

1. for a first, easy pass, just make sure all unit tests pass.
2. Run convergence or regression testing on a beefier set of machines (@marcvanzee pointed out that maybe not all developers can easily do this locally) possibly less frequently.
3. Run performance tests (to flag memory issues or bad optimizations)

avital:
Closing this as it's being taken care of by #231

PR #225, opened 2020-04-26
Refactor mnist/train_test.py from testing single epoch to single train step

mohitreddy1996:
This change is to make the tests across examples consistent.

- GNNs: https://github.com/google/flax/pull/224
- Issue: https://github.com/google/flax/issues/176

### DIFF
--- a/examples/mnist/train_test.py
+++ b/examples/mnist/train_test.py
@@ -28,17 +28,22 @@ jax.config.parse_flags_with_absl()
 class TrainTest(absltest.TestCase):
 
   def test_single_train_step(self):
-    train_ds, _ = train.get_datasets()
+    train_ds, test_ds = train.get_datasets()
     batch_size = 32
     model = train.create_model(random.PRNGKey(0))
     optimizer = train.create_optimizer(model, 0.1, 0.9)
 
-    _, train_metrics = \
-      train.train_step(optimizer=optimizer,
-                       batch={k: v[:batch_size] for k, v in train_ds.items()})
+    # test single train step.
+    optimizer, train_metrics = train.train_step(
+        optimizer=optimizer,
+        batch={k: v[:batch_size] for k, v in train_ds.items()})
     self.assertLessEqual(train_metrics['loss'], 2.302)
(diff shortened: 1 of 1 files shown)


examples/mnist/train_test.py
```diff
@@ -21,26 +21,23 @@
...
+  def test_single_train_step(self):
+    train_ds, _ = train.get_datasets()
+    batch_size = 32
     model = train.create_model(random.PRNGKey(0))
     optimizer = train.create_optimizer(model, 0.1, 0.9)
-    optimizer, train_metrics = train.train_epoch(optimizer, train_ds, 128, 0,
-                                                 input_rng)
-    self.assertLessEqual(train_metrics['loss'], 0.27)
-    self.assertGreaterEqual(train_metrics['accuracy'], 0.92)
-    loss, accuracy = train.eval_model(optimizer.target, test_ds)
```
avital, line 41: I think this test should still execute the `eval_model` step even thought we're only training for one step.

examples/mnist/train_test.py
```diff
@@ -21,26 +21,23 @@
...
-    self.assertLessEqual(loss, 0.06)
-    self.assertGreaterEqual(accuracy, 0.98)
+
+    _, train_metrics = \
```
avital, line 36: I know this should be fixed with a linter and not a human :) but I don't think we use backslashes anywhere else in our code like this for newline. I would do

```
_, train_metrics = train.train_step(optimizer,
                                    batch=...)
```

examples/mnist/train_test.py
avital: (For MNIST it should be fine. For larger datasets that would be too slow, so we should at least test `eval_model` on a subset of the dataset -- say the first 32 elements)

mohitreddy1996: Yup that sounds good. Done!

examples/mnist/train_test.py
mohitreddy1996: Done, my bad. Made the Indentation consistent with train.py

Issue #231, opened 2020-04-30
Guidelines for Examples


Marvin182:
Following some discussion I wrote down a set of guidelines for Flax examples inside the google/flax repository. It will take some work to update the current set of examples but I believe that the examples will be even better afterwards and easier to maintain.

## Guidelines
- Each example should have at least 2 owners . Creating examples cost time and we want to recognize this. We also would like to have a point of contact in case the example needs to be updated to a new API.
- Every example should have a `README.md` which specifies at least:
  - Command for running the example with a link to http://tensorboard.dev for a successful run. Optionally one can also include some output logged on the command line.
  - Link to related references (paper for the model or a great blogpost).
- Keep the example focused.
  - Each example should support a single model on a single dataset.
  - Keep the number of configurable hyper-parameters basic (<10) and use `absl.flags`. We do not want to overwhelm readers.
- Default hyperparameter configuration should be provided as ConfigDict in `configs/default.py`. An additional `configs/test.py` for a CPU friendly unit test case are ok.
- To make it easier to test, maintain and reuse code structure the code as follows:
  - `main.py` contains the flags and calls a method from `train.py` to run the training loop. This should be the only file defining flags! This can be almost identical for all examples, please copy from `linen_examples/wmt/main.py`. If possible stick with the 2 flags (`--workdir` (not `model_dir`) and `--config`).
  - `train.py` contains classes and methods for training and evaluating the model.
  - `train_test.py` for test cases of the training code. At a minimum the test should run a single training step but more fine grained unit tests are a bonus. You can use `tfds.testing.mock_data` to avoid real data from disk.
  - Additional files for more complex architectures or `input_pipeline.py`.
- Only use public datasets available in [TensorFlow Datasets](https://www.tensorflow.org/datasets). TensorFlow Datasets standardizes how datasets downloaded and prepared. Synthetic data is also allowed.
- Follow the [Google Python Style Guide](http://google.github.io/styleguide/pyguide.html) (no relative imports, use type annotation)
- `requirements.txt` if running the example requires additional Python packages that are not installed by Flax or it's dependencies.
- Use the `clu` package where appropriate. This helps us to de-duplicate common functionality and make the code agnostic to the platform. We would like to see examples that work equally well on Google's internal systems as on Cloud or locally.
- 

## Future Work
Examples should also be covered by regression tests, see #144.

We should also add general instructions for running on examples (on GPU and TPU) using Google Cloud.

danielsuo:
[...]

Issue #230, opened 2020-04-29
Run tests with pytest directly instead of via tests/run_all_tests.sh


danielsuo:
After some discussion with @avital, I started a PR, but saw that the change may be a bit more involved. Opening an issue here for feedback.

# Introduction
**Where we are today**: We use `tests/run_all_tests.sh` because our `example` tests have conflicting names (`train_test.py`) and there are some additional checks (e.g., warn if example doesn't have tests). Perhaps by design, `tests/run_all_tests.sh` also lets us have very self-contained examples by collecting/running each one with different `pytest` calls.

**Why that might be suboptimal**: Testing can quickly become brittle. As an example, I wanted to avoid a second call to `pytest tests` while creating coverage reports, so I made [these changes](https://github.com/google/flax/pull/227/files#diff-bcf3e47075e6daafc409adc7a3c0787a) to `tests/run_all_tests.sh`. Not my best work...

**What's the blocker**: `absl` maintains a global `FLAGS` object that raises a `DuplicateFlagError` exception because multiple examples use, for instance, `learning_rate` as a flag. Because `pytest` collects tests before distributing to `pytest-xdist`, we fail.

**Possible solution**: @avital mentioned that @Marvin182 suggested splitting

> all examples into, e.g., `mnist_train.py` (just a train function, no flags, no main) and `mnist_main.py` (with `FLAGS`).

At the end of the day, hopefully this minimally impacts both example maintainers (i.e., follow some documentation to set up an example in the expected way) and users (can still `cd` in and run).

# Proposal
- Standardize example format into `{example}_train.py`, `{example}_main.py`, and any supplementary files, as noted above
- Turn each example into a module (i.e., add `__init__.py`) and switch to relative imports

# Transition
- One PR / example with reviews from example maintainers
- Add notes to `flax/examples/README.md` describing the new example structure
- Remove `tests/run_all_tests.sh`. Given the philosophy re examples, I don't think we need an automated check. If we want all examples to have tests, we just make them have tests.

Marvin182:
I just created #231 to add guidelines for examples. As part of this I'm also advocating to split examples into a library without flags and a small main file with the flags.

I'm not sure I agree with the idea to add a `__init__.py` and switch to relative imports. What benefit would that provide?

danielsuo:
I believe `pytest` will complain without since `examples` is not a package and `pytest` is running from the root repository directory, but I haven't looked very carefully into this. Do you have suggestions?

danielsuo:
[...]

PR #234, opened 2020-05-04
Prototype of removing running `pytest`

danielsuo:
This isn't a fully working example because of the `absl.FLAG` issue in #230, but demonstrates that we can run `pytest` by making otherwise simple modifications:

- Changing example file names according to #231.
- ...ignore the changed version number

These changes allow:
- Running `pytest` from the root `flax` directory
- Running `pytest` from each example directory
- Running each example from the root `flax` directory as a script
- Running each example from the example directory as a script (e.g., `pytest train.py`)
- Each example to be run independently outside of the `flax` source directory

And disallow:
- Importing `examples` or any submodule (i.e., `examples` is not part of `flax`'s exported namespace)

@Marvin182


danielsuo:
Also tagging @levskaya since he knows the rationale for `tests/run_all_tests.sh` best and can point out things I may have missed.

examples/mnist/mnist_lib_test.py
```diff
@@ -17,7 +17,7 @@
 
 from absl.testing import absltest
 
-import train
+from examples.mnist import train
```
avital, line 20: I believe that with this change, we lose the property that "you can easily just fork an example and use it independently" -- how would tests work in that repo? (This code would no longer work, I think)

danielsuo: Yeah, that’s right; one benefit of the existing script is it allows each example to be fully self-contained. If someone were to copy just a single example directory to somewhere outside the repo, Python would get confused by `example.`. I’ll have a think if there’s a way around this issue, but nothing comes immediately to mind.

danielsuo:
My bad. Just needed to _actually_ rename according to #231. Before I just renamed the test files, but needed to rename the files that test files call (e.g., we have many `train.py`). Brain fail.

Confirmed that the previous properties retained + `mnist` can run as a self-contained example.

Anyway, to clarify for @levskaya, this PR is purely for demonstrating what an `mnist` example might look like. Hopefully once all the heavy lifting is done re standardizing examples (e.g., #231), this would be a tiny change.

levskaya:
The latest version just looks like a rename of the mnist example - which seems totally reasonable to me if that follows the newer ideas around naming style.  I'm a teensy bit confused by the last comment - is this just an illustrative PR of future changes or should I review it for merging?

The "run_all_tests.sh" just deals w. treating each example completely separately during CI, and not treating them as part of the flax package - again the idea is to make it abundantly clear that each example is a forkable starting-point, not an api-stable part of the library - we reserve the right to do serious refactors of examples to improve clarity/speed/performance and don't want people depending on subparts of them directly.

danielsuo:
1. Yup, just a rename; I had mistakenly thought replacing the
run_all_tests.sh script with just pytest would be more involved, but after
minimizing the number changes required, that’s clearly not the case (I
hadn’t realized I needed to clear out several places that were keeping
cached versions of flax).

2. No review needed. Should’ve closed once the change became so trivial. If
you’re ok with #231, then when the time comes, #230 is a tiny change
(understanding that we’d lose some self-documentation re “forkable”-ness).

Thanks for taking the time and apologies for the confusion!

levskaya:
no worries!! all thoughts and help most welcome!

PR #235, opened 2020-05-04
RAdam

vballoli:
This PR adds Rectified Adam.


avital:
@levskaya, @jheek -- Looking to solicit some opinions -- should we merge this and maintain in Flax core? We could alternatively link to additional optimizers maintained in external repositories in `flax/optim/README.md` (like we have starting doing with `flax/examples`)

avital:
Assigning to @levskaya just to make sure someone owns the decision on the best path forward.

danielsuo:
Also curious! We are trying many optimizers / variations of optimizers (e.g., per-parameter gradient clipping thresholds) and have so far been keeping them in a separate repository.

Definitely get that we don't want the burden of maintaining so many optimizers, made doubly so by the philosophy to keep code simple, even if the code is duplicated.

Maybe there is a middle path? E.g., have an `experimental` where no guarantees / tests are made (besides maybe some auto-generated tests to make sure the optimizers actually conform to the API). It seems like it will be relatively clear when an optimizer is important enough to pull into `optim` proper on a case-by-case basis?

flax/optim/radam.py
```diff
@@ -0,0 +1,111 @@
...
+      new_param = param - ((grad_ema / grad_sq_ema_) *
+                           step_size * jnp.sqrt(1 - (beta2 ** t)))
+      return new_param
+
+    def low_variance(*args):
+      step_size = hyper_params.learning_rate / (1 - (beta1 ** t))
+      new_param = param - step_size
+      return new_param
+
+    new_param = lax.cond(n_sma_t >= 5,
```
jheek, line 99: this will be verry slow on GPUs because cond will result in a sync between host and gpu for every parameter

flax/optim/radam.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+@struct.dataclass
+class _RAdamParamState:
+  grad_ema: onp.ndarray
+  grad_sq_ema: onp.ndarray
+
+
+class RAdam(OptimizerDef):
+  """RAdam optimizer."""
```
jheek, line 43: In particular for more niche optimizers I thing it is good to have a better description of the algorithm.
Explaining the algorithm and a link to the paper

flax/optim/radam.py
jheek: Also the constant here seems very arbitrary. Should it be configurable?

vballoli: It can be configurable, just followed the original author's implementation and since it was mentioned as a request in one of the optimizer feature request issues, I thought this would be a helpful contribution.

flax/optim/radam.py
vballoli: I'll definitely add more clear documentation. Thanks a lot for the reviews. I apologize for the really late reply.

flax/optim/radam.py
```diff
@@ -0,0 +1,111 @@
...
+    beta1 = hyper_params.beta1
+    beta2 = hyper_params.beta2
+    weight_decay = hyper_params.weight_decay
+    grad_sq = lax.square(grad)
+    grad_ema = beta1 * state.grad_ema + (1. - beta1) * grad
+    grad_sq_ema = beta2 * state.grad_sq_ema + (1. - beta2) * grad_sq
+
+    t = step + 1.
+    n_sma_inf = 2. / (1 - beta2) - 1.
+    n_sma_t = n_sma_inf - 2. * step * (beta2 ** t) / (1. - (beta2 ** t))
```
nestordemeure, line 82: I believe there is a small mistake in your formula, `2.*step` should be `2.*t`.

See line 8 of the algorithm description given page 5 of the [RAdam paper](https://arxiv.org/pdf/1908.03265.pdf).

avital:
RAdam is now in @nestordemeure's [flaxOptimizers](https://github.com/nestordemeure/flaxOptimizers) so I'll close this issue.

Issue #208, opened 2020-04-17
FLIP: Make module instances semantically meaningful by not overriding `module.__new__`


avital:
# Introduction

Currently, while Flax modules are defined by subclassing `flax.nn.Module`, those modules don't behave the same way that normal Python objects behave.

One of the large differences is that Flax Modules override `__new__`, meaning that module instances aren't a semantically meaningful thing in Flax at the moment. Right now, in Flax, what looks like module construction (`nn.Dense(x, features=10)`) actually does two things:
1. Construct an object of type `nn.Dense` (using the non-documented API `module.new_instance()`)
2. Call the `apply` method on that instance and return it.

Some upsides of the current approach are:
1. Modules are defined as a single function, as opposed to, e.g. the style of other libraries, such as Haiku, where you need to scroll up and down between `__init__` and `__call__` to fully understand what a module does.
2. Calls to submodules are very concise, e.g. `nn.Dense(x, features=10)`.

Some downsides of the current approach are:
1. In order to reuse a module, you must use the `module.shared()` abstraction which has a confusing mental model -- what does `module.shared()` return? A module class? A module instance? Moreover, which arguments must be passed into `module.shared()` in order for the shared module to be usable? (Behind the scenes `shared` is implemented on top of `partial`)
2. You can't instantiate a module directly, outside of another module. This leads to surprising things like `new nn.Model(nn.Dense.partial(features=10), params)` -- why do we need to use `partial` to instantiate a Model? What type does the first argument to `nn.Model` have? Is it a module class? Module instance?
3. In a few spots in `flax/nn/base.py` there is code that does "kwarg mangling". Some of these code had bugs before. It would be nice to reduce the need for kwarg mangling.
4. In order to support multiple methods on a module, the `module_method` decorator turns methods that aren't `apply` into new Modules. This is surprising, for example how would I do the equivalent of `module.call(params, *args)` but to call a method `foo` that's not `apply`? That would be `module.foo.call(params, *args)`. That's a pretty surprising mental model.
5. Wanting to define shared parameters or submodules that work across multiple methods requires either using non-traditional patterns and/or with more complexity in Flax core (see discussion on https://github.com/google/flax/issues/161)
6. `apply` was a special-cased method on modules.

# Proposal

1. No longer override `__new__` in Modules
2. Eliminate `.partial()`
3. Potentially eliminate `.shared()` (though we may choose to keep it as a safeguard -- see below)
[...]

PR #237, opened 2020-05-04
Adds Alpha dropout

vballoli:


avital:
Hi @vballoli, thanks for the contribution! I've never heard of Alpha dropout personally, can you share more about where it's used? We'd like to be careful to merge new functionality into Flax core unless we're sure that the Flax core team can adequately maintain the code. I know in case your PR is small but we should still think carefully because it's easy to accidentally accumulate maintenance burden without noticing.

Generally, it's best to clearly describe the purpose of a PR, in this case it's hard to judge.

vballoli:
Hi, sorry for the really late reply! There aren't many use cases per se, but i thought it'd be good to have a non-random dropout as a feature. But I do agree with you, regarding the maintenance and new functionality. If this isn't a good enough contribution, I'll gladly close this PR. 

On a side note, is there a dedicated list of core features required to better help contributions ? I'd really love if there is one to streamline new additions from keen contributors like me!

avital:
Hi @vballoli -- yeah, probably at this point best to close this PR, but if you put your code on your own GitHub repository we are happy to link to it.

Thanks for asking about contributing!

We don't yet have a good list of needed core features, partly because we are in the process of a major API change (initial discussion in #208, though the concrete proposal at the moment has diverged a bit from the exact details there).

After that point, we will try to find time to better organize our thoughts for where the ecosystem is headed and will be more open about possible contributions.

PR #239, opened 2020-05-05
Change RMSProp beta2 default to 0.9

lucasb-eyer:
Looks like the 0.999 is meant to mirror Adam's. But I don't recall seeing RMSProp ever used with such high value. In [the original slides introducing it they use 0.9](http://www.cs.toronto.edu/~tijmen/csc321/slides/lecture_slides_lec6.pdf), which is also the default in both tf.train and tf.keras. [PyTorch defaults to 0.99](https://pytorch.org/docs/stable/optim.html#torch.optim.RMSprop) which I haven't seen before either, and [Alex Grave's seminal paper](https://arxiv.org/abs/1308.0850) that introduces it uses 0.95, although it  is actually a slightly different formulation.

My point is that 0.999 as default is a bad idea here.

### DIFF
--- a/flax/optim/rmsprop.py
+++ b/flax/optim/rmsprop.py
@@ -37,13 +37,13 @@ class _RMSPropParamState:
 
 class RMSProp(OptimizerDef):
   """RMSProp optimizer"""
-  def __init__(self, learning_rate: float = None, beta2=0.999, eps=1e-8):
+  def __init__(self, learning_rate: float = None, beta2=0.9, eps=1e-8):
     """Constructor for the RMSProp optimizer
     
     Args:
       learning_rate: the step size used to update the parameters.
       beta2: the coefficient used for the moving average of the
-        gradient magnitude (default: 0.999).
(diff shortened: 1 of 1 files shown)


danielsuo:
The default actually wasn't meant to mirror Adam's; it was just an experimentally useful value for things we were working on outside of `flax`. Certainly no religion, so no objections from me (the original contributor).

avital:
We don't use RMSProp in any of our example, so we don't need to run any end-to-end testing for this. Happy to merge, generally but I have always assumed that as a default, following PyTorch is probably a good first-order approximation.

@lucasb-eyer -- why did you choose 0.9 here rather than 0.99?

lucasb-eyer:
Chose 0.9 based on 1) the original slides and 2) my past experience from Theano times. I had that default, so did Lasagne and Keras also value comes from back then.

PyTorch inherited 0.99 from Torch7, which got it from [here](https://github.com/torch/optim/pull/49) and that looks completely random to me.

avital:
OK :)

PR #241, opened 2020-05-07
Run unit tests for each affected example in a howto

danielsuo:
From an offline discussion with @avital:

> HOWTOs can fail for two reasons -- the diff can be stale and no longer apply without conflicts, or it can apply and unit tests won't pass. Ideally, both of these should be tested on every commit in CI.

It looks like the first is already tested for during the `apply-howto-branches.yml` workflow. This line checks which files were touched by the `howto` and runs `pytest` on those directories.

It seems that the changes in this PR very narrowly resolve the issue @avital raised; the check assumes:

1. Files that are touched are accompanied by tests in the same directory (examples fall into this category)
2. The only changes a `howto` can make are testable via the tests in the same directory (example tests _should_ fall into this category)

Let me know if/how I misjudged the scope of the issue and I'll be happy to address!

### DIFF
--- a/howtos/scripts/apply_howto_diffs.sh
+++ b/howtos/scripts/apply_howto_diffs.sh
@@ -45,6 +45,10 @@ for howto in $howtos; do
   git apply $diff_file
   git commit -am "Added howto branch ${howto_branch}"
   git push -u origin $howto_branch
+
+  # Run unit test on affected examples only
+  git diff --name-only $master_branch | xargs dirname | xargs pytest
+
   # Make sure to checkout the master branch, otherwise the next diff branch
   # will be branched off of the current diff branch.
   git checkout $master_branch
--- a/howtos/scripts/apply_howto_diffs.sh
+++ b/howtos/scripts/apply_howto_diffs.sh
@@ -43,11 +43,13 @@ for howto in $howtos; do
     exit 1
   fi
   git apply $diff_file
-  git commit -am "Added howto branch ${howto_branch}"
-  git push -u origin $howto_branch
 
   # Run unit test on affected examples only
   git diff --name-only $master_branch | xargs dirname | xargs pytest
+
+  # Once we are satisfied with the howto, commit and push
+  git commit -am "Added howto branch ${howto_branch}"
+  git push -u origin $howto_branch
 
   # Make sure to checkout the master branch, otherwise the next diff branch
   # will be branched off of the current diff branch.


danielsuo:
This errored out with `FutureWarning: jax.random.shuffle is deprecated and will be removed in a future release. Use jax.random.permutation` in the `graph` example test.

danielsuo:
This fails because of #242. Note that the first commit passed and the second didn't without substantive changes between the two.

danielsuo:
Rebased onto #242 after @jekbradbury kindly merged.

avital:
Thanks @danielsuo, it's great to see how short this turned out to be!

I think this isn't a narrow solutions -- HOWTOs are designed for examples only. I'll merge this now.

avital:
Actually, this is good but not sufficient -- if someone breaks tests on a HOWTO in their commit, they should not be able to merge the commit. So this should run as part of unit tests, not only when pushing to master.

avital:
Meaning also editing `.github/workflows/build.yml` appropriately.

danielsuo:
Ah! My fault; I mistakenly thought (and didn't check) that the `apply-howto` workflow was part of the checks on github. I'll put in another PR to fix.

For the record (and also to make sure I didn't miss this second bit), the change in this PR will cause `apply_howto_diffs.sh` to fail when `pytest` returns `1` if a `howto` fails.

avital:
In fact looks like HOWTOs are already broken on master due to the lack of this unit test (not sure since when): https://github.com/google/flax/runs/656295527?check_suite_focus=true

danielsuo:
Also, to clarify, should we edit the unit tests themselves (i.e., `tests/run_all_tests.sh`) so that the `howto` checks run when a user runs unit tests locally or just in the github workflow `build.yml`?

The latter, as you suggested, makes more sense to me; I'm a bit reluctant to have users run a script that alters their local work space.

danielsuo:
Actually, would the simplest thing be to just add the `apply-howto` workflow to the list of checks for merging in?

avital:
The apply-howto does more than running tests -- it actually creates new howto branches and pushes them. That should definitely only be done after merging to master is complete.

So I do think we need a new script that /just/ apply HOWTO diffs and runs tests. Then perhaps we don't need to do that as part of `apply-howto`

danielsuo:
We'd need one modification; we'd need to modify this a little bit to check each howto first before applying any commits (i.e., make the process atomic with respect to `howto`s).

danielsuo:
Haha, crossed wires. Let me know if you still want a separate script or modify `apply-howto-diffs.sh` to be "atomic". Either way is fine by me

danielsuo:
We can see the breaking commit [here](https://github.com/google/flax/actions?query=workflow%3A%22Apply+HOWTO+diffs+to+branches%22).

avital:
> Edit: I'll just make a new script howtos/scripts/check_howto_diffs.sh

Yeah, that sounds right to me.

Once that's in we can theoretically get rid of the change introduced in this PR as master shouldn't be able to reach a state where unit tests fail on top of HOWTOs

avital:
(Actually, I'm not sure I follow what you meant by atomic `apply-howto-diffs.sh`)

danielsuo:
Re this PR: agree; check checks and apply applies. GitHub workflows shouldn't have race conditions, right?

Re atomic: ah, just meant that either all howtos branches are created or none are

Re breaking commit: I'll have a look to see what needs to be done to fix that commit...otherwise we won't be able to merge anything else in

avital:
> Re this PR: agree; check checks and apply applies. GitHub workflows shouldn't have race conditions, right?
Unless someone force pushes to master, then yes as long as tests passed on a PR and then merged to master, master should be clean.

> Re atomic: ah, just meant that either all howtos branches are created or none are
Actually, now that you point this out we do have an atomicity problem right now. If two PRs are merged quickly in succession into master, then two runs of the `apply-howtos` workflow are run in parallel and they fail. This isn't a huge deal in practice, but if there's an easy fix (is there a flag for github actions to not allow them to run in parallel?) that could be nice. But orthogonal to this conversation.

PR #244, opened 2020-05-07
Allow unfrozen dataclasses.

malmaud:
Sometimes it's convenient to have unfrozen dataclasses so that you can modify them before passing them to a Flax module. If the dataclasses are deeply nested, in can be impractical to use `replace` to set deeply-nested values. 

Hacks like turning the dataclass into a dict, modifying the dict, and turning it back into a dataclass can work, but then all the static type safety of dataclasses is lost within the code that mutates the dict. That defeats one of the main reasons of using a dataclass in the first place.

Here I propose adding a `frozen` keyword arg to `@dataclass`, just as the base Python dataclass has. However, I set the default to `frozen=True` and add a stern warning in the docstring (feel free to tweak). I do think if the user wants to opt in to using a mutable dataclass and knows the caveats, they should have the option of doing so.

### DIFF
--- a/flax/struct.py
+++ b/flax/struct.py
@@ -70,10 +70,9 @@ def dataclass(clz=None, *, frozen=True):
 
   Args:
     clz: the class that will be transformed by the decorator.
-    frozen: whether to freeze the dataclass (default=True). WARNING: mutating
-      instance of the dataclass from within a module is not allowed and can
-      silently cause incorrect behavior. Only change this to `False` if you
-      want to mutate the dataclass from *outside* a module.
+    frozen: whether to freeze the dataclass (default=True). WARNING: mutations
+      that occur to an unfrozen dataset within a function that has been transformed
+      by Jax (influxing Flax modules) are not visible outside of that function.
   Returns:
     The new class.
   """
--- a/flax/struct.py
+++ b/flax/struct.py
@@ -71,8 +71,8 @@ def dataclass(clz=None, *, frozen=True):
   Args:
     clz: the class that will be transformed by the decorator.
     frozen: whether to freeze the dataclass (default=True). WARNING: mutations
-      that occur to an unfrozen dataset within a function that has been transformed
-      by Jax (influxing Flax modules) are not visible outside of that function.
+      that occur to an unfrozen dataclass within a function that has been transformed
+      by Jax (including Flax modules) are not visible outside of that function.
   Returns:
     The new class.
   """
--- a/flax/struct.py
+++ b/flax/struct.py
@@ -38,7 +38,7 @@ import dataclasses
 import jax
 
 
-def dataclass(clz):
+def dataclass(clz=None, *, frozen=True):
   """Create a class which can be passed to functional transformations.
 
   Jax transformations such as `jax.jit` and `jax.grad` require objects that are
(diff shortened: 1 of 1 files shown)


flax/struct.py
```diff
@@ -70,66 +70,75 @@ def __apply__(self, *args):
 
   Args:
     clz: the class that will be transformed by the decorator.
+    frozen: whether to freeze the dataclass (default=True). WARNING: mutating
+      instance of the dataclass from within a module is not allowed and can
```
hexahedria, line 74: I'm not sure the warning should be specifically about modules? Seems like you might run into problems with mutation even outside a module if you were doing some other JAX operation. Conversely, doing this inside a module might be safe as long as you knew what you were doing (i.e. locally creating, then mutating, a `flax.struct.dataclass` object you define yourself).

It seems to me the issue is more about the scope of your mutations, in the sense that you shouldn't expect mutating a dataclass inside any (jax-traced) function to have any effect outside that function. Roughly I think the semantics are something like "assume any argument value in your function might be a (possibly abstract) copy of the actual input argument (so changes will be lost), and don't ever put local traced values into global objects, instead make sure to return the object with your modifications".

(I might be missing something regarding module vs not-module, though; someone please correct me if I'm wrong!)

malmaud: Ya you're right, I wrote the docstring thinking of a Flax neural net user but ofc its Jax that's the issue. I just wanted to include some kind of warning in the docstring, but happy to just change it to some generic warning about only setting if you are aware of the caveats of mutation.

malmaud: Updated.

flax/struct.py
```diff
@@ -70,66 +70,74 @@ def __apply__(self, *args):
...
+        names = ','.join(state.keys())
+        raise ValueError(f'Unknown field(s) "{names}" in state dict while'
+                        f' restoring an instance of {clz.__name__}')
+      return x.replace(**updates)
+
+    serialization.register_serialization_state(
+        data_clz, to_state_dict, from_state_dict)
+
+    return data_clz
+  if clz is None:
```
levskaya, line 137: sorry could you explain what this new conditional on clz is for?  I thought this was a required argument?

malmaud: It's just to let `dataclass` be called with or without the `frozen` argument when used as a decorator. eg,

```
@dataclass(frozen=False)
class X:
  ...
```

desugars to `X=dataclass(frozen=True)(X)` so `clz` is `None` in that call to `dataclass`. Meanwhile 

```
@dataclass
class X:
   ...
```

desguars to `X=dataclass(X)`, and there `clz` is `X`.

This is the same pattern used the original dataclasses from https://github.com/ericvsmith/dataclasses/blob/master/dataclasses.py#L935.

flax/struct.py
```diff
@@ -70,66 +70,78 @@ def __apply__(self, *args):
 
   Args:
     clz: the class that will be transformed by the decorator.
+    frozen: whether to freeze the dataclass (default=True). WARNING: mutations
+      that occur to an unfrozen dataset within a function that has been transformed
```
jheek, line 74: dataset -> dataclass

flax/struct.py
```diff
@@ -70,66 +70,78 @@ def __apply__(self, *args):
...
+      by Jax (influxing Flax modules) are not visible outside of that function.
```
jheek, line 75: including

malmaud:
I definitely get that concern, but that is equally a limitation of the original Python `dataclass` that the Flax wrapper is merely inheriting (and which others have complained about on the Python forums). Since this wrapper is just a Jax-compatible dataclass, I'm wondering if it's too ambitious (or even possible) to try to remedy that here.

jheek:
I wonder if we shouldn't just allow arbitrary classes that have fields defined to be registered as pytree nodes. This way the flax dataclass api remains a safe subset of dataclasses.
Otherwise a simple boolean flag changes the mental model and safety of flax dataclasses completely and we need to really on the warning in the docstring for users to be aware.

malmaud:
Ah ya, that sounds like a good idea. IIUC, then we could just use a native Python dataclass, conveniently register it as a pytree node, pass it into Jax-jitted functions, and not have to debate how `flax.struct.dataclass` should work.

How do you imagine that looking like? Would it be a new decorator defied in Jax and Flax, or something Jax would do automatically for any instance of a class that has attributes?

jheek:
For starters the part of the decorator that registers a dataclass as pytree can be factored out into a seperate public api. This way it becomes a backward compatible change. So we could have somthing like `flax.struct.register_as_pytree`

jheek:
@malmaud this issue has been inactive for a while, do you still intent to push this further?

malmaud:
I think this PR can be closed since the solution you're imagining doesn't involve altering flax.struct.dataclass, so it wouldn't really have any code in common with this PR. 
I'd hope to get around to implementing the solution you suggested, but not sure when it would be and would be happy for anyone else to get around to it first.

PR #248, opened 2020-05-08
Split check and apply steps for howtos

danielsuo:
Update: ready for review.

### DIFF
--- a/howtos/scripts/apply_howto_diffs.sh
+++ b/howtos/scripts/apply_howto_diffs.sh
@@ -32,20 +32,13 @@ cd $howto_diff_path
 howtos=$(ls *.diff | sed -e 's/.diff//')
 cd $top_dir
 
-printf "Applying HOWTO diffs to branches..\n"
+printf "Applying HOWTO diffs to branches...\n"
 
 for howto in $howtos; do
   howto_branch="howto/${howto}"
   git checkout -b $howto_branch
   diff_file="${howto_diff_path}/${howto}.diff"
-  if [[ -n $(git apply --check "${diff_file}") ]]; then
-    printf "\nERROR: Cannot apply howto ${howto}! ==> PLEASE FIX HOWTO\n"
-    exit 1
-  fi
   git apply $diff_file
-
-  # Run unit test on affected examples only
-  git diff --name-only $master_branch | xargs dirname | xargs pytest
 
   # Once we are satisfied with the howto, commit and push
   git commit -am "Added howto branch ${howto_branch}"
--- a/howtos/scripts/check_howto_diffs.sh
+++ b/howtos/scripts/check_howto_diffs.sh
@@ -1,0 +1,57 @@
+#!/bin/bash
+#
+# Checks the following for each HOWTO diff in howtos/diffs
+# 1. Check if diffs are stale and can no longer be applied to master
+# 2. Apply each diff in turn and run appropriate unit tests
+
+# https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/
+set -euxo
+
+old_pwd=$(pwd)
+top_dir=$(git rev-parse --show-toplevel)
+howto_diff_path="${top_dir}/howtos/diffs"
+master_branch="origin/master"
+
+# Get names of howtos from diff files.
+cd $howto_diff_path
+howtos=$(ls *.diff | sed -e 's/.diff//')
+cd $top_dir
+
+# Check for stale diffs first before running any tests. This duplicates some
+# logic, but should speed up testing.
+printf "Verify all HOWTO diffs can be applied...\n"
+
+for howto in $howtos; do
+  diff_file="${howto_diff_path}/${howto}.diff"
+
+  # Check if command fails
+  # See: https://stackoverflow.com/q/26675681/
+  if ! git apply --check "${diff_file}"; then
+    printf "\nERROR: Cannot apply howto ${howto}! ==> PLEASE FIX HOWTO\n"
+    exit 1
+  fi
+done
+
+printf "Run unit tests for each diff...\n"
+
+# Fetch all branches
+git fetch --all
+
+for howto in $howtos; do
+  diff_file="${howto_diff_path}/${howto}.diff"
+  git apply $diff_file
+
+  # Run unit test on affected examples only
+  if ! git diff --name-only $master_branch | xargs dirname | xargs pytest; then
+    printf "\nERROR: Tests failed for howto ${howto}! ==> PLEASE FIX HOWTO\n"
+
+    # Undo patch in case we're running locally
+    git apply -R $diff_file
+    exit 1
+  fi
+
+  # Undo patch so we can run the next test
+  git apply -R $diff_file
+done
+
+cd $old_pwd
--- a/howtos/scripts/check_howto_diffs.sh
+++ b/howtos/scripts/check_howto_diffs.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 #
 # Checks the following for each HOWTO diff in howtos/diffs
-# 1. Check if diffs are stale and can no longer be applied to master
-# 2. Apply each diff in turn and run appropriate unit tests
+# 1. Check if diffs are stale and can no longer be applied to *this branch*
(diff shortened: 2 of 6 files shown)


howtos/scripts/apply_howto_diffs.sh
```diff
@@ -32,16 +32,12 @@ cd $howto_diff_path
...
 cd $top_dir
 
-printf "Applying HOWTO diffs to branches..\n"
+printf "Applying HOWTO diffs to branches...\n"
 
 for howto in $howtos; do
   howto_branch="howto/${howto}"
   git checkout -b $howto_branch
   diff_file="${howto_diff_path}/${howto}.diff"
-  if [[ -n $(git apply --check "${diff_file}") ]]; then
```
danielsuo, line 41: This line was always returning false, so we never hit this branch.

danielsuo:
One comment: as the number of howtos increases, we should really consider parallelizing the check step for howtos (one easy way to do this is to use GNU `parallel` and feed it a script that takes list of HOWTO names). They currently take ~40% of the PR testing.

The other thing to do, of course, is to improve the `mnist` example (#225 seems like the right idea).

danielsuo:
@marcvanzee I hate to be a bother, but would you mind having a look at this? Without this issue, we will continue to fail the `apply-howto` workflow and as folks continue to update examples/howtos, this PR will get stale (it currently fails because someone changed the `mnist` example, but didn't update the howtos).

I am paying particular attention to the issue since I'm the reason the `apply-howto` workflow is failing in the first place!

danielsuo:
Just updated the HOWTO diffs so that they can both be applied and pass unit tests. #225 broke `distributed-training`, `ensembling`, and `polyak-averaging`.

Not great that I'm including the howto fix-ups in this PR, but wanted to show that this PR does indeed do the correct thing.

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+howto_diff_path="${top_dir}/howtos/diffs"
+
+# Get names of howtos from diff files.
+cd $howto_diff_path
+howtos=$(ls *.diff | sed -e 's/.diff//')
+cd $top_dir
+
+# Check for stale diffs first before running any tests. This duplicates some
+# logic, but should speed up testing.
+printf "Verify all HOWTO diffs can be applied...\n"
```
marcvanzee, line 21:
```suggestion
printf "Verifying all HOWTO diffs can be applied...\n"
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+
+  # Check if command fails
+  # See: https://stackoverflow.com/q/26675681/
+  if ! git apply --check "${diff_file}"; then
+    printf "\nERROR: Cannot apply howto ${howto}! ==> PLEASE FIX HOWTO\n"
+    exit 1
+  fi
+done
+
+printf "Run unit tests for each diff...\n"
```
marcvanzee, line 34:
```suggestion
printf "Running unit tests for each diff...\n"
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+#
+# Checks the following for each HOWTO diff in howtos/diffs
```
marcvanzee, line 3:
```suggestion
# Checks the following for each HOWTO diff in howtos/diffs:
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+  # See: https://stackoverflow.com/q/26675681/
+  if ! git apply --check "${diff_file}"; then
+    printf "\nERROR: Cannot apply howto ${howto}! ==> PLEASE FIX HOWTO\n"
+    exit 1
+  fi
+done
+
+printf "Run unit tests for each diff...\n"
+
+# Fetch all branches
```
marcvanzee, line 36:
```suggestion
# Fetch all branches.
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+git fetch --all
+
+for howto in $howtos; do
+  diff_file="${howto_diff_path}/${howto}.diff"
+  curr_branch="$(git rev-parse --abbrev-ref HEAD)"
+
+  # Delete local howto branch if already exists in case we're running locally
```
marcvanzee, line 43:
```suggestion
  # Delete local howto branch if it already exists (in case we're running locally).
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+  git branch -D $howto || true
+  git checkout -b $howto
+  git apply $diff_file
+
+  # Run unit test on affected examples only
```
marcvanzee, line 48:
```suggestion
  # Run unit test on affected examples only.
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+  if ! git diff --name-only $curr_branch | xargs dirname | xargs pytest; then
+    printf "\nERROR: Tests failed for howto ${howto}! ==> PLEASE FIX HOWTO\n"
+
+    # Undo patch in case we're running locally
```
marcvanzee, line 52:
```suggestion
    # Undo patch in case we're running locally.
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+done
+
+printf "Run unit tests for each diff...\n"
+
+# Fetch all branches
+git fetch --all
+
+for howto in $howtos; do
+  diff_file="${howto_diff_path}/${howto}.diff"
+  curr_branch="$(git rev-parse --abbrev-ref HEAD)"
```
marcvanzee, line 41:
```suggestion
curr_branch="$(git rev-parse --abbrev-ref HEAD)"
for howto in $howtos; do
  diff_file="${howto_diff_path}/${howto}.diff"
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+  if ! git diff --name-only $curr_branch | xargs dirname | xargs pytest; then
+    printf "\nERROR: Tests failed for howto ${howto}! ==> PLEASE FIX HOWTO\n"
+
+    # Undo patch in case we're running locally
+    git apply -R $diff_file
+    git checkout $curr_branch
+    exit 1
+  fi
+
+  # Undo patch so we can run the next test
```
marcvanzee, line 58:
```suggestion
  # Undo patch so we can run the next test.
```

howtos/scripts/check_howto_diffs.sh
```diff
@@ -0,0 +1,63 @@
...
+set -euxo
+
+old_pwd=$(pwd)
+top_dir=$(git rev-parse --show-toplevel)
+howto_diff_path="${top_dir}/howtos/diffs"
+
+# Get names of howtos from diff files.
+cd $howto_diff_path
+howtos=$(ls *.diff | sed -e 's/.diff//')
+cd $top_dir
```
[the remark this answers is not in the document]
danielsuo: Can do! What do you think we might refactor?

For example, `old_pwd` (line 10) to `cd $top_dir` (line 17) is setup code. Maybe that could be sourced from a separate file? Or a function to loop through `howto`s?

Anyway, I saw that there was already some duplication between `apply_howto_diffs` and `pack_howto_diffs` (though not as much) so thought I should adhere duplicated code for sake of readability.

danielsuo:
Thanks for the super fast reply! I think I got everything (at least all the requested changes show `outdated` now).

howtos/scripts/check_howto_diffs.sh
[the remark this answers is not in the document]
marcvanzee: Sounds good, perhaps let's forget about factoring out stuff in Bash scripts. If they turn out to be very complex, I guess we better switch to another programming language.

PR #251, opened 2020-05-08
Update howtos so all diffs can be applied and all pass example tests

danielsuo:
Some notes:
- Just updated tests to match whatever was in `train.train` 
- Had to change loss/accuracy thresholds slightly for the ensembling example

Hopefully nothing wrong or controversial.

Once this is in, we can verify that #248 is working (confirmed that #248 with this PR works locally).

### DIFF
--- a/howtos/diffs/ensembling.diff
+++ b/howtos/diffs/ensembling.diff
@@ -1,5 +1,5 @@
 diff --git a/examples/mnist/train.py b/examples/mnist/train.py
-index 51d2fde..0ca3365 100644
+index 362b9b2..fe44d34 100644
 --- a/examples/mnist/train.py
 +++ b/examples/mnist/train.py
 @@ -23,6 +23,9 @@ from absl import app
@@ -11,8 +11,8 @@ index 51d2fde..0ca3365 100644
 +from flax import jax_utils
  from flax import nn
  from flax import optim
- 
-@@ -80,12 +83,14 @@ class CNN(nn.Module):
+ from flax.metrics import tensorboard
+@@ -79,12 +82,14 @@ class CNN(nn.Module):
      return x
  
  
@@ -27,7 +27,7 @@ index 51d2fde..0ca3365 100644
  def create_optimizer(model, learning_rate, beta):
    optimizer_def = optim.Momentum(learning_rate=learning_rate, beta=beta)
    optimizer = optimizer_def.create(model)
-@@ -111,7 +116,7 @@ def compute_metrics(logits, labels):
+@@ -110,7 +115,7 @@ def compute_metrics(logits, labels):
    return metrics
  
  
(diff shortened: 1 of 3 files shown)


howtos/diffs/ensembling.diff
```diff
@@ -94,26 +94,43 @@ index 51d2fde..0ca3365 100644
...
+     optimizer, train_metrics = train_epoch(
          optimizer, train_ds, batch_size, epoch, input_rng)
      loss, accuracy = eval_model(optimizer.target, test_ds)
 -    logging.info('eval epoch: %d, loss: %.4f, accuracy: %.2f',
 +    # `loss` and `accuracy` are now 1D arrays of length `jax.device_count()`
 +    logging.info('eval epoch: %d, loss: %s, accuracy: %s',
                   epoch, loss, accuracy * 100)
-   return optimizer
- 
+     summary_writer.scalar('train_loss', train_metrics['loss'], epoch)
```
avital, line 104: I know this isn't a problem in this PR, but I have confused myself now -- what is the "train loss" scalar for an ensemble training?

I'm just putting it here to share my confusing but it shouldn't block merging this PR.

howtos/diffs/polyak-averaging.diff
```diff
@@ -61,6 +61,21 @@ index 51d2fde..eab7983 100644
...
+index fb02770..4cfec2a 100644
+--- a/examples/mnist/train_test.py
++++ b/examples/mnist/train_test.py
+@@ -34,8 +34,8 @@ class TrainTest(absltest.TestCase):
+     input_rng = onp.random.RandomState(0)
+     model = train.create_model(random.PRNGKey(0))
+     optimizer = train.create_optimizer(model, 0.1, 0.9)
+-    optimizer, train_metrics = train.train_epoch(optimizer, train_ds, 128, 0,
+-                                                 input_rng)
++    optimizer, train_metrics = train.train_epoch(optimizer, model.params, train_ds,
```
avital, line 77: This seems wrong? `train_epoch` does not take a params 2nd argument?

avital: (Do tests pass on this HOWTO? I'm surprised)

danielsuo: I think this diff of diff might be confusing to read; the signature of train_epoch that this diff of diff references is with respect to the original example, but if you look at the original diff file itself (ie polyak-averaging.diff), the signature of train_epoch changes.

(Ah shoot, on mobile and copied the line not the link to the line):

+def train_epoch(optimizer, params_ema, train_ds, batch_size, epoch, rng):

danielsuo: Link: https://github.com/google/flax/blob/dea96db3fa779b25c32ab8c40f4b55a7d26fec93/howtos/diffs/polyak-averaging.diff#L32

avital: Ah, I see. This is why the tests weren't passing originally.

Issue #249, opened 2020-05-08
`./howtos/scripts/pack_howto_diffs.sh` adds all untracked files, not just files shown to user


danielsuo:
Issue is straightforward:
- `./howtos/scripts/pack_howto_diffs.sh` gives a warning stating which files will be packed into a diff
- Later, it runs `git add .` on all untracked files

Was surprised when my `.vim` and `environment.yml` files got packed in.

Not sure what folks would like the solution to be since this gets to what is fair game for a `howto` diff (e.g., which directories can in touch).

danielsuo:
Er, started a bug template, but then deleted; didn't realize it would add a label. That's a bit aggressive IMO, but I can't remove.

danielsuo:
It looks like at minimum, packing the `examples`. It didn't sound like folks wanted examples or howtos to modify core code, but I'm not sure. I could imagine also packing `tests` in case we want unit tests for `howto`s / we need to modify existing tests (that might be a no-no).

avital:
> That's a bit aggressive IMO, but I can't remove the bug label.
Yeah, we need other types of labels...

avital:
It's OK to assume that HOWTOs only affect examples/. In fact, we already
assume that in `pack_howto_diffs.sh`.

avital:
`tests/` can't depend on examples (by design) so it should only be necessary to run the particular tests for the example being changed by the howto.

(Right now each HOWTO modifies a single example, and I think that's probably the way we'll go, in case that assumption helps the scripting in any way)

danielsuo:
> > That's a bit aggressive IMO, but I can't remove the bug label.
> 
> Yeah, we need other types of issue templates...

Ah, oops, I meant this particular issue is too aggressively labeled a bug; I think the template is fine.

> It's OK to assume that HOWTOs only affect examples/. In fact, we already assume that in `pack_howto_diffs.sh`.

Yep! Just meant that `git add .` then adds more stuff into the commit besides just the changes to the example.

> `tests/` can't depend on examples (by design) so it should only be necessary to run the particular tests for the example being changed by the howto.
> 
> (Right now each HOWTO modifies a single example, and I think that's probably the way we'll go, in case that assumption helps the scripting in any way)

Perfect!

PR #255, opened 2020-05-12
Only pack examples directories when packing howtos

danielsuo:
Addresses #249. After reviewing the blame history, however, I can't figure out if there was a reason why we add all untracked files and want to be sure I didn't miss something important.

@marcvanzee originally had `git add *`, which @avital then changed to `git add .` to [bring back the correct ensembling HOWTO](https://github.com/danielsuo/flax/commit/57fce50c7be2b14d17beecb7b2e0f666bde8b7f3).

Hopefully the change really is this simple, but let me know.


avital:
Thanks Daniel!

PR #257, opened 2020-05-13
support axis index groups for distributed batchnorm

jekbradbury:
Now that https://github.com/google/jax/pull/2382 is in upstream, I'm resuscitating #42. As before, the target use case is a situation where you want your batchnorm reductions to take place over a "batch" larger than each individual device but smaller than the whole cluster.

In the meantime I spent some time exploring whether we could use nested `pmap` instead (part of the original intent of Flax batchnorm taking an axis name is that it wouldn't necessarily have to be the same axis name used in the gradient reduction, and I've long thought of distributed batchnorm as a good use case for nested `pmap`). But I decided that wasn't the right approach for now, even though it's very appealing in principle, for two reasons:
- The training loop changes needed to use nested `pmap` for distributed batchnorm are somewhat larger than those needed to use `axis_index_groups`. This is because `pmap` axes are orthogonal, so the overall gradient reduction axis would need to be split into `"inner"` and `"outer"` axes (where batchnorm `psum`s would just use `"inner"` and gradient `psum`s would use `("inner", "outer")`) and intermediate arrays in the training loop would need to have two leading mapped axes.
- Users of distributed batchnorm want more control over how the subgroups are laid out physically (for locality, or to choose a particular topology), and the XLA:TPU compiler is currently sensitive to even the order of replicas within each `replica_groups` group. While some combination of JAX and XLA could eventually make good choices here automatically, we're pretty far away from either that or a comprehensive manual logical->physical mapping API for nested pmaps.

This is testable in OSS, but I'm not sure how much tests here would add over upstream (since we're just passing a kwarg through).

### DIFF
--- a/flax/nn/normalization.py
+++ b/flax/nn/normalization.py
@@ -43,7 +43,8 @@ class BatchNorm(base.Module):
             scale=True,
             bias_init=initializers.zeros,
             scale_init=initializers.ones,
-            axis_name=None):
+            axis_name=None,
+            axis_index_groups=None):
     """Normalizes the input using batch statistics.
 
     Args:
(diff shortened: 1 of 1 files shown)


jekbradbury:
Merging to get a version of this internally; happy to make changes later (or e.g. drop the API in favor of a future iteration of nested pmap when we can).

PR #258, opened 2020-05-13
Migrate TF1 compat Summary writer to TF2 in metrics/tensorboard.py

mohitreddy1996:
Migrate TF1 based event files/summaries to TF2 based summaries. Adding tests for the same.

NOTE: I am not sure how to show the outputs of the logged tensorboard. I am able to check them locally and they look fine.

tensorboard.dev for scalar - https://tensorboard.dev/experiment/fNiwW36JR96ekMTRVmPLrg/#scalars

- tensorboard.dev supports only scalars so majority cannot be validated using tensorboard.dev
- uploading the event files? - difficult to validate by the reviewers.
- screenshots?

### DIFF
--- a/flax/metrics/tensorboard.py
+++ b/flax/metrics/tensorboard.py
@@ -75,6 +75,8 @@ class SummaryWriter(object):
     Args:
       tag: str: label for this data
       image: ndarray: [H,W], [H,W,1], [H,W,3] save image in greyscale or colors.
+        Pixel values could be either uint8 or float.
+        Floating point values should be in range [0, 1).
       step: int: training step
     """
     image = onp.array(image)
@@ -85,6 +87,9 @@ class SummaryWriter(object):
     # tf.summary.image expects image to have shape [k, h, w, c] where,
     # k = number of samples, h = height, w = width, c = number of channels.
     image = image[onp.newaxis, :, :, :]
+
+    # Convert to tensor value as tf.summary.image expects data to be a tensor.
+    image = tf.convert_to_tensor(image)
     with self._event_writer.as_default():
       tf.summary.image(name=tag, data=image, step=step)
 
@@ -105,9 +110,12 @@ class SummaryWriter(object):
     # tf.summary.audio expects the audio data to have floating values in
     # [-1.0, 1.0].
     audiodata = onp.clip(onp.array(audiodata), -1, 1)
+
+    # Convert to tensor value as tf.summary.audio expects data to be a tensor.
+    audio = tf.convert_to_tensor(audiodata, dtype=tf.float32)
     with self._event_writer.as_default():
       tf.summary.audio(
-          name=tag, data=audiodata, sample_rate=sample_rate, step=step,
+          name=tag, data=audio, sample_rate=sample_rate, step=step,
           max_outputs=max_outputs, encoding='wav')
 
   def histogram(self, tag, values, step, bins=None):
--- a/flax/metrics/tensorboard.py
+++ b/flax/metrics/tensorboard.py
@@ -32,11 +32,8 @@ with warnings.catch_warnings():
 # pylint: disable=g-import-not-at-top
 import matplotlib.pyplot as plt
 import numpy as onp
-import tensorflow.compat.v1 as tf
-from tensorflow.compat.v1 import HistogramProto
-from tensorflow.compat.v1 import Summary
-from tensorflow.compat.v1 import SummaryMetadata
-from tensorflow.compat.v1.io import gfile
+import tensorflow.compat.v2 as tf
+from tensorflow.compat.v2.io import gfile
 
 from tensorflow.core.util import event_pb2  # pylint: disable=g-direct-tensorflow-import
 from tensorflow.python.summary.writer.event_file_writer import EventFileWriter  # pylint: disable=g-direct-tensorflow-import
@@ -58,11 +55,9 @@ class SummaryWriter(object):
     self._event_writer = EventFileWriter(log_dir, 10, 120, None)
     self._closed = False
 
-  def _add_summary(self, summary, step):
+  def _add_summary(self, summary):
     event = event_pb2.Event(summary=summary)
     event.wall_time = time.time()
-    if step is not None:
-      event.step = int(step)
     self._event_writer.add_event(event)
 
   def close(self):
@@ -84,8 +79,8 @@ class SummaryWriter(object):
       step: int: training step
     """
     value = float(onp.array(value))
-    summary = Summary(value=[Summary.Value(tag=tag, simple_value=value)])
-    self._add_summary(summary, step)
+    summary = tf.summary.scalar(tag, value, step=step)
+    self._add_summary(summary)
 
   def image(self, tag, image, step):
     """Saves RGB image summary from onp.ndarray [H,W], [H,W,1], or [H,W,3].
@@ -103,13 +98,9 @@ class SummaryWriter(object):
       image = onp.repeat(image, 3, axis=-1)
     image_strio = io.BytesIO()
     plt.imsave(image_strio, image, vmin=0., vmax=1., format='png')
-    image_summary = Summary.Image(
-        encoded_image_string=image_strio.getvalue(),
-        colorspace=3,
-        height=image.shape[0],
-        width=image.shape[1])
-    summary = Summary(value=[Summary.Value(tag=tag, image=image_summary)])
-    self._add_summary(summary, step)
+    image_summary = tf.summary.image(name=tag,
+        data=[1, image.shape[0], image.shape[1], 3], step=step)
+    self._add_summary(image_summary)
 
   def audio(self, tag, audiodata, step, sample_rate=44100):
     """Saves audio.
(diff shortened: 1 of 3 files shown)


tests/tensorboard_test.py
```diff
@@ -0,0 +1,204 @@
...
+    summary_value = self.parse_and_return_summary_value(path=log_dir)
+    self.assertEqual(summary_value.tag, 'text_test')
+    self.assertEqual(
+        tensor_util.make_ndarray(summary_value.tensor).item().decode('utf-8'),
+        text)
+
+  def test_summarywriter_image(self):
+    log_dir = tempfile.mkdtemp()
+    summary_writer = SummaryWriter(log_dir=log_dir)
+    img = tf.random.uniform(shape=[30, 30, 3])
```
AlexeyG, line 97: Why are these TF tensors and not numpy arrays?

AlexeyG: Same comment also below. Shouldn't the summary writer get numpy arrays or JAX DeviceArrays?

tests/tensorboard_test.py
```diff
@@ -0,0 +1,204 @@
...
+    summary_writer.image(tag='2dimage_test', image=img, step=1)
+    summary_value = self.parse_and_return_summary_value(path=log_dir)
+
+    self.assertEqual(summary_value.tag, '2dimage_test')
+    expected_img = tf.image.decode_image(summary_value.tensor.string_val[2])
+    # assert the image was increased in dimension
+    self.assertEqual(expected_img.shape, (30, 30, 3))
+
+  def test_summarywriter_single_channel_image_scaled(self):
+    log_dir = tempfile.mkdtemp()
```
AlexeyG, line 120: Seems like you could move `log_dir` and `summary_writer` creation into a setUp method.

tests/tensorboard_test.py
```diff
@@ -0,0 +1,204 @@
...
+        map(_process_event, event_file_generator))
+    for value_dict in event_values:
+      event_value_list.append(value_dict)
+
+    self.assertLen(event_value_list, 1)
+    self.assertEqual(event_value_list[0]['step'], 1)
+    self.assertGreater(event_value_list[0]['wall_time'], 0.0)
+    return event_value_list[0]['value']
+
+  def assert_encoded_audio_files_equal(self, actual_audio, expected_audio):
```
AlexeyG, line 53: I am somewhat in favour of inlining assert and trim functions form this method into the actual tests using this helper method. It would not duplicate a lot of code, but make the tests easier to read.

tests/tensorboard_test.py
```diff
@@ -0,0 +1,204 @@
...
+
+    expected_audio = \
+        tf.audio.decode_wav(summary_value.tensor.string_val[2]).audio
+    self.assert_encoded_audio_files_equal(
+        actual_audio=audio[1], expected_audio=expected_audio)
+
+  def test_summarywriter_audio_sampled_output(self):
+    log_dir = tempfile.mkdtemp()
+    summary_writer = SummaryWriter(log_dir=log_dir)
+    audio = tf.random.uniform(shape=[2, 48000, 2], minval=-1.0, maxval=1.0)
```
AlexeyG, line 158: Maybe also have a test where the audio is clipped by the writer?

tests/tensorboard_test.py
```diff
@@ -0,0 +1,204 @@
...
+  def test_summarywriter_histogram(self):
+    log_dir = tempfile.mkdtemp()
+    summary_writer = SummaryWriter(log_dir=log_dir)
+    histogram = onp.arange(1000)
+    summary_writer.histogram(tag='histogram_test', values=histogram, step=1)
+
+    summary_value = self.parse_and_return_summary_value(path=log_dir)
+    self.assertEqual(summary_value.tag, 'histogram_test')
+    expected_histogram = tensor_util.make_ndarray(summary_value.tensor)
+    self.assertTrue(expected_histogram.shape, (30, 3))
```
AlexeyG, line 183: I'm puzzled by the shape. How does one arrive to (30, 3) ? And what are the entries of `expected_histogram` ?

flax/metrics/tensorboard.py
```diff
@@ -84,134 +66,74 @@ def scalar(self, tag, value, step):
...
-        tensor = tf.make_tensor_proto(
-            values=[
-                td.encode(encoding='utf_8') for td in onp.reshape(textdata, -1)
-            ],
-            shape=(datashape[0], datashape[1]))
-    summary = Summary(
-        value=[Summary.Value(tag=tag, metadata=smd, tensor=tensor)])
-    self._add_summary(summary, step)
+    if not isinstance(textdata, (str, bytes)):
+      raise ValueError('textdata should be of the type `str` or `bytes`.')
```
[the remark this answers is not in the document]
mohitreddy1996: Hey Alexey, I am sorry but github review is messed up. Which line is this comment for?

tests/tensorboard_test.py
mohitreddy1996: Yeah I think that makes sense. Inlined the necessary assertions. Thanks!

tests/tensorboard_test.py
mohitreddy1996: Thanks for catching this. Yes, these should be ndarrays as mentioned in the docstrings of the methods. 

Changed the input for image and audio both.

tests/tensorboard_test.py
mohitreddy1996: Yeah I had something like that initially. Just wasn't sure if it actually adds a lot of value v/s being explicit and verbose about how the test is being setup (temp file + summary writer). Felt the latter was more readable.. WDTY?

tests/tensorboard_test.py
mohitreddy1996: Done, thanks for the pointer!

tests/tensorboard_test.py
mohitreddy1996: Yeah my bad, the default value of bins assumed by the tf.summary.histogram is 30. Added the comment and adjusted the test name accordingly.

flax/metrics/tensorboard.py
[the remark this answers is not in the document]
AlexeyG: Line 137. The comment did not render correctly, but I was trying to ask wether it makes sense to have single quotes around the variable name `textdata` in the error message.

tests/tensorboard_test.py
AlexeyG: Personal preference would be to move the boilerplate out of the test methods, but don't have a strong opinion. Please go with what you prefer / have already implemented.

tests/tensorboard_test.py
AlexeyG: Thank you for the clarification!

tests/tensorboard_test.py
```diff
@@ -0,0 +1,233 @@
...
+    summary_writer = SummaryWriter(log_dir=log_dir)
+    img = onp.random.uniform(low=0., high=1., size=(30, 30, 3))
+    summary_writer.image(tag='image_test', image=img, step=1)
+    summary_value = self.parse_and_return_summary_value(path=log_dir)
+
+    self.assertEqual(summary_value.tag, 'image_test')
+    expected_img = tf.image.decode_image(summary_value.tensor.string_val[2])
+    # convert img to uint8 as the decoded image is of type uint8.
+    actual_img = tf.image.convert_image_dtype(image=img, dtype=onp.uint8)
+    self.assertTrue(onp.allclose(actual_img, expected_img.numpy()))
```
AlexeyG, line 108: `actual_img` is a TF tensor, isn't it? Shouldn't there be a `.numpy()` for it too?

tests/tensorboard_test.py
```diff
@@ -0,0 +1,233 @@
...
+    summary_writer.audio(tag='audio_test', audiodata=audio, step=1)
+
+    summary_value = self.parse_and_return_summary_value(path=log_dir)
+    self.assertEqual(summary_value.tag, 'audio_test')
+
+    # Assert two audio files are parsed.
+    self.assertLen(summary_value.tensor.string_val, 4)
+
+    # Assert values.
+    expected_audio = \
```
AlexeyG, line 149: Are the names `audio` and `expected_audio` swapped maybe?

`expected_audio` should be the one we generate and write (so line 139) and the audio should be the one we read from summaries (so line 150). Or I am really misreading this?

I think this also applies to some other tests.

flax/metrics/tensorboard.py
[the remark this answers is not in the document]
mohitreddy1996: Yup, makes sense. Made the change

tests/tensorboard_test.py
mohitreddy1996: Since expected_img is a numpy of dtype float32 and in the range of 0.0 to 1.0, summary writer internally scales up the pixel values accordingly. So here we need to do the same for assertion. So we scale up the values and also set dtype as numpy uint8 (this dictates the dtype of the output object).

tests/tensorboard_test.py
mohitreddy1996: yup, that's true. My bad, made the changes in rest of the occurrences as well. Thanks for catching this!

tests/tensorboard_test.py
AlexeyG: I am not sure I follow. I think what I was trying to say is that `actual_img` and `expected_img` are both TF tensors, so it seem natural that we should convert both (or not cover either) to numpy ndarrays before the assertion in line 108, i.e. should it not be

```
self.assertTrue(onp.allclose(actual_img.numpy(), expected_img.numpy()))
```

If it works in the current form, then my comment is just a small nit an can be ignored.

mohitreddy1996: Thanks for catching this alexey!

I think in the initial comments when I was using tensors for inputs (instead of ndarray or devicearray), I needed to extract the numpy ndarrays, that doesn't seem to be necessary now. Resolved in other occurrences as well!

PR #259, opened 2020-05-13
Make LayerNorm dtype behavior consistent with other layers

jheek:


jheek:
@levskaya have you tried this with the mlperf models yet?

levskaya:
Hi, sorry for the delay - I just tried this on the main wmt example and noticed significant training performance degradation in bfloat16 mode using this change... so I'm slightly hesitant to adopt this change as-is without understanding why it's so much worse... is there an application or user that needed the new implementation?

jheek:
Closing because of unintended performance regression

PR #267, opened 2020-05-16
Learning rate schedule howto

### LINKED ISSUE #222: Learning rate schedule HOWTO
Is there interest in a PR showing how to use learning rate schedules?

joaogui1:
Closes #222 
I used a triangular learning schedule, as it's a simple schedule


danielsuo:
@joaogui1 FYI you'll now have to pass `pytest examples/mnist` when your diff is applied before the build will succeed. This means that the `howto` diff will also need to modify the mnist test.

joaogui1:
Huh, didn't know that, thanks @danielsuo !
It's pretty late in Brazil, so tomorrow I will be pushing a fix to this

danielsuo:
Yeah, new change introduced in #248. Let me know if you need any help! Easy final check is to just run `./howtos/scripts/check_howto_diffs.sh`

joaogui1:
Pinging @danielsuo  and @avital

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,89 @@
...
+   return model
+ 
++def create_triangular_schedule(lr_min, lr_max, epochs):
++  """return a triangular learning rate schedule"""
++  top = epochs // 2
++  def learning_rate_fn(epoch):
++    if epoch < top:
++      lr = lr_min + epoch/top * (lr_max - lr_min)
++    else:
++      lr = lr_max - (epoch/top - 1)*(lr_max - lr_min)
```
avital, line 16: 1. Please be consistent in your whitespace. Add spaces around `*`.
2. The `- 1` in the arithmetic confuses me (though it's correct). I would write:

`lr = lr_max - ((epoch - top) / top) * (lr_max - lr_min)`

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,89 @@
+diff --git a/examples/mnist/train.py b/examples/mnist/train.py
+index 5dc74db..11a5e50 100644
+--- a/examples/mnist/train.py
++++ b/examples/mnist/train.py
+@@ -84,6 +84,16 @@ def create_model(key):
+   model = nn.Model(CNN, initial_params)
+   return model
+ 
++def create_triangular_schedule(lr_min, lr_max, epochs):
++  """return a triangular learning rate schedule"""
```
avital, line 10:
```suggestion
+  """Return fn from epoch to LR, linearly interpolating between `lr_min`->`lr_max`->`lr_min`."""
```

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,89 @@
...
++  top = epochs // 2
++  def learning_rate_fn(epoch):
++    if epoch < top:
++      lr = lr_min + epoch/top * (lr_max - lr_min)
++    else:
++      lr = lr_max - (epoch/top - 1)*(lr_max - lr_min)
++    return jnp.asarray(lr, dtype=jnp.float32)
```
avital, line 17: Why is this needed here?

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,89 @@
...
+@@ -130,18 +140,18 @@ def eval_step(model, batch):
+   return compute_metrics(logits, batch['label'])
+ 
+ 
+-def train_epoch(optimizer, train_ds, batch_size, epoch, rng):
++def train_epoch(optimizer, train_ds, batch_size, epoch, rng, learning_rate_fn):
+   """Train for a single epoch."""
+   train_ds_size = len(train_ds['image'])
+   steps_per_epoch = train_ds_size // batch_size
+-
```
avital, line 49: Why did you remove the whitespace here?

joaogui1: added it back after the calling the lr schedule

avital: I think the problem is that the whitespace you added below has a training space. If you make it actually an empty line then the diff will be cleaner.

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,90 @@
...
++  optimizer = optimizer.apply_gradient(grad, learning_rate=lr)
+   metrics = compute_metrics(logits, batch['label'])
+   return optimizer, metrics
+ 
+@@ -130,18 +140,19 @@ def eval_step(model, batch):
+   return compute_metrics(logits, batch['label'])
+ 
+ 
+-def train_epoch(optimizer, train_ds, batch_size, epoch, rng):
++def train_epoch(optimizer, train_ds, batch_size, epoch, rng, learning_rate_fn):
```
avital, line 45: Actually sorry for just now pointing this out, but let's pass `lr` (`learning_rate_fn(epoch)`) as an argument here instead of the learning rate function.

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,182 @@
...
+     optimizer, train_metrics = train_epoch(
+-        optimizer, train_ds, batch_size, epoch, input_rng)
++        optimizer, train_ds, batch_size, epoch, input_rng, lr)
+     loss, accuracy = eval_model(optimizer.target, test_ds)
+     logging.info('eval epoch: %d, loss: %.4f, accuracy: %.2f',
+                  epoch, loss, accuracy * 100)
+diff --git a/examples/mnist/train.py.rej b/examples/mnist/train.py.rej
+deleted file mode 100644
+index 89fff40..0000000
+--- a/examples/mnist/train.py.rej
```
danielsuo, line 77: We should make sure this diff doesn't have any reference to `train.py.rej`

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,182 @@
...
+         optimizer=optimizer,
+-        batch={k: v[:batch_size] for k, v in train_ds.items()})
++        batch={k: v[:batch_size] for k, v in train_ds.items()}, lr=0.1)
+     self.assertLessEqual(train_metrics['loss'], 2.302)
+     self.assertGreaterEqual(train_metrics['accuracy'], 0.0625)
+ 
+diff --git a/examples/mnist/train_test.py.rej b/examples/mnist/train_test.py.rej
+deleted file mode 100644
+index 50c2e9a..0000000
+--- a/examples/mnist/train_test.py.rej
```
danielsuo, line 170: Same here re `*.rej` files.

danielsuo:
Looks like the test is failing for reasons outside of this PR; hopefully #298 resolves the issue.

howtos/diffs/lr-schedule.diff
joaogui1: I think I fixed this

howtos/diffs/lr-schedule.diff
joaogui1: Same

avital:
Hi @joaogui1, this change LGTM but the diff doesn't apply -- see the build failure: "ERROR: Cannot apply howto lr-schedule! ==> PLEASE FIX HOWTO".

joaogui1:
Hey @avital do you know anything else about the error? Just ran it locally and it worked normally :/

danielsuo:
I can confirm @joaogui1. Re-running the jobs to see if that fixes.

danielsuo:
Ah, @joaogui1, you need to rebase onto the latest version of master. `mnist` has changed a little.

joaogui1:
It seems things are working now @avital @danielsuo

danielsuo:
This PR might not quite do what was intended; you pass around a function , but never actually call it with the `epoch` argument. It passes the test because you pass a float to `train_step` instead of the function produced by `create_triangular_schedule`. If you run `train.py`, this failure shows up with

```bash
TypeError: Argument '<function create_triangular_schedule.<locals>.learning_rate_fn at 0x7fd970444c20>' of type <class 'function'> is not a valid JAX type
```

Left a note above that might resolve the issue.

danielsuo:
Separately, testing a single `train_step` might not actually be sufficient. @avital caught me making a similar mistake in another PR.

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,87 @@
...
+   # compute mean of metrics across each batch in epoch.
+@@ -186,11 +198,13 @@ def train(train_ds, test_ds):
+   rng, init_rng = random.split(rng)
+   model = create_model(init_rng)
+   optimizer = create_optimizer(model, FLAGS.learning_rate, FLAGS.momentum)
++  learning_rate_fn = create_triangular_schedule(FLAGS.learning_rate, 2.*FLAGS.learning_rate, num_epochs)
+ 
+   for epoch in range(1, num_epochs + 1):
+     rng, input_rng = random.split(rng)
++    lr = learning_rate_fn
```
danielsuo, line 68: Perhaps you meant the following?

```python
lr = learning_rate_fn(epoch)
```

danielsuo, line 68: Perhaps you meant the following?

```python
lr = learning_rate_fn(epoch)
```

joaogui1: yes, I think after the rebase I messed up copying the diff to rerun the script, thanks!

joaogui1:
> Separately, testing a single `train_step` might not actually be sufficient. @avital caught me making a similar mistake in another PR.

Right now train_test is a single step, I'm not sure it's a good idea to change it in this PR (sorry if I misunderstood what you meant)

danielsuo:
> > Separately, testing a single `train_step` might not actually be sufficient. @avital caught me making a similar mistake in another PR.
> 
> Right now train_test is a single step, I'm not sure it's a good idea to change it in this PR (sorry if I misunderstood what you meant)

Oh oops, this comment was not at all clear. I was making a more general comment about our move towards testing `train_step` while the `howto`s may change more than just `train_step`. I don't think anything for you to do now, but something for `flax` core folks to consider (i.e., `howto` testing should ideally test every diff if possible).

howtos/diffs/lr-schedule.diff
```diff
@@ -0,0 +1,87 @@
...
+ 
+ 
++def create_triangular_schedule(lr_min, lr_max, epochs):
++  """Return fn from epoch to LR, linearly interpolating between `lr_min`->`lr_max`->`lr_min`."""
++  top = epochs // 2
++  def learning_rate_fn(epoch):
++    if epoch < top:
++      lr = lr_min + epoch/top * (lr_max - lr_min)
++    else:
++      lr = lr_max - ((epoch - top)/top) * (lr_max - lr_min)
```
danielsuo, line 16: Minor nit: LR schedule could wind up negative here. Wasn't sure if that was intended behavior. Marking as a "changes requested", but mostly just want to make sure there's a brief discussion

joaogui1: Good point, I'm against that. So if I round up top, we get `epoch <= 2*top`, which leads to the minimum value of `lr` being `lr_min` and max being `lr_max`

danielsuo:
Looks good to me! @avital

joaogui1:
Hey @avital think it's good for merging?

joaogui1:
Uh, is there something left to do here @avital @danielsuo ?

avital:
Hi Joao, sorry -- we're refactoring the MNIST base example (#289) so we're waiting for that before merging any additional HOWTOs

PR #270, opened 2020-05-18
Only install dataclasses dep in python 3.6

### LINKED ISSUE #269: `typing._ClassVar` cannot be accessed in the iPython shell – dataclasses package seems to mess up on Python 3.7

I just installed flax and tried to import it from the iPython shell. But it raises an `AttributeError`.

```
In [1]: import flax

*snip*

~/.virtualenvs/flax2/lib/python3.7/site-packages/dataclasses.py in _is_classvar(a_type, typing)
    548     # This test uses a typing internal class, but it's the best way to
    549     # test if this is a ClassVar.
--> 550     return type(a_type) is typing._ClassVar
    551
    552

AttributeError: module 'typing' has no attribute '_ClassVar'
```
This does not happen in the normal interpreter, where everything goes fine. 

I expected the import to work the same in iPython and the normal python shell.

Full traceback in this gist: https://gist.github.com/bayerj/96f096c7fb09a7c9b758dabdbca32671

### Steps to reproduce:

On Mac OS X with Python 3.7.6, not anaconda, virtuelenvwrapper installed.

```
❯❯❯ mkvirtualenv flax2
❯❯❯ pip install jaxlib
*snip*
❯❯❯ pip install flax
*snip*
❯❯❯ ipython

*snip*

In [1]: import flax
```

### Workaround

The problem seems to be in the `dataclasses` package–not python's own one–from PyPI. If I uninstall it...

```
❯❯❯ pip uninstall dataclasses
Found existing installation: dataclasses 0.6
Uninstalling dataclasses-0.6:
  Would remove:
    /Users/bayerj/.virtualenvs/debug2/lib/python3.7/site-packages/dataclasses-0.6.dist-info/*
    /Users/bayerj/.virtualenvs/debug2/lib/python3.7/site-packages/dataclasses.py
Proceed (y/n)? y
  Successfully uninstalled dataclasses-0.6
❯❯❯ ipython
/usr/local/lib/python3.7/site-packages/IPython/core/interactiveshell.py:931: UserWarning: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
  warn("Attempting to work in a virtualenv. If you encounter problems, please "
Python 3.7.6 (default, Dec 30 2019, 19:38:28)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.9.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import flax
```
... this goes fine.

levskaya:
Use the [PEP508](https://www.python.org/dev/peps/pep-0508/) `python_version` environment marker in our setup.py file to correctly install the pypi `dataclasses` dependency _only_ for python3.6 and not for python3.7 and above where the builtin dataclasses library must be used.

Fixes #269

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,7 +30,7 @@ install_requires = [
     "numpy>=1.12",
     "jax>=0.1.59",
     "matplotlib",  # only needed for tensorboard export
-    "dataclasses",  # will only install on py3.6
+    "dataclasses;python_version<'3.7'", # will only install on py3.6
     "msgpack",
 ]
 


levskaya:
Note: external-only change so GH merge should be ok, shouldn't need to be dealt w. in internal copybara.

PR #278, opened 2020-05-21
Added checkpointing howto

rolandgvc:
This HOWTO showcases loading and saving models from checkpoints.

### DIFF
--- a/docs/howtos.rst
+++ b/docs/howtos.rst
@@ -45,6 +45,7 @@ Polyak averaging
 .. raw:: html
    :file: _formatted_howtos/polyak-averaging.diff.html
 
+
 Scheduled Sampling
 ----------------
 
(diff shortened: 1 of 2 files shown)


rolandgvc:
Should this be more "interesting" as in the imagenet example?

avital:
I think it's perfect that this is "uninteresting". Does this work though? I don't think you've imported `checkpoints`

avital:
Awesome, look good! Can you add a link from docs/howtos.rst? You can find more details on how to test docs locally in docs/README.md

rolandgvc:
After running `make html` and building the files in `_build/html` I don't see any changes in the `howtos.rst`. What should I be looking for?

avital:
You should change howto.rst to point to your new howto

avital:
Thanks!

PR #279, opened 2020-05-22
Accessing intermediate layer states howto

rolandgvc:
Let's discuss how we could allow layers to preserve inputs/outputs in states. 

This HOWTO exemplifies how we can extract the shapes of the inputs and outputs of the first conv layer in the MNIST example, logging:
/Conv_0
inputs: (1, 28, 28, 1)
outputs: (1, 28, 28, 32)

Another option could be extracting the layer in/out from the compiled graph, but it seems a bit less practical.


howtos/diffs/intermediate-layer-state.diff
```diff
@@ -0,0 +1,32 @@
+diff --git a/examples/mnist/train.py b/examples/mnist/train.py
+index 5dc74db..7affff9 100644
+--- a/examples/mnist/train.py
++++ b/examples/mnist/train.py
+@@ -65,7 +65,7 @@ class CNN(nn.Module):
+   """A simple CNN model."""
+ 
+   def apply(self, x):
+-    x = nn.Conv(x, features=32, kernel_size=(3, 3))
```
avital, line 9: Can we get an API that's like this:
```suggestion
    x = capture_shapes(nn.Conv)(x, features=32, kernel_size=(3, 3))
```

or like this:
```suggestion
    x = capture_shapes(nn.Conv(x, features=32, kernel_size=(3, 3)))
```

(the second seems cleaner as it doesn't require people to think about functions that take modules and return modules; but maybe you lose name information or something else that's helpful)

avital:
For context, I don't think we want to add this functionality directly to all of our layers. So I'd like to first see what a solution that doesn't change our existing layers at all looks like.

If we can get a general purpose wrapper around modules that keeps track of these shapes, we could perhaps use it in core. But then we'd have to think carefully about the fact that modules can have multiple methods on them, each returning different shapes. That's why to first order I'd prefer we first think about how to do this specifically for an existing model codebase without changing any layers.

Which I guess brings up the point that MNIST is too simple to try this on. It'd be good to see what this solution looks like in one of the more complex models. ImageNet or LM1b.

rolandgvc:
> For context, I don't think we want to add this functionality directly to all of our layers. So I'd like to first see what a solution that doesn't change our existing layers at all looks like.
> 
> If we can get a general purpose wrapper around modules that keeps track of these shapes, we could perhaps use it in core. But then we'd have to think carefully about the fact that modules can have multiple methods on them, each returning different shapes. That's why to first order I'd prefer we first think about how to do this specifically for an existing model codebase without changing any layers.
> 
> Which I guess brings up the point that MNIST is too simple to try this on. It'd be good to see what this solution looks like in one of the more complex models. ImageNet or LM1b.

Makes sense. Working on it 👍

jheek:
Can we not use `nn.capture_module_outputs()` for this?

avital:
@levskaya has advocated for us to remove that as it introduces additional complexity into core, e.g. transparent modules. I'll let him chine in for his opinion. In this case I wanted to explore what it would look like to get intermediate outputs without any special functionality from core.

PR #281, opened 2020-05-24
Add --model_dir flag as the training code requires this field to be set.

mohitreddy1996:
Set model_dir flag in README as the training code expects the flag to be set. Tensorboard summary writer otherwise fails with  

"TypeError: Expected binary or unicode string, got None"


avital:
Can we also make `model_dir` no longer have a default, or force it to have a value with `flags.flags.mark_flag_as_required('filename')`? The error message you got if you didn't set the `model_dir` flag isn't good.

PR #282, opened 2020-05-24
Refactor imagenet example.

mohitreddy1996:
- Remove no-op assignment of base learning rate in training code.
- In input pipeline, refactor to have repeat the tf dataset for both train and eval. 
  - Currently repeating of tf dataset is done for both, but is implemented as part of two different conditions.

### DIFF
--- a/examples/imagenet/input_pipeline.py
+++ b/examples/imagenet/input_pipeline.py
@@ -214,15 +214,13 @@ def load_split(batch_size, train, dtype=tf.float32, image_size=IMAGE_SIZE, cache
   if cache:
     ds = ds.cache()
 
+  ds = ds.repeat()
+
   if train:
-    ds = ds.repeat()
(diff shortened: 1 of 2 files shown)


avital:
The PR LGTM but I'd like us to manually verify that the results haven't changed. Can you start a run before the change and after (either on Cloud or on Google infrastructure) and share links to Tensorboard.dev?

(Hopefully soon we'll have full automated end-to-end testing so I won't have to ask to do this manually :D )

avital:
Or I guess, just curious if you're run this change and verified that it doesn't change behavior. You don't need to link to Tensorboard.dev if you're tested it.

marcvanzee:
Assigned this to Mohit, since Avital asked him to verify that this change doesn't change behavior.

mohitreddy1996:
Done as part of #287

PR #285, opened 2020-05-26
Fixing LICENSE info in setup.py

kshithijiyer:


setup.py
```diff
@@ -53,15 +53,15 @@
         "Development Status :: 3 - Alpha",
         "Intended Audience :: Developers",
         "Intended Audience :: Science/Research",
-        "License :: OSI Approved :: MIT License",
+        "License :: OSI Approved :: Apache v2.0",
```
danielsuo, line 56: This line should be `License :: OSI Approved :: Apache Software License`. Unfortunately, there is no v2 (and [these](https://github.com/pypa/pypi-legacy/issues/91) [issues](https://github.com/pypa/trove-classifiers/issues/17) note that this needs to be clarified). Small point, but want to make sure things are indexed correctly in `pypi`.

setup.py
```diff
@@ -53,15 +53,15 @@
...
         "Programming Language :: Python :: 3.7",
         "Topic :: Scientific/Engineering :: Artificial Intelligence",
         ],
     keywords="",
     author="Flax team",
     author_email="flax-dev@google.com",
     url="https://github.com/google/flax",
-    license="Apache",
+    license="Apache v2.0",
```
danielsuo, line 64: We can delete this line per [python docs](https://docs.python.org/3.7/distutils/setupscript.html#installing-additional-files).

> The `license` field is a text indicating the license covering the package where the license is not a selection from the “License” Trove classifiers. See the `Classifier` field. Notice that there’s a `licence` distribution option which is deprecated but still acts as an alias for license.

danielsuo:
Good catch, thanks! See minor comments.

avital:
Thanks!

PR #286, opened 2020-05-26
Refactor mnist

mohitreddy1996:


danielsuo:
@mohitreddy1996 let me know when this has settled and I can fix up the how-tos so the PR passes all checks

mohitreddy1996:
Closing this in favor a different PR. 

We want to keep the MNIST example as simple as possible. Therefore I will revert few changes to single host training.

PR #287, opened 2020-05-26
Refactor ImageNet example aligning with Issue #231

mohitreddy1996:
Refactor Imagenet example aligning with #231 

Major changes:
- Refactor files to example_lib (containing training and eval loop code), example_main (script which parses flags and starts the training loop), model_name (containing necessary model definition), input_pipeline (bucketing the entire preprocessing code).

- Introduce a test using tfds.testing.mock_data which executes a single step of training and eval.
  - In order to do so, copy the metadata files from TFDS and store in examples/testing/datasets.
  - Document necessary changes required to be made by other examples.


examples/imagenet/input_pipeline.py
```diff
@@ -57,13 +56,21 @@ def distorted_bounding_box_crop(image_bytes,
     max_attempts: An optional `int`. Number of attempts at generating a cropped
         region of the image of the specified constraints. After `max_attempts`
         failures, return the entire image.
+    rng: RNG used for generating seeds for reproducibility of
```
avital, line 59: Instead of this, add a comment that we need to resolve reproducibility issues, linking to a GitHub issue with details

mohitreddy1996: Created the issue and mentioned it in the README file.

examples/imagenet/imagenet_benchmark.py
```diff
@@ -41,14 +32,12 @@ class ImagenetBenchmark(Benchmark):
...
     model_dir = tempfile.mkdtemp()
-    FLAGS.batch_size = 2048
-    FLAGS.half_precision = True
-    FLAGS.loss_scaling = 256.
-    FLAGS.num_epochs = 2
-    FLAGS.model_dir = model_dir
 
     start_time = time.time()
-    train.main([])
+    imagenet_lib.train_and_evaluate(model_dir=model_dir, batch_size=2048,
```
avital, line 37: I originally suggested to @AlexeyG to do the benchmarks this way but he persuaded me that it would be good to have the benchmarks exactly line up with the arguments you'd pass in the command line to the main train python file.

What do you think?

examples/imagenet/imagenet_lib.py
```diff
@@ -44,56 +31,37 @@
...
-
-flags.DEFINE_bool(
-    'half_precision', default=False,
-    help=('If bfloat16/float16 should be used instead of float32.'))
+import input_pipeline
+from resnet_v1 import ResNet
 
 
-def create_model(key, batch_size, image_size, model_dtype):
+def _create_model(key, batch_size, image_size, model_dtype):
```
avital, line 48: Let's keep these public methods.

examples/imagenet/imagenet_lib.py
```diff
@@ -191,58 +159,78 @@ def _prepare(x):
   return jax.tree_map(_prepare, xs)
 
 
-def create_input_iter(batch_size, image_size, dtype, train, cache):
-  ds = input_pipeline.load_split(
-      batch_size, image_size=image_size, dtype=dtype, train=train, cache=cache)
-  it = map(prepare_tf_data, ds)
+def _create_input_iter(dataset_builder, batch_size, image_size, dtype, train,
+                       cache):
+  ds = input_pipeline.create_split(
```
avital, line 164: Can you tell me more about this TFDS API change? What are "dataset builders"?

examples/imagenet/imagenet_lib.py
```diff
@@ -191,58 +159,78 @@ def _prepare(x):
...
+  it = map(_prepare_tf_data, ds)
   it = jax_utils.prefetch_to_device(it, 2)
   return it
 
 
 # flax.struct.dataclass enables instances of this class to be passed into jax
 # transformations like tree_map and pmap.
 @flax.struct.dataclass
-class TrainState:
+class _TrainState:
```
avital, line 175: ditto on this and all the others, we don't need to add underscores to any of these

examples/imagenet/imagenet_lib.py
```diff
@@ -191,58 +159,78 @@ def _prepare(x):
...
+                       num_train_and_eval_steps: int = -1,
+                       disable_checkpointing: bool = False):
+  """Runs model training and evaluation loop.
+
+  Args:
+    model_dir: Directory where the checkpoints and tensorboard summaries
+      should be written to.
+    batch_size: Batch size of the input.
+    num_epochs: Number of epochs to cycle through before stopping.
+    learning_rate: Learning rate for the momentum optimizer.
```
avital, line 212: This isn't accurate. This argument is "the learning rate in case you have batch size 256. The effective learning rate is scaled linearly to the batch size"

examples/imagenet/imagenet_lib.py
```diff
@@ -191,58 +159,78 @@ def _prepare(x):
...
+    momentum: Momentum value for the momentum optimizer.
+    cache: Determines whether the dataset should be cached.
+    half_precision: Determines whether bfloat16/float16 should be used
+      instead of float32.
+    num_train_and_eval_steps: Number of steps for training and eval.
```
avital, line 217: Hmm... Do we need this? It's a bit off to see an argument designed just for a test. Are there any other ways to achieve this?

examples/imagenet/imagenet_lib.py
```diff
@@ -191,58 +159,78 @@ def _prepare(x):
...
+      This is used for testing (default: -1 i.e use the entire dataset).
+    disable_checkpointing: Determines whether the training and evaluation
```
avital, line 219: I'm also not sure about this. Our official examples need to be as short as possible. What's wrong with always saving checkpoints? We should also test that...

examples/imagenet/imagenet_lib_test.py
```diff
@@ -0,0 +1,84 @@
...
+# utility class for parsing event files and extracting scalar summaries. 
+def _process_event(event):
+  for value in event.summary.value:
+    yield value
+
+
+def _parse_and_return_summary_values(path):
+  """Parses event file in given `path` and returns scalar summaries logged."""
+  tag_event_value_dict = {}
+  event_file_generator = directory_watcher.DirectoryWatcher(
```
avital, line 42: Why do we need this? Are we really watching the directory? Isn't there a more straightforward API to just read the latest observations? (What did Alexey do in his original Benchmark PR?)

examples/imagenet/imagenet_lib_test.py
```diff
@@ -0,0 +1,84 @@
...
+      path, event_file_loader.EventFileLoader).Load()
+  event_values = itertools.chain.from_iterable(
+      map(_process_event, event_file_generator))
+  for value in event_values:
+    tag_event_value_dict[value.tag] = tensor_util.make_ndarray(value.tensor)
+
+  return tag_event_value_dict
+
+
+class ImageNetLibTest(absltest.TestCase):
```
avital, line 52: Just call it `ImageNetTest`

examples/imagenet/imagenet_lib_test.py
```diff
@@ -0,0 +1,84 @@
...
+    data_dir = parent_dir + '/testing/datasets'
+
+    # Create a temporary directory where tensorboard metrics are written.
+    model_dir = tempfile.mkdtemp()
+
+    with tfds.testing.mock_data(num_examples=8, data_dir=data_dir):
+      imagenet_lib.train_and_evaluate(
+          model_dir=model_dir, batch_size=8, num_epochs=1,
+          learning_rate=0.1, momentum=0.9, cache=False, half_precision=False,
+          num_train_and_eval_steps=1, disable_checkpointing=True)
```
avital, line 71: I see that this is the place that you need `num_train_and_eval_steps=1`. Instead, we should consider refactoring `train_and_evaluate` into smaller functions and then just call into those directly instead of having this override. That would also clarify the different parts of the ImageNet train code.

examples/imagenet/imagenet_lib.py
avital: (I wrote another comment on this below where you call this function from your test)

examples/imagenet/imagenet_lib_test.py
```diff
@@ -0,0 +1,84 @@
...
+
+    summary_values_dict = _parse_and_return_summary_values(path=model_dir)
+
+    # Since the values could change due to stochasticity in input processing
```
avital, line 75: If we're not actually looking at the numbers at all there's no point in these assertions. I don't mind saying that this only tests that nothing breaks after one step of training, and the other full end-to-end tests test the meaningful part -- that training gets to the expected numbers.

examples/imagenet/input_pipeline.py
```diff
@@ -21,22 +21,18 @@
...
 STDDEV_RGB = [0.229 * 255, 0.224 * 255, 0.225 * 255]
 
 
-def distorted_bounding_box_crop(image_bytes,
-                                bbox,
-                                min_object_covered=0.1,
-                                aspect_ratio_range=(0.75, 1.33),
-                                area_range=(0.05, 1.0),
-                                max_attempts=100):
+def _distorted_bounding_box_crop(image_bytes,
```
avital, line 30: No underscores before function names

examples/imagenet/input_pipeline.py
```diff
@@ -126,21 +122,21 @@ def _decode_and_center_crop(image_bytes, image_size):
 
   offset_height = ((image_height - padded_center_crop_size) + 1) // 2
   offset_width = ((image_width - padded_center_crop_size) + 1) // 2
-  crop_window = tf.stack([offset_height, offset_width,
-                          padded_center_crop_size, padded_center_crop_size])
-  image = tf.image.decode_and_crop_jpeg(image_bytes, crop_window, channels=3)
+  crop_window = tf.stack([offset_height, offset_width, padded_center_crop_size,
```
avital, line 125: nit but please don't change whitespace unless there's a reason -- it makes merge conflicts unnecessarily likely

examples/imagenet/input_pipeline.py
```diff
@@ -172,42 +168,50 @@ def preprocess_for_eval(image_bytes, dtype=tf.float32, image_size=IMAGE_SIZE):
...
 
 
-def load_split(batch_size, train, dtype=tf.float32, image_size=IMAGE_SIZE, cache=False):
+def create_split(dataset_builder: tfds.core.DatasetBuilder, batch_size: int,
+                 train: bool, dtype: tf.DType = tf.float32,
+                 image_size: int = IMAGE_SIZE, cache: bool = False):
   """Creates a split from the ImageNet dataset using TensorFlow Datasets.
 
   Args:
+    dataset_builder: TFDS dataset builder for ImageNet.
```
avital, line 182: Can't the dataset builder just be constructed here rather than at the top level of the training loop?

examples/imagenet/input_pipeline.py
```diff
@@ -216,9 +220,11 @@ def decode_example(example):
 
   if train:
     ds = ds.repeat()
-    ds = ds.shuffle(16 * batch_size, seed=0)
+    ds = ds.shuffle(16 * batch_size)
```
avital, line 223: Why get rid of the seed here?

examples/imagenet/resnet_v1.py
```diff
@@ -13,18 +13,17 @@
...
 
+import jax.numpy as jnp
 
 from flax import nn
 
-import jax.numpy as jnp
-
 
-class ResidualBlock(nn.Module):
+class _ResidualBlock(nn.Module):
```
avital, line 23: No underscores

examples/imagenet/resnet_v1.py
```diff
@@ -13,18 +13,17 @@
...
   """Bottleneck ResNet block."""
 
+  # pylint: disable=arguments-differ
```
avital, line 26: Oh? What's this?

examples/imagenet/resnet_v1.py
```diff
@@ -51,10 +50,25 @@ def apply(self, x, filters, strides=(1, 1), train=True, dtype=jnp.float32):
 
 
 class ResNet(nn.Module):
-  """ResNetV1."""
+  """ResNet V1."""
+
+  # pylint: disable=arguments-differ
```
avital, line 55: Oh?

examples/testing/datasets/imagenet2012/5.0.0/image.image.json
```diff
@@ -0,0 +1 @@
+{"encoding_format": "jpeg", "shape": [-1, -1, 3]}
```
avital, line 1: Have you tried to see if we can avoid this by loading these files directly via

```
import tfds
tfds.__path__
```

examples/testing/datasets/README.md
```diff
@@ -0,0 +1,34 @@
...
+parent_dir = os.path.dirname(current_dir)
+data_dir = parent_dir + '/testing/datasets'
+
+with tfds.testing.mock_data(num_examples=num_examples, data_dir=data_dir):
+  # code executing tfds.load() or equivalent.
+```
+
+Example test: `flax/examples/imagenet_lib_test.py` 
+
+PR introducing the above mentioned changes: [PR](https://github.com/google/flax/pull/287).
```
avital, line 34: I don't think this should be here. (And in general people can always `git blame`)

examples/imagenet/imagenet_lib.py
mohitreddy1996: Done. 

Since we wanted to have type annotation for public methods with appropriate docstrings, I felt that the ones which are local/private, we could stick with the standard.

examples/imagenet/imagenet_lib.py
mohitreddy1996: done.

examples/imagenet/imagenet_lib_test.py
mohitreddy1996: done.

examples/imagenet/input_pipeline.py
mohitreddy1996: done.

examples/imagenet/resnet_v1.py
mohitreddy1996: done.

examples/testing/datasets/README.md
mohitreddy1996: Yup that makes sense.

examples/testing/datasets/imagenet2012/5.0.0/image.image.json
mohitreddy1996: Yup, I gave that a shot. testing/metadata not exposed as part of the library in the python packages sadly. 

I spoke with the TFDS team, they recommended to copy the metadata files for now but a better solution is WIP (ETA: later this year).

examples/imagenet/resnet_v1.py
mohitreddy1996: aah my bad, I have a linter in VS code and it was complaining about it. I guess I will remove it.

examples/imagenet/input_pipeline.py
mohitreddy1996: sorry my bad. I was trying to introduce the seeds using jax.random.split(seed) for reproducibility. Removed it assuming I introduced it. 

Thanks for catching it.

examples/imagenet/resnet_v1.py
mohitreddy1996: Same as below. Removed.

examples/imagenet/input_pipeline.py
mohitreddy1996: reverted the changes.

examples/imagenet/imagenet_lib.py
mohitreddy1996: oh no... Thanks for catching (will make the necessary changes in #289 as well)

examples/imagenet/imagenet_lib.py
mohitreddy1996: The reason for this was majorly because to get rid of previously globally declared vars:
TRAIN_IMAGES = 1281167
EVAL_IMAGES = 50000

To have an extendable solution, we shouldn't have hardcoded values of something which is controlled by a different library. In order to fetch this information using the TDFS library itself, I modified the existing code to use DatasetBuilder (an abstract base class for every dataset). DatasetBuilder contains information around dataset such as: feature names, types, shapes, version, splits etc.

Also the dataset_builder is initialized in the main train loop to avoid multiple initializations which increases the chances of inconsistent changes.

examples/imagenet/imagenet_lib.py
mohitreddy1996: So I had seen multiple instances where checkpointing took quite a bit of wall time in test execution during development (although I ran the test multiple times now, no such issues :/). So I decided to disable it, given how `checkpoints` is a different library and should have testing of its own. 

But I guess we still need to ensure that after the last (and only) step, checkpoints are saved in tests. Removed this flag.

examples/imagenet/imagenet_lib_test.py
mohitreddy1996: This is required to test the recorded metrics as an assertion that the model training indeed happened.
So this code is a motivation from that PR. This is mostly copy pasted from there and I have #290 to address this duplication.

examples/imagenet/input_pipeline.py
mohitreddy1996: Discussed in the previous comment. Basically to avoid multiple initializations of the dataset_builder. 

The values of number of train and eval images are required in main train loop to compute the steps per epoch and here as well to compute split. We definitely don't want to have hardcoded numbers. So initialized it in the main train loop and passing it on as argument.

examples/imagenet/imagenet_lib_test.py
mohitreddy1996: Hmm, that's true. Without reproducibility of the example, having these assertions are meaningless :/ 

Although we could just put values like loss greater than 6.0 or something. I believe the accuracy would stay at 0.0 since the input used is always the same and the remaining randomness (is in the input preprocessing logic with stateful TF Image ops). 

WDYT? I am TBH inclining towards removing it. Just a thought.

examples/imagenet/imagenet_lib.py
mohitreddy1996: Although specifying the number of examples for train and test/eval in tfds.testing.mockdata, the base info such as num_examples in a split (train/eval) is same as the original dataset. It is documented as WARNING in [documentation](https://www.tensorflow.org/datasets/api_docs/python/tfds/testing/mock_data).

Without this flag, we would do a lot more than a single iteration :/ I couldn't think of something better.. 

We can discuss further in the below comment if refactoring is the way to go.

examples/imagenet/imagenet_benchmark.py
mohitreddy1996: Most of the cases, just set a few flag. 
Worst case, same as passing them to the example_lib.method().

I think I like calling the main idea better, Thanks for noticing this!

avital:
@jheek As the "owner" of the ImageNet example, I'd like to get your review on this as well.

examples/imagenet/imagenet_lib_test.py
marcvanzee: Remove them makes sense to me.

examples/imagenet/imagenet_lib.py
marcvanzee: The WARNING in the documentation says that `info.split['train'].num_examples` doesn't match `len(list(ds_train))`. So how about you use `len(list(ds_train))` below instead of `.num_examples` to determine the size of the dataset? In that case I think you won't need this argument.

examples/imagenet/imagenet_lib.py
```diff
@@ -123,7 +91,7 @@ def step_fn(step):
 
 def train_step(state, batch, learning_rate_fn):
   """Perform a single training step."""
-  def loss_fn(model):
+  def _loss_fn(model):
```
marcvanzee, line 94: I think up to now we haven't used underscores for functions in our codebase, so I would avoid it here as well for consistency.

avital: Or more generally, a pull request should only change things that pertain to that particular pull request.

If we think we should change our naming convention, we can discuss that but we shouldn't combine that with a pull request refactoring an example for sake of testing.

examples/imagenet/README.md
```diff
@@ -18,7 +18,10 @@ The model should run with other configurations and hardware, but explicitely tes
...
 #### 8 x Nvidia V100 (16GB)
-`python train.py --batch_size=512 --cache=True --model_dir=./imagenet_fp32_bs512`
+`python imagenet_main.py --batch_size=512 --cache=True --model_dir=./imagenet_fp32_bs512`
 
 #### 8 x Nvidia V100 (16GB), mixed precision
-`python train.py --batch_size=2048 --cache=True --model_dir=./imagenet_fp16_bs2048 --half_precision=True --loss_scaling=256.`
+`python imagenet_main.py --batch_size=2048 --cache=True --model_dir=./imagenet_fp16_bs2048 --half_precision=True --loss_scaling=256.`
+
+### Reproducibility
+Making the ImageNet classification example reproducible in WIP. For more details, follow [#291](https://github.com/google/flax/issues/291).
```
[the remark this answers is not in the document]
mohitreddy1996: Thanks!

examples/imagenet/imagenet_lib.py
mohitreddy1996: I used this convention as this file had a similar naming convention (def _prepare(x)) in `prepare_tf_data`. Changed it back and also changed _prepare -> prepare. 

@avital yeah, I think that makes sense. Since this was a huge refactoring, I just put them the smaller ones in as well, probably wasn't a good idea :)

examples/imagenet/imagenet_lib_test.py
mohitreddy1996: Removed the assertions altogether.

examples/imagenet/imagenet_lib_test.py
mohitreddy1996: Thanks! Removed it.

mohitreddy1996:
@marcvanzee 

> Re "The WARNING in the documentation says that info.split['train'].num_examples doesn't match len(list(ds_train)). So how about you use len(list(ds_train)) below instead of .num_examples to determine the size of the dataset? In that case I think you won't need this argument."

Please correct me if I am wrong, but `info.split['train'].num_examples` is basically reading a field which I believe is present in tfds metadata. But when we do len(list(ds_train)), do you think we might be doing additional transformations? The main reason for introducing a `test-only` field was primarily that. WDYT? @avital what are your thoughts on this?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "editor.tabSize": 2
```
Marvin182: Do you want this to be part of this CL?

examples/imagenet/imagenet_lib.py
```diff
@@ -44,42 +32,23 @@
 import jax.numpy as jnp
 
 import tensorflow.compat.v2 as tf
```
Marvin182, line 33: /tensorflow.compat.v2/tensorflow

examples/imagenet/imagenet_lib_test.py
```diff
@@ -0,0 +1,54 @@
...
+import tensorflow.compat.v2 as tf
+import tensorflow_datasets as tfds
+
+import flax.testing as testing
+
+import imagenet_lib
+
+
+class ImageNetLibTest(absltest.TestCase):
+
```
Marvin182, line 31: You might want to add:
```
  def setUp(self):
    super().setUp()
    tf.config.experimental.set_visible_devices([], "GPU")
```
This avoids some OOM error when the test case runs on GPUs.

examples/imagenet/imagenet_main.py
```diff
@@ -0,0 +1,73 @@
...
+"""ImageNet example.
+
+This script trains a ResNet-50 on the ImageNet dataset.
+The data is loaded using tensorflow_datasets.
+"""
+
+from absl import app
+from absl import flags
+
+import tensorflow.compat.v2 as tf
```
Marvin182: /tensorflow.compat.v2/tensorflow

examples/imagenet/input_pipeline_test.py
```diff
@@ -0,0 +1,54 @@
...
+import tensorflow_datasets as tfds
+
+import input_pipeline
+
+
+class InputPipelineTest(absltest.TestCase):
+
+  def test_train_and_evaluate(self):
+    # Get testing directory.
+    tf.enable_v2_behavior()
```
Marvin182: not needed. I think we don't support TF1 in Flax.

Marvin182 (approved):
Looks good. We could do more changes (like switching to ml_collections.ConfigDict for the hyperparameters), but let's do it in separate PRs and in coordination with Andreas.

.vscode/settings.json
mohitreddy1996: My bad, this was accidentally added as one of the commit. This should be an outdated file now.

examples/imagenet/imagenet_main.py
mohitreddy1996: Again outdated file :) 

Replaced it in all instances of imagenet example.

examples/imagenet/input_pipeline_test.py
mohitreddy1996: Done! Thanks for pointing this out.

examples/imagenet/imagenet_lib_test.py
mohitreddy1996: Great suggestion, thanks!

avital (approved):

PR #288, opened 2020-05-27
Param info util: dict and string summary

rolandgvc:
This is a proposal for two parameter summary utils, one that returns the information as a dict and one as a string.

### DIFF
--- a/flax/nn/utils.py
+++ b/flax/nn/utils.py
@@ -19,7 +19,7 @@ import contextlib
 import threading
 import jax
 import numpy as onp
-from collections import namedtuple
+from typing import Dict
 
 class CallStack(object):
   """Utility for tracking data across a call stack."""
@@ -98,7 +98,49 @@ def _level_of_value(xs):
   return max_level
 
 
-def model_summary(model):
+def get_param_count(params: Dict[str, onp.ndarray]) -> int:
+  """Returns the count of variables for the module or parameter dictionary."""
+  flatten_params = flatten_dict(params)
+  return onp.sum(param.size for param in flatten_params.values())
+
+
+def _get_name_idx(name: str):
+  """Returns the layer index of the parameter name."""
+  index = name[name.find('_') + 1 : name.find('/')]
+  return int(index) if index[0] in '0123456789' else -1
+
+
+def get_param_info(param: Dict[str, onp.ndarray]):
+  """Returns dictionary with parameter shapes, numbers, types and bytes."""
+
+  if not isinstance(param, dict):
+    raise ValueError('Please provide a dictionary of parameters.')
+
+  param = flatten_dict(param)
+  # Sort parameter names by the order of layers in the module 
+  param_names, param_values = map(list, tuple(zip(*sorted(
+    param.items(), key=lambda item: _get_name_idx(item[0])))))
+  
+  param_info = {}
+  total_number = total_bytes = 0
+  for idx in range(len(param_values)):
+    name = param_names[idx]
+    value = param_values[idx]
+    param_info[name] = {
+      "shape": onp.shape(value), 
+      "number": value.size,
+      "type": value.dtype, 
+      "bytes": value.dtype.itemsize * value.size
+    }
+    total_number += value.size
+    total_bytes += value.dtype.itemsize * value.size
+  param_info["totals"] = {"number": total_number, "bytes": total_bytes}
+  
+  return param_info
+
+
+
+def params_summary(model_params):
   """Returns a summary of the model's parameters.
 
   Args:
@@ -127,13 +169,13 @@ def model_summary(model):
 
   # Get parameter names
   param_names = []
-  for layer in model.params.keys():
-    for params in list(model.params[layer].keys()):
+  for layer in model_params.keys():
+    for params in list(model_params[layer].keys()):
       param_names.append("{}/{}".format(layer, params))
 
   # Get parameter shapes, numbers, types, sizes
   param_info = []
-  for params in jax.tree_flatten(model.params)[0]:
+  for params in jax.tree_flatten(model_params)[0]:
     param_info.append(Parameters(
       onp.shape(params),
       onp.prod(onp.shape(params)),
--- a/flax/nn/utils.py
+++ b/flax/nn/utils.py
@@ -152,10 +152,10 @@ def show_param_info(params: Dict[str, onp.ndarray], max_lines: int = None):
   ---------------------------------------------------------
   Parameters                     Shape     Number      Type
   =========================================================
-  BatchNorm_1/bias               (32,)         32    float32
-  BatchNorm_1/scale              (32,)         32    float32
   Conv_0/bias                    (32,)         32    float32
   Conv_0/kernel          (3, 3, 1, 32)        288    float32
+  BatchNorm_1/bias               (32,)         32    float32
+  BatchNorm_1/scale              (32,)         32    float32
   Conv_2/bias                    (64,)         64    float32
   Conv_2/kernel         (3, 3, 32, 64)      18432    float32
   Dense_3/bias                  (256,)        256    float32
@@ -189,8 +189,8 @@ def show_param_info(params: Dict[str, onp.ndarray], max_lines: int = None):
     _Column("Type", [str(v["type"]) for v in values]),
   ]
 
-  offset = 2 # between columns
-  name_format = f"{{: <{columns[0].width + offset}s}}" # align parameters on the left
+  offset = 2 # set distance between columns
+  name_format = f"{{: <{columns[0].width + offset}s}}" # align parameters to the left
   value_format = name_format + "".join(f"{{: >{c.width + offset}s}}" for c in columns[1:])
   header = value_format.format(*[c.name for c in columns])
 
--- a/flax/nn/utils.py
+++ b/flax/nn/utils.py
@@ -18,7 +18,8 @@
 import contextlib
 import threading
 import jax
-
+import numpy as onp
+from collections import namedtuple
 
 class CallStack(object):
   """Utility for tracking data across a call stack."""
@@ -97,7 +98,7 @@ def _level_of_value(xs):
   return max_level
 
 
-def model_summary(model: nn.Model):
+def model_summary(model):
   """Returns a summary of the model's parameters.
 
   Args:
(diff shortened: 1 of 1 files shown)


flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
       level = _trace_level(x._trace.master)
       max_level = max(level, max_level)
   return max_level
+
+
+def model_summary(model: nn.Model):
```
Marvin182, line 100: Can wee use something more generic than nn.Model? dict o list with the parameteres?

flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
...
+      param_names.append("{}/{}".format(layer, params))
+
+  # Get parameter shapes, numbers, types, sizes
+  param_info = []
+  for params in jax.tree_flatten(model.params)[0]:
+    param_info.append((
+      onp.shape(params),
+      onp.prod(onp.shape(params)),
+      params.dtype,
+      params.dtype.itemsize * params.size / 1000000
```
Marvin182, line 140: Should this be 2**20 and MiB below? A bit more precise.

flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
...
+  param_names = []
+  for layer in model.params.keys():
+    for params in list(model.params[layer].keys()):
+      param_names.append("{}/{}".format(layer, params))
+
+  # Get parameter shapes, numbers, types, sizes
+  param_info = []
+  for params in jax.tree_flatten(model.params)[0]:
+    param_info.append((
+      onp.shape(params),
```
Marvin182, line 137: Can we use a namedtuple here?

flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
...
+      onp.prod(onp.shape(params)),
```
Marvin182, line 138: Can we print the number of parameters with thousand delimiter?

flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
...
+      params.dtype,
+      params.dtype.itemsize * params.size / 1000000
+    ))
+
+  summary_str = "----------------------------------------------------------\n"
+  summary_str += "{:<20} {:>15} {:>10} {:>10}\n".format("Parameters", "Shape", "Number", "Type")
```
Marvin182, line 144: What happens when parameter names are longer than 20 characters?

flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
...
+  
+  # Get parameter names
+  param_names = []
+  for layer in model.params.keys():
+    for params in list(model.params[layer].keys()):
+      param_names.append("{}/{}".format(layer, params))
+
+  # Get parameter shapes, numbers, types, sizes
+  param_info = []
+  for params in jax.tree_flatten(model.params)[0]:
```
Marvin182, line 135: Design question: Do we want to sort parameters by name?

flax/nn/utils.py
```diff
@@ -95,3 +95,66 @@ def _level_of_value(xs):
...
+  # Totals
+  total_params = total_size = 0
+  for i in range(len(param_names)):
+    summary_str += "{:<20} {:>15} {:>10} {:>10}\n".format(param_names[i],
+      str(param_info[i][0]), str(param_info[i][1]), str(param_info[i][2]))
+    total_params += param_info[i][1]
+    total_size += param_info[i][3]
+
+  summary_str += "==========================================================\n"
+  summary_str += f"Total Parameters: {total_params}\n"
```
Marvin182, line 156: Replace {total_params} with {total_params:,} to add thousand delimiters.

mohitreddy1996:
This is awesome! 

I have a small request. Would it be possible to refactor out logic which return the number/count of parameters (something similar to @Marvin182 's implemented method 'count_parameters' in https://gist.github.com/Marvin182/4c87f14b01aa1bf481d312e36e32332e).

This would be helpful in testing model definition in examples. I currently have TODOs in #287 and #289

flax/nn/utils.py
rolandgvc: Correct, that would be cleaner.

flax/nn/utils.py
rolandgvc: Not necessarily. Is there a way of flattening the tree while keeping the layer order?

rolandgvc:
> This is awesome!
> 
> I have a small request. Would it be possible to refactor out logic which return the number/count of parameters (something similar to @Marvin182 's implemented method 'count_parameters' in https://gist.github.com/Marvin182/4c87f14b01aa1bf481d312e36e32332e).
> 
> This would be helpful in testing model definition in examples. I currently have TODOs in #287 and #289

Thanks! Avital wanted that as a HOWTO, which I have here #277. I'll consult with him :)

flax/nn/utils.py
rolandgvc: Would get deformed :( I'll look into making everything adaptable for the next commit 👍

flax/nn/utils.py
```diff
@@ -95,3 +96,66 @@ def _level_of_value(xs):
       level = _trace_level(x._trace.master)
       max_level = max(level, max_level)
   return max_level
+
+
+def model_summary(model):
```
avital, line 101: Better to take `params` as an argument rather than `model`.

flax/nn/utils.py
Marvin182: I think this could be done using an OrderedDict.

flax/nn/utils.py
```diff
@@ -95,3 +96,66 @@ def _level_of_value(xs):
...
+  Total Parameters: 824,522
+  Total Size: 3.1 MB
+  ----------------------------------------------------------
+  """
+  Parameters = namedtuple("Parameters", ["shape", "number", "type", "size"])
+
+  # Get parameter names
+  param_names = []
+  for layer in model.params.keys():
+    for params in list(model.params[layer].keys()):
```
avital, line 131: Please test this on the ResNet example, where have multiply-nested modules. There, the param dict might look like:

{
  "Block_0": {
    "Conv_0": {
      "weight": ...
    }
  }
}

flax/nn/utils.py
```diff
@@ -95,3 +96,66 @@ def _level_of_value(xs):
...
+      param_names.append("{}/{}".format(layer, params))
+
+  # Get parameter shapes, numbers, types, sizes
+  param_info = []
+  for params in jax.tree_flatten(model.params)[0]:
```
avital, line 136: If we ever have `[0]`, it should either instead call into a utility that is semantically meaningful, or if for some reason that's not possible at the very least a comment. 

In this case I think calling `tree_leaves()` resolves the issue.

flax/nn/utils.py
```diff
@@ -95,3 +96,66 @@ def _level_of_value(xs):
...
+    param_info.append(Parameters(
+      onp.shape(params),
+      onp.prod(onp.shape(params)),
+      params.dtype,
+      params.dtype.itemsize * params.size / 2**20
+    ))
+
```
avital, line 143: I propose we split this into two functions: One that just returns `param_info`, and another separate function that does the pretty-printing.

That also helps resolve Marvin's concern below, it would be easy for a user to fork the pretty printer and replace it with their own.

flax/nn/utils.py
```diff
@@ -95,3 +96,120 @@ def _level_of_value(xs):
...
+    if isinstance(value, dict):
+      output_dict.update(flatten_dict(value, prefix=nested_key))
+    else:
+      output_dict[nested_key] = value
+  return output_dict
+
+
+def _name_idx(name: str):
+  """Returns the layer index of the parameter name."""
+  index = name[name.find('_') + 1 : name.find('/')]
```
avital, line 115: cc @levskaya @jheek weird stuff that happens when we autoincrement layer names as strings -- "everything becomes part of the API"

Marvin182: I think this is bad style. New layers might not follow this convention and the param info shouldn't rely on it. I actually have good experience with just sorting alphabetical by name.

levskaya: @rolandgvc - I wouldn't try to sort parameters by the "layer order", in general the "order" of layers is a partial order given that arbitrary dataflow can happen in complex modules.  Also this ordering information is lost for manually named layers.

@avital - the alternative to names is an opaque tree-structure-based serialization (which we used in trax) and it is a complete nightmare to work with when debugging.  Pragmatically, I am happy to suffer hyrum's law if it enables vastly more pleasant debugging.  I want a human-navigable "at rest" representation of models.

avital:
@levskaya I believe we can get "the best of both worlds" with easy
introspection while allowing people to look through ordered lists when
those are semantically meaningful.

flax/nn/utils.py
avital: I believe we can get "the best of both worlds" with easy
introspection while allowing people to look through ordered lists when
those are semantically meaningful.

In the meanwhile I propose we add a strong TODO comment here saying that 
we should use this as a use-case for the ongoing API rewrite considerations.

rolandgvc: This could also be resolved by having the params as OderedDicts instead so they can keep the order of the applied modules by default.

avital: Let's just add a comment for now and I think we can merge this.

avital:
See CLU (Common Loop Utils) for another implementation: https://github.com/google/CommonLoopUtils/blob/master/clu/parameter_overview.py

cgarciae:
Hey! If adding a dependency is not too much of a burden, maybe we could leverage the [rich.tables](https://rich.readthedocs.io/en/latest/tables.html), they provide a lot of format options, colors, and other stuff like properly handing multi-line rows. 

Here is an example of rich in action:

rich

avital:
@cgarciae that looks really good. But I wonder if we can/should separate the needs: Part 1 is a function that give a Flax module (and variables and inputs?) generates some simple dict output that describes the module heirarchy. Then part 2 is a small piece of code that reads this dict and calls into rich.tables. Then we could replace part 2 with any other renderer. WDYT? We could even put the "renderer" part in a separate pip package to remove the dependency of flax on it.

It's also worth comparing and contrasting with https://dm-haiku.readthedocs.io/en/latest/notebooks/visualization.html

cgarciae:
I think splitting the functionality makes a lot of sense, the current proposed implementation in the PR more or less does this via the separation between `param_info` and `show_param_info`. I think we could expand the current idea with the following:

* [optional] Rename `param_info` to `module_info`, this would take in a Module, variables, sample inputs, and would use `capture_intermediates` to add input/output information about the layers to the dict structure. 
* Create a `render_info` function that could render the structure, internally it would use rich.tables.
* `show_{module,param}_info` could have a `render_fn` that defaults to `render_info`, it would simply call `{module,param}_info` and subsequently call `render_fn`. Users could provide their own `render_fn`. 

Later on `render_info` could be extracted if deemed generally useful.

avital:
Closing in favor of #1844

PR #289, opened 2020-05-27
Refactor MNIST example aligning with Issue #231.

mohitreddy1996:
Refactor MNIST example aligning with #231

Major changes:

Refactor files to example_lib (containing training and eval loop code), example_main (script which parses flags and starts the training loop).

Introduce a test using tfds.testing.mock_data which executes a single step of training and eval.

In order to do so, copy the metadata files from TFDS and store in examples/testing/datasets.

Documentation for testing with mock TFDS in #287


danielsuo:
Let me know if this is ready for me to update the `howto`s!

danielsuo:
@googlebot I consent

danielsuo:
Updated the branch so it could be merged/updated two howtos to understand scope. Since this branch will probably have changes, I'll stop here until we're happy with the PR.

howtos/diffs/distributed-training.diff
```diff
@@ -1,26 +1,33 @@
...
+@@ -60,6 +63,7 @@ def create_model(key):
  def create_optimizer(model, learning_rate, beta):
    optimizer_def = optim.Momentum(learning_rate=learning_rate, beta=beta)
    optimizer = optimizer_def.create(model)
 +  optimizer = jax_utils.replicate(optimizer)
    return optimizer
  
  
-@@ -99,6 +103,11 @@ def cross_entropy_loss(logits, labels):
+@@ -72,6 +76,11 @@ def cross_entropy_loss(logits, labels):
```
avital, line 30: I'm curious about these -- looks like when base examples change, many of the HOWTO diff changes lead to these essentially informationless changes, just because line numbers change.

Does `git apply` even look at line numbers or just the context around them? Could we perhaps somehow not include line numbers in our diffs?

howtos/diffs/distributed-training.diff
```diff
@@ -62,50 +69,21 @@ index 9c21985..fed5421 100644
...
      optimizer, train_metrics = train_epoch(
          optimizer, train_ds, batch_size, epoch, input_rng)
 -    loss, accuracy = eval_model(optimizer.target, test_ds)
 +    model = jax_utils.unreplicate(optimizer.target)  # Fetch from 1st device
 +    loss, accuracy = eval_model(model, test_ds)
+ 
      logging.info('eval epoch: %d, loss: %.4f, accuracy: %.2f',
                   epoch, loss, accuracy * 100)
-     summary_writer.scalar('train_loss', train_metrics['loss'], epoch)
-diff --git a/examples/mnist/train_test.py b/examples/mnist/train_test.py
```
avital, line 83: Is this expected? We no longer need test changes for this HOWTO?

danielsuo:
This is something I’m looking into. The current refactor pushes these
changes into mnist_lib, but verifying it’s actually doing a test.

Re only changing line numbers, I think this is just the diff of diff view
ie the howto diff itself won’t have this.

avital:
> Re only changing line numbers, I think this is just the diff of diff view
ie the howto diff itself won’t have this.
Right. It's just that it's annoying to review change to base examples if it
requires scanning through a bunch of essentially empty diff chunks like
these...

danielsuo:
Ah yeah. I think the line numbers might be fundamental, but couldn't find
documentation to the contrary. I know you can play around with whitespace
settings (i.e., does this kind of whitespace count as a change), but didn't
see anything about line numbers.

examples/mnist/mnist_lib.py
```diff
@@ -36,34 +29,15 @@
...
-
+from flax import nn
+from flax import optim
+from flax.metrics import tensorboard
 
 
 class CNN(nn.Module):
   """A simple CNN model."""
 
+  # pylint: disable=arguments-differ
```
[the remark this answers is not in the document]
mohitreddy1996: done.

examples/mnist/mnist_lib.py
```diff
@@ -79,29 +53,29 @@ def apply(self, x):
     return x
 
 
-def create_model(key):
+def _create_model(key):
```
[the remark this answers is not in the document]
mohitreddy1996: done.

examples/mnist/mnist_lib_test.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+# TODO(#290): Refactor logic in testing/benchmark.py to create a
+# utility class for parsing event files and extracting scalar summaries.
+def process_event(event):
+  for value in event.summary.value:
+    yield value
+
+
+def parse_and_return_summary_values(path):
```
avital, line 44: Can we avoid having this function and instead of parsing the tensorboard log files, just look at the return value from the train functions?

avital: Actually no need to return anything, because we can't reliably test the metrics anyways. Let's just test that nothing breaks.

examples/testing/datasets/mnist/3.0.1/dataset_info.json
```diff
@@ -0,0 +1,105 @@
...
+        ],
+        "numExamples": "60000"
+      }
+    }
+  ],
+  "supervisedKeys": {
+    "input": "image",
+    "output": "label"
+  },
+  "version": "3.0.1"
```
avital, line 104: IF we do use this we should lock mnist to version 3.0.1 in the data loading code as well

examples/mnist/mnist_lib.py
```diff
@@ -173,39 +146,39 @@ def get_datasets():
   return train_ds, test_ds
```
avital, line 146: Re the line above, `tfds.builder('mnist')`: Looks like we're hard-coding the MNIST dataset version in the tfds JSON file we have committed in this PR, but this code here doesn't specify a version. We should change the call to `tfds.builder` to have an explicit version of the MNIST TFDS dataset.

Moreover, we should maybe try to find a way to guarantee that nobody can accidentally bump this version without also updating the JSON file

avital: @mohitreddy1996 presuaded me that this should already break because `tfds.mock_data` wouldn't be able to find the JSON file otherwise. Worth doing a quick check by setting version explicitly here and seeing the test fail.

examples/testing/datasets/mnist/3.0.1/dataset_info.json
```diff
@@ -0,0 +1,105 @@
+{
+  "citation": "@article{lecun2010mnist,\n  title={MNIST handwritten digit database},\n  author={LeCun, Yann and Cortes, Corinna and Burges, CJ},\n  journal={ATT Labs [Online]. Available: http://yann. lecun. com/exdb/mnist},\n  volume={2},\n  year={2010}\n}\n",
```
avital, line 2: We should add a README to examples/testing/datasets explaining when and how to add entries here.

avital: @mohitreddy1996 tells me that this already exists in #287

tests/run_all_tests.sh
```diff
@@ -1,5 +1,22 @@
 #!/bin/bash
 
+# Download TFDS metadata to flax/.tdfs/metadata directory. 
+# This allows the tests to specify the `data_dir` when using tfds.testing.mock_data().
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd)"
```
avital, line 6: I think `dirname "$0"` works.

avital: Instead of setting the DIR env var, you can start the script with `cd $(dirname "$0")`

tests/run_all_tests.sh
```diff
@@ -1,5 +1,22 @@
...
+DIR="$( dirname "${DIR}" )"
+
+if [ -d "${DIR}/.tfds/metadata" ]; then 
+  echo 'Exists'; 
+else 
+  echo 'TFDS metadata doesnt exist. Downloading...';
+  # subversion checkout to the `trunk` branch which corresponds to `tree/master`.
+  # To download from branch `foo`, replace `trunk` with `branches/foo`.
+  svn checkout \
```
avital, line 15: Add a comment before saying, e.g. "We use subversion for faster downloads of only the metadata subdirectory with no history. If this script fails for you because you don't have subversion installed, please let us know by opening a GitHub issue"

examples/mnist/mnist_lib_test.py
```diff
@@ -75,9 +55,8 @@ def test_train_and_evaluate(self):
        mocked data for MNIST dataset.
     """
     # Get datasets testing-metadata directory.
-    current_dir = os.path.dirname(os.path.realpath(__file__))
-    parent_dir = os.path.dirname(current_dir)
-    data_dir = parent_dir + '/testing/datasets'
+    flax_root_dir = Path(__file__).parents[2]
```
avital, line 58: Add a small comment, e.g. "Go two directories up to the root of the flax repository"

tests/run_all_tests.sh
```diff
@@ -1,5 +1,22 @@
 #!/bin/bash
 
+# Download TFDS metadata to flax/.tdfs/metadata directory.
```
avital, line 3: Let's extract this code into a separate script called `download_dataset_metadata.sh`

tests/run_all_tests.sh
mohitreddy1996: Done.

examples/mnist/mnist_lib_test.py
mohitreddy1996: Done.

tests/run_all_tests.sh
mohitreddy1996: Done.

tests/run_all_tests.sh
mohitreddy1996: Done.

avital:
@googlebot I consent.

PR #294, opened 2020-05-28
Use three-way merge when HOWTO diffs don't cleanly apply

avital:
I believe this should resolve the failing test in #293

### DIFF
--- a/howtos/scripts/apply_howto_diffs.sh
+++ b/howtos/scripts/apply_howto_diffs.sh
@@ -38,7 +38,7 @@ for howto in $howtos; do
   howto_branch="howto/${howto}"
   git checkout -b $howto_branch
   diff_file="${howto_diff_path}/${howto}.diff"
-  git apply $diff_file
+  git apply -3 $diff_file
 
   # Once we are satisfied with the howto, commit and push
   git commit -am "Added howto branch ${howto_branch}"
--- a/howtos/scripts/check_howto_diffs.sh
+++ b/howtos/scripts/check_howto_diffs.sh
@@ -27,7 +27,7 @@ for howto in $howtos; do
 
   # Check if command fails
   # See: https://stackoverflow.com/q/26675681/
-  if ! git apply --check "${diff_file}"; then
+  if ! git apply -3 --check "${diff_file}"; then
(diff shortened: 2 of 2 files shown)


avital:
@danielsuo Could you take a look at this?

On a related note, I'm curious why we ended up creating local branches for every HOWTO for running tests. Couldn't we just apply each HOWTO separately and then revert the patch? Probably there was a need for making the process on iterating on broken HOWTOs easier? Can you give us a reminder?

danielsuo:
Some comments:

1. We should update the howto for howtos with this `git apply -3`; three-way merge, even with conflicts, would be helpful when updating examples

2. Of course, `git apply -3` isn't a silver bullet; we can pass `--check`, but looks like applying can still cause conflicts

3. Any three-way merge will introduce new changes that need to be committed. How do we feel about modifying the PR automatically? It may be better to just fail?

danielsuo:
> On a related note, I'm curious why we ended up creating local branches for every HOWTO for running tests. Couldn't we just apply each HOWTO separately and then revert the patch? Probably there was a need for making the process on iterating on broken HOWTOs easier? Can you give us a reminder?

Spent some time trying to retrace why I have checking out the branch, but seems like tests pass and I can't remember what I was trying to resolve (vaguely something about the GitHub workflow checking out some headless branch). Changed my dummy PR into a real one here #295

avital:
@danielsuo can you explain "Any three-way merge will introduce new changes that need to be committed. How do we feel about modifying the PR automatically? It may be better to just fail?"

Why would three-way merging introduce new changes?

avital:
@danielsuo is it right that you're replacing this with a new PR that solves the problem more correctly? If so I'll close this one.

danielsuo:
Yeah, I'm taking a look at the easiest way to check if applying a patch will cause a merge conflict without actually applying the patch.

This way, we can get the benefits of `--3way`, but still have a potentially dirty local copy.

Forcing a clean copy turns out to have a suboptimal workflow since you wind up creating a new commit for each incremental problem you fix (can only find the next problem with `check_howtos_diff.sh`, but can only run that script if your local copy is clean).

avital:
Cool, sounds good -- thanks for investigating!

PR #296, opened 2020-05-28
Use Jax RNG for data shuffling in MNIST example.

danielsuo:
This PR contains commit history from #293 and fixes up `howto`s.

### DIFF
--- a/howtos/diffs/checkpointing.diff
+++ b/howtos/diffs/checkpointing.diff
@@ -25,8 +25,8 @@ index 9c21985..45311c7 100644
      summary_writer.scalar('eval_loss', loss, epoch)
      summary_writer.scalar('eval_accuracy', accuracy, epoch)
 +
-+  # Save model at every epoch
-+  checkpoints.save_checkpoint(FLAGS.model_dir, optimizer, epoch, keep=3)
++    # Save model at every epoch
++    checkpoints.save_checkpoint(FLAGS.model_dir, optimizer, epoch, keep=3)
 +
    summary_writer.flush()
    return optimizer
--- a/howtos/diffs/checkpointing.diff
+++ b/howtos/diffs/checkpointing.diff
@@ -1,5 +1,5 @@
 diff --git a/examples/mnist/train.py b/examples/mnist/train.py
-index 5dc74db..975e50f 100644
+index 9c21985..45311c7 100644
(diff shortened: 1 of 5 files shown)


howtos/diffs/checkpointing.diff
```diff
@@ -11,23 +11,23 @@ index 5dc74db..975e50f 100644
...
+@@ -198,6 +202,10 @@ def train(train_ds, test_ds):
      summary_writer.scalar('train_accuracy', train_metrics['accuracy'], epoch)
      summary_writer.scalar('eval_loss', loss, epoch)
      summary_writer.scalar('eval_accuracy', accuracy, epoch)
-+    
-+    # Save model at every epoch
-+    checkpoints.save_checkpoint(FLAGS.model_dir, optimizer, epoch, keep=3)
-+  
++
++  # Save model at every epoch
```
avital, line 28: Is this correct? The indentation look to have changed -- doesn't this now only save the model at the end of training?

howtos/diffs/distributed-training.diff
```diff
@@ -62,21 +62,21 @@ index 5dc74db..77b49fa 100644
...
-@@ -192,7 +207,8 @@ def train(train_ds, test_ds):
-   for epoch in range(1, num_epochs + 1):
+@@ -191,7 +206,8 @@ def train(train_ds, test_ds):
+     rng, input_rng = random.split(rng)
      optimizer, train_metrics = train_epoch(
          optimizer, train_ds, batch_size, epoch, input_rng)
 -    loss, accuracy = eval_model(optimizer.target, test_ds)
 +    model = jax_utils.unreplicate(optimizer.target)  # Fetch from 1st device
-+    loss, accuracy = eval_model(model, test_ds)
++    loss_accuracy = eval_model(model, test_ds)
```
avital, line 79: Is this correct? Seem like a typo maybe? (Do HOWTO tests pass here? -- looks like yes, but maybe the tests don't yet cover this step?)

danielsuo: Yes, this is a typo! The test tests `train_step`, but not `train`.

howtos/diffs/checkpointing.diff
danielsuo: Yes, that sounds right; fixed.

PR #297, opened 2020-05-29
Add *.rej to .gitignore

danielsuo:
Tiny housekeeping thing; now that the [HOWTOs HOWTO](https://flax.readthedocs.io/en/latest/howtos-howto.html) suggests using `git apply --reject`, adding `*.rej` files to `.gitignore`. I noticed that `*.rej` files were making their way into commits.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ docs/_build
 dist/
 build/
 *.egg-info
+*.rej


danielsuo:
Looks like there's an issue with `tensorboard`. They just updated to v `2.2.2` today and something broke. Looking into it.

danielsuo:
Waiting on #298.

avital:
Thanks for catching this!

PR #298, opened 2020-05-29
Update tensorboard tests

danielsuo:
`tensorboard` just pushed a new release today (`2.2.2`) and broke `tests/tensorboard_test.py`. Looks like they're indexing differently, but didn't have time to look into more detail.

@mohitreddy1996 would you mind taking a look at the [release](https://github.com/tensorflow/tensorboard/releases) and confirming the changes I've made are semantically correct?

Unfortunately, short of pinning `tensorboard`, this PR needs to get in before any others will pass.

### DIFF
--- a/tests/tensorboard_test.py
+++ b/tests/tensorboard_test.py
@@ -147,7 +147,7 @@ class TensorboardTest(absltest.TestCase):
     self.assertEqual(summary_value.tag, 'audio_test')
 
     # Assert two audio files are parsed.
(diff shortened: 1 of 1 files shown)


mohitreddy1996:
Wow finally this change was released :) 

I was confused when I was trying to debug the issue of why there were '4' entries when I wrote summaries for 2 entries. 

Thanks for the fix!

danielsuo:
Good! I was reading the code and I was like...two audio files...length 4...what??

PR #299, opened 2020-05-29
Use three-way merge for checking/applying howtos

danielsuo:
After going through the exercise of implementing, I actually lean towards rejecting this PR (though a very weakly held opinion):

1. Increases burden on users (need `git apply -3` or `git apply --3way` instead of just `git apply`) without obvious benefit to them
2. Using `--3way` may actually complicate life a little for maintainers of these scripts and howto authors because we may cause conflicts in their local copies as they are working on other things. There are some hacky ways to check if a merge will cause conflicts before actually merging, which makes `--3way` more palatable.

`--3way` does save us from some howto-editing, but if we expect conflicts to happen infrequently anyway, maybe not a huge benefit.

Anyway, look forward to comments / suggestions!

### DIFF
--- a/docs/howtos.rst
+++ b/docs/howtos.rst
@@ -17,6 +17,19 @@ Note that these HOWTOs do not require special library support, they just
 demonstate how assembling the JAX and Flax primitives in different ways
 allow you to make various training loop modifications.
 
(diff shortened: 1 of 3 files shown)


avital:
Let's keep this open for a while and assess how frequently we need to resolve HOWTO merge conflicts before deciding.

avital:
I have updated the HOWTO HOWTO docs to show how to use 3 way merge in case the normal merge fails.

PR #300, opened 2020-05-30
Optimize repeated function calls in serialization.py

kshithijiyer:


danielsuo:
Thanks for this! If I understand the PR correctly, is this refactoring two calls to `register_serializationg_state` into a `for` loop?

I think we'd prefer to keep the original for easier reading since it's just two lines and the change doesn't improve performance.

Please let me know if I'm mistaken!

AlexeyG:
Ping @kshithijiyer

kshithijiyer:
> Ping @kshithijiyer 

Pong @AlexeyG

kshithijiyer:
> Thanks for this! If I understand the PR correctly, is this refactoring two calls to `register_serializationg_state` into a `for` loop?
> 
> I think we'd prefer to keep the original for easier reading since it's just two lines and the change doesn't improve performance.
> 
> Please let me know if I'm mistaken!

Feel free to close the PR if you prefer the original one.

jheek:
I don't think the for loop makes the code more readable. I'll close this issue for now.

Issue #214, opened 2020-04-19
BatchNorm error message


david-waterworth:
The `flax.nn.normalization.BatchNorm` error message "batch_stats should be provided if use_running_averages is True" is slightly misleading. 

I think would be clearer if it say something like "when use_running_averages is True either use a stateful context or provide batch_stats"

It wasn't clear to me why the `imagenet` example worked without providing batch_stats and my layer failed until I looked closer at the source and realised that it was using a stateful context.

avital:
Good point. We're reconsidering our use of context managers (perhaps in favor of function decorators), and perhaps that would simplify the mental model.

In the meanwhile, would you like to give a shot at improving the error message in a pull request? The language in the new error should really direct the user towards the right solution, so it's important to think carefully about the wording.

avital:
Hi @david-waterworth -- any interesting in submitting a pull request improving the error message?

ayush-1506:
@avital I've opened a PR that should fix this: #301. Would love a review. Thanks

PR #301, opened 2020-05-31
Add more meaningful error message to batchnorm

ayush-1506:
Fixes: #214 

Replaced the error message with a more specific message as mentioned in the issue thread. Kindly let me know if it's better to have some other message here.

### DIFF
--- a/flax/nn/normalization.py
+++ b/flax/nn/normalization.py
@@ -91,8 +91,8 @@ class BatchNorm(base.Module):
 
     if use_running_average:
       if ra_mean is None:
-        raise ValueError('batch_stats should be provided if '
-                         'use_running_averages is True')
+        raise ValueError('when use_running_averages is True '
+                         'either use a stateful context or provide batch_stats')
       mean, var = ra_mean.value, ra_var.value
     else:
       mean = jnp.mean(x, axis=reduction_axis, keepdims=False)


danielsuo:
As @avital mentioned in #214, this is a good opportunity to rethink the error message (understanding that the API for stateful contexts may change).

What would help guide you to doing the right thing? Perhaps a link to the `state` documentation in the guided tour ([link](https://flax.readthedocs.io/en/latest/notebooks/flax_guided_tour.html#State
))? Maybe flesh out the `BatchNorm` documentation more with an example and link to it as well?

ayush-1506:
@danielsuo Thanks for the input. I could add an example of BatchNorm usage in the batchnorm docstring and link that in the error message (along with adding the link to flax). Would that be better?

ayush-1506:
ping @danielsuo @avital

danielsuo:
I think that would be a reasonable start, but this might need @avital's attention since he may have an opinion. I could imagine more of our error messages using this one as a template.

ayush-1506:
Ok, I guess I'll probably wait for @avital 's input here.

AlexeyG:
The proposed error message is strictly better than the current behaviour. Moving forward, once the FLIPs land, error messages like that would no longer be possible.

PR #306, opened 2020-06-03
HOWTO: Per example gradient

rolandgvc:
This howto exemplifies how per-example gradients can be retrieved using `jax.vmap`.

### DIFF
--- a/howtos/diffs/per-example-gradient.diff
+++ b/howtos/diffs/per-example-gradient.diff
@@ -1,0 +1,31 @@
+diff --git a/examples/mnist/train.py b/examples/mnist/train.py
+index 9c21985..3fad33c 100644
+--- a/examples/mnist/train.py
++++ b/examples/mnist/train.py
+@@ -112,12 +112,20 @@ def compute_metrics(logits, labels):
+ @jax.jit
+ def train_step(optimizer, batch):
+   """Train for a single step."""
+-  def loss_fn(model):
+-    logits = model(batch['image'])
+-    loss = cross_entropy_loss(logits, batch['label'])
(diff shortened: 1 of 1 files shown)


howtos/diffs/per-example-gradient.diff
```diff
@@ -0,0 +1,31 @@
...
++    return cross_entropy_loss(model(inputs), labels)
++  vmap_loss_grad = jax.vmap(jax.grad(vmap_loss_fn), in_axes=(None, 0, 0))
++  grads = vmap_loss_grad(optimizer.target, batch['image'], batch['label'])
++  
++  # Average across examples
++  mean_fn = lambda x: jnp.mean(x, axis=0)
++  grad = jax.tree_map(mean_fn, grads) 
++
+   optimizer = optimizer.apply_gradient(grad)
+   metrics = compute_metrics(logits, batch['label'])
```
avital, line 30: Need to check compute_metrics. It probably computes the loss a second time. Instead, we should use `jax.value_and_grad` where vmap_loss_grad is computed.

avital:
We've redone our HOWTO system. Here's an example of the new format:

Rendered in ReadTheDocs: https://flax.readthedocs.io/en/latest/howtos/ensembling.html
Document source: https://github.com/google/flax/blob/master/docs/howtos/ensembling.rst (see the `:testsetup`, `:testcode` and `:codediff` directives)

We'd happy take pull requests revamping this HOWTO to the new system.

avital:
Looks like this PR is inactive, so I'll close it. We're tracking the HOWTO request here: #858

PR #307, opened 2020-06-04
HOWTO: Access intermediate layer output shapes

rolandgvc:
Capture intermediate layer output shapes with an `nn.Module`.

### DIFF
--- a/howtos/diffs/layer-output-shapes.diff
+++ b/howtos/diffs/layer-output-shapes.diff
@@ -1,0 +1,62 @@
+diff --git a/examples/mnist/train.py b/examples/mnist/train.py
+index 9c21985..c63f6fa 100644
+--- a/examples/mnist/train.py
++++ b/examples/mnist/train.py
+@@ -60,24 +60,42 @@ flags.DEFINE_string(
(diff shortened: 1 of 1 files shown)


rolandgvc:
The extra modules added slightly decrease the accuracy of the model, resulting in a build fail. @avital, as @danielsuo mentioned, we should be looking into bypassing some of the inference assertions when the howtos are addressing debugging.

avital:
We've redone our HOWTO system. Here's an example of the new format:

Rendered in ReadTheDocs: https://flax.readthedocs.io/en/latest/howtos/ensembling.html
Document source: https://github.com/google/flax/blob/master/docs/howtos/ensembling.rst (see the `:testsetup`, `:testcode` and `:codediff` directives)

We'd happy take pull requests revamping this HOWTO to the new system.

marcvanzee:
I'm closing this PR due to inactivity.

Note we added a HOWTO called [Extracting intermediate values](https://flax.readthedocs.io/en/latest/howtos/extracting_intermediates.html), which should allow one to extract layer output shapes as well.

PR #310, opened 2020-06-04
HOWTO: Selective optimization

rolandgvc:
Modify the `train_step` function to compute per-example gradients.


avital:
Looks like this is a dupe of #306 ?

PR #311, opened 2020-06-04
HOWTO: Params info

rolandgvc:
Log number and size of trainable parameters.


rolandgvc:
@marcvanzee I've closed the other branch due to a rebasing mistake :) Please have a look at this one 👍

rolandgvc:
I believe #288 does a greater job at extracting parameter info. Maybe this howto could instead demonstrate how the util works?

avital:
I'll close this and keep the idea tracked at #288 -- if we have a utility function we just need to document it, not show how to use it in a HOWTO.

PR #312, opened 2020-06-04
Hide diffs from test files in howto documentation

danielsuo:
We parse the diff file. Another way to do this would be to
1. Apply the howto diff
2. Undo any diffs on test files
3. Update the howto
4. Format the updated howto

This alternative way is better in the build environment, but makes building documentation locally more difficult (e.g., what if my local copy has changes?).

### DIFF
--- a/docs/format_howtos.py
+++ b/docs/format_howtos.py
@@ -61,28 +61,22 @@ def format_howto(input_file, output_file):
   # - By compiling a grouped expression, we ensure re.split returns both
   # delimiter and delimited strings (i.e., an array of alternating delimiters
   # and delimited strings)
-  # - We don't assume any particular file extension or form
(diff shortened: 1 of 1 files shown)


marcvanzee:
I don't really get the advantage of this approach over the current one.

> ..makes building documentation locally more difficult (e.g., what if my local copy has changes?).

I don't see how your changes does this. it seems like you mostly process the input file differently with some regexes. Could you please elaborate?

marcvanzee (changes requested):
I don't really get the advantage of this approach over the current one.

> ..makes building documentation locally more difficult (e.g., what if my local copy has changes?).

I don't see how your changes does this. it seems like you mostly process the input file differently with some regexes. Could you please elaborate?

marcvanzee:
Closing this since we have simplified our HOWTO system.

PR #313, opened 2020-06-04
Give each howto its own page

danielsuo:
Preview of what docs look like: https://dsuo-flax.readthedocs.io/en/dsuo-howto-individual-pages

FYI for @rolandgvc.

Some notes:
- I added a check to make sure the `howto` author adds a documentation file. It's not a perfect check, but hopefully it reminds reviewers/authors that there is some plumbing to add for documentation
- Right now, the individual pages just show the diff and nothing else. We will add some context to these in the future


avital:
Thank you Daniel!

PR #314, opened 2020-06-04
Add examples to read-the-docs

danielsuo:
Preview: https://dsuo-flax.readthedocs.io/en/dsuo-examples-to-rtd/

I updated the main `README.md` file and put the examples in the same structure as in the main `README.md`.

Because we already have `README.md`s for each example, I thought we may as well reuse. I copy over the relevant files (currently whitelisted extensions `*.py` and `*.png`) into a `.gitignore`d directory. There are downsides to this approach (what if we have `*.jpg`?), so happy to hear suggestions.

Note, this would increase the maintenance burden a little bit since we add a second place that requires updating.


avital:
We completely lost track of this issue a long time ago, but since then we have indeed moved examples to RTD: https://flax.readthedocs.io/en/latest/examples.html

PR #315, opened 2020-06-04
Remove loss_scaling flag from imagenet benchmark

mohitreddy1996:
Flag removed in #220

### DIFF
--- a/examples/imagenet/train_benchmark.py
+++ b/examples/imagenet/train_benchmark.py
@@ -43,7 +43,6 @@ class ImagenetBenchmark(Benchmark):
     model_dir = tempfile.mkdtemp()
     FLAGS.batch_size = 2048
     FLAGS.half_precision = True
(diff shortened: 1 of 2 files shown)


examples/imagenet/train_benchmark.py
```diff
@@ -43,7 +43,6 @@ def test_8x_v100_half_precision(self):
     model_dir = tempfile.mkdtemp()
     FLAGS.batch_size = 2048
     FLAGS.half_precision = True
-    FLAGS.loss_scaling = 256.
```
AlexeyG, line 46: I know that it's technically not a part of your PR, but you could also update the README fo the imagenet example? It currently lists `--loss_scaling=256.` as a command line argument.

mohitreddy1996: Good point. Thanks for catching this.

Issue #166, opened 2020-04-01
Re-shuffle data each iteration


AlexeyG:
Some example (e.g. `nlp_seq`) show clear patterns in their training loss (see TensorBoard).

This is probably because the training data isn't shuffled enough. We should consider adding `shuffle` with `reshuffle_each_iteration=True` to the train IO pipelines of existing examples.

jheek:
`reshuffle_each_iteration` is `False` by default? I guess this is a great example of why implicit random seeds are evil. 

In the past I have also noticed that creating iterators in `tf.data` tends to come with significant overhead so I prefer to create one infinite iterator and use a manual `for step in range(steps_per_epoch)` to loop for an epoch. This also avoids some other issues like partial batches.

AlexeyG:
Just checked, it actually does default to `True`:

```
reshuffle_each_iteration: (Optional.) A boolean, which if true indicates that the dataset should be pseudorandomly reshuffled each time it is iterated over. (Defaults to True.)
```

Perhaps the issue is that the `nlp_seq` example does not shuffle the data?

AlexeyG:
Marking as PR welcome. A successful PR would also include (a link to) new TensorBoard logs verifying the metrics (see `README.md` in `examples/nlp_seq` for expected numbers).

gan3sh500:
It looks like dataset.shuffle was just missing. I've made the change. I'll train on a V100 similar to the README and submit a PR with updated logs and code.

avital:
Awesome @gan3sh500 ! Indeed if you can reproduce the same results on a V100 and share a link to tensorboard.dev we'll merge your PR.

gan3sh500:
Sorry for slow turnaround. I didn't get the V100 but I now have a 2080Ti locally to run and have ran the same batch_size and iterations. The results are 0.23% lower than the one shown in the repo. Please see for yourself. [TB Dev](https://tensorboard.dev/experiment/1kp6O2xbRDWCJCmrmnLBgA) #321

marcvanzee:
It looks like @bohnetbd felt this wasn't worth adding due to the added complexity (given that this is an educational example). So I am closing this.

PR #321, opened 2020-06-11
Fixing pattern in examples/nlp_seq due to lack of shuffling

gan3sh500:
As mentioned in this [issue](https://github.com/google/flax/issues/166) nlp__seq was missing shuffling in the tf dataset. I have added it and re ran the code with the same batch size and training iterations and reproduced a similar accuracy.

### DIFF
--- a/examples/nlp_seq/README.md
+++ b/examples/nlp_seq/README.md
@@ -21,7 +21,7 @@ The model should run with other configurations and hardware, but explicitly test
 
 | Hardware |  Batch size  | Learning rate | Training time | Accuracy  | TensorBoard.dev |
 |:---:|:---:|:---:|:---:|:---:|:---:|
-| Nvidia V100 (16GB) | 64  |  0.05 | 5h 15m | 72.20% | [2020-03-22](https://tensorboard.dev/experiment/YkUAdwYaQ9OtYl2IVe3MvA/) |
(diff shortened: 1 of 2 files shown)


AlexeyG:
@gan3sh500 Just to make sure - could you please sign the CLA? This is required for us to be able to accept your PR. Thank you!

gan3sh500:
Sorry forgot about it. Signed now.

avital:
Hi @gan3sh500 -- I'm so sorry for this laying unmerged for so long. In the meanwhile we moved to `linen_examples/` for all of our new examples. Do you want to want to submit a new PR (without changing the README) to `linen_examples/nlp_seq/`? Alternatively we could just apply your change and reference this old PR for acknowledgement.

(cc @bohnetbd @bastings as the owners of the nlp_seq example)

bohnetbd:
Thanks for the suggestion. With this example, we do not go for high accuracy and to keep it simple. In addition and more important maybe, I see also that there was no accuracy gain from shuffling. So that I would hesitate to take this change in before we get clear picture that this would help for other languages. I would rather close this PR if fine with all.

gan3sh500:
@avital @bohnetbd Shall I take it up. I need to go through the code. I've forgotten it kind of.

avital:
Hi @bohnetbd -- I found this old improvement to nlp_seq while cleaning up my GitHub issues. Can you please review it?

bohnetbd:
Hi @avital and @gan3sh500,

I would suggest to not take this up as the accuracy improvements are not give and we want to focus only a simple example here. In case, we would like to work on accuracy, we could use pertained embeddings or character embeddings which would lead to very high accuracy gains.

so lets close this. Thank you all!

Issue #264, opened 2020-05-14
LM1B example does not handle partial batches in evals


georgedahl:
Provide as much information as possible. At least, this should include a description of your issue and steps to reproduce the problem. If possible also provide a summary of what steps or workarounds you have already tried.

### Problem you have encountered:

The LM1B example calls input_poipeline.get_lm1b_datasets https://github.com/google/flax/blob/master/examples/lm1b/train.py#L350 with drop_remainder at its default of True, dropping partial batches in the validation data caused by the number of examples being evenly divisible by the batch size. This will ignore some validation data and if a user modifies this example to evaluate on their own test set it will produce incorrect results.

### What you expected to happen:

The example should be able to correctly handle any number of validation examples and drop none of them by default. This might require padding and masking. It is fine to have slightly different behavior for the training set, but for evals every example in the eval set should be processed exactly once and the metrics should be aggregated across all the eval batches correctly.

jheek:
It's quite common to drop the remainder in for example the official TF examples. But I agree we should really do better. For a single host the fix is trivial but for multiple hosts we need to be careful about feeding the same number of batches and padding up to the same size on each host.

georgedahl:
It isn't safe to do work intended for scientific publication with code that doesn't compute correct evals. It is a shame incorrect evals have become normalized by bad examples in other frameworks.

I know you aren't really disagreeing with me here, but I worry your comment could be read as minimizing this issue. This is a severe correctness problem even though it might be a common problem. The longer it exists in an example the more risk there is someone will put wrong eval numbers in their publications and it will spread to more codebases.

I also think this is a great example of where our tools push us to write bad code because writing correct code is inconvenient and uglier. It should be easier to handle these things which is why I additionally filed https://github.com/google/jax/issues/3088

PR #323, opened 2020-06-13
Correct evaluation in CIFAR10 example for incomplete batches.

### LINKED ISSUE #262: CIFAR10/ImageNet examples incorrectly calculate test metrics with default batch size
If I understand the current example code for CIFAR10 and ImageNet, it calculates the test metrics by first taking the mean over the mini-batch in each step, and then taking the mean over all steps, i.e. calculating the mean of means:

https://github.com/google/flax/blob/58e340bee4dc12415ea9a46e59056c6f4f1b5527/examples/cifar10/train.py#L119-L127
https://github.com/google/flax/blob/58e340bee4dc12415ea9a46e59056c6f4f1b5527/examples/cifar10/train.py#L265-L276

When all mini-batches are of the same size this is equivalent to the mean of the metrics over all data samples, however in the case where mini-batches differ in size this is not equivalent to the mean of the samples. This is of course the case typically when test data set is not evenly divisible by the batch size — and this is the case in both the CIFAR and ImageNet examples currently.

If my understanding of this is correct, both the CIFAR10/ImageNet examples the test metrics are likely currently being calculated incorrectly.

There are a number of solutions, the simplest is probably to instead use a weighted mean, i.e. weighting the metrics by the batch size for each step, and then taking the mean over the weighted metrics over all batches:

```diff
diff --git a/examples/cifar10/train.py b/examples/cifar10/train.py
index b9e0b2d..8643939 100644
--- a/examples/cifar10/train.py
+++ b/examples/cifar10/train.py
@@ -270,10 +270,10 @@ def train(module, model_dir, batch_size,
         eval_batch = load_and_shard_tf_batch(eval_batch)
         # Step
         metrics = p_eval_step(optimizer.target, state, eval_batch)
-        eval_metrics.append(metrics)
+        eval_metrics.append(jax.tree_map(lambda x: len(eval_batch) * x, metrics))
       eval_metrics = common_utils.get_metrics(eval_metrics)
       # Get eval epoch summary for logging
-      eval_summary = jax.tree_map(lambda x: x.mean(), eval_metrics)
+      eval_summary = jax.tree_map(lambda x: x.sum()/data_source.EVAL_IMAGES, eval_metrics)
 
       # Log epoch summary
       logging.info(
```
Alternatively the batch vector of metrics could be returned and accumulated (probably a performance hit?) or the CIFAR10 example could be changed to use a different compatible test batch size to prevent this, but this would not work for the ImageNet example in general over multiple devices, and doesn't serve the primary purpose of this code which is to be a good starter/example of training.

yanii:
This patch fixes #262 for CIFAR10, by using a batchsize weighted average instead of the mean, allowing eval to use arbitrarily-sized batch sizes. A similar patch could address #262 for ImageNet and potentially also #264 .

### DIFF
--- a/examples/cifar10/train.py
+++ b/examples/cifar10/train.py
@@ -266,14 +266,15 @@ def train(module, model_dir, batch_size,
       eval_metrics = []
       for _ in range(steps_per_eval):
         eval_batch = next(eval_iter)
+        eval_batch_len = eval_batch['image'].shape[0]
         # Load and shard the TF batch
         eval_batch = load_and_shard_tf_batch(eval_batch)
         # Step
         metrics = p_eval_step(optimizer.target, state, eval_batch)
-        eval_metrics.append(metrics)
+        eval_metrics.append(jax.tree_map(lambda x: eval_batch_len * x, metrics))
       eval_metrics = common_utils.get_metrics(eval_metrics)
       # Get eval epoch summary for logging
-      eval_summary = jax.tree_map(lambda x: x.mean(), eval_metrics)
+      eval_summary = jax.tree_map(lambda x: x.sum()/data_source.EVAL_IMAGES, eval_metrics)
 
       # Log epoch summary
       logging.info(


jheek:
The solution looks correct to me. Thanks for your contribution!
Googlebot is still complaining about signing the CLA though. Could you check this?

yanii:
Hi Jonathan - yes, it's because I'm TVC and there are some hoops to jump through to sign the CLA, it's in process though :-)

I will probably try adding the code for this to the ImageNet example too, but I'm not sure I'll have the resources to test that.

avital:
Hi @yanii we actually have an internal performance regression monitoring setup (still in early stages but already functional) -- @mohitreddy1996 has been the main person driving that. We're going use this PR as a way to test that and see whether there are changes to the eval numbers.

yanii:
Sounds good!

avital:
Hi @yanii, is there any progress on signing the CLA?

yanii:
@avital CCed you on the internal bug tracking the CLA request.

jheek:
@googlebot refresh

yanii:
Hey Jonathan, I'll CC you on the internal bug too, and then you'll know why this CLA still isn't signed.

yanii:
@googlebot refresh

yanii:
@avital @jheek  Finally able to sign the CLA as an individual now my Google contract has ended. Let's merge this fix :-)

avital:
We have eliminated our CIFAR example in the meanwhile in favor of https://github.com/google-research/google-research/tree/master/flax_models/cifar

FYI @andsteing and @Marvin182 -- should we follow this pattern, or some other pattern to resolve the "end of batch" issues for our examples?

(Closing because we no longer have a CIFAR example). Sorry it took so long for you to be able to sign the CLA ;)

andsteing:
Created #527 for resolving "end of batch" and other issues. Please have a look and comment there.

PR #328, opened 2020-06-18
Add link to DotGeneral in DenseGeneral doc.

lucasb-eyer:


MostafaDehghani:
Isn't it better to link to the `lax.dot_general` in JAX instead of XLA?
https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.dot_general.html

lucasb-eyer:
I thought about that first, but the only useful thing in that documentation is the link to XLA's doc, so why two clicks when one can do? :)

lucasb-eyer:
But I guess it doesn't hurt to have both links in case the `lax.dot_general` becomes more useful in the future, so added that too. (Sorry, 2 commits as I use github's UI for this PR)

MostafaDehghani:
Thanks you @lucasb-eyer.

PR #337, opened 2020-06-25
Have README.md link to GitHub discussions

avital:


avital:
> maybe worth having links to issues as well and guidance on putting actual concrete bugs there? or do we want to filter everything through dscussions?

My thinking is that people already know how to find the "Issues" tab, I am sure people will still know how to file bugs.

But we should pull people towards discussions, and better to start a discussion before filing a feature request, for example.

PR #343, opened 2020-06-29
ResNet v1 is missing ReLU in root-block

lucasb-eyer:
See [reference implementation](https://github.com/KaimingHe/deep-residual-networks/blob/master/prototxt/ResNet-50-deploy.prototxt#L31-L58) and (more readable) [torchvision implementation](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py#L205). If however this omission is intentional, it needs a comment to explain it.


lucasb-eyer:
ping

avital:
Hi Lucas, we'll try to merge this in this week but we need to run our
end-to-end tests first to see whether this affects the reported metrics.

PR #358, opened 2020-07-15
WMT example hyperparameters to flags

juliakreutzer:
Previously hard-coded model and training hyperparameters are now exposed as flags for `train.py`. 
[repeating #329 but from Github account]


juliakreutzer:
@levskaya CLA is cleared.

PR #362, opened 2020-07-18
fix pooling ops with explicit padding

### LINKED ISSUE #356: Pooling: passing "sequence of `n` `(low, high)` integer pairs" resulting in TypeError
Trying to pass a tuple or list of tuples to a pool operation's padding parameter gives out the following errors: 
`TypeError: Unknown padding type: (1, 1).`
`TypeError : unhashable type: 'list' `

Sample code for reproducing the bug:
```python3
from flax import nn
from jax import random

class FlaxModel(nn.Module):
    def apply(self, x):
        x = nn.max_pool(x, (3, 3), strides=(2, 2), padding=[(1, 1), (1, 1)])
        return x

rng = random.PRNGKey(0)
model, _ = FlaxModel.init_by_shape(rng, [(1, 100, 100, 1)])
```

levskaya:
previous max_pool and avg_pool didn't work with explicit (e.g. `((1,1),(1,1))` padding specs) padding for the reduce window. 
fixes #356

### DIFF
--- a/flax/nn/pooling.py
+++ b/flax/nn/pooling.py
@@ -44,6 +44,14 @@ def pool(inputs, init, reduce_fn, window_shape, strides, padding):
   strides = strides or (1,) * len(window_shape)
   strides = (1,) + strides + (1,)
   dims = (1,) + window_shape + (1,)
+  if not isinstance(padding, str):
+    padding = tuple(map(tuple, padding))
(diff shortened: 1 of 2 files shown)


levskaya:
Note we need to wait for JAX to cut a new release to test and use this modification.

levskaya:
Ok we can finally merge this!

PR #369, opened 2020-07-20
Rewriting seq2seq to new API

marcvanzee:
The model trains correctly and the tests pass. The initialization of the rngs in the modules is wrong, but I'm not sure how to do this nicely. What I tried is self.make_rng('...'), but then I have to pass all these rngs every time I want to create a model, which also feels a bit cumbersome. I guess the best way is to use `lift.scan`, but I haven't figured out yet how to do that.


examples/seq2seq/train.py
```diff
@@ -325,16 +325,17 @@ def compute_metrics(logits, labels):
...
+def train_step(optimizer, batch):
   """Train one step."""
   labels = batch['answer'][:, 1:]  # remove '=' start token
 
-  def loss_fn(model):
+  def loss_fn(params):
     """Compute cross-entropy loss."""
-    with nn.stochastic(rng):
-      logits, _ = model(batch['query'], batch['answer'])
+    logits, _ = model().apply({'param': params}, batch['query'],
```
avital, line 334: I am surprised that this works without passing in an rng as an option to apply. Was the original code not actually using dropout here?

marcvanzee:
Closing this since it is outdated.

PR #373, opened 2020-07-21
seq2seq example in new API

marcvanzee:
The model trains correctly and the tests pass. The initialization of the rngs in the modules is wrong, but I'm not sure how to do this nicely. What I tried is self.make_rng('...'), but then I have to pass all these rngs every time I want to create a model, which also feels a bit cumbersome. I guess the best way is to use lift.scan, but I haven't figured out yet how to do that.


marcvanzee:
Closing this since I have a new version now.

PR #383, opened 2020-08-04
Add overview to examples/README.md

andsteing:
A step towards streamlining the Flax examples (see also issue #231) - this PR gives a short overview over the different examples that allows users to quickly find an example they might be interested in (e.g. finding an example implementing multi host SPMD with custom preprocessing vs. a simple example with inline data).

Note that the table is deliberately kept short, both in terms of description/features, as well as in number of rows (@avital mentioned `nlp_seq`, `sst`, and `wip/moco` examples were not a priority).


avital:
Thanks, this looks great.

Going forward, we probably need to generally split up the examples we currently maintain into separate repositories (and we'll have to individually decide on a case-by-case basis).

PR #384, opened 2020-08-04
Add .style.yapf config that matches the Google internal style guide.

Marvin182:
Unfortunately the Google internal formatting is different from PEP8
(column limit 79 -> 80 and indent width 4 -> 2, a few things around when
and how to split lines). YAPFs own style matches this but is not the default.

YAPF will automatically pick up the config when run anywhere inside the
repository.

### DIFF
--- a/.style.yapf
+++ b/.style.yapf
@@ -1,0 +1,2 @@
+[style]
+based_on_style: yapf


.style.yapf
```diff
@@ -0,0 +1,2 @@
+[style]
+based_on_style: yapf
```
levskaya, line 2: huh. is it really this simple a config file?

Marvin182: Yes, I think our internal configs for Python formatting corresponded to the "chromium" config in  yapf and "chromium" was recently renamed to "yapf".

avital:
I think this is great! A few questions:

1. Is yapf both a formatter and a linter?
2. Does it integrate with common IDEs like VSCode?
3. Is the plan to add yapf to some kind of pre-commit hook?

Marvin182:
> I think this is great! A few questions:
> 
> 1. Is yapf both a formatter and a linter?
just a formatter. It can fix some things that a linter might complain about but not all.

We probably want 3 things:
1) Linter. I don't use any.
2) Formatter that makes it a bit easier to be conform with the linter and have consistent style.
3) Type checking (internally we use pytype, I think JAX uses mypy)

> 2. Does it integrate with common IDEs like VSCode?
Yes, Dan is using it with VSCode

> 3. Is the plan to add yapf to some kind of pre-commit hook?
No, I think a linter and type checking would be the way to go

avital:
Wouldn't we want to make sure all commits adhere to the same style guide? That way blame always works because you don't catch other people just reformatting a file that they are editing.

PR #402, opened 2020-08-11
seq2seq example in the Linen API

marcvanzee:
Anselm managed to get `transforms.scan` working in this PR, very cool!

I also added a fix related to Jax omnistaging to `jax_utils.py`, which is in the master branch but not in linen2.

### DIFF
--- a/examples/seq2seq/train.py
+++ b/examples/seq2seq/train.py
@@ -19,7 +19,7 @@ import random
 from absl import app
 from absl import flags
 from absl import logging
-from functools import partial
+import functools
 
 from flax import jax_utils
 from flax import linen as nn
@@ -30,8 +30,11 @@ from typing import Any, Callable, Tuple
 
 import jax
 import jax.numpy as jnp
+from jax.config import config
+config.enable_omnistaging()
 
 import numpy as np
+
 
 FLAGS = flags.FLAGS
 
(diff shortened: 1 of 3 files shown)


examples/seq2seq/train.py
```diff
@@ -305,7 +319,7 @@ def get_examples(num_examples):
 def get_batch(batch_size):
   """Returns a batch of example of size @batch_size."""
   inputs, outputs = zip(*get_examples(batch_size))
-
+  #print(inputs[0], '|', outputs[0])
```
levskaya, line 322: perhaps remove this old debugging line?

flax/jax_utils.py
```diff
@@ -117,7 +117,8 @@ def partial_eval_by_shape(fn, input_spec, *args, **kwargs):
   if config.omnistaging_enabled:
     _, out_pvals, _ = pe.trace_to_jaxpr(f_flat, in_pvals)
   else:
-    _, out_pvals, _ = pe.trace_to_jaxpr(f_flat, in_pvals, stage_out=True)
+    with jax.core.initial_style_staging():
```
levskaya, line 120: thanks for adding this, we'll also do a review of all recent merges into master's old "nn" lib and rebase onto master again as well to make sure we don't miss any important commits.

marcvanzee: Sounds good!

PR #410, opened 2020-08-14
Fix broken links to the readme.md of all examples

ludgerpaehler:
Links for examples pointed to <..>/example/example/.., which lead to an "page not found" error. Links were fixed to point to the correct URLs again.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,7 +13,7 @@ Each examples is designed to be self-contained and easily forkable, while reprod
 
 | Example | Description | Features |
 | ------- | ----------- | -------- |
-| [cifar10](examples/cifar10/README.md) | Wideresnet & pyramidnet implementations with shake-shake / shake-drop regularization for CIFAR-10 classification | Single host SPMD, tfds `cifar10`, custom preprocessing |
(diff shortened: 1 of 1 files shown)


ludgerpaehler:
@googlebot I signed it!

examples/README.md
```diff
@@ -14,14 +14,14 @@ Each examples is designed to be self-contained and easily forkable, while reprod
 | Example | Description | Features |
 | ------- | ----------- | -------- |
 | [cifar10](examples/cifar10/README.md) | Wideresnet & pyramidnet implementations with shake-shake / shake-drop regularization for CIFAR-10 classification | Single host SPMD, tfds `cifar10`, custom preprocessing |
```
avital, line 16: I think you missed this?

avital:
Thanks! One comment

examples/README.md
ludgerpaehler: I did.. - my bad. The updated commit should now fix this as well.

levskaya:
@ludgerpaehler thanks so much for alerting us to this problem.  When investigating why it was happening I discovered that it was due to a mis-configured internal<-->external repo-sync script.  When I fixed this it automatically applied the same fixes that you have here as PR #413 .  (I hope you'll forgive me for fixing it this way instead of merging this PR first, but there was a complicated sequencing issue that required me to do the config fix first.)  I'll close this issue for now, but thank you again so much for pointing out the problem to us!

PR #411, opened 2020-08-14
Linen2 template

jheek:


avital:
Do all tests pass? I don't remember if we added tests for init/apply

PR #415, opened 2020-08-17
add transform tests.

levskaya:
Added some missing basic tests for transforms, removed now-pointless Module class default for setup fn.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -27,8 +27,8 @@ import numpy as np
 
 from flax import traverse_util
 from flax import serialization
-from flax.core import Scope, apply, init
-from flax.core.scope import _unfreeze_variables, Variable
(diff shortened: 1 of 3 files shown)


flax/linen/module.py
```diff
@@ -341,17 +348,9 @@ def __post_init__(self):
...
-
-  def setup(self):
-    """Called when Module instance receives variables and PRNGs.
-
-    If you want to use PRNGs and/or read or write variables during module
-    instance construction, override this method. It will be automatically
-    called from the dataclass `__post_init__`.
-    """
-    pass
+    # Call the user-defined initialization setup() function.
```
avital, line 351: I don't like this change. Where would we put the docstring for setup()?

levskaya: restored it.

PR #416, opened 2020-08-17
PixelCNN++ In Linen API

marcvanzee:
(Creating a new PR for this since branch `linen2` is merged with master. Old PR: https://github.com/google/flax/pull/408)

With lots of help from @j-towns, thanks!

Side-by-side diff of change: https://github.com/marcvanzee/flax/compare/linen2..pixelcnn2


linen_examples/pixelcnn/train.py
```diff
@@ -0,0 +1,298 @@
...
+  # pmap the train and eval functions
+  p_train_step = jax.pmap(
+      partial(train_step, learning_rate_fn=learning_rate_fn), axis_name='batch')
+  p_eval_step = jax.pmap(eval_step, axis_name='batch')
+
+  # Gather metrics
+  train_metrics = []
+
+  for step, batch in zip(range(step_offset, num_steps), train_iter):
+    logging.info('Train step %d', step)
```
j-towns, line 239: Maybe remove this?

linen_examples/pixelcnn/sample.py
```diff
@@ -0,0 +1,126 @@
...
+  rng_mix, rng_logistic = random.split(rng)
+  # Add channel dimension to one-hot mixture indicator
+  mix_indicator = _categorical_onehot(rng_mix, logit_probs)[..., jnp.newaxis]
+  # Use the mixture indicator to select the mean and inverse scale
+  mean      = jnp.sum(means      * mix_indicator, -4)
+  inv_scale = jnp.sum(inv_scales * mix_indicator, -4)
+  sample = mean + random.logistic(rng_logistic, mean.shape) / inv_scale
+  return snap_to_grid(sample)
+
+@partial(jax.pmap, static_broadcasted_argnums=1)
```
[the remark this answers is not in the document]
j-towns: > This means you will recompile for each set of parameters.

That doesn't appear to be happening, I wonder if that's because the static_broadcasted args are hashed on their `id`, and I don't think they're treated as pytrees, and the model is stateful/mutable? I could be wrong, this is just a guess :).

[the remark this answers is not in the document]
j-towns: Ah sorry my bad, I think `sample.py` hasn't been tested and you're probably right that it's broken as you describe (I thought that the comment was on `train.py`).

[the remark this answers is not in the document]
j-towns: Actually, I was wrong both times: because this is in the sampling code, not training, it _is_ ok to treat the model as static.

j-towns:
I think the eval is happening on the model parameters, but should be happening on the `ema` (exponential moving average) parameters. In the old version this was done by these lines:
https://github.com/google/flax/blob/1f4b841af5c00b573d3de6a71ce82fd18f73b08d/examples/pixelcnn/train.py#L269

https://github.com/google/flax/blob/1f4b841af5c00b573d3de6a71ce82fd18f73b08d/examples/pixelcnn/train.py#L276

that might explain why we're seeing 'rougher' eval curves during training.

linen_examples/pixelcnn/train.py
marcvanzee: Done.

linen_examples/pixelcnn/model_test.py
```diff
@@ -0,0 +1,142 @@
...
+
+import pixelcnn
+
+
+class ModelTest(absltest.TestCase):
+
+  def test_conv(self):
+    rng = random.PRNGKey(0)
+    x = np.arange(24).reshape(1, 4, 3, 2)
+    conv_module = pixelcnn.Conv.partial(features=4, kernel_size=(3, 2))
```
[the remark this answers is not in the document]
marcvanzee: I've updated the tests, but now two of them are failing:
```
FAILED linen_examples/pixelcnn/model_test.py::ModelTest::test_conv_down_right - AssertionError: 
FAILED linen_examples/pixelcnn/model_test.py::ModelTest::test_conv_transpose_down - AssertionError: 
```

The problem is that the mean and variance are no longer resp. 0 and 1. Example of an assertion error:
```
>     onp_testing.assert_allclose(np.mean(out, (0, 1, 2)), 0., atol=1e-5)
E     AssertionError: 
E     Not equal to tolerance rtol=1e-07, atol=1e-05
E     
E     Mismatched elements: 4 / 4 (100%)
E     Max absolute difference: 0.15736276
E     Max relative difference: inf
E      x: array([ 0.021286, -0.157363, -0.045011,  0.001434], dtype=float32)
E      y: array(0.)
```

Perhaps this is because the logic is a bit different now: in the old API we only init the model and we get an output, but in the new API we first init it with zeros, and then apply it with the input data. Do you think this may be the problem?

avital:
Hi @marcvanzee, we're looking through this week's Pull Requests.

Looks like this is almost ready to go but there are a few last comments that need to be acted on, is that right? Do you have a sense for when you'll be able to finish the last few issues before we can merge it?

(I know you're focusing on showcasing automasking now so I'm not proposing we push on this, I was curious about your plan)

avital:
(Oh, I just saw commits in the past few hours so you seem to actively be working on this!)

linen_examples/pixelcnn/model_test.py
```diff
@@ -0,0 +1,154 @@
...
+
+    self.assertEqual(direction.shape, (3, 2, 2, 4))
+    self.assertEqual(scale.shape, (4,))
+    self.assertEqual(bias.shape, (4,))
+    self.assertEqual(out.shape, (1, 2, 2, 4))
+    self.assert_close_to_zero(out)
+
+
+  def test_conv_down(self):
+    module_fn = lambda: pixelcnn.ConvDown(features=4)
```
avital, line 75: I'm curious why we need to use lambdas here. I think it's unnecessary and we can create instances of `pixelcnn.ConvDown(features=4)` and pass that in to `init_module` (and change `init_module` and `apply_module`) to not call the function passed in, instead to accept a module instance.

linen_examples/pixelcnn/model_test.py
```diff
@@ -0,0 +1,154 @@
...
+    params = self.init_module(module_fn)
+    out = self.apply_module(module_fn, params)
+    weightnorm_params = params['Conv_0']['weightnorm_params']
+    direction, scale, bias = self.get_weightnorm(weightnorm_params)
+
+    self.assertEqual(direction.shape, (2, 3, 2, 4))
+    self.assertEqual(scale.shape, (4,))
+    self.assertEqual(bias.shape, (4,))
+    self.assertEqual(out.shape, (1, 4, 3, 4))
+    self.assert_out_close_to_zero(out)
```
avital, line 85: Per your other comment about this test failing, you can change the logic here to use `module.init_and_output` that reproduces the behavior we had before.

That being said I'm not sure why this would have behaved differently, it would be a bit slower than taking the output during initialization but shouldn't lead to different outcomes.

avital: Ah yes I see. In this code you initialize with a different batch of data than what you predict with. So indeed this assertion should fail in this case. Data dependent init only guarantees that the mean and variance are set to 0 and 1 respectively for the same initial batch that you initialized with.

I think if you replace `setUp` so that `self.x` and `self.init_input` are the same it should work, I think (why is there a separation anyways?). But also probably good to use `module.init_and_output` that reproduces what we originally had.

linen_examples/pixelcnn/pixelcnn.py
```diff
@@ -0,0 +1,449 @@
...
+    # Note init_scale=0.1 on this layer was not in the original implementation,
+    # but seems to make training more stable.
+    return ConvOneByOne(10 * self.logistic_components, 
+                        init_scale=0.1)(nn.elu(down_right))
+
+
+# General utils
+def centre(images):
+  """Mapping from {0, 1, ..., 255} to {-1, -1 + 1/127.5, ..., 1}."""
+  return images / 127.5 - 1
```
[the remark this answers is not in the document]
marcvanzee: Ah it is not even used it. Removed it, nice catch!

linen_examples/pixelcnn/pixelcnn.py
```diff
@@ -0,0 +1,449 @@
...
+
+def _make_kernel(direction, scale):
+  """Maps weightnorm parameterization (direction, scale) to standard
+  parameterization. The direction has shape (spatial..., in_features, 
+  out_features), scale has shape (out_features,)."""
+  return scale * _l2_normalize(direction)
+
+
+# 2D convolution Modules with weightnorm
+class Conv(nn.Module):
```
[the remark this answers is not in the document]
marcvanzee: I agree Conv is not very informative, but ConvWN is also not very clear. I've renamed it to ConvWeightNorm now.

linen_examples/pixelcnn/pixelcnn.py
```diff
@@ -0,0 +1,449 @@
...
+
+  @nn.compact
+  def __call__(self, inputs):
+    out_h, out_w = onp.multiply(self.strides, inputs.shape[1:3])
+    return ConvTranspose(self.features, self.kernel_size,
+                         self.strides)(inputs)[:, :out_h, :out_w]
+
+
+class GatedResnet(nn.Module):
+  conv_module: Callable[..., Any] = None
```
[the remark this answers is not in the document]
marcvanzee: Good point, removed it.

linen_examples/pixelcnn/sample.py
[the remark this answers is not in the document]
marcvanzee: Ok so assuming Jamie is correct I'm resolving this :-).

linen_examples/pixelcnn/model_test.py
marcvanzee: Good point, updated.

linen_examples/pixelcnn/model_test.py
marcvanzee: Thanks for the tip! I'm now using `Module.init_with_output`. The code looks nicer but the two tests are still failing. I'm happy to dig deeper into this, but maybe you can take a quick look and see whether you think I'm using it correctly?

marcvanzee:
> I think the eval is happening on the model parameters, but should be happening on the `ema` (exponential moving average) parameters. In the old version this was done by these lines:
> https://github.com/google/flax/blob/1f4b841af5c00b573d3de6a71ce82fd18f73b08d/examples/pixelcnn/train.py#L269
> 
> https://github.com/google/flax/blob/1f4b841af5c00b573d3de6a71ce82fd18f73b08d/examples/pixelcnn/train.py#L276
> 
> that might explain why we're seeing 'rougher' eval curves during training.

Thanks Jamie! I ran a new experiment and the results look much more stable now. The loss is still a bit worse than on GPU, but the difference is very small, so I think it is acceptable. TPU does take much longer to run though, but I guess that could be optimized in some way which is probably beyond the scope of this PR.

linen_examples/pixelcnn/model_test.py
marcvanzee: To clarify, `test_conv_down_right` and `test_conv_transpose_down` are still failing. Example error:

```
>     onp_testing.assert_allclose(np.mean(out, (0, 1, 2)), 0., atol=1e-5)
E     AssertionError: 
E     Not equal to tolerance rtol=1e-07, atol=1e-05
E     
E     Mismatched elements: 4 / 4 (100%)
E     Max absolute difference: 0.13608912
E     Max relative difference: inf
E      x: array([ 0.047518, -0.103525,  0.02776 , -0.136089], dtype=float32)
E      y: array(0.)
```

linen_examples/pixelcnn/model_test.py
```diff
@@ -0,0 +1,136 @@
...
+  def assert_mean_and_variance(self, out):
+    # Weightnorm should ensure that, at initialization time, the outputs of the
+    # module have mean 0 and variance 1 over the non-feature dimensions.
+    onp_testing.assert_allclose(np.mean(out, (0, 1, 2)), 0., atol=1e-5)
+    onp_testing.assert_allclose(np.var(out, (0, 1, 2)), 1., atol=1e-5)
+
+
+  def test_conv(self):
+    model = pixelcnn.ConvWeightNorm(features=4, kernel_size=(3, 2))
+    out, params = model.init_with_output(self.rng, self.x)
```
avital, line 52: `params` should probably be named `variables`

marcvanzee: Ok, updated everywhere.

linen_examples/pixelcnn/model_test.py
marcvanzee: The tests are passing now! There was a bug in the weight normalization.

linen_examples/pixelcnn/model_test.py
```diff
@@ -0,0 +1,134 @@
...
+  def test_conv_transpose_down(self):
+    model = pixelcnn.ConvTransposeDown(features=4)
+    out, variables = model.init_with_output(self.rng, self.x)
+    params = variables["param"]["ConvWeightNorm_0"]["weightnorm_params"]
+    direction, scale, bias = self.get_weightnorm(params)
+
+    self.assertEqual(direction.shape, (2, 3, 2, 4))
+    self.assertEqual(scale.shape, (4,))
+    self.assertEqual(bias.shape, (4,))
+    self.assertEqual(out.shape, (1, 8, 6, 4))
```
avital, line 111: Looks like you accidentally removed the test for actual mean and stddev here?

avital:
@marcvanzee is re-running after fixing the weight norm init scale from 0.1 back to 1.0, but I'll merge this in the meanwhile and we can make incremental improvements afterwards.

linen_examples/pixelcnn/model_test.py
marcvanzee: Actually it wasn't meant to be in the test, so remove it is correct.

avital (approved):

PR #419, opened 2020-08-20
Add unittest for LM1B example.

mohitreddy1996:
* Being done as part of #176 

* Minor refactoring in train.py. This will be helpful in future refactoring as proposed in #231 . Also it would be easier to replace the Flag values with dictionary like container.

* The unit test has a total time of ~40 seconds on local CPU machine.

### DIFF
--- a/examples/lm1b/train.py
+++ b/examples/lm1b/train.py
@@ -109,7 +109,7 @@ flags.DEFINE_bool(
 
 flags.DEFINE_integer(
     'checkpoint_freq', default=10000,
-    help='Whether to restore from existing model checkpoints.')
+    help='Save a checkpoint every these number of steps. Default=10000.')
 
 flags.DEFINE_integer(
     'random_seed', default=0,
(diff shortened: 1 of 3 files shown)


avital (approved):

examples/lm1b/train.py
```diff
@@ -324,32 +326,46 @@ def tohost(x):
...
-    raise app.UsageError('Too many command-line arguments.')
-
+def train_and_evaluate(
+  random_seed, batch_size, learning_rate, num_train_steps, num_eval_steps,
+  eval_freq, max_target_length, max_eval_target_length, weight_decay, data_dir,
+  model_dir, restore_checkpoints, save_checkpoints, checkpoint_freq,
+  max_predict_token_length, sampling_temperature, sampling_top_k, prompt_str):
+  """Executes model training and evaluation loop.
+  
+  Args:
```
andsteing, line 336: Could we instead use the newly open-sourced `ml_collections.ConfigDict` ?

avital: @mohitreddy1996 has been refactoring all of our examples, enabling unit tests. My vote is to decouple this process with the adoption of ConfigDict. (And then refactoring the newly written unit tests becomes in itself part of the ConfigDict-adoption change)

PR #421, opened 2020-08-20
Add unittest for WMT example.

mohitreddy1996:
* Being done as part of #176 

* Refactor train.py: Introduce train_and_evaluate method with flags as arguments. This is originally proposed in #231 Further refactoring to align the example to guidelines in #231 will be done as a follow up.
* Add tensorflow-text and sentencepiece as test and examples requirements.
* input_pipeline should break out if enough chars are not accepted - This is required in the unit test as data is small.
* Remove unused flags.

### DIFF
--- a/examples/wmt/train.py
+++ b/examples/wmt/train.py
@@ -81,10 +81,6 @@ flags.DEFINE_string(
     'eval_dataset_name', default='wmt14_translate/de-en:test',
     help='Optional name of TFDS translation dataset to use for evaluation.')
 
-flags.DEFINE_bool(
-    'reverse_translation', default=False,
-    help='Reverse the direction of translation.')
-
 flags.DEFINE_integer(
     'batch_size', default=256,
     help='Per host batch size for training.')
@@ -160,14 +156,6 @@ flags.DEFINE_integer(
 flags.DEFINE_integer(
     'num_heads', default=16,
     help='Number of attention heads.')
-
-flags.DEFINE_float(
-    'dropout_rate', default=0.1,
-    help='Dropout rate.')
-
-flags.DEFINE_float(
-    'attention_dropout_rate', default=0.1,
-    help='Attention dropout rate.')
 
 flags.DEFINE_integer(
     'random_seed', default=0,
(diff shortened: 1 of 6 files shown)


avital (approved):

examples/wmt/train.py
```diff
@@ -497,13 +485,53 @@ def tohost(x):
...
+  random_seed, jax_backend_target, model_dir, data_dir, vocab_path, vocab_size,
+  dataset_name, eval_dataset_name, batch_size, beam_size, eval_frequency,
+  num_train_steps, num_eval_steps, learning_rate, warmup_steps,
+  label_smoothing, weight_decay, max_target_length, max_eval_target_length,
+  max_predict_length, emb_dim, num_heads, num_layers, qkv_dim, mlp_dim,
+  share_embeddings, logits_via_embedding, use_bfloat16, restore_checkpoints,
+  save_checkpoints, checkpoint_freq):
+  """Executes model training and evaluation loop.
+  
+  Args:
```
andsteing, line 498: Could we instead use the newly open-sourced `ml_collections.ConfigDict` ?

mohitreddy1996: I think it will be better to decouple the changes. Similar thoughts by Avital on #419

andsteing: Decoupling SGTM. I'm not a big fan of these huge signatures. If we refactored the examples first to use `ConfigDict` and then added the unit tests, we would never have to introduce them. Feel free to decide yourself in which order you prefer to do the two changes...

mohitreddy1996: Yeah agreed. My thinking behind doing this before using ConfigDicts was that we will have a unit test in place as a sanity check for the transition from flags to configs. WDYT? 

I was leaning towards this solution unit test -> follow up PR with ConfigDicts.

avital:
This is for the pre-Linen examples, so I'll close this PR.

Looks like the Linen WMT example has a test now: https://github.com/google/flax/blob/master/linen_examples/wmt/train_test.py

PR #423, opened 2020-08-24
Add unittest for Cifar10 example. Follow up of #414

mohitreddy1996:
This is a follow up to PR #414 

* Being done as part of #176
* Refactor train.py: Introduce train_and_evaluate method with flags as arguments. This is originally proposed in #231 Further refactoring to align the example to guidelines in #231 will be done as a follow up.

### DIFF
--- a/examples/cifar10/input_pipeline.py
+++ b/examples/cifar10/input_pipeline.py
@@ -59,9 +59,6 @@ def augment(image, crop_padding=4, flip_lr=True):
 
 class CIFAR10DataSource(object):
   """CIFAR-10 data source."""
-  TRAIN_IMAGES = 50000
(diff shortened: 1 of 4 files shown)


avital:
LGTM (you may need to rebase your branch or recreate it after #414 is merged)

avital (approved):

avital:
(Though we'll soon remove much of the code here in CIFAR and replace with https://github.com/google/flax/pull/394/files -- @andsteing can advise)

andsteing:
+1 on what Avital said : The repo linked in #394 seems identical (but somewhat more verbose) to the current CIFAR10 example, so I think it would be better to remove it from `examples/` given that we already added the link to `examples/README.md`.

PR #424, opened 2020-08-24
Add unittest for PixelCNN example.

mohitreddy1996:
* Being done as part of #176 and #231 
* Configure HPs using `ml_collections.ConfigDict`.


examples/pixelcnn/configs/default.py
```diff
@@ -0,0 +1,42 @@
...
+  config = ml_collections.ConfigDict()
+
+  config.learning_rate = 0.001
+  config.lr_decay = 0.999995
+  config.init_batch_size = 16
+  config.batch_size = 64
+  config.num_epochs = 200
+  config.dropout_rate = 0.5
+
+  config.rng = 0
```
andsteing, line 31: Maybe rename it to `random_seed` ?

examples/pixelcnn/configs/default.py
```diff
@@ -0,0 +1,42 @@
...
+
+  config.n_resnet = 5
+  config.n_feature = 160
+  config.n_logistic_mix = 10
+
+  config.polyak_decay = 0.9995
+
+  config.num_train_steps = -1
```
andsteing, line 39: Maybe add a comment how this relates to `num_epochs`

(See also comment below in `train.py`)

examples/pixelcnn/train.py
```diff
@@ -144,7 +109,7 @@ def loss_fn(model):
   optimizer = optimizer.apply_gradient(grad, learning_rate=lr)
 
   # Compute exponential moving average (aka Polyak decay)
-  ema_decay = FLAGS.polyak_decay
+  ema_decay = polyak_decay
```
andsteing, line 112: Consider using `polyak_decay` directly, or renaming the config attribute to `ema_decay`.

examples/pixelcnn/train.py
```diff
@@ -168,40 +133,51 @@ def _prepare(x):
...
+  Args:
+    config: Hyperparameter configuration for training and evaluation.
+    model_dir: Directory to store model data.
+  """
   if jax.host_count() > 1:
     raise ValueError('PixelCNN++ example should not be run on more than 1 host'
                      ' (for now)')
 
+  # Make sure tf does not allocate gpu memory.
+  tf.config.experimental.set_visible_devices([], 'GPU')
```
andsteing, line 160: Could we instead move this to the `main()` and `test.setUp()`?
See #472

examples/pixelcnn/train.py
```diff
@@ -210,32 +186,46 @@ def train(pcnn_module, model_dir, batch_size, init_batch_size, num_epochs,
...
-  steps_per_epoch = data_source.TRAIN_IMAGES // batch_size
-  steps_per_eval = data_source.EVAL_IMAGES // batch_size
+  if config.num_train_steps == -1:
+    steps_per_epoch = (
+      data_source.info.splits['train'].num_examples // config.batch_size
+    )
+  else:
+    steps_per_epoch = config.num_train_steps
+
+  if config.num_eval_steps:
```
andsteing, line 196: should it be `if config.num_eval_steps == -1:` ?

examples/pixelcnn/train.py
```diff
@@ -210,32 +186,46 @@ def train(pcnn_module, model_dir, batch_size, init_batch_size, num_epochs,
...
 
   # Compute steps per epoch and nb of eval steps
-  steps_per_epoch = data_source.TRAIN_IMAGES // batch_size
-  steps_per_eval = data_source.EVAL_IMAGES // batch_size
+  if config.num_train_steps == -1:
+    steps_per_epoch = (
+      data_source.info.splits['train'].num_examples // config.batch_size
+    )
+  else:
+    steps_per_epoch = config.num_train_steps
```
andsteing, line 194: So it is not possible to have `num_train_steps > num_train_examples` ?
I would have assumed that you can specify e.g. `num_train_steps = 2 * num_train_examples` which would result in the same as `num_epochs = 2`.

(I think that's also what I have seen in similar configurations in other projects, though AFAIK we don't use this in Flax examples yet.)

examples/pixelcnn/train_test.py
```diff
@@ -0,0 +1,49 @@
...
+  """Test cases for train."""
+
+  def test_train_and_evaluate(self):
+    """Tests pixelcnn training and evaluation loop."""
+    # Create a temporary directory where tensorboard metrics are written.
+    model_dir = tempfile.mkdtemp()
+
+    # Go two directories up to the root of the flax directory.
+    flax_root_dir = pathlib.Path(__file__).parents[2]
+    data_dir = str(flax_root_dir) + '/.tfds/metadata'
```
andsteing, line 39: Wouldn't it be better to read this from local path `'./.tfds/metadata'` ?

In case tests are run in Github action etc : it's the same, since they're run from the Flax root directory.

In case examples are downloaded independently / forked : There wouldn't be a `../../.tfds/metadata` directory; the test should rather complain with an error message how to get these files - not sure if this is an important usecase, though.

pytest.ini
```diff
@@ -3,7 +3,6 @@
 # UserWarning is due to statement: tensorflow.compat.v2.io import gfile
 # DeprecationWarning is due to statement: tensorflow.compat.v2.io import gfile
 # inspect.getargspec() is invoked in tensorboard/backend/event_processing/event_file_loader.py:61
-# jnp.sum([]) in examples/sst2/train.py:L152 should be replaced with an ndarray.
```
andsteing, line 6: Is this change related?
(It's fine to keep it, just wanted to make sure it's on purpose)

pytest.ini
```diff
@@ -13,6 +12,5 @@ filterwarnings =
     ignore:can't resolve package from.*:ImportWarning
     ignore:the imp module is deprecated.*:DeprecationWarning
     ignore:inspect.getargspec() is deprecated since Python 3.0.*:DeprecationWarning
-    ignore:jax.numpy reductions won't accept lists and tuples in future versions, only scalars and ndarrays.*:FutureWarning
```
andsteing, line 16: ditto.

pytest.ini
```diff
@@ -13,6 +12,5 @@ filterwarnings =
...
     ignore:tostring() is deprecated. Use tobytes() instead.*:DeprecationWarning
-    ignore:Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated.*:DeprecationWarning
```
andsteing, line 18: ditto.

andsteing (changes requested):

examples/pixelcnn/configs/default.py
mohitreddy1996: Done.

examples/pixelcnn/configs/default.py
mohitreddy1996: Done, PTAL.

examples/pixelcnn/train.py
mohitreddy1996: Done.

examples/pixelcnn/train.py
mohitreddy1996: Done made the changes, thanks for pointing out. 

Just curious, why can't we have this in `train_and_evaluate`? Since the unit tests, benchmark and even the main would just call that method.

examples/pixelcnn/train.py
mohitreddy1996: Done, thanks!

pytest.ini
mohitreddy1996: Sorry I think this got added back in one of the PRs which had merge conflicts and I accidentally did a bad merge.

examples/pixelcnn/train_test.py
mohitreddy1996: So the idea behind this having it this was to have:
- TFDS metadata is downloaded exclusively for these unit tests (we download it automatically in `run_all_tests.sh`) and we place it under flax/.tfds/.. If somebody runs a test individually, they would need to run `download_tfds_metadata` script. Since the downloaded content is specific to Flax, I thought may be it would be best to have in the same directory locality i.e somewhere under flax.

I think we can move this discussion to another forum may be as this will need to be changed (if we agree on something) for other unit tests as well.

examples/pixelcnn/train.py
mohitreddy1996: Hmm TBH I actually didn't think of this scenario at all.

Some thoughts:

by `num_train_examples` do you mean `data_source.info.splits['train'].num_examples`? So even with mock data and explicitly specifying `num_examples`, `data_source.info.splits['train'].num_examples` would still return the number of examples in the split information from original metadata i.e 50000 for train split. See warning here: https://www.tensorflow.org/datasets/api_docs/python/tfds/testing/mock_data. 

Also since we do `ds.repeat()` in the input_pipeline, so we can't really do `len(list(ds))`. 

So IMO this shouldn't really effect the training loop?

examples/pixelcnn/train.py
andsteing: We could also have it here, but I think it's cleaner to have this kind of "setup code" in the calling functions.
Don't benchmarks usually call the `main()` function?

examples/pixelcnn/train.py
andsteing: Yes, 'data_source.info.splits['train'].num_examples == num_train_examples` (please forgive my laziness :-)

IMO it would be great to have `config.num_train_steps` for total number of training step. This makes it easy to set a target for total number of examples. Alternatively, if `config.num_train_steps == -1`, it would use `num_train_steps = config.num_epochs * steps_per_epoch`. This would allow flexibly to define either in the config, and would also make it possible to stop after a single step in the config regardless of the value of `data_source.info.splits['train'].num_examples`.

(Otherwise, if you prefer keeping the current implementation, I would suggest to rename to `config.num_steps_per_epoch` - also, if it's only used for testing, then you might even use a single config param for this.)

examples/pixelcnn/train_test.py
andsteing: SGTM, let's move this to a separate discussion.

examples/pixelcnn/train.py
mohitreddy1996: Cool, thanks for the suggestion! Made the necessary changes.

andsteing (approved):

examples/pixelcnn/README.md
```diff
@@ -24,10 +24,26 @@ The model should run with other configurations and hardware, but was tested on t
...
+```shell
+python train.py \
+--model_dir=./pixelcnn --config=configs/default.py \
+--config.batch_size=320
 ```
-python train.py --batch_size=320
+
+#### 8 x Nvidia V100 (16GB)
+```shell
+python train.py --config=configs/default.py --config.batch_size=320
```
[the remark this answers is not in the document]
mohitreddy1996: Done, thanks for catching this!

examples/pixelcnn/README.md
```diff
@@ -24,10 +24,26 @@ The model should run with other configurations and hardware, but was tested on t
...
 
 ### How to run
-#### 8 x Nvidia V100 (16GB)
-To run training
+
+`python train.py --model_dir=./pixelcnn`
+
+#### Overriding Hyperparameter configurations
+
+MNIST example allows specifying a hyperparameter configuration by the means of
```
avital, line 32: This isn't MNIST

examples/pixelcnn/README.md
```diff
@@ -24,10 +24,26 @@ The model should run with other configurations and hardware, but was tested on t
...
+setting `--config` flag. Configuration flag is defined using
+[config_flags](https://github.com/google/ml_collections/tree/master#config-flags).
+`config_flags` allows overriding configuration fields. This can be done as
+follows:
+
+```shell
+python train.py \
+--model_dir=./pixelcnn --config=configs/default.py \
```
avital, line 40: Is `--config` mandatory?

examples/pixelcnn/README.md
```diff
@@ -24,10 +24,26 @@ The model should run with other configurations and hardware, but was tested on t
...
+--config.num_epochs=100
 ```
-python train.py --batch_size=320
+
+#### 8 x Nvidia V100 (16GB)
+```shell
+python train.py --model_dir=./pixelcnn --config=configs/default.py --config.batch_size=320
```
avital, line 46: I think we should not have `configs/default.py`. We should only have `configs/v100_x8.py` (then we can add `configs/v100.py`, ...

examples/pixelcnn/configs/testing.py
```diff
@@ -0,0 +1,39 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Testing Hyperparameter configuration."""
+
+from configs import default as default_lib
+
+
+def get_config():
```
avital, line 20: I like that we have a separate file for a testing config.

examples/pixelcnn/README.md
mohitreddy1996: Great suggestion, done!

examples/pixelcnn/README.md
mohitreddy1996: Done.

examples/pixelcnn/configs/default.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_config():
+  """Get the default hyperparameter configuration."""
+  config = ml_collections.ConfigDict()
+
+  config.learning_rate = 0.001
+  config.lr_decay = 0.999995
+  config.init_batch_size = 16
+  config.batch_size = 64
```
avital, line 27: cc @levskaya what do you think about configuration with defaults that are overriden like this, vs having less (or nothing) in a default config and specifying more explicitly in each separate config file?

avital:
This PR is on the old pre-Linen example, so I'll close it.

I see that linen_examples/pixelcnn has a test: https://github.com/google/flax/blob/master/linen_examples/pixelcnn/train_test.py#L43, is that sufficient?

PR #425, opened 2020-08-24
harmonized scan

jheek:


jheek:
Oops, Didn't see your comment so I rebased this branch already

PR #431, opened 2020-08-27
Update common_utils.py

jpuigcerver:
If the forest contains only one tree (e.g. metrics from a single batch are passed to get_metrics()), the former code would crash because jax.tree_multimap expects more than one tree.


avital:
Thanks @jpuigcerver! Can you please add a test for the originally failing case (and make sure it indeed failed)?

avital:
When does this fail? When you only report a single metric?

flax/training/common_utils.py
```diff
@@ -41,6 +41,10 @@ def onehot(labels, num_classes, on_value=1.0, off_value=0.0):
 
 
 def stack_forest(forest):
+  if not forest:
```
avital, line 44: It would help to add a docstring to this function at this point, I think.

flax/training/common_utils.py
```diff
@@ -41,6 +41,10 @@ def onehot(labels, num_classes, on_value=1.0, off_value=0.0):
...
+    return forest
+  if len(forest) == 1:
```
avital, line 46: Maybe add a comment that this code is needed to work around the limitation of tree_multimap.

jheek:
Hi, @jpuigcerver this issue has been inactive for a while. do you still intent to submit this PR?

avital:
Hi @jpuigcerver, looks like this PR has been inactive for a while so I'll close it for now. Feel free to reopen whenever you'd like to get this ready for submission

PR #441, opened 2020-08-30
Update attention layers with simpler flat forms, update wmt example.

levskaya:
refactoring wmt example in progress.

### DIFF
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -33,99 +33,78 @@ from . import initializers
 
 PRNGKey = Any
 Shape = Tuple[int]
-Dtype = Any  # this could be a real type?
+Dtype = Any
 Array = Any
 
 
-def dot_product_attention(query,
-                          key,
-                          value,
-                          dtype=jnp.float32,
-                          bias=None,
-                          axis=None,
-                          broadcast_dropout=True,
-                          dropout_rng=None,
-                          dropout_rate=0.,
-                          deterministic=False,
-                          precision=None):
+def dot_product_attention(query: Array,
+                          key: Array,
+                          value: Array,
+                          bias: Optional[Array] = None,
+                          broadcast_dropout: bool = True,
+                          dropout_rng: Optional[PRNGKey] = None,
+                          dropout_rate: float = 0.,
+                          deterministic: bool = False,
+                          dtype: Dtype = jnp.float32,
+                          precision: Optional[lax.Precision] = None):
   """Computes dot-product attention given query, key, and value.
 
   This is the core function for applying attention based on
   https://arxiv.org/abs/1706.03762. It calculates the attention weights given
-  query and key and combines the values using the attention weights. This
-  function supports multi-dimensional inputs.
+  query and key and combines the values using the attention weights.
 
+  Note: query, key, value needn't have any batch dimensions.
 
   Args:
-    query: queries for calculating attention with shape of `[batch_size, dim1,
-      dim2, ..., dimN, num_heads, mem_channels]`.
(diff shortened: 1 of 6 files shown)


flax/linen/attention.py
```diff
@@ -363,14 +250,13 @@ def __call__(self,
...
         dropout_rng=dropout_rng,
         dropout_rate=self.dropout_rate,
         broadcast_dropout=self.broadcast_dropout,
-        deterministic=self.deterministic)
+        deterministic=self.deterministic,
+        dtype=self.dtype,
+        precision=self.precision)
 
     # back to the original inputs dimensions
     out = DenseGeneral(features=features,
```
MostafaDehghani, line 262: now that things are efficient, clean and simple, don't we want to also get rid of  `DenseGeneral`s as well?

MostafaDehghani (changes requested):

flax/linen/attention.py
```diff
@@ -33,147 +33,99 @@
...
-                          broadcast_dropout=True,
-                          dropout_rng=None,
-                          dropout_rate=0.,
-                          deterministic=False,
-                          precision=None):
+def dot_product_attention(query: Array,
+                          key: Array,
+                          value: Array,
+                          bias: Optional[Array] = None,
+                          broadcast_dropout: bool = True,
```
jheek, line 44: Do you think we should get rid of the "inlined dropout" too?
One way is to make the attention function dynamic like in the design tests.

jheek (commented):
Looks good to me! I have one small question

flax/linen/attention.py
levskaya: Hmm, I think the use of DenseGeneral here and above is nice, actually.  What you often do otherwise is define two "split_heads", "join_heads" util functions that do the two reshapes... and DenseGeneral is sort of just a nice concise way to do that.  Since we used it in mlperf I also know that it performs fine.

flax/linen/attention.py
levskaya: I agree that we should decouple it more.  it's already a dynamic attribute but the expected signature of the Callable is too specialized, it should just take q,k,v,bias.  I think I'd rather do that change in a separate PR though.

flax/linen/attention.py
MostafaDehghani: Totally agree that `DenseGeneral` is quite nice.  I was just referring to a tiny bit of improvement we can get from removing it, which might not be true given the extra reshape/transpose stuff we add with "split_heads", "join_heads" (which also makes the code a bit ugly).

p.s.
Just for the record, I profiled these two cases (which is wat is happening in DenseGeneral as well, the flattening applying dot_general on on axis was slightly better than applying dot_general on multiple axis:

```pyhton
rng = random.PRNGKey(0)
rng1, rng2 = random.split(rng)

inputs_shape = (8, 32, 32, 16, 16)  # bs, h, w, heads, features
inputs = random.normal(rng1, inputs_shape)
kernel = random.normal(rng2, (1, 16, 16))

inputs_flatten = inputs.reshape((8, 32, 32, 256))
kernel_flatten = kernel.reshape((1, 256))

@jax.jit
def dot_2d(inputs, kernel):
  return jax.lax.dot_general(inputs,  kernel,
                          (((3, 4), (1, 2)), ((), ())),
                          precision=jax.lax.Precision.HIGHEST)
@jax.jit
def dot_flatten(inputs, kernel):
  return jax.lax.dot_general(inputs,  kernel,
                          (((3,), (1,)), ((), ())),
                          precision=jax.lax.Precision.HIGHEST)

```

MostafaDehghani (approved):

Issue #440, opened 2020-08-30
Use Config Dict and Config Flags to configure Flax Examples


mohitreddy1996:
Both Config Dict and Config Flags are open sourced as part of [ML Collections](https://github.com/google/ml_collections).

I will create a PR for the MNIST example. Please feel free to change other examples and communicate through comments on this issue.

mohitreddy1996:
This is already requested in #231 by @andsteing 

Closing this issue as a duplicate. 

Refer to: https://github.com/google/flax/issues/231#issuecomment-700599090 for more information.

Refer to #492 as a reference.

PR #442, opened 2020-08-31
Add ml_collection to VAE

vballoli:
This PR adds ml_collection to the VAE example in reference to this [issue](https://github.com/google/flax/issues/440)

### DIFF
--- a/examples/vae/README.md
+++ b/examples/vae/README.md
@@ -7,6 +7,12 @@ This code follows [pytorch/examples/vae](https://github.com/pytorch/examples/blo
 pip install -r requirements.txt
 python train.py --config.batch_size=batch_size --config.learning_rate=lr 
 ```
+or
+set all hyperparameters manually(see `configs/default.py`) and pass them directly
+```bash
+pip install -r requirements.txt
+python train.py --config=configs/custom.py
+```
 
 ## Examples
 
--- a/examples/vae/README.md
+++ b/examples/vae/README.md
@@ -5,7 +5,7 @@ This code follows [pytorch/examples/vae](https://github.com/pytorch/examples/blo
 
 ```bash
 pip install -r requirements.txt
-python main.py
+python train.py --config.batch_size=batch_size --config.learning_rate=lr 
 ```
 
 ## Examples
--- a/examples/vae/configs/default.py
+++ b/examples/vae/configs/default.py
@@ -1,0 +1,28 @@
+# Copyright 2020 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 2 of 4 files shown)


mohitreddy1996:
Hi, 

Thank you for the PR. 

I would prefer using `config_flags.DEFINE_config_file` instead of `config_flags.DEFINE_config_dict` as the former accepts a file path to a python file which has `get_config()` method which returns an instance of `ConfigDict`. This allows having multiple python files defining multiple configurations (say depending on the hardware on which you train your model or platform (cloud, locally)) and makes defining hyper parameters more pythonic. 

See https://github.com/google/flax/pull/463 for reference.

vballoli:
Ah, I did think of that initially, but since the number of values weren't high, I decided to go with the dict implementation. I'll revert to the file and fix the PR. Thank you for the feedback !

examples/vae/configs/default.py
```diff
@@ -0,0 +1,28 @@
...
+"""Default Hyperparameter configuration."""
+
+import ml_collections
+
+
+def get_config():
+  """Get the default hyperparameter configuration."""
+  config = ml_collections.ConfigDict()
+
+  config.learning_rate = 0.1
```
mohitreddy1996, line 24: Any reason to have the values here different than flag default values?

examples/vae/requirements.txt
```diff
@@ -1,4 +1,6 @@
 jax
+jaxlib
 flax
 tensorflow
 pillow
+git+https://github.com/google/ml_collections
```
mohitreddy1996, line 6: ml-collections is available as a pip package :) 

https://pypi.org/project/ml-collections/

examples/vae/train.py
```diff
@@ -12,45 +12,32 @@
...
 import jax
 from jax import random
 
-from flax import nn
-from flax import optim
-
 import tensorflow as tf
 import tensorflow_datasets as tfds
 
+from flax import nn
```
mohitreddy1996, line 31: I would recommend reverting this change as it is not related to the problem being solved. 

There are such linter occurrences in almost all the examples, but it would be nice to have a wider fix or follow up PR for the same.

examples/vae/README.md
```diff
@@ -5,7 +5,13 @@ This code follows [pytorch/examples/vae](https://github.com/pytorch/examples/blo
 
 ```bash
 pip install -r requirements.txt
-python main.py
+python train.py --config.batch_size=batch_size --config.learning_rate=lr
```
mohitreddy1996, line 8: Actually it would be nice to have the README consistent across examples. Could you rephrase to have something similar to: https://github.com/google/flax/blob/master/examples/mnist/README.md

mohitreddy1996 (changes requested):

mohitreddy1996:
ping :)

vballoli:
Sorry about the delay, will fix the issues in some time !!!

avital:
Hi @vballoli -- this PR has been open for a while, I'll close it for now but feel free to re-open if you'd like to see it through!

Issue #375, opened 2020-07-22
vmap-friendly Conv


levskaya:
We should make a vmap-friendly conv, i.e. a Conv that can work on single-examples (XLA's Conv assumes a batch dimension).

Perhaps something like this modified version of nn.Conv that detects the single-example case, adds a singleton batch dimension, then removes it at the end:
```python
from jax import lax
class Conv(nn.base.Module):
  def apply(self,
            inputs,
            features,
            kernel_size,
            strides=None,
            padding='SAME',
            input_dilation=None,
            kernel_dilation=None,
            feature_group_count=1,
            bias=True,
            dtype=jnp.float32,
            precision=None,
            kernel_init=nn.linear.default_kernel_init,
            bias_init=nn.initializers.zeros):
    inputs = jnp.asarray(inputs, dtype)
    # For single-example inputs, fake the batch dimension.
    single_example_conv = False
    if inputs.ndim == len(kernel_size) + 1:
      single_example_conv = True
      inputs = jnp.expand_dims(inputs, axis=0)

    if strides is None:
      strides = (1,) * (inputs.ndim - 2)

    in_features = inputs.shape[-1]
    assert in_features % feature_group_count == 0
    kernel_shape = kernel_size + (in_features // feature_group_count, features)
    kernel = self.param('kernel', kernel_shape, kernel_init)
    kernel = jnp.asarray(kernel, dtype)

    dimension_numbers = nn.linear._conv_dimension_numbers(inputs.shape)

    y = lax.conv_general_dilated(
        inputs,
        kernel,
        strides,
        padding,
        lhs_dilation=input_dilation,
        rhs_dilation=kernel_dilation,
        dimension_numbers=dimension_numbers,
        feature_group_count=feature_group_count,
        precision=precision)
    # Remove the added batch-dimension for single-example inputs.
    if single_example_conv:
      y = jnp.squeeze(y, axis=0)
    if bias:
      bias = self.param('bias', (features,), bias_init)
      bias = jnp.asarray(bias, dtype)
      y = y + bias
    return y
```

PR #444, opened 2020-09-01
Add single input conv and conv_transpose support

vballoli:
This PR adds support and relevant test cases for single example input to conv and conv-transpose. The code changes provided by @levskaya [here](https://github.com/google/flax/issues/375) is perfect and is directly adapted into this PR.


jheek (approved):
Thanks, this looks nice!

PR #453, opened 2020-09-08
Exposing more Conv flags from ConvLSTM.

mbz:
The alternative is to allow a kwargs to be passed directly into Convs. Please let me know if you think the other one is better.


jheek (approved):
Looks good to me!

PR #454, opened 2020-09-08
Use automasking in seq2seq example

marcvanzee:
Use automasking to create ragged batches for the seq2seq example. This ends up simplifying especially the Encoder significantly since we do no longer have to keep track of hidden state of all `eos_id` tokens.

In this PR, I only use automasking during training and evaluation, but we can extend this for decoding. I plan to do this in a follow-up step.

Thanks to @mattjj and @j-towns for making this great technology possible! Especially thanks to @j-towns for guiding me through all the bugs and error messages.


marcvanzee:
**Note:** the accuracy is not as good as before yet (it only reaches about 70%), so I am debugging the reason why. I expect there is some bug in padding, or in computing the loss, since the loss doesn't go below 0.2 (before it was 0.005). Moreover, decoding never seems to give a correct answer. 

Example output of current run:
```
I0908 20:25:31.590295 140558216381824 train.py:384] train step: 7200, loss: 0.2025, accuracy: 61.72
I0908 20:25:31.604599 140558216381824 train.py:345] DECODE: 81+807 = 870 (INCORRECT) correct=888
I0908 20:25:31.604734 140558216381824 train.py:345] DECODE: 51+98 = 1006 (INCORRECT) correct=149
I0908 20:25:31.604794 140558216381824 train.py:345] DECODE: 50+847 = 878 (INCORRECT) correct=897
I0908 20:25:31.604840 140558216381824 train.py:345] DECODE: 6+856 = 648 (INCORRECT) correct=862
I0908 20:25:31.604883 140558216381824 train.py:345] DECODE: 53+715 = 746 (INCORRECT) correct=768
```

linen_examples/seq2seq/train.py
```diff
@@ -130,12 +130,13 @@ def encode_onehot(batch_inputs, max_len):
...
-    if len(tokens) > max_len:
+    unpadded_len = len(tokens)
+    if unpadded_len > max_len:
       raise ValueError(f'Sequence too long ({len(tokens)}>{max_len}): \'{s}\'')
     tokens = np.pad(tokens, [(0, max_len-len(tokens))], mode='constant')
-    return onehot(tokens, CTABLE.vocab_size)
+    return onehot(tokens, CTABLE.vocab_size), unpadded_len
 
-  return np.array([encode_str(inp) for inp in batch_inputs])
+  return np.array([encode_str(inp) for inp in batch_inputs], dtype=object)
```
levskaya, line 139: What is the point of wrapping this in a `np.array(..., dtype=object)`?  Don't the below uses work just as well if this were just a list?

linen_examples/seq2seq/train.py
```diff
@@ -352,24 +302,35 @@ def compute_metrics(logits, labels):
   return metrics
 
 
+IN_SHAPES = [{'query': '(n, _)', 'answer': '(m, _)'}]
+OUT_ELEM = f'(m +- 1, {CTABLE.vocab_size})'
```
levskaya, line 306: curious: what does `m +- 1` syntax mean here?

linen_examples/seq2seq/train.py
```diff
@@ -223,31 +179,26 @@ def __call__(self, carry, x):
...
-    projection = nn.Dense(features=self.vocab_size, name='projection')
-    lstm_state, y = lstm_cell(lstm_state, x)
-    logits = projection(y)
-    predicted_tokens = jax.random.categorical(categorical_rng, logits)
-    prediction = onehot(predicted_tokens, self.vocab_size)
+    lstm_state, y = nn.LSTMCell()(lstm_state, x)
+    logits = nn.Dense(features=CTABLE.vocab_size)(y)
+    predicted_token = jax.random.categorical(categorical_rng, logits)
+    prediction = jnp.array(predicted_token == jnp.arange(CTABLE.vocab_size), 
+                           dtype=jnp.float32)
```
levskaya, line 186: just curious: why return things in one-hot form instead of just integer ids?  It's probably obvious I'm just dumb about this.

levskaya (approved):
Everything looks pretty clear to me, just a few points of curiosity.

linen_examples/seq2seq/train.py
marcvanzee: If I don't add this then I get the following warning:

```
train.py:139: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  return np.array([encode_str(inp) for inp in batch_inputs])
```

linen_examples/seq2seq/train.py
marcvanzee: This is the somewhat funny notation to denote "m-1". The decoder outputs "m-1" tokens since the input also contains the start token "=".

linen_examples/seq2seq/train.py
levskaya: but why wrap this in an ndarray at all??  why not just keep it as a list?

linen_examples/seq2seq/train.py
levskaya: ok cool that makes sense.

linen_examples/seq2seq/train.py
marcvanzee: I think the whole logic of outputting both logits and predictions is a bit funny. I think it would be easier to only output the logits, and then use them to get the predictions, if necessary. However, in this PR I tried to keep the logic mostly the same except for the automasking, so I think that would be nice as a follow-up. WDYT?

linen_examples/seq2seq/train.py
andsteing: Minor nit : Could we make it `m + -1` then? Then it's clearer that we add `-1` as opposed to inventing a new operator `+-` with mysterious semantics :)

linen_examples/seq2seq/train.py
levskaya: Oh sure, I was just curious about it!

marcvanzee: No problem! I agree with your observation :-)

linen_examples/seq2seq/train.py
marcvanzee: Ahh yes that actually works! Nice, changed it.

linen_examples/seq2seq/train.py
marcvanzee: Sure, good idea! Done.

PR #456, opened 2020-09-10
Fix frozen dict

jheek:
reimplement frozen dict to support flatten_up_to correctly.

Also moves the core tests into the right directory because tests weren't running.


flax/core/frozen_dict.py
```diff
@@ -22,50 +22,42 @@
...
-      self._leaves = None
+    xs = dict(*args, **kwargs)
+    self._dict = _freeze(xs)
     self._hash = None
     
   def __getitem__(self, key):
-    self._ensure_dict()
-    return self._dict[key]
+    v = self._dict[key]
+    if isinstance(v, dict):
```
avital, line 38: Is this necessary? Looks like _freeze already recursively freezes all subducts?

avital (approved):
Made a small suggestion

flax/core/frozen_dict.py
jheek: I added a comment in _freeze

flax/core/frozen_dict.py
```diff
@@ -74,40 +66,54 @@ def copy(self, **add_or_replace):
...
 
-def freeze(x: Dict[K, V]) -> FrozenDict[K, V]:
-  """Freeze a nested dict."""
-  if not isinstance(x, dict):
-    return x
-  temp = {}
-  for key, value in x.items():
-    temp[key] = freeze(value)
-  return FrozenDict(temp)
+def _freeze(xs: Mapping[K, V]) -> FrozenDict[K, V]:
```
avital, line 83: Ah. I see now. So this method name doesn't sound right. It's actually `_unfreeze`?

jheek: Thanks your comment made me realize `_freeze` is redundant

PR #461, opened 2020-09-14
replace kinds with collections

jheek:


flax/core/scope.py
```diff
@@ -34,12 +34,13 @@
 PRNGKey = Any
 Array = Any
 
+Filter = Union[bool, str, Sequence[str]]
+CollectionFilter = Filter
+RNGFilter = Filter
```
avital, line 39: Have you considered naming this `RNGCollectionFilter`?

flax/core/lift.py
```diff
@@ -149,21 +149,21 @@ def repack(inner_scope_tree):
...
+        for col_name, collection in out_variable_group.items():
+          for name, value in collection.items():
+            scope.put_variable(col_name, name, value)
     return y
   return wrapper
 
 id_fn = lambda x: x
 
 def transform_module(fn: Callable[..., Any],
-                     target: KindFilter = 'param',
```
avital, line 161: Should we also considering making the "standard" collections into constance, e.g. `flax.linen.PARAMS = 'params'` and `flax.linen.normalization.BATCH_STATS = 'batch_stats'`?

flax/core/lift.py
```diff
@@ -447,14 +446,14 @@ def jitted(variable_groups_xs, rng_groups_xs, *args):
...
 
 
 def remat_scan(body_fn: Callable[..., Any], scope: Scope, carry: Any,
                lengths: Sequence[int],
-               variable_carry: KindFilter = False,
-               variable_axes: Mapping[KindFilter, InOutScanAxis] = {},
-               split_rngs: Mapping[KindFilter, bool] = {}):
+               variable_carry: CollectionFilter = False,
+               variable_axes: Mapping[CollectionFilter, InOutScanAxis] = {},
+               split_rngs: Mapping[RNGFilter, bool] = {}):
```
avital, line 456: It's cool to have these indeed be different types (even if they're the same under the hood) for sake of self-documenting these functions.

flax/core/scope.py
```diff
@@ -205,21 +206,21 @@ def wrapper(*args, **kwargs):
       return fn(scope.rewound(), *args, **kwargs)
     return wrapper
 
-  def get_kind(self, kind: str, mutable: bool = False) -> MaybeFrozenKind:
-    """Returns all variable of a given kind."""
-    if kind not in self._variables:
+  def collection(self, col: str, mutable: bool = False) -> MaybeFrozenCollection:
```
avital, line 209: The return value type here is also much clearer -- before this function returned a "kind"?

avital (approved):
LGTM!

We need to also fix all of the content that's linked to from https://github.com/google/flax/blob/master/flax/linen/README.md

flax/core/lift.py
jheek: probably will do in seperate PR though.

flax/core/lift.py
jheek: I agree, there's a weird balancing act between keeping things consistent and abstract in the implementation and separated in the public api for clarity.

flax/core/scope.py
jheek: I don't think collection is a good name here (it's a prng sequence not really a collection). I have considered `PRNGSequenceFilter`, It's a bit long though.

avital: Ah oops yes I meant `PRNGSequenceFilter`. OK I trust your judgement

PR #462, opened 2020-09-14
Avoid passing python collections to jax reductions in SST2 example.

mohitreddy1996:
* Avoid passing python collections to jax reductions in SST2 example.

* Remove pytest.ini entry for the same. The example has unit test coverage which executes a single step of training and evaluation.


avital (approved):

avital:
Thank you for "leaving things better than you found them!"

PR #463, opened 2020-09-15
Use ConfigDict to represent hyperparameter configurations for MNIST example 

mohitreddy1996:
* Use `ConfigDict` and `config_flags` from `ml_collections` to represent hyper parameter configurations for MNIST example. Similar changes should be made in further examples as well.

* Add writing hparams to TB summaries support.

* Change mnist_lib_test and mnist_benchmark to reflect accepting ConfigDict instance.

Being done as part of #440 

TB.dev link for local execution: https://tensorboard.dev/experiment/wFqxlO1IROG3fpLJ5wwwBg/#scalars

### DIFF
--- a/examples/mnist/README.md
+++ b/examples/mnist/README.md
@@ -16,6 +16,16 @@ I0527 09:46:50.714237 139989059204928 mnist_lib.py:178] eval epoch: 10, loss: 0.
 
 `python mnist_main.py --model_dir=/tmp/mnist`
 
-Optionally set `--config` flag value to a python file path which returns
-a hyperparameter configuration. This usually is an instance of [ConfigDict](https://github.com/google/ml_collections/tree/master#configdict).
-By default, this is set to `configs/default.py`.
+#### Overriding Hyperparameter configurations
+
+MNIST example allows specifying a hyperparameter configuration by the means of
+setting `--config` flag. Configuration flag is defined using
+[config_flags](https://github.com/google/ml_collections/tree/master#config-flags).
+`config_flags` allows overriding configuration fields. This can be done as
+follows:
+
+```shell
+python mnist_main.py \
+--model_dir=/tmp/mnist --config=configs/default.py \
+--config.learning_rate=0.05 --config.num_epochs=5
+```
--- a/examples/mnist/README.md
+++ b/examples/mnist/README.md
@@ -14,4 +14,8 @@ I0527 09:46:50.714237 139989059204928 mnist_lib.py:178] eval epoch: 10, loss: 0.
 
 ### How to run
 
-`python mnist_main.py --config=configs/default.py --model_dir=/tmp/mnist`
+`python mnist_main.py --model_dir=/tmp/mnist`
+
+Optionally set `--config` flag value to a python file path which returns
+a hyperparameter configuration. This usually is an instance of [ConfigDict](https://github.com/google/ml_collections/tree/master#configdict).
+By default, this is set to `configs/default.py`.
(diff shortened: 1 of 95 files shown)


examples/mnist/README.md
```diff
@@ -14,4 +14,4 @@ I0527 09:46:50.714237 139989059204928 mnist_lib.py:178] eval epoch: 10, loss: 0.
 
 ### How to run
 
-`python mnist_main.py --model_dir=/tmp/mnist`
```
avital, line 17: Can we make there be a default value for `--config` so that it's not passed in explicitly every time?

flax/metrics/tensorboard.py
```diff
@@ -163,3 +165,12 @@ def write(self, tag, tensor, step, metadata=None):
           tensor=tensor,
           step=step,
           metadata=metadata)
+
+  def hparams(self, hparams):
```
avital, line 169: Cool!

howtos/diffs/checkpointing.diff
```diff
@@ -10,17 +10,17 @@ index 63af6e8..59c0d1c 100644
  
  class CNN(nn.Module):
    """A simple CNN model."""
-@@ -166,6 +167,9 @@ def train_and_evaluate(model_dir: str, num_epochs: int, batch_size: int,
+@@ -165,6 +166,9 @@ def train_and_evaluate(config: ml_collections.ConfigDict, model_dir: str):
    model = create_model(init_rng)
-   optimizer = create_optimizer(model, learning_rate, momentum)
+   optimizer = create_optimizer(model, config.learning_rate, config.momentum)
  
 +  # Restore saved model
```
avital, line 17: How hard was it to figure out how to fix the HOWTO?

avital (approved):
Much nicer this way!

examples/mnist/README.md
andsteing: +1

andsteing (approved):

examples/mnist/README.md
mohitreddy1996: Passing a default value for --config is definitely possible. But since the config paths work with relative paths based on the current directory of execution, I don't think we can fix a reasonable default value. 

i.e if someone runs the mnist example from mnist/ dir, the path will be `configs/default.py`. Let's say the Main is executed from flax/, the config path becomes `examples/mnist/configs/default.py`.

howtos/diffs/checkpointing.diff
mohitreddy1996: It took me 5 mins to figure out how the HOWTOs were generated. Read up `git apply`, then it was easy from there. 

But I had one thing to say about git apply, the error message doesn't help at all. It could have been a bit more verbose and provide few more details.

examples/mnist/mnist_benchmark.py
```diff
@@ -14,34 +14,30 @@
...
 
-  @flagsaver
   def test_cpu(self):
     """Run full training for MNIST CPU training."""
     model_dir = self.get_tmp_model_dir()
-    FLAGS.model_dir = model_dir
     start_time = time.time()
-    mnist_main.main([])
+    mnist_lib.train_and_evaluate(
+      config=config_lib.get_config(),model_dir=model_dir)
```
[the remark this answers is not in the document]
mohitreddy1996: done. thanks for the catch!

examples/mnist/README.md
andsteing: What about `os.path.join(os.path.dirname(__file__), 'configs/default.py'))` ?

howtos/diffs/checkpointing.diff
andsteing: I also updated diffs recently and tried to improve the documentation a little bit - feel free to comment additional ideas to #473

examples/mnist/README.md
mohitreddy1996: Thanks for the suggestion. Done!

andsteing (approved):

PR #468, opened 2020-09-17
Remove unnecessary shared() from seq2seq example

marcvanzee:
Also removes the method `_create_modules`


jheek (approved):

howtos/diffs/scheduled-sampling.diff
```diff
@@ -123,16 +123,10 @@ index 1597079..77449b7 100644
...
-+          optimizer, batch, 0.5, nn.make_rng())
+@@ -390,4 +402,4 @@ def main(_):
  
-     self.assertLessEqual(train_metrics['loss'], 5)
-     self.assertGreaterEqual(train_metrics['accuracy'], 0)
+ 
+ if __name__ == '__main__':
+-  app.run(main)
++  app.run(main)
+\ No newline at end of file
```
avital, line 132: Can we fix this (re-add the newline, or just remove this line from the diff file?) Otherwise this will appear in the HOWTO view

avital: Er actually you need to remove this whole block -- this part of the code shouldn't change at all in the HOWTO

marcvanzee: I removed it.

howtos/diffs/scheduled-sampling.diff
```diff
@@ -123,3 +122,22 @@ index 60b79b5..b1b8f18 100644
...
+ 
+diff --git a/examples/seq2seq/train_test.py b/examples/seq2seq/train_test.py
+index b638d55..26fc7b8 100644
+--- a/examples/seq2seq/train_test.py
++++ b/examples/seq2seq/train_test.py
+@@ -101,7 +101,7 @@ class TrainTest(absltest.TestCase):
+       model = train.create_model()
+       optimizer = train.create_optimizer(model, 0.003)
+       optimizer, train_metrics = train.train_step(
+-          optimizer, batch, nn.make_rng())
```
avital: This doesn't look right to me. Have the arguments to `train_step` changed? I don't see that in the rest of the HOWTO diff above.

avital: OH SORRY MY BAD YOU'RE RIGHT.

PR #469, opened 2020-09-17
Add linen intro notebook

jheek:


avital (approved):
I assume this is just a copy of the other Colab?

avital:
Oh, I see you also updated param -> params, is that right?

jheek:
Yes fixed the issues and re-ran all the cells

Issue #445, opened 2020-09-01
Recent TF breaks JAX on GPU by eating all memory


levskaya:
Some update to TF is again causing it to preemptively grab all memory on GPU.  This manifests with strange errors such as:
`CUBLAS_STATUS_NOT_INITIALIZED`

the workaround requires declaring explicitly to TF that GPUs are off-limits, we should add this to all of our examples:
```
import tensorflow as tf
tf.config.experimental.set_visible_devices([], "GPU")
```

PR #472, opened 2020-09-17
Examples : Hide GPUs from TensorFlow and use default v2.

andsteing:
This change hides GPUs from TensorFlow to avoid it from grabbing GPU
memory (see issue #445).

Also bumps to TensorFlow 2 has been the default for a while.

### DIFF
--- a/examples/imagenet/imagenet_lib.py
+++ b/examples/imagenet/imagenet_lib.py
@@ -197,30 +197,9 @@ def sync_batch_stats(state):
   return state.replace(model_state=avg(state.model_state))
 
 
-def train_and_evaluate(model_dir: str, batch_size: int, num_epochs: int,
-                       learning_rate: float, momentum: float, cache: bool,
-                       half_precision: bool, num_train_steps: int = -1,
-                       num_eval_steps: int = -1):
-  """Runs model training and evaluation loop.
-
-  Args:
-    model_dir: Directory where the checkpoints and tensorboard summaries
-      should be written to.
-    batch_size: Batch size of the input.
-    num_epochs: Number of epochs to cycle through before stopping.
-    learning_rate: The learning rate in case you have batch size 256.
-      The effective learning rate is scaled linearly to the batch size.
-    momentum: Momentum value for the momentum optimizer.
-    cache: Determines whether the dataset should be cached.
-    half_precision: Determines whether bfloat16/float16 should be used
-      instead of float32.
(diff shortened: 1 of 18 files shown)


levskaya (approved):
Thanks so much for fixing this!!

examples/imagenet/imagenet_lib.py
```diff
@@ -30,6 +30,7 @@
 import jax.numpy as jnp
 
 import tensorflow as tf
+tf.config.experimental.set_visible_devices([], "GPU")
```
levskaya, line 33: Realized after seeing some jax2tf test failures - We have to only call this after absl flag parsing / absl lib init happens so this should instead be put in "imagenet_main.py" and not a the example training lib file.

examples/mnist/mnist_lib.py
```diff
@@ -27,6 +27,8 @@
 
 import numpy as onp
 
+import tensorflow as tf
+tf.config.experimental.set_visible_devices([], "GPU")
```
levskaya, line 31: same comment as above - needs to go in main file, not lib file.

linen_examples/mnist/mnist_lib.py
```diff
@@ -28,6 +28,8 @@
 
 import numpy as onp
 
+import tensorflow as tf
+tf.config.experimental.set_visible_devices([], "GPU")
```
levskaya, line 32: ditto

levskaya (changes requested):
noticed that we need to move the GPU hint away from a few lib files to "main" files so that we don't break other tests that import these lib files before absl config.

examples/imagenet/imagenet_lib.py
andsteing: Thanks for highlighting this. I have moved th `tf.config.experimental.set_visible_devices([], 'GPU')` into the `*_main.py` (or `train.py`) and `_lib_test.py` files (I could have put them into the `train_and_evaluate()` methods but I thought it cleaner to leave this setup to the main/test setup).

examples/mnist/mnist_lib.py
andsteing: ditto.

linen_examples/mnist/mnist_lib.py
andsteing: ditto.

levskaya (approved):

PR #474, opened 2020-09-18
Fix scan broadcast issue

jheek:
broadcast variables were re-computed during init in each iterations causing excessive slow down.


levskaya (approved):
looks great, thanks for looking into this!

PR #475, opened 2020-09-18
Expose __version__ attribute

wrzadkow:
This PR aims at flax exposing `__version__` attribute such that users can easily check their version of flax using
```
import flax
print(flax.__version__)
```
in the same way it can be done for many Python packages, including jax. In particular, this PR follows the pattern used by jax.


avital (approved):

avital:
Thanks!

PR #486, opened 2020-09-23
Pass flag values to input pipeline and model definition correctly.

mohitreddy1996:
- We currently have a few FLAGs defined which control the behavior of input pipelines and model definitions, but they are not used instead hardcoded constants are being used.


mohitreddy1996:
Waiting on @levskaya who is working on refactoring WMT example.

avital:
We use ConfigDict instead of FLAGS now so this pull requests is oboslete.

Issue #480, opened 2020-09-21
Updating examples/imagenet


andsteing:
This issue tracks updates the `imagenet` example to follow practices outlined in #231.

- [x]  Port to linen API - once ported all subsequent changes should be done in `linen_examples/imagenet`
- [x]  Update `README.md`
- [x]  Add `requirements.txt`
- [x]  Update file structure
- [x]  Use `ml_collections.ConfigDict`
- [x]  Add benchmark test
- [x]  Add unit test for training/eval step
- [x]  Add Colab
- [x]  Adhere to Google Python style
- [x]  Add mypy annotations
- [x]  Use `clu`
- [x]  Shorten/beautify training loop

Note to above list:

- Only some type annotations

andsteing:
Note that "file structure" in #231 got changed:

> * To make it easier to test, maintain and reuse code structure the code as follows:
>   
>   * `main.py` contains the flags and calls a method from `train.py` to run the training loop. This should be the only file defining flags! This can be almost identical for all examples, please copy from `linen_examples/wmt/main.py`.
>   * `train.py` contains classes and methods for training and evaluating the model.
>   * `train_test.py` for test cases of the training code. At a minimum the test should run a single training step but more fine grained unit tests are a bonus. You can use `tfds.testing.mock_data` to avoid real data from disk.

PR #490, opened 2020-09-23
Use ml_collections.ConfigDict for configuring Hyperparameters in Imagenet example.

mohitreddy1996:
* Use ConfigDict to configure HPs.
* Introduce a single config - default.
 ** In unit test, import the default config and override the necessary fields.


mohitreddy1996:
It's probably better to make the changes directly in Linen Examples as part #480 as the number of hyper parameters for the existing Flax example are low hence defeats the purpose of ConfigDict.

Issue #479, opened 2020-09-21
Updating examples/mnist


andsteing:
This issue tracks updates the `mnist` example to follow practices outlined in #231.

- [x] Port to linen API - once ported all subsequent changes should be done in `linen_examples/mnist`
- [x]  Update `README.md`
- [x]  Add `requirements.txt`
- [x]  Update file structure
- [x]  Use `ml_collections.ConfigDict`
- [x]  Add benchmark test
- [x]  Add unit test for training/eval step
- [x]  Add Colab
- [x]  Adhere to Google Python style
- [x]  Shorten training loop

Removed list:

- Add mypy annotations - not needed.
- Use `clu` - to keep example more concise

avital:
People look at MNIST in particular for a very concise example. I think some of the items on this checklist make the code more concise, and others make the code more verbose. I think it would be good to lean on the direction of keeping this example short.

andsteing:
Good point!

I removed two points that I think make the example more complicated.

I think it's still good practice to have tests. We could discuss if `ConfigDict` is necessary (especially in this case where we have only few parameters), and whether we want to have mypy annotations (in the sake of consistency I would say yes).

PR #492, opened 2020-09-24
Add configdict for mnist linen example to configure HP.

mohitreddy1996:
- Done as part of #231 & #479 
- HP configuration and definition now matches examples/mnist.

### DIFF
--- a/linen_examples/mnist/mnist_lib.py
+++ b/linen_examples/mnist/mnist_lib.py
@@ -23,8 +23,8 @@ from absl import logging
 import jax
 from jax import random
 import jax.numpy as jnp
-from jax.config import config
-config.enable_omnistaging()
+
+jax.config.enable_omnistaging()
 
 import ml_collections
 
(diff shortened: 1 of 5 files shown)


linen_examples/mnist/mnist_lib.py
```diff
@@ -148,30 +150,30 @@ def get_datasets():
   return train_ds, test_ds
 
 
-def train_and_evaluate(model_dir: str, num_epochs: int, batch_size: int,
-                       learning_rate: float, momentum: float):
+def train_and_evaluate(hp_config: ml_collections.ConfigDict, model_dir: str):
```
andsteing, line 153: I would vote for just `config`

and instead write `jax.config.enable_omnistaging()`

andsteing (approved):

linen_examples/mnist/mnist_lib.py
mohitreddy1996: Done.

PR #499, opened 2020-09-28
design unit tests

jheek:
Makes the core design tests actual unit tests.


levskaya (changes requested):
This all looks good, but are you sure you actually ran these?  you need these lines:
```python
if __name__ == '__main__':
  absltest.main()
```
at the end of the file to get these tests to run.

jheek:
> This all looks good, but are you sure you actually ran these? you need these lines:
> 
> ```python
> if __name__ == '__main__':
>   absltest.main()
> ```
> 
> at the end of the file to get these tests to run.

Actually if you use `pytest` you don't need that (It doesn't load modules as '__main__'. But you're right it will not run the test now if you run just `python [test].py` so I'll add those

levskaya:
Ah you're right, I think there's some other test-runner that I used that actually runs things as main that bit me once a long while ago. ;)

jheek:
Actually it does turn out the files have to end with _test to be found by pytest. I fixed that now. The annoying things is pytest doesn't complain when you pass a file explicitly but doesn't find the file when you pass a directory.

levskaya (approved):

PR #503, opened 2020-09-30
Improve error message `.apply()` with invalid arguments (variable structure, unknown rngs etc)

mohitreddy1996:
Initializing a variable lazily should fail in the following scenarios:

- If the collection is frozen and if the variable doesn't exist
- If the collection is mutable, if the variable doesn't exist and there doesn't exist any RNGs associated with the variable.

Then either the variable exists or can be initialized lazily.


mohitreddy1996:
Closing PR.

Issue #489, opened 2020-09-23
Linen: Misleading "Need RNG for kind params" error in `.apply()` with wrong variable structure


avital:
If you call `.apply` on a Module with a variable dict structure that doesn't match the expected one, you get this error: `AssertionError: Need RNG for kind params`.

The reason, I believe, is that the code is saying: "oh, you defined a variable with lazy init, and I can't find that variable, so let me try to initialize it -- oops, I don't have an RNG so I can't initialize it". But the variable dict is a frozen dict anyways, so even if we did have an RNG we couldn't initialize it. So I think we should fix the check to first see if the variables are frozen (for that kind) and if so report a better error, e.g. "can't find variable FOO in kind BAR, and the kind BAR is not set to mutable". (There's probably a much better way to word that error...)

cc @marcvanzee and @rolandgvc who experienced this.

Here's a Colab reproducing the error: https://colab.research.google.com/drive/1RXv8vRI0gq7o8wHkBMHY8iN0_uHQvhqU

mohitreddy1996:
Hi Avital, 

Quick question here though. May be when we apply an empty dictionary, this is expected? 
Since providing an empty dict as a variable and since initialization of variable is mutable #504 , it will be mutable (only a FrozenDict would be considered immutable) so given the right dict of PRNGs, we could initialize the variable [1]. 

May be the error message you described is valid when the variable structure has a frozen collection?  

[1] https://github.com/google/flax/blob/9015cc26d1d4a8b086e1bffacd157f863988fc4d/flax/core/scope.py#L301

avital:
Here's how I read the logic of `apply` -- it doesn't matter if you pass in a (mutable) dict or a FrozenDict. Because the `mutable` argument overrides whatever you passed in. (Or, that's how I understand it as a naive end-user).

So I think if a user calls `module.apply({}, x, mutable=False)` then the user shouldn't have to know how `apply` is implemented -- they specified their intent -- which is "don't make anything mutable". That should be sufficient to deduce that the error raised should say "can't add a new variable to an immutable variable dict". (I don't know off the top of my head how the code would have to change -- I'd start with a failing test case and work from there).

jheek:
At head the colab now raises:
```
ScopeParamNotFoundError: No parameter named "foo" exists in "/". (https://flax.readthedocs.io/en/latest/flax.errors.html#flax.errors.ScopeParamNotFoundError)
```

So I think this issue no longer occurs

PR #504, opened 2020-09-30
Fail with a better error message on invalid variable structure in `Module.apply()`

mohitreddy1996:
#489 - if the collection of variables of a kind is immutable and the variable doesn't exist in the collection, we cannot initialize it (even if the RNGs are provided).

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -297,7 +297,15 @@ class Scope:
                *init_args) -> Variable[T]:
     """Create a Variable."""
     self.reserve(name)
-    if not self.has_variable(col, name):
+    collection = self.collection(col=col, mutable=True)
+
+    # If the collection is immutable and the variable does not exist
+    # in it, we cannot initialize it.
+    if isinstance(collection, FrozenDict) and (name not in collection):
+      raise ValueError(f"Kind: {col} is immutable and "
+                       f"variable: {name} not found in it.")
+
+    if name not in collection:
       init_value = init_fn(*init_args)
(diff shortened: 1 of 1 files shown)


avital:
Can you please add a test?

Also I'd like a second review from @jheek to see if I didn't miss any edge case.

avital:
Wow, I had PR comments from months ago that were unsubmitted :( Apologies @mohitreddy1996 -- do you think you want to get this PR updated and merged? Alternatively, if you're busy maybe best to close the PR in the meanwhile

jheek:
@mohitreddy1996 -- this PR has a merge conflict now. Could you take a look?

flax/core/scope.py
```diff
@@ -297,7 +297,15 @@ def variable(self, col: str, name: str, init_fn: Callable[..., T],
                *init_args) -> Variable[T]:
     """Create a Variable."""
     self.reserve(name)
-    if not self.has_variable(col, name):
+    collection = self.collection(col=col, mutable=True)
+
+    # If the collection is immutable and the variable does not exist
+    # in it, we cannot initialize it.
+    if isinstance(collection, FrozenDict) and (name not in collection):
+      raise ValueError(f"Kind: {col} is immutable and "
```
marcvanzee, line 305: Could you use a custom error message instead? https://github.com/google/flax/blob/master/flax/errors.py

marcvanzee (changes requested):

avital:
This PR is inactive, and the original issue (#489) has been resolved. So I'm closing it.

PR #507, opened 2020-10-02
Updates README "Getting involved".

andsteing:
The concommitant changes to the `.sh` files make sure that
`run_all_tests.sh` fails if `svn` is not installed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,18 @@ When working with large-scale input data, it is important to create large enough
 TODO: Add an example for running on Google Cloud.
 
 ## Getting involved
+
+Currently, you need to install Python 3.6 and `svn` for developing Flax. After installing these prerequisites, you can clone the repository, set up your local environement, and run all tests with the following commands:
+
+```
+git clone https://github.com/google/flax
+cd flax
+python3.6 -m virtualenv env
+. env/bin/activate
+pip install -e . .[testing]
+./tests/run_all_tests.sh
+```
+
 We welcome pull requests, in particular for those issues [marked as PR-ready](https://github.com/google/flax/issues?q=is%3Aopen+is%3Aissue+label%3A%22Status%3A+pull+requests+welcome%22). For other proposals, we ask that you first open an Issue to discuss your planned contribution.
 
 ## Note
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ TODO: Add an example for running on Google Cloud.
 
 ## Getting involved
 
-Currently, you need to install Python 3.6 and `svn` for developing Flax. After installing these prerequisites, you can clone the repository, set up your local environment, and run all tests with the following commands:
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -222,6 +222,18 @@ When working with large-scale input data, it is important to create large enough
 TODO: Add an example for running on Google Cloud.
 
 ## Getting involved
+
+Currently, you need to install Python 3.6 and `svn` for developing Flax. After installing these prerequisites, you can clone the repository, set up your local environement, and run all tests with the following commands:
```
marcvanzee, line 226:
```suggestion
Currently, you need to install Python 3.6 and `svn` for developing Flax. After installing these prerequisites, you can clone the repository, set up your local environment, and run all tests with the following commands:
```

marcvanzee, line 226: Sorry for the silly question, but why do we need svn?

marcvanzee (approved):

README.md
andsteing: For downloading tfds metadata (a directory inside the tensorflow/datasets repo).

Using git for downloading the directory would require us to clone the entire tensorflow/datasets repo (currently ~226M when cloning with `--depth=1`).

marcvanzee:
Some tests are failing, could you check if this is related to your `-e` change?

marcvanzee (changes requested):
Some tests are failing, could you check if this is related to your `-e` change?

README.md
andsteing: huh - I thought I had a spell checker installed ?!
apparently not. installed one just now :)

andsteing:
Fixed failing test, PTAL.

README.md
marcvanzee: Hmm interesting. In my VMs I never install svn and everything seems to work fine. Are you sure this is needed for all Flax development, or only if you want to use TFDS? (which is not a requirement for developing with Flax, I think)

andsteing: It's for running the examples.
If your VMs succeeded the `./tests/run_all_test.sh` then they either had a `.tfds` directory in the root or `svn` was installed on them by default.

README.md
```diff
@@ -222,6 +222,18 @@ When working with large-scale input data, it is important to create large enough
 TODO: Add an example for running on Google Cloud.
 
 ## Getting involved
+
+Currently, you need to install Python 3.6 and `svn` for developing Flax. After installing these prerequisites, you can clone the repository, set up your local environment, and run all tests with the following commands:
```
marcvanzee, line 226:
```suggestion
Currently, you need to install Python 3.6 for developing Flax, and `svn` for running the `run_all_tests.sh` script. After installing these prerequisites, you can clone the repository, set up your local environment, and run all tests with the following commands:
```

marcvanzee (approved):

PR #522, opened 2020-10-09
Create new bug template "feature request"

marcvanzee:


avital:
I actually prefer users not file feature requests as issues -- they should start with a discussion. (I believe @jheek and I also discussed this a while back). After a discussion if there's agreement on the request then we can make an issue for them but otherwise it gets odd/rude to "close" user feature requests, and really every new feature request should have a short discussion first.

jheek:
@avital I'm not sure if I follow. I think before we said people should create an issue before a PR to discuss the feature addition (like adding a new optimizer). It seems to me this template would actually help that workflow.

jheek:
Ah you mean a GH discussion. Sorry, totally misunderstood your comment. I would say an issue should be well-defined and a GH discusion can be more open-ended (read vague). In discussion we have the ideas category which captures that open-endedness pretty well I think.  We could make the template more explicit to point that issues should have a clear end goals instead of something vague like "I want to export my model to production" but something like "I want a arg for the broadcast axes in `nn.dropout`" is specific.

avital:
I could see the upside of a template if it introduced more structure, but given this pull request really just makes a template that says "describe your feature request here", I think I'd rather point people in the README to the discussions if they want to file a feature request. I'll file a PR to the README.

avital:
https://github.com/google/flax/pull/550

PR #526, opened 2020-10-11
Add DCGAN example

bkkaggle:


bkkaggle:
@googlebot I signed it!

andsteing (approved):
Thanks for adding this!

I have added "Colab" to the README to highlight that your implementation is inside a Colab.

Maybe you could update `https://github.com/bkkaggle/jax-dcgan/README` to add some explanations of your implementation?

You could also consider adding it to https://paperswithcode.com/paper/unsupervised-representation-learning-with-1#code

bkkaggle:
Hi again,

I added a little more information to the readme about the DCGAN, lmk if I should add a more detailed summary of my implementation

andsteing:
That looks good, thanks!

andsteing:
Note that when using Colab TPUs, you could distribute every batch onto 8 cores:
https://flax.readthedocs.io/en/latest/howtos/distributed-training.html

bkkaggle:
Hmm interesting, I'll add this to the notebook sometime tonight. I guess the way the code is right now, the gradients from all the cores have to be collected to a single TPU device to be optimized?

andsteing:
The code seems correct (I think results are also like in the TF example?)
If you don't specify anything, then JAX will simply use the best device to process the computations (in this case a single TPU core). If you want to make use of multiple accelerators (a Colab TPU runtime is connected to 8 cores as returned by `jax.devices()`) you need to instruct JAX how to do so. The simplest is to use the `pmap()` SPMD directive. Note that this requires you to reshape the inputs to have a leading dimension that matches the number of devices (see the HOWTO for details).

bkkaggle:
Hey, I added pmap to the notebook (https://github.com/bkkaggle/jax-dcgan/blob/main/dcgan.ipynb), I'm pretty sure I handled replicating the batchnorm context managers and splitting the rngs for each core correctly but lmk if there's something wrong with it :)

andsteing:
Hi Bilal

The code looks fine, congrats to your first data parallelization using JAX :-)

A couple of comments
- You don't need any of the `jit()` - since you wrap the function `train_steop()` with a `@pmap`, everything below will get compiled.
- We usually add `print(jax.devices())` to the first cell so readers of the notebook see how it was run.
- You're currently feeding every TPU core with a tensor of size `(8, 32, 32, 1)` - not the best use of the 8 GB HBM ... You could try increase the batch size (and change learning rate etc accordingly; probably needs some experimentation) and get to the same result in like no time at all ... Or change the dataset to something a bit more compute intensive. Or you can just leave it, but then maybe put a comment that it is a toy example and mention the per-device batch size to make this clear.

Cheers,
Andreas

bkkaggle:
Hi, thanks for the feedback. I updated the notebook with your comments

PR #529, opened 2020-10-12
Improved mutable collection handling

jheek:
Refactor collection creation to fail early when modifying an undefined collection.

Before a mutation to a non-existing non-mutable collection would just be silently ignored.


levskaya (approved):
This all looks great, sorry for the delay!

PR #531, opened 2020-10-13
Fix and simplify pixelcnn learning rate

j-towns:
On linen/master the PixelCNN++ training is crashing with
```text
TypeError: train_step() got multiple values for argument 'learning_rate_fn'
```
This is due to the way `learning_rate_fn` is fed via `functools.partial` [here](https://github.com/google/flax/blob/717d50489d46efb92fa846878d51c4c06936b7cd/linen_examples/pixelcnn/train.py#L237-L238). This pr fixes this trivial issue.

### DIFF
--- a/linen_examples/pixelcnn/train.py
+++ b/linen_examples/pixelcnn/train.py
@@ -129,7 +129,7 @@ def neg_log_likelihood_loss(nn_out, images):
   return -jnp.mean(log_likelihoods) / (jnp.log(2) * np.prod(images.shape[-3:]))
 
 
-def train_step(optimizer, ema, batch, learning_rate_fn, dropout_rng):
+def train_step(learning_rate_fn, optimizer, ema, batch, dropout_rng):
   """Perform a single training step."""
 
   def loss_fn(params):
@@ -220,8 +220,7 @@ def train():
       'params': init_rng,
       'dropout': dropout_rng
   }, init_batch)['params']
-  optimizer_def = optim.Adam(
-      learning_rate=FLAGS.learning_rate, beta1=0.95, beta2=0.9995)
+  optimizer_def = optim.Adam(beta1=0.95, beta2=0.9995)
   optimizer = optimizer_def.create(initial_variables)
 
   optimizer, ema = restore_checkpoint(optimizer, initial_variables)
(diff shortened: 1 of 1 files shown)


j-towns:
The reason that this didn't show up before _might_ be that it is Python 3.8+ specific (it's showing up for me on Python 3.8).

marcvanzee:
Ah yes I actually hit this as well and fixed it locally. Thanks for the fix!

linen_examples/pixelcnn/train.py
```diff
@@ -220,8 +220,7 @@ def train():
       'params': init_rng,
       'dropout': dropout_rng
   }, init_batch)['params']
-  optimizer_def = optim.Adam(
-      learning_rate=FLAGS.learning_rate, beta1=0.95, beta2=0.9995)
+  optimizer_def = optim.Adam(beta1=0.95, beta2=0.9995)
```
marcvanzee, line 223: Wait, why did you remove the learning rate here?

marcvanzee (changes requested):

linen_examples/pixelcnn/train.py
j-towns: Because it's over-ridden on [this line](https://github.com/google/flax/blob/4dcad3a1953f37e0a519bdefbe1867f6100d3edb/examples/pixelcnn/train.py#L144), so unnecessary to feed it in here.

j-towns:
There's also an issue when you reload from a checkpoint, I don't know if you've noticed this already. It's caused by a pytree mismatch, and I think the reason is that Flax expects all parameters to be FrozenDicts, changing the weightnorm params from dict to FrozenDict [here](https://github.com/google/flax/blob/4dcad3a1953f37e0a519bdefbe1867f6100d3edb/examples/pixelcnn/pixelcnn.py#L216) seems to fix the problem.

linen_examples/pixelcnn/train.py
marcvanzee: Oh right, makes sense.

marcvanzee:
> There's also an issue when you reload from a checkpoint, I don't know if you've noticed this already. It's caused by a pytree mismatch, and I think the reason is that Flax expects all parameters to be FrozenDicts, changing the weightnorm params from dict to FrozenDict [here](https://github.com/google/flax/blob/4dcad3a1953f37e0a519bdefbe1867f6100d3edb/examples/pixelcnn/pixelcnn.py#L216) seems to fix the problem.

No I haven't noticed that, I haven't been restoring from checkpoints for quite a while. Do you want to send a fix?

PR #534, opened 2020-10-14
Allow assignment of mixed Module pytrees in setup.

levskaya:
Previously we forced assignment of Module lists in setup() to only allow
lists containing -only- Module subclasses.  But people would like to
specify lists containing functions like nn.relu as well, and there's
absolutely no reason to force this strict convention, so it is relaxed
here.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -84,30 +84,15 @@ def disable_named_call():
 
 # Utilities for autonaming pytrees of Modules defined inside setup()
 # -----------------------------------------------------------------------------
-def is_module_tree(in_tree: Any) -> bool:
-  """Determine if in_tree is a pytree of subclasses of Module.
-
-  Args:
-    in_tree: python object, typically a python tree.
-
-  Returns:
-    False in_tree is empty or if any leaf is not a Module, True otherwise.
-  """
-  # reject trivial pytrees, {}, [], (), etc.
-  if not tree_util.tree_leaves(in_tree):
-    return False
-  reduce_fn = lambda prev, cur: prev or isinstance(cur, Module)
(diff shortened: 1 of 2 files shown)


levskaya (commented):
Errare humanum est, perseverare diabolicum

flax/linen/module.py
```diff
@@ -96,8 +96,8 @@ def is_module_tree(in_tree: Any) -> bool:
   # reject trivial pytrees, {}, [], (), etc.
```
jheek, line 96: docstring is no longer valid

flax/linen/module.py
```diff
@@ -96,8 +96,8 @@ def is_module_tree(in_tree: Any) -> bool:
...
   if not tree_util.tree_leaves(in_tree):
     return False
-  reduce_fn = lambda prev, cur: prev and isinstance(cur, Module)
-  return jax.tree_util.tree_reduce(reduce_fn, in_tree, True)
+  reduce_fn = lambda prev, cur: prev or isinstance(cur, Module)
```
jheek, line 99: Is this function really useful still? It seems like a useless extra traversal of each pytree?
Why not just tree_map every pytree in assignment and register the leaves that are modules?

jheek (changes requested):

flax/linen/module.py
levskaya: yeah this seems super dumb now, I'll try to remove this nonsense tomorrow.

flax/linen/module.py
levskaya: removed this fn.

flax/linen/module.py
levskaya: ok, I cleaned all this up

jheek (approved):

PR #535, opened 2020-10-14
Fix Module dataclass handling of inheritance.

levskaya:
The dataclass transform looks into the base class __dataclass_fields__
to set inherited dataclass args.  We were permanently mutating these
base class fields, which would screw up the handling of multiple separate
subclasses of a Module.  This fixes this edgecase.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -210,8 +210,7 @@ class Module:
     # functional transformation.  Instead of using a python metaclass, we
     # automatically transform Modules into dataclasses at subclass creation
     # time, and we set the last dataclass arguments to `parent` and `name`.
-    cls._add_parent_and_name_attrs()
-    dataclasses.dataclass(cls)
+    cls._customized_dataclass_transform()
     # We wrap user-defined methods including setup and __call__ to enforce
     # a number of different checks and to provide clear error messages.
     cls._verify_single_or_no_compact()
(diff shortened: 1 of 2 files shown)


flax/linen/module.py
```diff
@@ -221,25 +220,31 @@ def __init_subclass__(cls):
     cls.scope = None
 
   @classmethod
-  def _add_parent_and_name_attrs(cls):
-    """Add final optional dataclass attributes: `parent` and `name`."""
-    annotations = cls.__dict__.get('__annotations__', {})
+  def _customized_dataclass_transform(cls):
+    """Handle final optional dataclass attributes: `parent` and `name`."""
+    annotations = dict(cls.__dict__.get('__annotations__', {}))
```
marcvanzee, line 225:
```suggestion
    # Use cls.__dict__ to get annotations of cls itself (no parent class).
    annotations = dict(cls.__dict__.get('__annotations__', {}))
```

flax/linen/module.py
```diff
@@ -221,25 +220,31 @@ def __init_subclass__(cls):
...
       raise ValueError(
           f'properties `parent` and `name` are reserved: {annotations}')
     # Add `parent` and `name` default fields at end.
-    new_annotations = {}
-    new_annotations.update(annotations)
-    if 'parent' in getattr(cls, '__dataclass_fields__', {}):
+    # We temporarily modify base class __dataclass_fields__ to force desired
+    # argument behavior and ordering from dataclass class-transform.
+    parent_dataclass_fields = dict(getattr(cls, '__dataclass_fields__', {}))
+    if 'parent' in parent_dataclass_fields:
```
marcvanzee, line 233:
```suggestion
    # Remove 'parent' and 'name' from parents because we always want parent and
    # name to show up last in the dataclass args.
    if 'parent' in parent_dataclass_fields:
```

flax/linen/module.py
```diff
@@ -221,25 +220,31 @@ def __init_subclass__(cls):
...
+                                  Type["_Sentinel"], None]
+    cls.parent = dataclasses.field(repr=False, default=_unspecified_parent)
+    annotations['name'] = str
     cls.name = None  # default value of name is None.
+    cls.__annotations__ = annotations
+    # Now apply dataclass transform (which operates in-place).
+    dataclasses.dataclass(cls)
+    # Restore original base class __dataclass_fields__.
+    if dataclasses.is_dataclass(cls.__bases__[0]):
+     cls.__bases__[0].__dataclass_fields__ = parent_dataclass_fields
```
marcvanzee, line 247: Hmm are you sure this is correct? IIUC then `parent_dataclass_fields` contains dataclass fields of `cls` and its base class. Now you will set the dataclass fields of the base class to this, which means the base class seems to inherit the fields from the child class. That doesn't seem like the right behavior to me, or am I missing something?

marcvanzee (changes requested):

flax/linen/module.py
levskaya: When parent_dataclass_fields is grabbed and copied by `dict(getattr(cls, '__dataclass_fields__', {}))` we _know_ there is no `__dataclass_fields__` on the child, because the `@dataclass` transform hasn't been run on this class yet.  So the only fields that can show up come from direct inheritance.
(Note there's the wrinkle of trying to deal w. multiple inheritance in this code, but that is always a buggy nightmare in python and we should just explicitly forbid multiple inheritance - dataclasses themselves only barely work in a useable way w. multiple inheritance.)

marcvanzee (approved):

PR #538, opened 2020-10-16
use core reservation check in linen.

jheek:
Also solves a bug in the clone method
which was silently sharing PRNG sequences

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -169,17 +169,15 @@ class _ModuleInternalState:
   in_setup: bool = False
   last_varname: Optional[str] = None
   autoname_cursor: Optional[dict] = dataclasses.field(default_factory=dict)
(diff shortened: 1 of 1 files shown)


avital:
This looks correct to me but I don't remember why we were trying to disable the scope reservation check to begin with. @levskaya do you remember?

avital (approved):

levskaya:
Well we disabled it because we were doing our own name-reservation handling, perhaps at the time because we had to for some short-lived reason while we were prototyping things.

levskaya (approved):

levskaya (changes requested):
I think this PR needs to be rebased on master and an extra `self._state.reservations` reference needs to be removed from the updated `self.param()` fn at head.  I think that's what's causing the failing tests below.

levskaya (approved):

Issue #523, opened 2020-10-09
Add a Flax guided tour


BertrandRdp:
We want to add a guided tour of Flax, walking users through a real world workflow:
- [x] JAX basics
- [x] Core module abstraction
- [x] Storing and inspecting data
- [x] Different types of parameters
- [ ] More complicated layers like transformers
- [ ] Model and performance inspection
- [x] Creating your own layers

andsteing:
For reference : the current guided tour has the following structure (in a single notebook):

1.  JAX intro : mostly links to JAX documentation and tutorials
2.  NN in JAX without Flax : init/apply in two separate functions
3.  Flax modules : starting with an example of `Dense` and then going into more details:
    - composition
    - parameter sharing
    - shape inference
    - `Model`
    - state
4.  Optimizer : plots of sampled 2d dots, then approximating with linear regression, and then with momentum optimizer
5.  Serialization : converting objects to dictionaries of arrays, serializing with messagepack
6.  Advanced features
    - `MultiOptimizer`
    - Multiple methods
    - Transforming sub module parameters.

marcvanzee:
I'm closing this issue since I think the guided tour is actually in a great state. For the two remaining points:

- Model inspection:  see the [Model surgery HOWTO](https://flax.readthedocs.io/en/latest/howtos/model_surgery.html)
- More complicated layers: I am not sure if this is necessary for the guided tour. We can add those as HOWTOs if necessary, but right now I don't see a real need, I guess if someone would like to understand a Transformer they can read one of the many great "Annotated Transformer" tutorials.

Issue #524, opened 2020-10-09
Assigning a list of Modules to `self` in `nn.Module.setup()` is confusing


marcvanzee:
In `setup()`, submodules are expected to "know their name" when they are assigned to `self`, which is implemented by  overriding `__setattr__`. This can cause problems when appending modules to a list. Consider the following code.

```
class Test(nn.Module):
  def setup(self):
    self.layers = []
    for i in range(3):
      self.layers.append(nn.Dense(5))

  def __call__(self, x):
    for layer in self.layers:
      x = layer(x)  # <--- ValueError: Can't call methods on orphaned modules
    return x
```

The code is failing because `self.layers.append()` does not assign anything to `self`, so the overwritten `__setattr__` isn't triggered. As a result, the `Dense` modules are not properly initialized and they are recognized as "orphaned" (no parent scope). The canonical way to fix this code is:
```
def setup(self):
  self.layers = [nn.Dense(5) for _ in range(3)]
```

Requiring the user to understand this distinction causes mental overload, especially since the error message is cryptic.

Some suggestions for improvements:

1. Improve the error message by explaining how modules should be assigned to `self` in `setup`.

2. Create a new abstraction `ModuleList`, and only allow lists of Modules here (similar to PyTorch).

3. Convert the list to a some data structure that tracks mutation properly.

Option (1) is a quick fix, but doesn't seem like a great long-term solution, since this error can appear in other cases as well, so we want to make sure the error message also covers the other cases.

Option (2) doesn't require much mental overload, but the user should remember that this abstraction exists.

Option (3) would also work, but I fear that this magic may lead to new special cases in the future that we should address.

I personally feel option (2) is most consistent and robust in the long-term. Since it doesn't have to be used in a particular way, it seems also easiest to understand for users.

avital:
I am sympathetic to option (2), but here are a few questions:
1. Could/should we raise an error if you assign a "vanilla" list of Modules?
2. What happens if you want to store a list that has both Modules and other things (e.g. the `relu` function, which is a common need for some `Sequential` patterns)

Also curious to hear from @levskaya who wrote the original "module list logic" in Modules.

rwightman:
Just came across the (lack of) these module containers and searched my way here. With the setup() approach it's definitely useful for more than just the setattr issue mentioned here.

[...]

PR #542, opened 2020-10-19
Creating getting started notebooks for JAX + Flax basics

BertrandRdp:
Progress on #523 including JAX and Flax basics for now.

For review:
https://colab.sandbox.google.com/github/BertrandRdp/flax/blob/master/docs/notebooks/jax_for_the_impatient.ipynb
https://colab.sandbox.google.com/github/BertrandRdp/flax/blob/master/docs/notebooks/flax_basics.ipynb


docs/notebooks/JAX_for_the_impatient.ipynb
```diff
@@ -0,0 +1,1293 @@
+{
```
avital, line 1: <p>Maybe add, "if you deal with DeviceArrays then computation is done on device so no additional data transfer is required, e.g. <code>jnp.dot(long_vector, long_vector)</code>  will only transfer a single scalar back from device to host.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

avital, line 1: <p>Maybe add a comment after line 3 stating that <code>x</code> is unchanged. (And then you'd need to add another last line so that the Colab output contains the changed array), e.g.: (maybe there's a better way to do this)</p><p> </p><pre class="ql-syntax" spellcheck="false">x = jnp.array([[1.0, 2.0, 3.0, 4.0],
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;[5.0, 6.0, 7.0, 8.0]])
updated = jax.ops.index_update(x,(0,0), 3.0) # whereas x[0,0] = 3.0 would fail

print("x", x)
print("updated", updated)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

avital, line 1: <p>Maybe add "(there is no special meaning to the two unsigned int32s -- it's just a way of representing a uint64)" [I <em>think</em> this is true]</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

avital, line 1: <p>Replace "you will not get any randomness" to "you'll get the same random output each time"? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

avital, line 1: <p>Maybe show what normal_pseudorandom you'd get for the original key as well? (And you can show a second line showcasing how when you use <code>key</code>  twice you get the same normal random?)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

avital, line 1: <p>Re: TODO does splitting by 2 instead of 4 give the same RNGs as the normal split?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

avital, line 1: <p>I'd think it's better to not have random variables in this example, so that people can eye-ball the correctness of the result. I'd just do <code>v = jnp.array([1.0, 2.0])</code> or something.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

avital, line 1: <p>I think someone who knows more than me should review the vjp and jvp sections. (And add a reference to forward-mode and reverse-mode autograd describing in short the computational tradeoff)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

avital, line 1: <p>I don't think the note at the end is necessary -- the expectation in this case /is/ the average as you wrote above.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='44'/>

avital, line 1: <p>I would split line 14 to two lines:</p><pre class="ql-syntax" spellcheck="false">y_pred = jnp.dot(W, x) + b
return jnp.inner(y - y_pred, y - y_pred) / 2.0
</pre><p>(any why <code>jnp.inner</code> instead of <code>jnp.dot</code>?)</p><p> </p><p>Or perhaps better:</p><p> </p><pre class="ql-syntax" spellcheck="false">y_pred = make_predict(W, b)(x)
return jnp.inner(y - y_pred, y - y_pred) / 2.0
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='45'/>

avital, line 1: <p>What's approximate about this? Sounds pretty complete to me? (For 1-D data at least)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='48'/>

avital, line 1: <p>Maybe instead of "3 leaves", "5 leaves", write precisely what the leaves are, e.g. "3 leaves: 1, (2, 3), ()"; "5 leaves: 1, 2, (3, 4), 5"</p><p> </p><p>Oh wait, I think the second example is wrong, doesn't it have 4 leaves? (2 and 3 are distinct?)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='50'/>

avital, line 1: <p>Same comment here re: line 14, compute the prediction first.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='57'/>

andsteing:
What do you think about keeping the filenames lowercased?
I think it's always preferable to have lowercase paths everywhere unless there is a very good reason not to.

docs/notebooks/JAX_for_the_impatient.ipynb
BertrandRdp: <p>Good point indeed, will add!</p>

BertrandRdp: <p>Added !</p>

BertrandRdp: <p>I'll trust you on that. At least I didn't read otherwise anywhere! Added</p>

BertrandRdp: <p>Yes rereading it's indeed a weird sentence</p>

BertrandRdp: <p>Added the normal_pseudorandom for the 3 keys (old, key, subkey). Also added one cell showing that multiple calls with same key gives same output.</p>

BertrandRdp: <p>No it doesn't (unless I missed something).</p>

BertrandRdp: <p>Ah yes it's a good point, also because it's harder to read. Changed</p>

BertrandRdp: <p>Yes, the only thing I can tell is that this is a wide vs tall matrix cost overhead but I'm probably missing some piece of context to make that clearer.</p>

BertrandRdp: <p>Well it depends on how you cast the problem, the expectation of the mean here is the "true" expectation. However, the other way is not true unless you consider that your sample fits the actual data distribution. The reason I made the note is that usually you start from a probability distribution (from which you sample come from), and you cast the optimization problem on the expectation, then solve it for the sampled one (that we have here). </p><p> </p><p>The reason I made the note is that I don't want people to nitpick on exactly that, since depending on how you usually cast the problem you might disagree with this one here. I'll rephrase the note a bit though.</p>

BertrandRdp: <p>Hm I don't remember why I ended up not using the make_predict (which was the intent intially). As for inner in fact i don't know, it was kind of my first idea as dot doesn't work in higher dimensions. But reading your question in fact inner doesn't work either. So I don't know :D. In 1D at least it's the same.</p>

BertrandRdp: <p>It's approximate because it's a convex objective and here we are stopping the algorithm "randomly". Given that you have bounds on the convergence, you should monitor those and perform backtracking linesearch or similar to choose the gradient step sizes, and stop at a given precision. But it's nitpicking agreed, this is just to say that we don't consider this as a proper full fledged solution for linear regression.</p>

BertrandRdp: <p>Unless I didn't understand something, the second example is correct: leaves are 1,2,3 as 2, 3 are packed in a tuple which is also a container.</p>

BertrandRdp: <p>Changed also</p>

docs/notebooks/flax_basics.ipynb
```diff
@@ -0,0 +1,911 @@
+{
```
avital, line 1: <blockquote><code>n</code>  the size of the input </blockquote><p>I'd say "n" is the dimensionality of the input</p><p> </p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

avital, line 1: <p>Maybe we should explain <code>FrozenDict</code>  here -- basically say that because JAX is purely functional we help you "not shoot yourself in the foot" (a common theme across Flax) so that you're aware that you can't modify these dicts directly -- the wrong thing will happen.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

avital, line 1: <p>Line 11: Why not just initialize W in the right shape instead of having a transpose?</p><p>Line 16: I don't like how everything is crowded. Good to add more whitespace like you would everywhere around commas and other operators. (maybe this should be two lines)</p><p>Line 16: I don't know if it's good to use vmap here. Readers are probably confused about it still (if they even understand it at all). I know the original JAX intro covered it but you can't assume readers read in the correct order. WDYT? You could use <code>jnp.dot</code>  just fine without vmap, it all still works (because then dot becomes a matrix product instead)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

avital, line 1: <p>Line 1: I'd call this <code>make_loss_func</code>  -- specifically it doesn't make an MSE, it makes a function. (SO or <code>make_mse_func</code>)  <span class="ql-cursor">﻿</span></p><p>Line 1: Add space after comma</p><p>Line 6: ditto</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

avital, line 1: <p>Line 10: Add space after comma</p><p>Line 11: No parenthesis, add space around all operators</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

avital, line 1: <p>Line 4: ditto (no parenthesis and spaces)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

avital, line 1: <p>First point is not true. Even in your example notice that you created <code>Dense(3)</code>  instance with only the output shape. With <code>setup()</code>  you can either have shape inference or not.</p><p> </p><p>An important point is that if you want more than one method on a module, then you <em>must</em> use <code>setup</code>  (we disallow two methods using <code>nn.compact</code>)  -- this is to avoid many hairy issues about whether a module is reused or not, or how they are named if you call one method from another</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

avital, line 1: <blockquote>ut as mentionned above, it won't be able to use shape inference as Flax is using lazy initialization at first call site.</blockquote><p>In this case with our current API you're right. If you define parameters during setup then you can't use shape inference. But it's not like with submodules.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='30'/>

avital, line 1: <p>I wouldn't show this case -- people should really treat params as immutable if they're going to take gradients w.r.t. them.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='31'/>

avital, line 1: <p>Looks like there's more to add for the state section. Maybe it should actually be a separate doc because there's a lot to talk about here, such as how you use an optimizer when you have both parameters as well as non-trainable variables. LMK if you want to chat more about this, or when you have a draft that you'd like to have reviewed.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

andsteing, line 1: <p>I would add a first cell with a main title and a link to open in Colab, e.g. something similar to</p><p> </p><p>(and then also change all other titles one level down to <code>##</code> and <code>###</code> )</p><p> </p><p>It would also be helpful to set the scope of the Colab in the beginning. What can readers expect to learn from this Colab? How does it compare to other Colabs/docs? Does it make most sense to start with this Colab and then move to the readthedocs and then the examples or is there another suggested order?</p><p> </p><pre class="ql-syntax" spellcheck="false"># Flax Basics


&lt;a href="https://colab.research.google.com/github/google/flax/blob/master/docs/notebooks/flax_basics.ipynb" &gt;&lt;img src="https://colab.research.google.com/assets/colab-badge.svg" alt="Open In Colab"/&gt;&lt;/a&gt;


Demonstration notebook for Flax.

Code at https://github.com/google/flax
Docs at https://flax.readthedocs.io/
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='0'/>

andsteing, line 1: <p>(The output seems stale. I usually try to remember to factory reset the notebook and run end to end before submitting.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

andsteing, line 1: <p>Consider adding a link to the other notebook. Is it expected that readers have read the previous notebook first? Then consider mentioning it here and getting rid of all the content that better fits in the first notebook (like below "Note: JAX and Flax, like numpy are row-based systems, ...")</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

andsteing, line 1: <p>In the spirit of making things shorter : You are mentioning here that you will pass a PRNGKey and dummy input parameters, then you have the code cell where you pass the PRNGKey and dummy input parameters (where you also added a comment that it is a dummy input), and then you have a long explanation afterwards why you do this.</p><p> </p><p>In general I think it would be great if the text could be made shorter. Long text cells significantly slow down the progress through the notebook and I suspect that fewer people finish it if they have to read too much (we're living in a information-supersaturated world and uninterrupted focus is a scarce resource ;-)</p><p> </p><p> </p><p>An extreme version would be something like this:</p><p> </p><p>Text : Note that the module does not contain it's parameters, but instead they first have to be generated by calling <code>.init()</code> :</p><p> </p><pre class="ql-syntax" spellcheck="false">key0 = random.PRNGKey(0) # Random seed for weight initialization.
params = model.init(key0, jnp.zeros(10)) # Shape inference from fake input.
jax.tree_map(jnp.shape, params)
</pre><p> </p><p>Text : (talking about the next cell - when combining text &amp; code cells I find it useful to have a clear relationship between text and code cells, e.g. always having a text talking about the next code cell, but YMMV)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

andsteing, line 1: <p>Why using random dummy input here? Makes the cell more complicated that it need to be...</p><p> </p><p>Why don't we call them simply <code>params</code>  ?</p><p> </p><p>Having <code>init_params</code>  and alter <code>params</code>  makes it very easy to forget to <code>del init_params</code>  then for large models we have a lot of RAM wasted for no good reason.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

andsteing, line 1: <p>(Additionally to what I said in above comment)</p><p> </p><p>Nit : "not column" -&gt; "not column vectors".</p><p> </p><p>Mentioning <code>kernel_init</code> and <code>bias_init</code> here without using it is a bit confusing. I would either remove it or add it to the code.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

andsteing, line 1: <p>(In the interest of shortness : I'm not sure if the comment adds anything here.)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

andsteing, line 1: <p>There is <strong>a lot</strong> of repetition between this section and</p><p>https://colab.sandbox.google.com/github/BertrandRdp/flax/blob/master/docs/notebooks/jax_for_the_impatient.ipynb#scrollTo=S2BcA8wm2_FW</p><p> </p><p>Why don't you ask readers to go to that other Colab (you can have a link like the above to the Colab on the master branch) instead of repeating it here? Also, the code below has not really a lot to do with Flax, so having all this content in "Flax basics" might be confusing and distract from the essentials.</p><p> </p><p>(not adding more comments on this section)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='9'/>

andsteing, line 1: <p>Do we need the <code>optimizer_def</code>  for anything? Why not directly </p><p><code>optimizer =&nbsp;optim.GradientDescent(learning_rate=alpha).create(init_params)</code></p><p> </p><p>Maybe <code>grad_fn</code>  -&gt; <code>loss_grad_fn</code>  (I would expect the <code>grad_fn</code>  to simply return the gradient)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='16'/>

andsteing, line 1: <p>Nit : Why 101 steps? We usully tart at step 1 with <code>range(1, total_steps + 1)</code></p><p> </p><p>Pystyle : <code>if i % 10 == 0:</code> </p><p> </p><p>Nit (applies to all cells) : Please use single or double quotes consistently.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

andsteing, line 1: <p>It's not clear what is the purpose of <code>serialization.to_state_dict()</code>  here - could you elaborate?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='19'/>

andsteing, line 1: <p>TODO: Yes, it would help to clarify this.</p><p> </p><p>Text below : "use as a template <strong>de</strong> <span class="ql-cursor">﻿</span>model parameter" </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='20'/>

andsteing, line 1: <p>In terms of structure, why do first have <code>nn.Module</code>, then a lot of stuff that's really JAX and not Flax, then optimization, then serialization, and then we come back to <code>nn.Module</code> ?</p><p> </p><p>Consider having all <code>nn.Module</code> stuff together. Personally, I also feel that optimization &amp; serialization should come after this, but YMMV.</p><p> </p><p>Maybe somewhere here it's a good point to stress again that we imported <code>linen as nn</code>,  not to be confounded with the old API.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

andsteing, line 1: <p>wrt lists : maybe mention that you should modify lists after assigning them, otherwise we don't handle them correctly (that might be fixed by now, though)</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='24'/>

andsteing, line 1: <p>Could you explain a bit what happens behind the scenes when w call <code>.apply()</code>  and why we can't call <code>module()</code>  directly?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='25'/>

andsteing, line 1: <p>Could you explain why the two names <code>counter</code> and <code>count</code> ?</p><p> </p><p>Maybe it would make more sense to use a "standard" name for the collection name? Do we want to motivate users to come up with their own collections?</p><p> </p><p>I would find it more straightforward to do an example with a real application for state (like batchnorm) to make clear what collections are useful for. Creating a Counter as a module feels a bit contrived.</p><p> </p><p>WDYT about using the same name for the value returned by <code>model.init()</code> everywhere? I would vote for simply <code>variables</code> (though pydoc says <code>data</code>, note that in the example above you use <code>params</code>  for the returned values) and then use for <code>y, mutable_variables = model.apply(variables, x, mutable=['counter'])</code> and <code>optimizer_def.create(variables['params'])</code> - i.e. <code>variables</code> instead of <code>params</code> (in the example above), and never using the <code>init_variables</code> (I don't see why we need to give them another name, makes it very easy to forget deleting them). Calling the update variables <code>state</code>  (or <code>updated_state</code>)  makes it less clear that the selection there is directly determined by the argument <code>mutable=</code>. <span class="ql-cursor">﻿</span></p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='33'/>

andsteing, line 1: <p>Should we mention <code>FrozenDict</code>  explicitly somewhere?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/542/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='35'/>

BertrandRdp: <p>I'm pretty sure that's what I did. Will check, but I think it's because I switched from colab to local with vscode (hence packets stick around).</p>

BertrandRdp: <p>Indeed, that's way more accurate</p>

BertrandRdp: <p>That's a good question in fact. My assumption when writing this was that some folks might directly start with the Flax part, but that might not be the right approach. We probably need some balance between self contained and redundancy. @avital any opinion ?</p>

BertrandRdp: <p>About the note, I've put it there because for me I was surprised to see that the kernel was the transpose of what I expected in the params, so I still think it's worth reminding people here.</p>

BertrandRdp: <p>I generally agree with the points being made here, and I'll definitely revisit the test to make it shorter and straighter to the point. I wrote this in one go, so I'm not surprised it's not "length-optimized" but every point in the text is still useful I think because I added what I felt I was missing during the learning process. I'll address other comments and then rework on the text.</p><p> </p><p>About text talking about the next cell, I think this is generally what I try to do, and when I don't it's because I need the output of the previous cell. I'll revisit that as well when rewriting.</p>

BertrandRdp: <p>Yes, agreed, in fact I realize that at that point I asked Jonathan about it but didn't write anything here</p>

BertrandRdp: <p>Random versus <code>jnp.array(something)</code>you  mean ? I'm not sure it's that complicated though, but the point was to be able to change the size easily and then see that the shape inference does its job. But maybe that's not the best way to show it.</p><p> </p><p>Re: <code>init_params</code>vs <code>params</code> , I don't have any opinion about it, I just reused the same pattern than the other notebooks, but I don't mind keeping params. Although when I read it I thought it was emphasising the fact that those are only the initial one (and are going to be mutated).</p>

BertrandRdp: <p>Good point about the init functions, will remove I guess (otherwise would make the code more complicated than it need).</p><p> </p><p>Added the missing word.</p>

BertrandRdp: <p>True, it's saying the same as the previous sentence in the text block.</p>

BertrandRdp: <p>The point initially was to walk people through linear regression by adding more stuff in there along the way. So here we still have a basic regression with a Flax model that shows something.</p><p> </p><p>But the suggestion is good, I'll probably hide the redundant parts, and people will be able to expand it if needed. I'll find something.</p>

BertrandRdp: <p>L11: indeed, this comes from the fact that I always think about column vectors more than row so this seemed more natural to me, but there is no reason to do so, I'll change it.</p><p> </p><p>L16: adding space indeed, and I somehow just thought "I want to apply that to a batch" and used vmap but that's definitely an overkill. <code>jnp.dot</code>will  do the trick just fine.</p>

BertrandRdp: <p>Agreed, fixed spaces</p>

BertrandRdp: <p>Done</p>

BertrandRdp: <p>I don't know, this is a good question: the point of having two objects was to highlight the different concepts. Since those are different in the design of the API I thought it would be better to separate. Happy to consider other opinions.</p><p> </p><p>Agreed on the grad vs val + grad indeed!</p>

BertrandRdp: <p>Tbh because I was lazy and I wanted the 100th step to print ^^ and didn't add a line after the loop. I that's too ugly I can change it ^^. Noted for quotes, fixing</p>

BertrandRdp: <p>I wanted to demonstrate the two provided formats for serialization (dict and bytes). But I didn't have further thoughts on those, I'd take any suggestion to make it more interesting though!</p>

BertrandRdp: <p>Fixed. On the todo it seems(c.f. chat with Anselm) it's a way en enforce that you know which structure generated the data, and avoid shooting yourself in the foot. I'll add that note</p>

BertrandRdp: <p>Happy to discuss the structure, and the one you suggest also makes sense. The rationale for doing it this way was that it was more natural as a first time user, meaning that you are adding complexities on step at a time. Here the structure is more as a learning process, and I think that before wanting to define your own modules, you might want to start by understanding how to import, use, train models provided by others. After that you might be interested in defining your own models.</p><p> </p><p>I shared the notebooks with people who know nothing about Jax and Flax, and I'm interested in what they'll think, but my feeling was that this way was a bit more "progressive". We should get more opinion on this to see what's the best way.</p>

BertrandRdp: <p>I saw this in the issue on Thursday, I wasn't aware in fact but agreed we should say something about it. However I'm not sure I know what the right pattern is, what would you recommend we explain here?</p>

BertrandRdp: <p>Do you mean why something like:</p><p><code>model = nn.Dense(features=5)</code></p><p><code>model(x) # Error</code></p><p>doesn't work vs using apply?</p>

BertrandRdp: <p>Ah, I'm confused then because that was my understanding from the linen notebook (the alpha one) with this sentence:</p><blockquote>We can also declare variables in setup, though in doing so you can't take advantage of shape inference and have to provide explicit shape information at initialization. The syntax is a little repetitive in this case right now, but we do force agreement of the assigned names.</blockquote><p>Let's chat to make sure I understand correctly. For the second point on multiple method, added a note.</p>

BertrandRdp: <p>Didn't understand the distinction with submodules ? </p>

BertrandRdp: <p>Ah no indeed it shouldn't be there, I just tried to see what it would do, but definitely shouldn't be showed here.</p>

BertrandRdp: <p>Yes indeed, Avital had the same comment, I added a note but linking to RTD might be a good idea once it's up to date.</p>

BertrandRdp: <p>Let's chat indeed as I'm not sure about how we should tackle this. Also the @andsteing comment above.</p>

BertrandRdp: <p>I see, so I think I miss an explanation of the collections and params, I'll add something.</p><p> </p><p>I don't have an opinion on having standard collection names, I guess beyond the batch norm one it's hard to think about all the use cases. This shouldn't probably be part of the best practices for Flax.</p><p> </p><p>On the contrived nature of the example I agree, and I actually hesitated between this and using batchnorm directly. The issue I see with BatchNorm is that I wanted to have the full implementation here (and not too long), so that you could see how to use variables. Problem is that a proper batch norm is a bit too long (and not everyone knows on top of their head what BatchNorm exactly is). However, I think it would be good to have something in between, but I didn't have a better idea at this point. If you have one happy to change to something that makes more sense indeed!</p><p> </p><p>About the names yeah I see, actually I changed the naming for the very specific reasons you mentioned, but ended up also confusing indeed. Will change across the doc to make it clearer and more consistent</p><p> </p>

andsteing: If you only care about shape then you should use `jnp.ones()` or `jnp.zeros()`.

I would only use random where we really need random (e.g. for initializing weight) and then taking care to properly derive all random seeds from an initial seed.

@avital WDYT about `init_params`? I noticed this pattern in a couple of places, but for the reason I mentioned in my comment, I think it would be preferable to always call params `params`. But maybe there was a good reason to call them `init_params` the first time?

andsteing: Yes it would be great to get some unbiased input from first time users!

It would also be interesting to know what percentage of new Flax users are also new to JAX, vs. users who have already used JAX and now want to learn Flax.

andsteing: I meant : in this example `serialization.to_state_dict(optimizer.target)` is exactly the same as `optimizer.target.unfreeze()`. But `serialization.to_state_dict()` can do more than that (e.g. if you say `serialization.to_state_dict(optimizer)`.

andsteing: Let's see how #524 gets resolved. Maybe point to that issue for the interested reader in the meantime.

andsteing: Exactly. I think this is not obvious for first time users.

andsteing: What about using a "simplified version of batchnorm", e.g. simply keeping the running average and subtracting that from the inputs? That should be fairly short and you could say that IRL you would ues batchnorm & link to the Flax implementation for reference.

As for naming : Note that currently you say `print('updated variables:\n', updated_state)` (i.e. `variables` but then only print `state` which is a subset of the variables).

andsteing:
Some more replies on the comments. Please mention explicitly when you would like another reply from me on some of the open comments.

docs/notebooks/flax_basics.ipynb
BertrandRdp: <p>Done</p>

BertrandRdp: <p>Ah, oh! Got it now indeed, will change the example for something different!</p>

BertrandRdp: <p>Makes sense, let me revamp a bit that section and address those.</p>

BertrandRdp: <p>Going to add trying mutating a frozendict to show it</p>

BertrandRdp: <p>Add a reference to a doc needed for patterns with shape inference (or not)</p>

BertrandRdp: <p>Remove as mentionned above (diff from submodule)</p>

BertrandRdp: <p>Added the scope part, however I don't think links are useful since those will live in RTD in the end, not as standalone colabs!</p>

PR #549, opened 2020-10-22
Preserve nested dict in MultiOptimizer

jheek:
Adds an option to flatten_dict that replaces  empty dicts with a stub


MostafaDehghani (approved):
Thank you for the fix Jonathan. Looks perfect!

PR #551, opened 2020-10-22
kernel_size parameter can be an int for 1d Conv and ConvTranspose

### LINKED ISSUE #547: Helpful error when kernel_size is not an array in Conv
If you try to init a Conv module by setting `kernel_size` to an int, you get an unhelpful error message.

Helpful error message that explains I should set `kernel_size` to an array.

`TypeError: object of type 'int' has no len()`

### Steps to reproduce:

```python
from flax import nn
from jax import numpy as jnp, random

class CNN(nn.Module):

    def apply(self, x):
        x = nn.Conv(x, features=32, kernel_size=3)
        x = nn.relu(x)
        return x

cnn = CNN.init(random.PRNGKey(0), jnp.ones((1, 28, 28, 1)))
```

freddyaboulton:
Fix #547

### DIFF
--- a/flax/nn/linear.py
+++ b/flax/nn/linear.py
@@ -170,6 +170,14 @@ def _conv_dimension_numbers(input_shape):
   return lax.ConvDimensionNumbers(lhs_spec, rhs_spec, out_spec)
 
 
+def _make_correct_length(kernel_size, ndim):
+  """Make kernel_size parameter match the number of spatial dims."""
+  if len(kernel_size) == 1:
+    return (kernel_size[0],) * (ndim - 2)
+  else:
+    return kernel_size
+
+
 class Conv(base.Module):
   """Convolution Module wrapping lax.conv_general_dilated."""
 
@@ -219,12 +227,16 @@ class Conv(base.Module):
     """
 
     inputs = jnp.asarray(inputs, dtype)
+    if isinstance(kernel_size, int):
+      kernel_size = (kernel_size,)
 
     is_single_input = False
     if inputs.ndim == len(kernel_size) + 1:
       is_single_input = True
       inputs = jnp.expand_dims(inputs, axis=0)
-      
+
+    kernel_size = _make_correct_length(kernel_size, inputs.ndim)
+
     if strides is None:
       strides = (1,) * (inputs.ndim - 2)
 
@@ -296,12 +308,16 @@ class ConvTranspose(base.Module):
       The convolved data.
     """
     inputs = jnp.asarray(inputs, dtype)
+    if isinstance(kernel_size, int):
+      kernel_size = (kernel_size,)
     
     is_single_input = False
     if inputs.ndim == len(kernel_size) + 1:
       is_single_input = True
       inputs = jnp.expand_dims(inputs, axis=0)
-    
+
+    kernel_size = _make_correct_length(kernel_size, inputs.ndim)
+
     strides = strides or (1,) * (inputs.ndim - 2)
 
     in_features = inputs.shape[-1]
--- a/flax/nn/linear.py
+++ b/flax/nn/linear.py
@@ -170,14 +170,6 @@ def _conv_dimension_numbers(input_shape):
   return lax.ConvDimensionNumbers(lhs_spec, rhs_spec, out_spec)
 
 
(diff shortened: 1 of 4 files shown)


flax/nn/linear.py
```diff
@@ -219,12 +227,16 @@ def apply(self,
     """
 
     inputs = jnp.asarray(inputs, dtype)
+    if isinstance(kernel_size, int):
+      kernel_size = (kernel_size,)
```
avital, line 231: In this code path, users could also pass `(kernel_size, )` as the argument. Is that desirable? (I would imagine not, unless there's a strong reason)

I'd prefer to see the argument be an int OR a tuple of the right length.

flax/nn/linear.py
```diff
@@ -219,12 +227,16 @@ def apply(self,
     """
```
avital, line 227: Can you fix the docstring as well?

avital:
Thanks @freddyaboulton! We are switching soon to Linen (https://github.com/google/flax/tree/master/flax/linen/) -- could you also please make these changes there? (And add the relevant tests as well?)

flax/nn/linear.py
```diff
@@ -219,12 +227,16 @@ def apply(self,
...
+    if isinstance(kernel_size, int):
+      kernel_size = (kernel_size,)
 
     is_single_input = False
     if inputs.ndim == len(kernel_size) + 1:
       is_single_input = True
       inputs = jnp.expand_dims(inputs, axis=0)
-      
+
+    kernel_size = _make_correct_length(kernel_size, inputs.ndim)
```
jheek, line 238: The broadcasting magic is not compatible with the code above recognizing whether the input has a batch dimension. `if inputs.ndim == len(kernel_size) + 1` needs the true kernel size to work correctly.

flax/nn/linear.py
jheek: The standard way of allowing a int as a short hand for (int,) in the numpy, jax, and flax api is to allow both `i` and `(i,)`. Sometimes code creates tuples of dynamic size so having to "unpack" a size 1 tuple into an int would be annoying for users.

jheek (changes requested):

flax/nn/linear.py
avital: TIL something :)

flax/nn/linear.py
freddyaboulton: Good point! But don't we need to know whether there is a batch size to infer the true kernel size if an int is passed? It's tricky to guess what the user means since there aren't separate Conv1D, Conv2D, ... classes. Maybe we should just issue an error when an array isn't passed in?

jheek: So I think what we can do is interpret int as (int,) so basically int becomes a short-hand for 1D conv.
(I would prefer that for consistency with other int to tuple conversions anyway.)

freddyaboulton: Done! Good idea.

flax/nn/linear.py
freddyaboulton: Done!

levskaya:
Sorry I didn't quite follow the comment trail here - we have to make sure whatever convention we pick here still does work with the batch-dim-free / single-example special case (I think we still need to port this special-case to linen), since we need this to make Conv etc compatible with `vmap` of single-example code, something many users want to use in the near future.

freddyaboulton:
@levskaya Got it - I just noticed that linen does not yet support the batch-dim-free case! Should that be added in this PR? This PR was just to map an int to (int,) to avoid a confusing error message. For context, @jheek and I decided that an int should only be valid for 1d convolution. I don't think that interferes with supporting the batch-dim-free case.

levskaya:
@freddyaboulton - ah ok got it, this should be fine.  If you feel like adding the single-example special case code and tests from #444 to the linen Conv/ConvTranspose that would be most welcome, but don't feel obliged!  It might make more sense as a separate PR.

freddyaboulton:
@levskaya Sounds great! I'd be happy to do so in a separate PR :)

levskaya (approved):
@freddyaboulton - thanks again for this!

PR #555, opened 2020-10-24
Fix transient module-state handling by transforms.

levskaya:
This should fix the tests failing with default named_call profiling.


levskaya (commented):
Pulvis et umbra sumus.

PR #563, opened 2020-10-27
Allow restoring raw state_dict using checkpoints.restore_checkpoint().

jpuigcerver:
Currently, in order to restore a checkpoint, one needs to pass a pytree
target with the same structure as the one contained in the checkpoint.

This is problematic if one wants to restore only part of the checkpoint,
for instance, to initialize a new but slightly different model (e.g.
suppose that I just change the number of outputs in the last layer, and
want to re-use everything else). By restoring the raw state_dict, one
has more liberty to manipulate it in arbitrary ways, without needing to
instantiate a pytree with the same structure as the checkpoint.

If the raw state_dict is desired, just pass `target=None`.

### DIFF
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -121,7 +121,8 @@ def restore_checkpoint(ckpt_dir, target, step=None, prefix='checkpoint_'):
 
   Args:
     ckpt_dir: str: directory of checkpoints to restore from.
-    target: matching object to rebuild via deserialized state-dict.
+    target: matching object to rebuild via deserialized state-dict. If None,
+      the deserialized state-dict is returned as-is.
     step: int: step number to load or None to load latest.
     prefix: str: name prefix of checkpoint files.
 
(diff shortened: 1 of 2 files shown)


levskaya (approved):
@jpuigcerver - this is totally reasonable and quite useful to add.  My apologies for the delay getting to this!

flax/training/checkpoints.py
```diff
@@ -121,7 +121,8 @@ def restore_checkpoint(ckpt_dir, target, step=None, prefix='checkpoint_'):
 
   Args:
     ckpt_dir: str: directory of checkpoints to restore from.
-    target: matching object to rebuild via deserialized state-dict.
+    target: matching object to rebuild via deserialized state-dict. If None,
+      the deserialized state-dict is returned as-is.
```
andsteing, line 125: Maybe add a comment that usually you would want to pass a `target` to make sure the deserialized object has the expected form. At least I think that was the original intention of `target`.

PR #564, opened 2020-10-28
Linen Conv and ConvTranspose support batch free input

freddyaboulton:
This PR adds support for single-example (no batch dim) inputs to linen's Conv and ConvTranspose.


levskaya (approved):
@freddyaboulton - this looks great, thank you so much for adding this!

PR #578, opened 2020-10-30
improved Embed input type check

nestordemeure:
Previous check was not accepting some valid integer types such as int8.


jheek (approved):
Thanks! The `issubtype` looks nicer also

levskaya:
@nestordemeure - would you want to add that to this PR as well?

nestordemeure:
PR updated.

levskaya:
@nestordemeure - thanks so much for that!

PR #581, opened 2020-10-30
Graduate Linen out of alpha

avital:
Given that we are actually telling people to build new codebases in Linen, we shouldn't claim that it's alpha.

The other changes we're discussing are improvements that we can and will make, but I don't think we're serving our users by waiting any longer to flip the switch.


jheek (approved):
Looks good to me!

PR #582, opened 2020-10-30
Add Linen reference docs. Mark `flax.nn` as deprecated.

avital:
(There's a lot of work still but this isn't too bad and I added a note that we need to improve it)


avital:
See what these docs look like at https://flax.readthedocs.io/en/rtd-linen/flax.linen.html

BertrandRdp (approved):

PR #585, opened 2020-10-31
Fix jax bfloat16 array handling in msgpack serialization.

levskaya:
JAX's bfloat16 extension dtype doesn't report the correct `arr.dtype.str` but does report the correct `arr.dtype.name`.  We use that to special-case fetching the `jnp.bfloat16` dtype for restoration.

### DIFF
--- a/flax/serialization.py
+++ b/flax/serialization.py
@@ -178,7 +178,7 @@ register_serialization_state(_NamedTuple,
 # https://github.com/msgpack/msgpack/blob/master/spec.md
 #
 # - ndarrays and DeviceArrays are serialized to nested msgpack-encoded string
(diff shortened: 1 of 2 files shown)


levskaya (commented):
Fallaces sunt rerum species

jekbradbury:
Do you think there's something JAX is doing wrong here?

levskaya:
@jekbradbury - I think this is just hard since there's no native numpy `bfloat16` dtype.  About the only reasonable thing to change would be to make `jnp.bfloat16.dtype.kind` equal `f` rather than `V`, not that that's useful for anything -- the rest would requiring having canonical letter codes for a "bfloat".  This JAX dtype has the `name` attribute which should suffice I think for any practical problem in terms of identifying and handling this extension dtype.  I just hadn't been careful to check how custom numpy dtypes were serialized earlier.

PR #597, opened 2020-11-04
Add __repr__ to linen module

jheek:
Adds a pretty trace to linen module:

```python
def fn(module):
class MLP(nn.Module):
      act: Callable = nn.relu
      sizes: Iterable[int] = (3, 2)

      @nn.compact
      def __call__(self, x):
        for size in self.sizes:
          x = nn.Dense(size)(x)
          x = self.act(x)
        return repr(self)
    mlp = MLP()
   
    x = jnp.ones((1, 2))
    trace, variables = mlp.init_with_output(random.PRNGKey(0), x)
    print(trace)
```

prints:
```
MLP(
    # attributes
    act = relu
    sizes = (3, 2)
    # children
    Dense_0 = Dense(
        # attributes
        features = 3
        use_bias = True
        dtype = float32
        precision = None
        kernel_init = init
        bias_init = zeros
    )
    Dense_1 = Dense(
        # attributes
        features = 2
        use_bias = True
        dtype = float32
        precision = None
        kernel_init = init
        bias_init = zeros
    )
)
```

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -49,6 +49,45 @@ def _check_omnistaging():
         "  config.enable_omnistaging()")
 
 
+def _indent(x: str, num_spaces: int):
+  indent_str = ' ' * num_spaces
+  lines = x.split('\n')
+  # skip last line because it is always empty and should not be indented.
+  assert lines[-1] == ''
+  return '\n'.join(indent_str + line for line in lines[:-1]) + '\n'
+
+
+def _attr_repr(value: Any):
+  if isinstance(value, Callable) and getattr(value, '__name__', None):
+    value_rep = value.__name__
+  else:
+    value_rep = repr(value)
+  return value_rep
+
+def module_repr(module: 'Module', num_spaces: int = 4):
+  """Returns a pretty printed representation of the module"""
+  cls = type(module)
+  cls_name = cls.__name__
+  rep = ''
+  attributes = {k: v for k, v in cls.__annotations__.items() if k not in ('parent', 'name')}
+  child_modules = {k: v for k, v in module.children.items() if isinstance(v, Module)}
+  if attributes:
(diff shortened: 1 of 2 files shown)


avital:
Why return a string? Can't we return an instantiated (but frozen) module instance instead and then people can do what they want with that (serialize into a string or inspect via "normal" Python)?

jheek:
Actually, I'll backtrack on the trace feature. You could return a frozen module instance (although we don't have that functionality yet). But I don't like to add yet another variant of the module state just for this. Actually a user can easily print the repr during apply or manually return the string from apply. I guess we should just make a doc or howto that explains how to get an overview of the module.

I think frozen module instances could be interesting but they don't have to be part of this PR. This is just about having a really informative __repr__ impl. I'll remove the trace stuff for now

avital (approved):
LGTM but I have one question: What would happen if you pass a submodule as an attribute into another submodule? Something like

```
class Foo(nn.Module):
  def setup():
    self.submodule = Dense()
    self.other_submodule = SomeWrapperThatTakesInAModule(self.submodule)
```

Would it print the submodule twice? (Not entirely sure what the right answer even is here, but it should at least not do something obviously wrong)

flax/linen/module.py
```diff
@@ -49,6 +49,45 @@ def _check_omnistaging():
         "  config.enable_omnistaging()")
 
 
+def _indent(x: str, num_spaces: int):
+  indent_str = ' ' * num_spaces
+  lines = x.split('\n')
+  # skip last line because it is always empty and should not be indented.
```
Marvin182, line 55: Isn't this just because of the way we construct the string in module_repr()? I think it would be better to not make this assumption here and handle both cases.

jheek: Done

PR #599, opened 2020-11-05
Fix incorrect reference to DeprecationWarning

avital:


mohitreddy1996 (approved):

jheek:
Ha that's funny, so DeprecationWarning is also a global but `warn()` is not?

PR #588, opened 2020-11-03
Updates top-level doc-string of functional core Scopes

marcvanzee:
Thanks to @avital for some valuable input!

Issue #593, opened 2020-11-03
Increase Linen pydoc Coverage


andsteing:
As part of the Linen doc improvements, the pydoc coverage should be increased.

PR #604, opened 2020-11-06
Adds docstrings to methods in module.py

marcvanzee:
Once these are done I plan to write a top-level doc-string for module.py, similarly to the one I wrote for scope.py: https://github.com/google/flax/pull/588.


marcvanzee:
Linking this to #593

avital:
This was replaced with #625

PR #608, opened 2020-11-09
Update linen intro notebook for new mutated collection behavior of apply.

levskaya:
`apply()` now returns only the explicitly mutated collections, rather than a copy of all variable collections, we update the linen intro notebook to use examples showing this correctly.


avital (approved):
Thanks Anselm!

PR #609, opened 2020-11-09
nlp_seq to linen port

bohnetbd:
I carried out the changes to port the example to linen.

PR #610, opened 2020-11-09
nlp_seq port to linen

bohnetbd:
I ported the nlp_seq example to linen.


andsteing:
Could we review the changes in the original location `flax/examples/nlp_seq` (#609) and then move it in a second PR? That would make the review a bit easier because we see actual deltas instead of "all new".

PR #618, opened 2020-11-10
Raise an Exception when a user accesses empty variables

### LINKED ISSUE #513: Linen: Consider raising an error when reading variables from submodule before initialization?

```
conv = nn.Conv(features=3)
x = conv(x)
params = conv.params()
```

But if you instead do:
```
conv = nn.Conv(features=3)
params = conv.params()
x = conv(x)
```

freddyaboulton:
Fixes #513 .

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -534,7 +534,13 @@ class Module:
 
   @property
   def variables(self):
-    return self.scope.variables()
+    variables = self.scope.variables()
+    if not variables:
+      raise RuntimeError("Variables dictionary is empty! This is probably because you have accessed the variables "
+                         "before shape inference. Apply your module on its expected input to define the variables. "
+                         "If you expect your module to not have any variables, then you should not subclass "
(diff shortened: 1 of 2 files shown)


flax/linen/module.py
```diff
@@ -534,7 +534,13 @@ def init(self, rngs, *args, method=None, **kwargs):
 
   @property
   def variables(self):
-    return self.scope.variables()
```
freddyaboulton, line 537: It's possible the user also tries to access variables before the scope is defined. Should we check for that too?

avital:
Thanks @freddyaboulton for jumping on this, it's really nice how quickly you wrapped your head around this. (Please keep us in the loop about how your experience with Flax is going!)

In the meanwhile I'm going to close this because the discussion on #513 is now leaning towards a different approach.

avital:
(But please do feel free to play around with @jheek 's latest proposal on #513 and see how it can be implemented)

PR #622, opened 2020-11-11
Update setup.py

Marvin182:
Add tensorflow_text and clu to our test dependencies. tensorflow_text is required for WMT test cases (after #616) and clu will be used in the future in the examples.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ install_requires = [
 
 tests_require = [
     "atari-py",
+    "clu",
     "gym",
     "jaxlib",
     "ml-collections",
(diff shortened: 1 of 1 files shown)


avital:
Are you sure this is the right place to add these? I think these are the dependencies of the library unit tests. For examples we install these dependencies: https://github.com/google/flax/blob/master/examples/requirements.txt

avital:
Oh hmm I see that we install opencv-python and gym as part of tests_require so I guess other examples list dependencies here. I am not sure what the impact of adding something to tests_require in setup.py vs examples/requirements.txt is.

Maybe @jheek knows better?

I'll approve this in the meanwhile but we may want to change it afterwards.

avital (approved):

PR #625, opened 2020-11-12
Adds Module docstrings and updates readthedocs

marcvanzee:
@avital, @jheek and I worked on this together over the past few days.

- Factors out variables into `core/variables.py`.
- Renames `Variables` to `VariableDict`.
- Uses a simpler type alias for `VariableDict` than what is used in `scope.py`: `Dict[string, Dict[str, Any]]`. This is to encapsulate scope's types (e.g., `FrozenCollection`) as much as possible so it is not exposed to the user.
- Adds the top-level docstring of `variables.py` to readthedocs.
- Adds docstrings to various methods of `Module`.


flax/core/scope.py
```diff
@@ -493,15 +460,18 @@ def apply(fn: Callable[..., Any],
     `fn` with the scope partially applied.
   """
   @functools.wraps(fn)
-  def wrapper(variables: FrozenVariables, *args,
-              rngs: Optional[RNGSequences] = None, **kwargs) -> (Any, FrozenVariables):
+  def wrapper(variables: FrozenVariableDict, *args,
+              rngs: Optional[RNGSequences] = None, 
+              **kwargs) -> Union[Any, Tuple[Any, VariableDict]]:
```
levskaya, line 465: Are we sure this type signature is correct?  it's not `Tuple[Any, FrozenVariableDict]`?

flax/core/variables.py
```diff
@@ -0,0 +1,78 @@
...
+from typing import Generic, TypeVar, Any, Dict
+
+VariableCollection = Dict[str, Any]
+VariableDict = Dict[str, VariableCollection]
+
+T = TypeVar('T')
+class Variable(Generic[T]):
+  """A Variable object allows mutable access to a variable in a VariableDict.
+  
+  Variables are identified by a collection (e.g., "params") and a name (e.g.,
```
levskaya, line 53: it might be better to refer to an actually mutable collection, since we treat "params" specially by exposing their direct value as opposed to a Variable

flax/linen/module.py
```diff
@@ -133,9 +137,30 @@ def get_suffix_value_pairs(
     flat_dict = traverse_util.flatten_dict(dict_or_leaf)
     return [('_' + '_'.join(k), v) for k, v in flat_dict.items()]
 
-
-def all_names_on_object(obj: Any) -> Set[str]:
-  """Get all names of attributes on self and its classes throughout MRO."""
+def _is_module_tree(in_tree: Any) -> bool:
```
levskaya, line 140: It's a little worrying that this function is here at all - it was deleted a while ago during a simplification / fix.  Are you sure you based this commenting PR on a recent commit of flax?

flax/linen/module.py
```diff
@@ -407,52 +511,60 @@ def __post_init__(self):
...
-    """Called when Module instance receives variables and PRNGs.
-    If you want to use PRNGs and/or read or write variables during module
-    instance construction, override this method. It will be automatically
-    called from the dataclass `__post_init__`.
+    """Initializes a Module (similar to ``__init__`` for non-dataclass Python classes).
+
+    Override this method in your module subclasses to initialize submodules and
+    other attributes. This method is called after all dataclass attributes are
+    assigned and the module is ready for use. Variables and RNGs are guaranteed
+    to be available.
```
levskaya, line 522: what do you want to convey with this phrasing?

flax/linen/module.py
```diff
@@ -491,49 +608,103 @@ def param(self, name: str, init_fn: Callable[..., T], *init_args) -> T:
...
+            mutable: Union[bool, str, Sequence[str]] = False,
+            **kwargs) -> Union[Any, Tuple[Any, VariableDict]]:
+    """Applies a module method to variables and returns output and modified variables.
+    
+    Args:
+      variables: A dictionary containing variables keyed by variable 
+        collections. See :mod:`flax.core.variables` for more details
+        about variables.
+      rngs: The rngs for the variable collections.
+      method: An optional method. If provided, applies this method. If not
```
levskaya, line 652: we should clarify here and elsewhere that you literally pass the method i.e. `mymodule.mymethod` which people find weird as hell and may change, but we should at least be clear about it for now.

flax/linen/module.py
```diff
@@ -491,49 +608,103 @@ def param(self, name: str, init_fn: Callable[..., T], *init_args) -> T:
...
+           method: Optional[Callable[..., Any]] = None, 
+           **kwargs) -> VariableDict:
+    """Initializes a module method with variables and returns modified variables.
+
+      Args:
+      rngs: The rngs for the variable collections.
+      method: An optional method. If provided, applies this method. If not
+        provided, applies the ``__call__`` method.
+    Returns:
+      `(output, vars)``, where ``vars`` are is a dict of the modified
```
levskaya, line 699: this is wrong, we return only `vars` from `init`

levskaya (commented):
I left a number of comments / questions

andsteing:
Linking to #593

avital:
@googlebot I consent.

flax/core/scope.py
marcvanzee: I think it should indeed be `FrozenVariableDict` instead of `VariableDict`, but I think we should use `Union` since depending on whether `mutable` is true or false we will either return only the output (`Any`) or a tuple containing the output and the frozen variables.

flax/core/variables.py
marcvanzee: Done.

flax/linen/module.py
marcvanzee: NIce catch! :-) I started this PR with a wrong base and we then rebased it, but I guess we forgot to remove this function. I removed it now.

flax/linen/module.py
marcvanzee: I agree. I've updated the comment.

flax/linen/module.py
avital: I wrote this comment originally. Maybe we can do better but the point is that `setup` acts like `__init__` but not ... exactly. In this code:

```
class MyModule(nn.Module):
  def setup(self):
    self.submodule = MySubModule()
```

Then `setup` is only called after the assignment (because then we know the attribute name hence the RNG and the place parameters live in the variable dict).

So I was trying to somehow explain that. But maybe I did a poor job. Maybe instead of explaining the technicalities I should just say exactly when setup is called, like I explained in this comment.

flax/core/variables.py
avital: I don't see this change, did you maybe miss any files you changed when you pushed your latest commit?

flax/linen/module.py
avital: By "the assignment" I mean the line `self.submodule = MySubModule()` [since we override `__setattr__`]

marcvanzee: I've updated the comment!

PR #626, opened 2020-11-12
Changes due to review of andsteing no resolved PR #609

bohnetbd:
a number of small changes due to comments of andsteing which have been not resolved in [PR #609](https://github.com/google/flax/pull/609).

### DIFF
--- a/examples/nlp_seq/models.py
+++ b/examples/nlp_seq/models.py
@@ -194,8 +194,8 @@ class Transformer(nn.Module):
 
   @nn.compact
   def __call__(self,
-               *,
                inputs,
+               *,
                train):
     """Applies Transformer model on the inputs.
 
--- a/examples/nlp_seq/models.py
+++ b/examples/nlp_seq/models.py
@@ -14,14 +14,13 @@
 
 """Transformer-based langauge models."""
 
-from flax import linen as nn
-
 from typing import Callable, Any, Optional
 
 import jax
 import jax.numpy as jnp
 import numpy as np
 
+from flax import linen as nn
 from flax import struct
 
 
@@ -37,8 +36,8 @@ class TransformerConfig:
   qkv_dim: int = 512
   mlp_dim: int = 2048
   max_len: int = 2048
-  dropout_rate: float = 0.1
-  attention_dropout_rate: float = 0.1
+  dropout_rate: float = 0.3
+  attention_dropout_rate: float = 0.3
   kernel_init: Callable = nn.initializers.xavier_uniform()
   bias_init: Callable = nn.initializers.normal(stddev=1e-6)
   posemb_init: Optional[Callable] = None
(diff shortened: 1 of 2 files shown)


examples/nlp_seq/train.py
```diff
@@ -301,7 +294,7 @@ def main(argv):
   @jax.jit
```
andsteing, line 294: Let's change above comment to:

```
# jitting initialization avoids heap fragmentation
```

examples/nlp_seq/train.py
```diff
@@ -20,26 +20,25 @@
 import tensorflow as tf
```
andsteing, line 20: BTW : why did you remove `perplexity` ?

examples/nlp_seq/models.py
```diff
@@ -79,8 +78,7 @@ class AddPositionEmbs(nn.Module):
 
   @nn.compact
   def __call__(self,
-               inputs,
-               inputs_positions=None):
```
andsteing, line 83: wrt packed data : did it ever work?
(if yes: why removing this functionality?)

examples/nlp_seq/models.py
```diff
@@ -205,14 +194,14 @@ class Transformer(nn.Module):
 
   @nn.compact
   def __call__(self,
-            inputs,
-            train=False,
-            dropout_rate=0.3):
+               *,
+               inputs,
```
andsteing, line 198: we usually have `inputs` (or `x`) as the first and only positional argument, i.e.

```
def __call__(self, inputs, *, train):
  ...
```

examples/nlp_seq/train.py
```diff
@@ -301,7 +294,7 @@ def main(argv):
   @jax.jit
   def initialize_variables(init_rng):
     init_batch = jnp.ones((config.max_len, 1), jnp.float32)
-    init_variables = model.init(init_rng, init_batch)
+    init_variables = model.init(init_rng, inputs=init_batch, train=False)
```
andsteing, line 297: after my proposed change in `models.py`, you can revert here to

```
init_variables = model.init(init_rng, init_batch, train=True
```

(we usually say `train=True` for initialization)

andsteing (changes requested):

examples/nlp_seq/models.py
bohnetbd: I never tried and never used it basically - it got copied down from other transformer code.

examples/nlp_seq/models.py
bohnetbd: ah. thanks! changed that.

examples/nlp_seq/train.py
bohnetbd: It is unusual in part-of-speech tagging and many other tagging task. Further, I wanted the training loop more lean to make it more clear.

examples/nlp_seq/train.py
bohnetbd: done

examples/nlp_seq/train.py
bohnetbd: I changed this see, see new commit.

PR #628, opened 2020-11-12
Raise minimal version to 0.1.77

### LINKED ISSUE #627: After update from 0.2.0: AttributeError: module 'jax.core' has no attribute 'eval_context'
After updating from flax 0.2.0 to flax 0.2.2 I get the above error message. Downgrading to 0.2.0 solves this, so the error source is located. I'm working with the now deprecated flax.nn package if backward-compatibility might be the reason for this issue.
The Issue is encountered in a custom RNN, when using the init_by_shape function in conjunction with jax.lax.scan.

jheek:
Previous version constraint fails due to missing omnistaging api's.

Closes #627

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ except IOError:
 
 install_requires = [
     "numpy>=1.12",
-    "jax>=0.1.59",
+    "jax>=0.1.77",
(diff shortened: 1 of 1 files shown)


avital:
Thanks @jheek  for the quick fix!

PR #643, opened 2020-11-17
Add pytype

jheek:
Add pytype validation to the tests and fixes any issues with the current type annotations.

### DIFF
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -137,7 +137,9 @@ def _prepare_freeze(xs: Any) -> Any:
   return {key: _prepare_freeze(val) for key, val in xs.items()}
 
 
-def freeze(xs: Dict[K, V]) -> FrozenDict[K, V]:
+# Why is the signature Dict[K, V] -> FrozenDict[K, V] not accepted here?
+
+def freeze(xs: Mapping[Any, Any]) -> FrozenDict[Any, Any]:
   """Freeze a nested dict.
 
   Makes a nested `dict` immutable by transforming it into `FrozenDict`.
@@ -145,7 +147,7 @@ def freeze(xs: Dict[K, V]) -> FrozenDict[K, V]:
   return FrozenDict(xs)
 
 
-def unfreeze(x: FrozenDict[K, V]) -> Dict[K, V]:
+def unfreeze(x: FrozenDict[Any, Any]) -> Dict[Any, Any]:
   """Unfreeze a FrozenDict.
 
   Makes a mutable copy of a `FrozenDict` mutable by transforming
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -137,8 +137,6 @@ def _prepare_freeze(xs: Any) -> Any:
   return {key: _prepare_freeze(val) for key, val in xs.items()}
 
 
-# Why is the signature Dict[K, V] -> FrozenDict[K, V] not accepted here?
-
 def freeze(xs: Mapping[Any, Any]) -> FrozenDict[Any, Any]:
   """Freeze a nested dict.
 
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -259,12 +259,12 @@ def swap_collection(col_a: str, col_b: str):
 @dataclass(frozen=True)
 class In(Generic[T]):
   """Specifies a variable collection should only be lifted as input."""
-  axis: T
+  axis: Any  # pytype does not support generic variable annotation
 
 @dataclass(frozen=True)
 class Out(Generic[T]):
   """Specifies a variable collection should only be lifted as output."""
-  axis: T
+  axis: Any  # pytype does not support generic variable annotation
 
 
 def _split_in_out_axes(xs: Mapping[CollectionFilter, Any]):
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -28,19 +28,16 @@
 """Flax functional core: Scopes."""
 
 import contextlib
-import enum
 import functools
 import hashlib
-from typing import Any, Callable, Container, Dict, Set, Iterable, Optional, Sequence, Tuple, TypeVar, Union, Generic
+from typing import Any, Callable, Container, Dict, Set, Iterable, Optional, Sequence, Tuple, TypeVar, Union, Generic, Mapping
 
 from . import tracers
 from .frozen_dict import freeze
 from .frozen_dict import FrozenDict
 from .frozen_dict import unfreeze
-from .variables import Variable, VariableDict
 
 import jax
-from jax import lax
 from jax import random
 from jax import numpy as jnp
 
@@ -52,15 +49,15 @@ Array = Any
 
 RNGSequences = Dict[str, PRNGKey]
 
-Filter = Union[bool, str, Sequence[str]]
+Filter = Union[bool, str, Container[str]]
 CollectionFilter = Filter
 PRNGSequenceFilter = Filter
 
+Collection = Mapping[str, Any]
 MutableCollection = Dict[str, Any]
-FrozenCollection = FrozenDict[str, Any]
-MaybeFrozenCollection = Union[MutableCollection, FrozenCollection]
 
-FrozenVariableDict = Dict[str, FrozenCollection]
+VariableDict = Mapping[str, Collection]
+MutableVariableDict = Dict[str, MutableCollection]
 
 
 def _fold_in_str(rng: PRNGKey, data: str) -> PRNGKey:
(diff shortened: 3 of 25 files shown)


flax/core/scope.py
```diff
@@ -191,11 +188,43 @@ def group_collections(xs: VariableDict,
   return tuple(groups)
 
 
+T = TypeVar('T')
```
avital, line 191: Why move this back to scope? It was intentionally placed in a separate file so that we can refer to its documentation separately from the rest of scope. Users using the Module API directly will need to know about variables (thus may want to look at variable.py) but don't need to know about scopes.

jheek: The code was incorrect it had a reference 'Scope' but it wasn't imported. But importing creates a circular dependency.

flax/linen/attention.py
```diff
@@ -25,10 +25,10 @@
 import jax.numpy as jnp
 import numpy as np
 
-from .linear import default_kernel_init
-from .linear import DenseGeneral
-from .module import Module, compact
-from . import initializers
+from .linear import default_kernel_init  # pytype: disable=pyi-error
```
marcvanzee, line 28: Out of curiosity: what error does this give? It seems to trigger always for relative imports?

flax/nn/attention.py
```diff
@@ -19,17 +19,19 @@
...
 from .. import jax_utils
-from . import base
-from . import initializers
-from . import stochastic
+from . import activation  # pytype: disable=pyi-error
+from . import base  # pytype: disable=pyi-error
+from . import initializers  # pytype: disable=pyi-error
+from . import stochastic  # pytype: disable=pyi-error
+from .linear import default_kernel_init  # pytype: disable=pyi-error
+from .linear import DenseGeneral  # pytype: disable=pyi-error
```
marcvanzee, line 27: Perhaps doing it like this reads nicer? (Also in many other places.)

```suggestion
# pytype: disable=attribute-error
from . import activation
from . import base
from . import initializers
from . import stochastic
from .linear import default_kernel_init
from .linear import DenseGeneral
# pytype: enable=attribute-error
```

flax/nn/base.py
```diff
@@ -200,7 +202,7 @@ class _ModuleMeta(abc.ABCMeta):
 
   def __init__(cls, name, bases, attrs):
     super(_ModuleMeta, cls).__init__(name, bases, attrs)
-    apply_fn = cls.apply
+    apply_fn = cls.apply  # pytype: disable=attribute-error
```
marcvanzee, line 205: Would this error go away if we type `cls` with something like `Callable[..., Any]`? Then it should know the `appl` function exists.

flax/optim/base.py
```diff
@@ -215,13 +212,12 @@ class _NoAux:
...
 
-@struct.dataclass
-class Optimizer:
+class Optimizer(struct.PyTreeNode):
   """Wraps an optimizer with its hyper_params, state, and model parameters."""
 
   optimizer_def: OptimizerDef = struct.field(pytree_node=False)
-  state: Any
-  target: Any
+  state: Any = struct.field(pytree_node=True)
```
marcvanzee, line 219: `True` seems to be the default value so you can omit it if you want.

flax/struct.py
```diff
@@ -30,14 +30,16 @@
...
+
 from . import serialization
 
 import dataclasses
 
 import jax
 
 
-def dataclass(clz):
+def dataclass(clz: Type):
```
marcvanzee, line 42: I think you can also just write `type` instead of `Type`, saving you an import :-)

flax/struct.py
```diff
@@ -133,3 +135,20 @@ def from_state_dict(x, state):
...
+class PyTreeNode():
+
+  def __init_subclass__(cls):
+    dataclass(cls)  # pytype: disable=wrong-arg-types
+
+  def __init__(self, *args, **kwargs):
+    # stub for pytype
+    raise NotImplementedError
+
+  def replace(self: TNode, **overrides) -> TNode:
```
marcvanzee, line 152: Why do you use `TNode` instead of simply `PyTreeNode` here?

flax/struct.py
```diff
@@ -133,3 +135,20 @@ def from_state_dict(x, state):
 
 def field(pytree_node=True, **kwargs):
   return dataclasses.field(metadata={'pytree_node': pytree_node}, **kwargs)
+
+
+TNode = TypeVar('TNode', bound='PyTreeNode')
+
+
+class PyTreeNode():
```
marcvanzee, line 143: Could you explain the intuition behind this class?

marcvanzee (changes requested):

flax/core/lift.py
```diff
@@ -259,12 +259,12 @@ def swap(target):
 @dataclass(frozen=True)
 class In(Generic[T]):
   """Specifies a variable collection should only be lifted as input."""
-  axis: T
+  axis: Any  # pytype does not support generic variable annotation
```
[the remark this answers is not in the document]
jheek: Yes it's only with attributes

flax/core/scope.py
```diff
@@ -48,15 +45,15 @@
 
 RNGSequences = Dict[str, PRNGKey]
 
-Filter = Union[bool, str, Sequence[str]]
+Filter = Union[bool, str, Container[str]]
 CollectionFilter = Filter
 PRNGSequenceFilter = Filter
 
+Collection = Mapping[str, Any]
 MutableCollection = Dict[str, Any]
```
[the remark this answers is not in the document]
jheek: yes

flax/core/scope.py
```diff
@@ -191,11 +188,43 @@ def group_collections(xs: VariableDict,
...
+T = TypeVar('T')
+class Variable(Generic[T]):
+  """A Variable object allows mutable access to a variable in a VariableDict.
+
+  Variables are identified by a collection (e.g., "batch_stats") and a name
+  (e.g., "moving_mean"). The value property gives access to the variable's
+  content and can be assigned to for mutation.
+  """
+
+  def __init__(self, scope: 'Scope', collection: str, name: str):
```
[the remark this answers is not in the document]
jheek: No scope is defined below so you need the string syntax

flax/core/scope.py
```diff
@@ -48,15 +45,15 @@
 
 RNGSequences = Dict[str, PRNGKey]
 
-Filter = Union[bool, str, Sequence[str]]
+Filter = Union[bool, str, Container[str]]
 CollectionFilter = Filter
 PRNGSequenceFilter = Filter
 
+Collection = Mapping[str, Any]
```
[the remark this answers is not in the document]
jheek: No before it was FrozenCollection or MutableCollection. But Frozen is deceptive because it might be Frozen but it doesn't have to be

flax/linen/attention.py
jheek: This is some kind of weird pytype bug. I'll make a seperate tracking bug for this

flax/nn/attention.py
jheek: I kind of like the line-by-line thing. The regions makes it easy to be lazy and add imports in the section that don't actually need the escape hatch

flax/nn/base.py
jheek: Pytype refuses a type hint on the self argument (which is cls in this case) :(

flax/optim/base.py
jheek: Pytype things the first argument is optional due the struct.field and it then complains all the attributes that come after are required. I'll add a tracking issue for that as well

flax/struct.py
jheek: Done

flax/struct.py
jheek: Because that's not the correct type. The returned value has the same type as self.
So for example optimizer.replace() returns an Optimizer instance not just an arbitrary PyTreeNode.

flax/struct.py
jheek: added a docstring

flax/core/frozen_dict.py
```diff
@@ -137,15 +137,17 @@ def _prepare_freeze(xs: Any) -> Any:
   return {key: _prepare_freeze(val) for key, val in xs.items()}
 
 
-def freeze(xs: Dict[K, V]) -> FrozenDict[K, V]:
+# Why is the signature Dict[K, V] -> FrozenDict[K, V] not accepted here?
```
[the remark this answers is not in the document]
jheek: Yes there's not really any benefit but there is also pytype issues so I will just remove it for now

marcvanzee (approved):

Issue #636, opened 2020-11-15
__mul__ method between nn.Module classes raises error


thomaspinder:
I'd like to be able to implement a `__mul__(self, other)` method for classes inheriting the `nn.Module` class. I'm trying to implement a set of kernel functions, and we know that the product of two kernels gives rise to a new kernel, so having the ability to define a new kernel by the product of two existing kernels would be useful. 

A minimal working example that demonstrates the problem is as follows:

```python
class SimpleMLP(nn.Module):
    features: Sequence[int]
    
    @nn.compact
    def __call__(self, inputs):
        x = inputs
        for i, feat in enumerate(self.features):
            x = nn.Dense(feat, name=f'layers_{i}')(x)
            if i != len(self.features) - 1:
                x = nn.relu(x)
        return x

    def __mul__(self, other):
        return self.features + other.features

m1 = SimpleMLP(features=5)
m2 = SimpleMLP(features=10)
m1*m2
```

Clearly the `__mul__(self, other)` method here is only for demonstration purposes, but I would expect `15` to be outputted. Instead, the following error message is given:

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-69-ddca13716f29> in <module>
     18 m1 = SimpleMLP(features=5)
     19 m2 = SimpleMLP(features=10)
---> 20 m1*m2

~/Documents/GPBlocks/venv/lib/python3.8/site-packages/flax/linen/module.py in wrapped_module_method(self, *args, **kwargs)
    146 
    147     if self.scope is None:
--> 148       raise ValueError("Can't call methods on orphaned modules")
    149 
    150     if is_compact_method:

ValueError: Can't call methods on orphaned modules
```

Is there a correct way to implement the `__mul__` method for an `nn.Module` class?

avital:
Hi @thomaspinder, good to see you here!

The "Can't call methods on orphaned modules" was supposed to be a helpful error to guard against user error. But this is an interested example of how it may be overly restricting. We could consider alternatively disallowing anything that calls into `self.variables`, `self.variables()` or `self.param()`, which would make your use case work fine.

In the meanwhile, a workaround would be to define `mul` as a top-level function (not a method on your Module instance). I know this isn't ideal but just thought I'd bring it up. I think we should fix this.

avital:
Hi Thomas, here is a PR that (I believe) resolves your issue: https://github.com/google/flax/pull/653

[...]

PR #653, opened 2020-11-19
Relax limit on calling methods on unbound modules

avital:
Inspired by #636, relax the limitation on calling methods on unbound modules.
While at time users may be confused about when `setup` is run, hopefully
the new docstring we have on `setup` clarifies things. If users want
to write a method that doesn't depend on any attributes defined during
`setup` (e.g. submodules or variables), that should be fine.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -208,10 +208,10 @@ def wrap_method(fun: Callable[..., Any]) -> Callable[..., Any]:
     is_compact_method = hasattr(fun, 'compact')
     is_setup_method = fun.__name__ == 'setup'
 
-    if self.scope is None:
-      raise ValueError("Can't call methods on orphaned modules")
-
     if is_compact_method:
+      if self.scope is None:
+        raise ValueError("Can't call compact methods on unbound modules")
+
       self._state.in_compact_method = True
     elif is_setup_method:
       self._state.in_setup = True
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -628,6 +628,8 @@ class Module:
     Returns:
       True if the variable exists.
     """
+    if self.scope is None:
+      raise ValueError("Can't access variables on unbound modules")
     return self.scope.has_variable(col, name)
 
   def make_rng(self, col: str) -> PRNGKey:
@@ -644,6 +646,8 @@ class Module:
     Returns:
       The newly generated RNG key.
     """
+    if self.scope is None:
+      raise ValueError("Can't use RNGs on unbound modules")
     return self.scope.make_rng(col)
 
   def apply(self, variables: VariableDict, *args, rngs: RNGSequences = None,
@@ -712,6 +716,8 @@ class Module:
   @property
   def variables(self) -> VariableDict:
     """Returns the variables in this module."""
+    if self.scope is None:
+      raise ValueError("Can't access variables on unbound modules")
     return self.scope.variables()
 
 
--- a/tests/linen/module_test.py
+++ b/tests/linen/module_test.py
@@ -590,6 +590,40 @@ class ModuleTest(absltest.TestCase):
     trace = mlp.apply(variables, x)
     self.assertEqual(trace, expected_trace)
 
+  def test_call_unbound_compact_module_methods(self):
+    dense = Dense(3)
+    with self.assertRaisesRegex(ValueError, "compact.*unbound module"):
+      dense(jnp.ones((1, )))
+
+
+  def test_call_unbound_noncompact_module_methods(self):
+    class EmptyModule(nn.Module):
+      foo: int = 3
+
+      def bar(self):
+        return self.foo
+
+    empty = EmptyModule()
+    # It's fine to call methods of unbound methods that don't depend on
+    # attributes defined during `setup`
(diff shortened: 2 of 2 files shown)


tests/linen/module_test.py
```diff
@@ -590,6 +590,40 @@ def __call__(self, x):
...
+
+  def test_call_unbound_noncompact_module_method_without_setup(self):
+    class EmptyModule(nn.Module):
+      def setup(self):
+        self.setup_called = True
+
+      def bar(self):
+        return self.setup_called
+
+    empty = EmptyModule()
```
andsteing, line 620: What about raising an error when `empty.setup()` is called directly ?

tests/linen/module_test.py
```diff
@@ -590,6 +590,40 @@ def __call__(self, x):
...
+    self.assertEqual(empty.bar(), 3)
+
+
+  def test_call_unbound_noncompact_module_method_without_setup(self):
+    class EmptyModule(nn.Module):
+      def setup(self):
+        self.setup_called = True
+
+      def bar(self):
+        return self.setup_called
```
jheek, line 618: I think what we really want here is a test that does something like `self.param(...)` and throws a proper error

flax/linen/module.py
```diff
@@ -208,10 +208,10 @@ def wrapped_module_method(self, *args, **kwargs):
     is_compact_method = hasattr(fun, 'compact')
     is_setup_method = fun.__name__ == 'setup'
 
-    if self.scope is None:
```
jheek, line 211: None of the scope helper methods `make_rng`, `variables`, `param` etc. do not check whether `scope is None`. It will now throw a confusing errror cannot access member of None.

jheek (changes requested):

tests/linen/module_test.py
avital: Yeah good point, we should add another test for that and all the other methods.

tests/linen/module_test.py
avital: Maybe as a separate change? Sounds like it's different than the issue in this PR? Want to file an issue or PR?

tests/linen/module_test.py
avital: But the error you'd get there should say "can't define submodules or parameters in a non-compact method". But agreed we should have tests.

flax/linen/module.py
andsteing: Instead of checking for `scope is None` everywhere, could we initialize `scope` with a dummy object that throws an exception with intelligible error message when any of its attributes is accessed?

avital: See my comment above, I think you either do or should get an error about being in a non-compact method.

tests/linen/module_test.py
jheek: Ah yes for variables you're right. For `make_rng` and `variables()` you still get the None issue because they don't have tot be in compact or setup

avital: Ah you're right, I'll add tests for all these cases and make sure the errors are clear.

flax/linen/module.py
avital: > Instead of checking for `scope is None` everywhere, could we initialize scope with a dummy object that throws an exception with intelligible error message when any of its attributes is accessed?

I like that. I'll first add tests and then consider this change once the test coverage is comprehensive.

avital: Actually after looking at the code more, there are many methods in Scope and implementing a new "Dummy object with good errors" there would be more code than just making sure we check `if scope is None` in a few more places. I'll be adding the relevant tests regardless so if we really wanted to later we could switch to this proposal later.

andsteing: What about subclassing `Scope` with an object that dynamically modifies these methods to rise the exception when they're called?

tests/linen/module_test.py
andsteing: Mentioning it here, because before this PR calling `empty.setup()` directly would have risen an exception:

`AssertionError: "compact.*unbound module" does not match "Can't call methods on orphaned modules"`

Once it's submitted I can create an issue if you prefer.

avital: Yes please file a new issue after this is submitted.

flax/linen/module.py
avital: Perhaps, but then it's harder to make the error messages explicit... They would have to be automatically generated from the method name, which may be fine, but then again we have some methods and `variables` is not a method but a property... Just sounds like a bit too much magic and indirection for something that can be solved with 3 more if statements and explicit errors

tests/linen/module_test.py
avital: OK done -- pushed a new commit that fixes the three remaining cases where `if scope is None` wasn't checked.

flax/linen/module.py
andsteing: Might be more code, but would make the `Module` code leaner and logically separate these checks into a proxy. And also protect against any such access in the future... Anyways, not feeling strongly about this, just throwing around an idea ;) Both solutions LGTM

PR #655, opened 2020-11-20
Improve frozendict performance

jheek:
- avoid unnecesary deep copies
- use jax.tree_map for faster deep copies.

### DIFF
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -159,7 +159,8 @@ def unfreeze(x: FrozenDict[K, V]) -> Dict[K, V]:
   if isinstance(x, FrozenDict):
     # deep copy internal state of a FrozenDict
     # the dict branch would also work here but
-    # it is much less performant.
+    # it is much less performant because jax.tree_map
+    # uses an optimized C implementation.
     return jax.tree_map(lambda y: y, x._dict)
   elif isinstance(x, dict):
     ys = {}
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -37,10 +37,13 @@ class FrozenDict(Mapping[K, V]):
   """An immutable variant of the Python dict."""
   __slots__ = ('_dict', '_hash')
 
-  def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


flax/core/frozen_dict.py
```diff
@@ -151,12 +156,18 @@ def unfreeze(x: FrozenDict[K, V]) -> Dict[K, V]:
   Makes a mutable copy of a `FrozenDict` mutable by transforming
   it into (nested) dict.
   """
-  if not isinstance(x, (FrozenDict, dict)):
+  if isinstance(x, FrozenDict):
+    # deep copy internal state of a FrozenDict
+    # the dict branch would also work here but
+    # it is much less performant.
```
avital, line 162: Can you please add a tiny additional clarification -- this is due to special implementation in JAX core written in C for tree_map of specifically `dict` instances, is that right?

avital (approved):
Have you tested how this impacts the performance of the ImageNet example?

andsteing:
General remark on `FrozenDict`: The following test would fail because the `list` is not copied:

```python
  def test_frozen_dict_copies(self):
    xs = {'a': 1, 'b': {'c': [1, 2]}}
    frozen = freeze(xs)
    xs['a'] += 1
    xs['b']['c'][0] += 1
    self.assertEqual(unfreeze(frozen), {'a': 1, 'b': {'c': [1, 2]}})
```

Should we check at some point that no other mutable container than `dict` can be used?

jheek:
> General remark on `FrozenDict`: The following test would fail because the `list` is not copied:
> 
> ```python
>   def test_frozen_dict_copies(self):
>     xs = {'a': 1, 'b': {'c': [1, 2]}}
>     frozen = freeze(xs)
>     xs['a'] += 1
>     xs['b']['c'][0] += 1
>     self.assertEqual(unfreeze(frozen), {'a': 1, 'b': {'c': [1, 2]}})
> ```
> 
> Should we check at some point that no other mutable container than `dict` can be used?

Yes, we can make a separate PR that bans lists in particular I think.

flax/core/frozen_dict.py
jheek: done

PR #662, opened 2020-11-20
Add the centered version of RMSProp

psc-g:
This is based on the RMSProp implementation in TF.Keras: https://github.com/tensorflow/tensorflow/blob/v2.3.1/tensorflow/python/keras/optimizer_v2/rmsprop.py

Having a centered RMSProp is necessary for proper performance of certain algorithms, such as the original DQN: https://github.com/google/dopamine/blob/master/dopamine/agents/dqn/configs/dqn_nature.gin#L23

### DIFF
--- a/flax/optim/rmsprop.py
+++ b/flax/optim/rmsprop.py
@@ -58,7 +58,8 @@ class RMSProp(OptimizerDef):
 
   def init_param_state(self, param):
     """Initialize parameter state"""
-    return _RMSPropParamState(jnp.zeros_like(param), jnp.zeros_like(param))
+    mg = jnp.zeros_like(param) if self.hyper_params.centered else None
+    return _RMSPropParamState(jnp.zeros_like(param), mg)
 
   def apply_param_gradient(self, step, hyper_params, param, state, grad):
     """Apply per-parameter gradients"""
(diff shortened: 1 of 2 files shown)


flax/optim/rmsprop.py
```diff
@@ -25,42 +25,55 @@ class _RMSPropHyperParams:
...
+        expensive in terms of computation and memory. Defaults to `False`.
     """
-    hyper_params = _RMSPropHyperParams(learning_rate, beta2, eps)
+    hyper_params = _RMSPropHyperParams(learning_rate, beta2, eps, centered)
     super().__init__(hyper_params)
 
   def init_param_state(self, param):
     """Initialize parameter state"""
-    return _RMSPropParamState(jnp.zeros_like(param))
+    return _RMSPropParamState(jnp.zeros_like(param), jnp.zeros_like(param))
```
jheek, line 61: This seems wasteful to me. If centered is `False` I would assume we simple set `mg = None`

jheek (changes requested):
Hi thanks for your PR, I think the implementation is wasting some memory. Let me know if you think this makes sense

flax/optim/rmsprop.py
psc-g: Done.
I've also added tests for the centered version.
TALP?

jheek (approved):
Thanks LGTM

PR #663, opened 2020-11-20
avoiding lambda redefinition and recompilation

mbz:


linen_examples/imagenet/imagenet_lib.py
```diff
@@ -210,10 +210,12 @@ def save_checkpoint(state, model_dir):
     step = int(state.step)
     checkpoints.save_checkpoint(model_dir, state, step, keep=3)
 
+def avg_across_axis(x, axis_name='x'):
```
jheek, line 213: Why not pmap this function right away?

so

```
@functools.partial(jax.pmap, axis_name='x')
def avg_across_axis(x):
  return lax.pmean(x, 'x')
```

this way it's a bit more clear that the pmap will be compiled just once.
Otherwise we need to really on some smart caching behavior.

jheek (changes requested):
Thanks for your PR. I have a suggestion for making the solution even more explicit

avital:
@mbz do you have time to respond to Jonathan's suggestion?

mbz:
This has been fixed in the latest update. Closing it.

jheek:
I think this should not be closed. The PR is ready to merge right?

mbz:
It seems it is fixed here, no?
https://github.com/google/flax/blob/master/linen_examples/imagenet/train.py#L212

jheek:
no because the pmap and lambda are still created every time `sync_batch_stats` is called

mbz:
oh OK, let's merge it then :) apparently I cannot merge ...

jheek:
There's a merge conflict. If you could fix it we can approve the merge

jheek:
this was fixed as part of PR #1207

PR #666, opened 2020-11-23
Fixed typos and missing attribute in module docstrings.

afqueiruga:
s/bias/use_bias/ in Dense and DenseGeneral, add dtype attribute to Embed, and nit that every line is lower case.


andsteing:
Thanks!

andsteing (approved):

PR #671, opened 2020-11-24
Improves imagenet examples file structure

andsteing:
Adheres to new filestructure from #231 (#480).

### DIFF
--- a/linen_examples/imagenet/main.py
+++ b/linen_examples/imagenet/main.py
@@ -35,12 +35,8 @@ import train
 FLAGS = flags.FLAGS
 
 flags.DEFINE_string(
-    'model_dir', default=None,
+    'workdir', default=None,
     help=('Directory to store model data.'))
-
-flags.DEFINE_string(
-    'data_dir', default=None,
-    help='Tensorflow datasets directory.')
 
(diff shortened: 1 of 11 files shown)


linen_examples/imagenet/main.py
```diff
@@ -37,6 +38,10 @@
     'model_dir', default=None,
     help=('Directory to store model data.'))
 
+flags.DEFINE_string(
```
Marvin182, line 41: Can we use the TFDS_DATA_DIR environment variable instead?

linen_examples/imagenet/main.py
```diff
@@ -37,6 +38,10 @@
     'model_dir', default=None,
```
Marvin182, line 38: Other examples use --workdir instead of --model_dir. Can we converge to --workdir everywhere?

Marvin182 (changes requested):
LGTM apart from the flags.

linen_examples/imagenet/main.py
andsteing: Done. Also updated https://github.com/google/flax/issues/231#issuecomment-716568710

PR #672, opened 2020-11-24
Adding missing docs in RTD

BertrandRdp:
Added optimizers missing from RTD and a traverse_util doc.
(Also fixes a few formating issues ad added auto include if constructor docstring in autoclass directive).


docs/flax.optim.rst
```diff
@@ -1,6 +1,9 @@
 
 flax.optim package
-=================
+===================
+
+Optimizers utilities
```
avital, line 5:
```suggestion
Optimizer core
```

docs/flax.traverse_util.rst
```diff
@@ -0,0 +1,44 @@
...
+
+.. automodule:: flax.traverse_util
+
+Traversal objects
+-------------------
+
+.. autoclass:: Traversal
+    :members:
+
+.. autoclass::TraverseId
```
avital, line 14: Is this missing a space?

avital: But actually, I don't think we should document these -- I consider them internal implementation details. People really only use `ModelParamTraversal`

avital (approved):
Added some small comments, but LGTM

docs/conf.py
```diff
@@ -92,6 +92,8 @@
 
 autodoc_typehints = 'description'
 
+autoclass_content = 'both'
```
avital, line 95: Are you sure we want this? This would impact all of our docstrings, are you sure it's better like that?

docs/flax.traverse_util.rst
BertrandRdp: So I asked Jonathan before adding those and he thought it would be better to document those. I don't mind either way, I also see your point (this is why I asked in the first place).

docs/conf.py
BertrandRdp: It depends. The thing is that we aren't consistent (at least in the optim package but I expect it's the case elsewhere) in how we split between class and __init__ docstrings. We can not have this, but in some cases here we are missing some docs. Alternative is changing docstrings accordingly (can do for optim). Let me ask others

marcvanzee:
@BertrandRdp once this PR is merged, can you go over #733 with @8bitmp3 and see whether there are still changes in the PR that can be merged?

BertrandRdp:
Working on it right now, will merge and work on the other one!

PR #683, opened 2020-11-27
Add call override for use_running_average

jheek:


avital:
Should we just get rid of the `use_running_average` attribute entirely?

jheek:
That was my first guess but I left it there because we use this very handy pattern in ImageNet:
```
norm = partial(nn.BatchNorm, use_running_average=not train, momentum=, eps=, ..).
```

Typically you have 1 template for a batchnorm layer throughout a network with a bunch of hyper params. This way you don't have to propagate train all throughout the model.

avital:
Hmm. I haven't thought about this too carefully, but I'm uneasy about having two different way to specify `use_running_average` for Batch Norm. That seems not so clean, as well as a footgun for users ("but I specified use_running_average! why isn't it using my setting?")

If the `partial` use-case is strong, maybe that's an argument against this PR. What is the argument in favor of it? (I didn't see any motivation in the description of this PR, maybe we should make sure to have those)

jheek:
If you specify a BatchNorm in setup you have to pass `use_running_average` in call because you don't know whether `train` is True/False.

I don't think it's a footgun because it is usually passed as `use_running_average=not train` and you either can't pass it because you don't have train yet or you define it inside compact and you can pick either.

But it is definitely a little tricky. Partially applying modules kind of blur the line between hyper parameters and call arguments.

avital:
I see. The goal is to support top-level modules that: (1) use `setup` and (2) where the training code only instantiates one module instance for both train and test time, e.g.

```
class MyModule(nn.Module):
  def setup(self):
    self.bn1 = BatchNorm(...)

  def __call__(self, x, is_training):
    return self.bn1(x)  # how do you use `is_training` here?
```

So in the case where someone develops entirely with submodules defined in `setup`, does this imply that all attributes end up needing to be overridable in `__call__`?

---

The footgun note was more about something like the following (which we could solve by having the default attribute value for `use_running_average` be a sentinel object like `UNDECIDED`, and then only let you specify a value in `__call__` if the instance attribute value is `UNDECIDED`:

```
class MyModule(nn.Module):
  def setup(self):
    # a user may look at this and be surprised that the batch norm ends up not
    # using the running average (especially if the module declaration is long and it's
    # not easy to find where `self.bn1` is used (or maybe in multiple places)
    self.bn1 = BatchNorm(use_running_average=True)

  def __call__(self, x):
    return self.bn1(x, use_running_average=False)
```

avital:
BTW somehow PyTorch only lets you define the equivalent attribute in the constructor to BatchNorm (https://pytorch.org/docs/stable/_modules/torch/nn/modules/batchnorm.html#BatchNorm2d) -- and they "define all submodules during setup" (in Flax speak). How do they get around this problem? Do people just create two different module instances at top level training code for Train vs Test time?

jheek:
Pytorch has the implicit training=True/False argument in every `Module`. So it will still only track it if you are in training mode

avital:
Oh wow. They ask you to globally, mutably, set your module into eval mode (https://discuss.pytorch.org/t/example-on-how-to-use-batch-norm/216/19)...

avital:
I would like for us to get additional opinions about this before we go down this route. As I wrote in https://github.com/google/flax/pull/683#issuecomment-737182948, I am afraid that more and more attributes will end up be such that you want to be able to override them this way, and this pattern infests our module implementations.

If the reasoning is solely to let folks define a single module instance at their top-level that they can use for train or eval, then maybe there's a simpler solution which is to define, at the top-level:

```
def MyModule(nn.Module):
  train: bool
  num_features: int

  def setup(self):
    self.bn1 = BatchNorm(use_running_average=not self.train, [...])

  def __call__(self, x):
    [...]
    x = self.bn1(x)

MyModule = ResNet.partial(num_features=1000)
MyModuleTrain = MyModule(train=True)
MyModuleEval = MyModule(train=False)
```

jheek:
This way we go back to the hyper parameters are just partially applied arguments pattern.
Both in the original abstraction & linen we could a lot of negative feedback on that.

avital:
I just think we'll also get negative feedback about this override pattern (I myself don't like it). So if we are going down this path I want to hear opinions of more folks, showing concretely the two options on the table.

rwightman:
@jheek @avital as someone who's new to JAX / FLAX, after getting used to things, I think either approach can work I just feel it should be consistent. Consistent in that I'd like to pass my train/not train state either exclusively through __call__ args or exclusively through dataclass attr but not a mix of both as is the case between BatchNorm vs Dropout right now.

If for some reason it is deemed that stochastic layers like dropout are different, and deterministic/non det state may change with more granularity, couldn't that also apply to batchnorm in some usecases where the phases of using running stats vs not might be finer grained than train / not train?

Essentially, why this (where I've wound up right now). 
```
def __call__(x, training: bool)
  ...
  x = self.norm_layer(name='bn', use_running_average=not training)(x)
  x = self.act_fn(x)
  x = x.mean((1, 2))
  x = Dropout(rate=self.drop_rate)(x, determinisitc=not training)
  ...
```

and not either of these (ignoring potential use of partial binding on purpose)

```
def __call__(x, training: bool)
  ...
  x = self.norm_layer(name='bn')(x, use_running_average=not training)
  x = self.act_fn(x)
  x = x.mean((1, 2))
  x = Dropout(rate=self.drop_rate)(x, determinisitc=training)
  ...

def __call__(x)
  ...
  x = self.norm_layer(name='bn', use_running_average=not self.train)(x)
  x = self.act_fn(x)
  x = x.mean((1, 2))
  x = Dropout(rate=self.drop_rate, determinisitc=not self.train)(x)
  ...
```

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+
+```python
+from functools import partial
+from flax import linen as nn
+
+class ResidualModel(nn.Module):
+  drop_rate: float
+
+  @nn.compact
+  def __call__(self, x, *, train):
```
avital, line 33: Does this syntax work? I thought you need a default value for an kwarg that comes after `*` maybe? (I could be wrong)

avital: (and does it need to be `*args`?)

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+    dropout = partial(nn.Dropout, rate=self.drop_rate, deterministic=not train)
+    for i in range(10):
+      x += ResidualBlock(dropout=dropout, ...)(x)
+```
+
+It makes sense to pass `determinstic` to the constructor here because this way we can pass the dropout template to the sub-modules.
+Now the sub-module no longer needs to take care of train vs eval mode and can simply use the `dropout` argument.
+Note that because the dropout layer can only be constructed in the sub-module we can only partially `deterministic` to the constructor but not to `__call__`.
+
+However, if deterministic is just an attribute we run into trouble when using the setup pattern:
```
avital, line 43:
```suggestion
However, if deterministic is an attribute we run into trouble when using the setup pattern. We would **want** to write our module code like this:
```

docs/notes/parameters.md
avital: Oh, this is just, `...`? Like arbitrary other arguments?

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+    self.dropout = nn.Dropout(rate=self.drop_rate)
+
+  @nn.compact
+  def __call__(self, x, *, train):
+    # ...
+    x = self.dropout(x, deterministic=not train)
+    # ...
+```
+
+Here it makes sense to pass `deterministic` during `__call__` because it depends on the `train` argument.
```
avital, line 59:
```suggestion
But as defined above `deterministic` would be an attribute, so this doesn't work. Here it makes sense to pass `deterministic` during `__call__` because it depends on the `train` argument.
```

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Solution
+
+We can support both use cases described before by allowing certain paramaters to be passed
```
avital, line 63: parameters -> properties

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+as attribute or as method argument (but not both!).
+This can be implemented as follows:
+```python
+class MyDropout(nn.Module):
+  drop_rate: float
+  deterministic: Optional[bool] = None
+
+  @nn.compact
+  def __call__(self, x, deterministic=None):
+    deterministic = nn.merge_param('deterministic', self.deterministic, deterministic)
```
avital, line 73: What purpose does the first `'deterministic'` string argument serve?

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+    # ...
+```
+
+In this example `nn.merge_param` will ensure that either `self.deterministic` or `deterministic` is set but not both.
+An error is raised if both values are `None` or both values are not `None`.
+This avoids confusing behavior where 2 different parts of the code set the same paramater and one is overruled by the other.
```
avital, line 79: parameter (e not a)

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+It also avoids a default value which would problably cause either the train step or eval step of a training procedure to be broken by default.
+
+
+
+## Functional Core and flax.nn
+
+The old NN api and functional core define functions rather than classes.
+Therefore, there is no clear distinction between hyper paramaters and call time arguments.
```
avital, line 87: parameters

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+The only way to pre-determine the hyper paramaters is by using `partial`.
```
avital, line 88: parameters

docs/notes/parameters.md
```diff
@@ -0,0 +1,89 @@
...
+On the upside, there are no ambiguis cases where method arguments could also be attributes.
```
avital, line 89: Ambiguous

avital (commented):
Hi @jheek thanks for writing up https://github.com/google/flax/pull/683/files/5e1a7c24787f498cbd24f81d3516ef3a52e2adb8#diff-a4e69f05db605d788793032faac416daf270becbb42ef8abd8f08c69946bd373.

1. I suggested changes to the language to make the different terms "parameter" "property" "hyperparameter" "attribute" "argument" more precise, and other clarifications to improve readability. PTAL.
2. I think perhaps we should make these live in Read The Docs? And add a notes/ directory with a README that just points there?

docs/notes/parameters.md
jheek: This is the syntax for a required keyword argument.

docs/notes/parameters.md
jheek: That's for a helpful error message. I should maybe make that optional though.

docs/notes/parameters.md
avital: You learn something every day!

levskaya:
@rwightman - for myself I tend to agree with you that I prefer consistency in applying this sort of thing, and in practice so far haven't found myself wanting to exploit a finer degree of granularity for things like dropout / batchnorm, but I'm curious if there are usecases for such that demand it that I'm unaware of.

@jheek @avital - In a lot of recent model development for large language models I have adopted the pattern of defining several top-level configurations for a model:  TrainModel / EvalModel / PredictModel that sets these train/eval/infer-time options across the model.  (I do so by typically aggregating "global" model options in a single dataclass that's threaded through the entirety of the "user-defined" portions of the model and actually have separate train/eval/predict configs, but this is an implementation detail.)

rwightman:
@avital One way to do things, or at least a canonical way certainly makes for less confusion.

@levskaya Are any of those 'large language models', or models using that pattern open-source and available to look at? Helpful to digest more examples...

For the cases I'm thinking of where one might definitely need to pass something through __call__ args to decay dropout rates, they are dropout(ish) layers that'd need their own impl anyways. For the majority of use cases I can think of with a normal dropout / BN layers the 'phases' are fairly coarse, even if they don't align with train/not train -- different top-level modules would suffice.

One thing I did notice with existing layers / examples is that the `train`, `use_running_avg`, `deterministic` arguments or attr often have default values. I'd recommend they don't for the simple reason that it can prevent a lot of hard to track down bugs. One big gotcha in PyTorch that has generated a lot of forum / issue tracker activity over the years is forgetting to flip the train/eval state of your models. Requiring the explicit setting or binding of those values can prevent bugs where all (or more insiduously, just a few) such layers in the model are in the wrong mode. I feel the verbosity vs ARRG tradeoff there is worth it.

jheek:
I like the dataclass config pattern especially for larger models.

I'm not sure about having multiple top level models though. It's not great for the functional training loop patterns because you cannot easily write your training+eval loop with just an apply function now (important for interop with other Jax libraries).

The second issue is interactive mode and composability. What if I want to dropout or use_running_average one some inputs but not others (let's say in a GAN discriminator)? Or how would I switch the mode if I have an interactive module instance?

@rwightman I fully agree that there should be no default for these mode like arguments.

patrickvonplaten:
Hey everybody, could we maybe use this PR to add `deterministic` as an input param to `MultiHeadDotProductAttention`'s `__call__` method?

We could replace this line: https://github.com/google/flax/blob/36eaa1effc9b24aa14f09c217cdd92bf3dfc105e/flax/linen/attention.py#L154 

by 
```python
  def __call__(self,
               inputs_q: Array,
               inputs_kv: Array,
               mask: Optional[Array] = None
               deterministic: Optional[bool] = None):
  ...
  deterministic = nn.merge_param('deterministic', self.deterministic, deterministic)
```

? 

It would solve the problem described in https://github.com/google/flax/issues/821 we're having at the moment for Flax Transformers

CHANGELOG.md
```diff
@@ -15,7 +15,8 @@ vNext
  -
  -
  -
- -
+ - Some Module arguments should now be passed either as dataclass attribute or
```
avital, line 18:
```suggestion
 - Some Module arguments can now be passed either as dataclass attribute or
```

docs/design_notes/arguments.md
```diff
@@ -0,0 +1,91 @@
+# Dealing with Module Arguments
+
+## Introduction
+
+In Linen we can define `Module` arguments either as dataclass attributes or as arguments to methods (usually `__call__`).
+Typically the distinction is clear:
+* Completely fixed properties, such as the choice of kernel initializer or number of output features, are hyperparameters and should be defined as dataclass attributes. Typically two Module instances with different hyperparamaters cannot share in a meaningfull way.
```
avital, line 7:
```suggestion
* Completely fixed properties, such as the choice of kernel initializer or number of output features, are hyperparameters and should be defined as dataclass attributes. Typically two Module instances with different hyperparamaters cannot share in a meaningful way.
```

docs/design_notes/arguments.md
```diff
@@ -0,0 +1,91 @@
...
+  @nn.compact
+  def __call__(self, x, *, train):
+    dropout = partial(nn.Dropout, rate=self.drop_rate, deterministic=not train)
+    for i in range(10):
+      x += ResidualBlock(dropout=dropout, ...)(x)
+```
+
+It makes sense to pass `determinstic` to the constructor here because this way we can pass the dropout template to the sub-modules.
+Now the sub-module no longer needs to take care of train vs eval mode and can simply use the `dropout` argument.
+Note that because the dropout layer can only be constructed in the sub-module we can only partially `deterministic` to the constructor but not to `__call__`.
```
avital, line 42:
```suggestion
Note that because the dropout layer can only be constructed in the sub-module we can only partially apply `deterministic` to the constructor but not to `__call__`.
```

flax/linen/attention.py
```diff
@@ -142,7 +144,7 @@ class MultiHeadDotProductAttention(Module):
   out_features: Optional[int] = None
   broadcast_dropout: bool = True
   dropout_rate: float = 0.
-  deterministic: bool = False
+  deterministic: Optional[bool] = None
```
avital, line 147: Should this be a separate CHANGELOG entry? i.e. "Remove deterministic default for ..."

flax/linen/normalization.py
```diff
@@ -60,7 +60,7 @@ class BatchNorm(Module):
       the examples on the first two and last two devices. See `jax.lax.psum`
       for more details.
   """
-  use_running_average: bool = False
+  use_running_average: Optional[bool] = None
```
avital, line 63: ditto here around perhaps a new CHANGELOG entry?

flax/linen/stochastic.py
```diff
@@ -15,23 +15,31 @@
...
+        masked, whereas if true, no mask is applied and the inputs are returned as
+        is.
   """
   rate: float
+  broadcast_dims: Iterable[int] = ()
+  deterministic: Optional[bool] = None
 
   @compact
-  def __call__(self, inputs, deterministic=False, rng=None, broadcast_dims=()):
+  def __call__(self, inputs, deterministic: Optional[bool] = None, rng=None):
```
avital, line 42: I have no context here, but is it obviously fine to remove broadcast_dims from the arguments? Maybe worth checking if any users on GitHub or Google are using this argument?

avital (approved):
Added a few small comments/suggestions, but generally LGTM!

flax/linen/stochastic.py
jheek: Will check before merging

PR #684, opened 2020-11-27
Fix non-init module attributes

jheek:
This fixes the following:
```
class Foo(nn.Module):
      bar: nn.Dense = dataclasses.field(init=False)

      def setup(self):
        self.bar = nn.Dense(2)
      
      @nn.compact
      def __call__(self, x):
        return self.bar(x)
```

because bar is a dataclass attributes the `__setattr__` logic that binds the scope didn't fire

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -78,7 +78,7 @@ def _module_repr(module: 'Module', num_spaces: int = 4):
     rep += '# attributes\n'
     for attr in attributes.keys():
       # TODO(jheek): can we get a nice string representation of attribute types?
-      value = getattr(module, attr)
+      value = getattr(module, attr, None)
       value_rep = _attr_repr(value)
       rep += f'{attr} = {value_rep}\n'
   if child_modules:
@@ -412,7 +412,7 @@ class Module:
     if name == 'parent':
       pass
     # Modules have been passed in as dataclass args.
-    elif name in self.__dataclass_fields__.keys():  # pytype: disable=attribute-error
+    elif name in self.__dataclass_fields__ and self.__dataclass_fields__[name].init:  # pytype: disable=attribute-error
       pass
     # Submodules are being defined and attached in setup()
     else:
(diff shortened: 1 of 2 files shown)


flax/linen/module.py
```diff
@@ -409,10 +409,12 @@ def __setattr__(self, name: str, val: Any):
       val: Value of the attribute.
     """
     # We don't mess with the parent module.
+    if name in self.__dataclass_fields__:
+      print(name, self.__dataclass_fields__[name].init, val)
```
avital, line 413: This `print` seems out of place?

flax/linen/module.py
```diff
@@ -432,6 +434,7 @@ def __setattr__(self, name: str, val: Any):
                 "using keyword argument name=\"<name>\"")
           subvalue.name = f'{name}{suffix}'
           subvalue.__post_init__()
+          print('post_init')
```
avital, line 437: ditto

tests/linen/module_test.py
```diff
@@ -668,6 +670,23 @@ def bar(self):
...
       empty.bar()
 
+  def test_module_with_attrs(self):
+    class Foo(nn.Module):
+      bar: nn.Dense = dataclasses.field(init=False)
+
+      def setup(self):
+        self.bar = nn.Dense(2)
+      
+      @nn.compact
```
avital, line 680: probably better to remove nn.compact here as it's unrelated to the thing being tested

tests/linen/module_test.py
```diff
@@ -668,6 +670,23 @@ def bar(self):
...
+      def __call__(self, x):
+        return self.bar(x)
+
+    foo = Foo()
+    x = jnp.ones((2,))
+    variables = foo.init(random.PRNGKey(0), x)
+    self.assertEqual(variables['params']['bar']['kernel'].shape, (2, 2))
```
avital, line 687: FWIW I try to avoid expected shapes like (2, 2) because if you mixed the order it'd still be right. So I'd have `x` have shape `(3, )`. But this is a bit nit-picky

avital (commented):
Generally LGTM but take a look at my comments

flax/linen/module.py
jheek: oops

tests/linen/module_test.py
jheek: I agree

tests/linen/module_test.py
jheek: done

avital (approved):

PR #696, opened 2020-11-30
Fix setup in transformed class

jheek:
setup was called outside of transform
which made parameters inside setup
misbehave.

See test case for example


levskaya (approved):
This is great - thank you so much for the fix.

PR #699, opened 2020-12-01
Add Adadelta Optimizer

kamalesh0406:


kamalesh0406:
@googlebot I signed it!

PR #700, opened 2020-12-01
Add Adadelta Optimizer

kamalesh0406:
I have added the adadelta optimizer (https://arxiv.org/abs/1212.5701).


kamalesh0406:
@levskaya can you review my pull request? Thanks!

avital:
@levskaya do you have a minute to take a look at this?

flax/optim/adadelta.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+    rho = hyper_params.rho
+    eps = hyper_params.eps
+    weight_decay = hyper_params.weight_decay
+
+    if hyper_params.weight_decay != 0:
+        grad = param + weight_decay
+
+    sq_avg = rho * (state.sq_avg + jnp.multiply(grad, grad) * (1- rho))
```
levskaya, line 71: shouldn't this be `rho * state.sq_avg +  (1- rho) * jnp.multiply(grad, grad)`? (i.e. the first rho shouldn't be outside)

flax/optim/adadelta.py
```diff
@@ -0,0 +1,79 @@
...
+    std = jnp.sqrt(sq_avg + eps)
+    delta = jnp.divide(jnp.sqrt(state.acc_delta+eps)/std)
+
+    new_param = param - hyper_params.learning_rate * grad * delta
+    acc_delta = rho * (state.acc_delta + jnp.multiply(delta, delta)*(1 - rho))
```
levskaya, line 76: same here, this should just be a mixing expression, no?: `rho * state.acc_delta + (1 - rho) * jnp.multiply(delta, delta)`

flax/optim/adadelta.py
```diff
@@ -0,0 +1,79 @@
...
+    assert hyper_params.eps is not None, 'no eps provided.'
+    assert hyper_params.weight_decay is not None, 'no weight decay provided.'
+
+
+    rho = hyper_params.rho
+    eps = hyper_params.eps
+    weight_decay = hyper_params.weight_decay
+
+    if hyper_params.weight_decay != 0:
+        grad = param + weight_decay
```
levskaya, line 69: This can't be right, if the weight decay is nonzero we don't want to replace the gradient with the parameter value!

Typically weight decay is implemented _directly_ in nontrivial optimizers with accumulated variables, i.e. we'd move the weight decay to the very end, _after_ performing the adadelta update:
```
new_param -= hyper_params.learning_rate * weight_decay * param
```
here we make weight decay relative to the learning rate, rather than specifying its absolute magnitude, as this is the convention in other flax optimizers, eg. Adam.

flax/optim/adadelta.py
```diff
@@ -0,0 +1,79 @@
...
+
+    sq_avg = rho * (state.sq_avg + jnp.multiply(grad, grad) * (1- rho))
+    std = jnp.sqrt(sq_avg + eps)
+    delta = jnp.divide(jnp.sqrt(state.acc_delta+eps)/std)
```
levskaya, line 73: This won't work as written, `jnp.divide` takes two arguments.  `delta` could simply be written `jnp.sqrt(state.acc_delta+eps) / jnp.sqrt(sq_avg + eps)`

flax/optim/adadelta.py
```diff
@@ -0,0 +1,79 @@
...
+class Adadelta(OptimizerDef):
+  """Adadelta optimizer"""
+  def __init__(self, learning_rate: float = None, rho=0.9, eps=1e-8, weight_decay=0):
+    """Constructor for the Adadelta optimizer.
+        
+    Args:
+      learning_rate: the step size used to update the parameters.
+      rho: coefficient used for computing a running average
+      eps: term added to the denominator to improve numerical stability
+    """
```
levskaya, line 48: missing comment for weight decay.

flax/optim/adadelta.py
```diff
@@ -0,0 +1,79 @@
...
+  def init_param_state(self, param):
+    """Initialize parameter state"""
+    return _AdadeltaParamState(jnp.zeros_like(param), jnp.zeros_like(param))
+
+  def apply_param_gradient(self, step, hyper_params, param, state, grad):
+    """Apply per-parameter gradients"""
+    assert hyper_params.learning_rate is not None, 'no learning rate provided.'
+    assert hyper_params.rho is not None, 'no rho provided.'
+    assert hyper_params.eps is not None, 'no eps provided.'
+    assert hyper_params.weight_decay is not None, 'no weight decay provided.'
```
levskaya, line 61: Since defaults are provided for rho, eps, weight_decay I'm not sure if we really need these asserts against None.

levskaya (changes requested):
Hi, sorry for the slow review!

There's some issues with the implementation, I provided some detailed comments inline.  I'd also request that you add a unit test for this optimizer so that we know it actually runs.  Thanks!

flax/optim/adadelta.py
kamalesh0406: I am resolving these issues, thanks for checking it!

kamalesh0406:
@levskaya I have corrected the mistakes and added the unit tests. Can you review it now?

levskaya (approved):
I think this looks ok now.

levskaya:
@kamalesh0406 - btw, in the future it's good practice to make a github PR from a working branch in your forked repo so that maintainers can do merge commits in a PR rather than your master branch.

PR #702, opened 2020-12-02
Re-add flax install to doc requirements

jheek:


andsteing:
Note that we now get the following error:

```
$ cd docs; pip install -r requirements.txt 
ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
```

Would it work to specify `flax` instead of `.` ?

Otherwise could you add a note to `README.md` that for local installation on a developer machine one actually needs to temporarily comment that `.` line?

docs/requirements.txt
```diff
@@ -9,3 +9,5 @@ recommonmark
 # The next packages are for notebooks
 matplotlib
 sklearn
+# Must install flax itself for notebook execution and auotdocs to work
+.
```
andsteing, line 13: Note that we now get the following error:

```
$ cd docs; pip install -r requirements.txt 
ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found.
```

Would it work to specify `flax` instead of `.` ?

Otherwise could you add a note to `README.md` that for local installation on a developer machine one actually needs to temporarily comment that `.` line?

andsteing (changes requested):

docs/requirements.txt
jheek: The blessing of relative paths. You need to call `pip instatll -r docs/requirements.txt`. It has to be a local install otherwise you will not build the docs of the actual checkout

jheek: I have added the correct command in the README

andsteing: Ah ok maybe it's obvious then.

andsteing (approved):

Issue #705, opened 2020-12-04
Requesting support for arbitrary pytrees of Module params 


j-towns:
As mentioned in the docs, `Module.__setattr__` allows for assigning arbitrary pytrees of submodules in `setup`. I would like to assign a list of `param`s during setup. The following fails:
```python
from jax import random
import flax.linen as nn


class Foo(nn.Module):
    def setup(self):
        self.param_list = [
            self.param('param_list', nn.initializers.zeros, (1, 2, 3))
            for _ in range(4)
        ]

foo = Foo()
foo.init({'params': random.PRNGKey(0)})
```
with
```text
~/dev/flax/flax/linen/module.py in param(self, name, init_fn, *init_args)
    614           'wrapped in `@compact`')
    615     if self._name_taken(name):
--> 616       raise ValueError(
    617           f'Name {name} already in use in {self.__class__.__name__}.')
    618     # ephemeral state for setattr name-equality-check

ValueError: Name param_list already in use in Foo.
```
Setting the names of each list element manually also fails:
```python
class Bar(nn.Module):
    def setup(self):
        self.param_list = [
            self.param(f'param_list_{i}', nn.initializers.zeros, (1, 2, 3))
            for i in range(4)
        ]

bar = Bar()
bar.init({'params': random.PRNGKey(0)})
```
this gives a different error though:
```
~/dev/flax/flax/linen/module.py in __setattr__(self, name, val)
    439           # namecheck to ensure named variable matches self attribute name.
    440           if self._state.last_varname and self._state.last_varname != var_name:
--> 441             raise ValueError(f'Variable name {self._state.last_varname} must '
    442                              f'equal attribute name {var_name}.')
    443           self._state.last_varname = None

ValueError: Variable name param_list_3 must equal attribute name param_list_0.
```
Would it be possible for an appropriate name string to be automatically inferred rather than being supplied by the user? I'm sure there's a good reason that it is the way it is, apologies if this has already been brought up elsewhere and I've missed it, just curious.

j-towns:
A workaround for lists is to define a list of name strings, and then iterate over it, assigning params directly to the Module. This seems to work but obviously is quite cumbersome.
```python
class Bar(nn.Module):
    def setup(self):
        param_names = [f'param_list_{i}' for i in range(4)]
        for n in param_names:
            self.__setattr__(
                n, self.param(n, nn.initializers.zeros, (1, 2, 3)))
        self.param_names = param_names

    def __call__(self):
        return self.param_list_0

[...]

PR #706, opened 2020-12-04
Relax parameter attribute check

avital:
We currently try to enforce that a user can't do:

```py
class MyModule(nn.Module):
  def setup(self):
    self.name1 = self.param('name2', ...)
```

But the check is brittle and entirely disallows assigning dicts or lists of parameters. This check relaxes our heuristic such that it doesn't run at all for lists or dicts of parameters assigned to attributes during setup.

See https://github.com/google/flax/issues/705#issuecomment-738761853 for a bit more context.


jheek (approved):
LGTM

Issue #707, opened 2020-12-04
Publish pretrained models for examples


andsteing:
Publishing pretrained examples on a public GCS bucket would help users to quickly get started with trained models for finetuning, for using them directly for inference etc.

marcvanzee:
Closing this since this is complete.

PR #694, opened 2020-11-30
Adds linen_exampels/cloud/

andsteing:
This new directory contains scripts for launching examples "fire and forget" style on GCS. Verified with `MNIST` and `imagenet` examples so far.

See #707

PR #708, opened 2020-12-04
Updates imagenet,mnist examples with result tables

andsteing:
As part of #707

Examples were run with code from #694


myagues:
There is a typo in [`linen_examples/imagenet/train.py#L286`](https://github.com/google/flax/blob/9a0b06d24add079c9605e4ccb3cd36a370b56773/linen_examples/imagenet/train.py#L286), when defining `steps_per_eval`, that causes the model to iterate over validation data multiple times. This may explain the increased training time compared to [`examples/imagenet/imagenet_lib.py#L264`](https://github.com/google/flax/blob/9a0b06d24add079c9605e4ccb3cd36a370b56773/examples/imagenet/imagenet_lib.py#L264), where `steps_per_eval` is correctly set.

linen_examples/imagenet/README.md
```diff
@@ -10,10 +10,10 @@ This example uses linear learning rate warmup and cosine learning rate schedule.
...
+| Name                    |   Steps | Walltime   | Top-1 accuracy   | Metrics                                                                                                                               | Workdir                                                                                                                                                              |
+|:------------------------|--------:|:-----------|:-----------------|:--------------------------------------------------------------------------------------------------------------------------------------|:---------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| v100_x8                 |  250199 | 18.0h      | 76.62%           | [tfhub.dev](https://tensorboard.dev/experiment/6yEYynrdRS2HZqHAD2ejPQ/#scalars&_smoothingWeight=0&regexInput=v100_x8)                 | [gs://flax_public/examples/imagenet/v100_x8](https://console.cloud.google.com/storage/browser/flax_public/examples/imagenet/v100_x8)                                 |
```
avital, line 15: It really is concerning that this got so much slower. Do we have any lead on why?

avital (approved):
Please see my comment on the ImageNet training getting slower. Someone should investigate this. Do you know if it's happening? Or should we find someone to own this investigation

linen_examples/imagenet/README.md
andsteing: Rerunning it with #717 as suggested by @myagues above... Will update here when the results are ready.

jheek: A this explains why I could not find anything during profiling. The steps are fast there’s just too many

andsteing: Ok that fixed it :) We're now even faster & more accurate (at least in this run ;-)

andsteing:
> Oh! So do we match the pre-Linen walltime now?

We surpassed it !

PR #711, opened 2020-12-07
Add early stopping to flax.training

gmittal:
Basic early stopping object to prevent overfitting during training. The API could probably be improved and I'm open to suggestions.

Discussion for feature request: #728


jheek:
Hi, thanks for your proposal!
Because this introduces a completely new feature we would like you to open an issue to describe the Feature Request first.
This way we can first discuss whether it is in scope of `flax.training` and what the api should look like.

My suggestion would be to consider the following though:
We generally try to avoid "inversion of control" APIs (a somewhat vague term). In this case the EarlyStopping class takes control of checkpointing even though it only adds the logic to decide whether we should stop or not.
I think a proposal that decouples checkpointing from the early stopping criteria (I imagine there are many criteria like this) is more likely to be accepted.

I hope that helps!

flax/training/early_stopping.py
```diff
@@ -0,0 +1,76 @@
...
+      step: int or float: training step number or other metric number.
+      metric: int or float: metric (i.e. validation loss) to determine 
+          improvement. 
+
+    Returns:
+      Filename of saved checkpoint or None if there was no improvement.
+    """
+    output = None
+    if self.best_loss - metric > self.min_delta:
+        output = checkpoints.save_checkpoint(self.ckpt_dir,
```
jheek, line 64: this is not pre-emption safe. You would have to store the other internal state in the checkpoint as well `(best_loss, count)`.

gmittal:
Is `self.replace` no longer the idiomatic/safe way to return a new instance of a `dataclass`?

jheek:
> Is `self.replace` no longer the idiomatic/safe way to return a new instance of a `dataclass`?

I see you are running into trouble with the pytype check that was recently introduced. It doesn't understand that the `@struct.dataclass` introduces the `replace` method. There's now a new base class that doesn't have this issue:

```
class EarlyStopping(flax.struct.PyTreeNode):
  ...
```

gmittal:
@jheek @avital any other feedback on this?

flax/training/early_stopping.py
```diff
@@ -0,0 +1,58 @@
...
+
+  def update(self, metric):
+    """Update the state based on metric.
+    
+    Returns:
+      Whether there was an improvement greater than min_delta from
+          the previous best_metric and the updated EarlyStop object.
+    """
+
+    if math.isinf(self.best_metric) or self.best_metric - metric > self.min_delta:
```
jheek, line 52: should the training stop when the loss diverges (inf or NaN)?

gmittal: This line contains the criteria for considering `metric` as an improvement over the best value seen so far. Since we initialize `best_metric = float('inf')`, any finite value for metric will be considered an improvement.

avital:
@jheek -- can this be merged?

Another request (for @gmittal or anyone else) -- after this is merged, it'd be great to add a HOWTO showing a side-by-side diff view using this, such as in https://flax.readthedocs.io/en/latest/howtos/extracting_intermediates.html (I can file an issue after this is merged to keep track of it)

jheek (approved):
All looks good! Sorry for the delay, I lost track of this.

PR #723, opened 2020-12-10
Correct typo in linen/module.py

AntoinePlumerault:
In the documentation, there is an 'import nn as linen' while it should be 'import linen as nn'.


AntoinePlumerault:
@googlebot I signed it!

avital (approved):
Nice catch, thanks! Please let us know of any other feedback and confusions!

PR #736, opened 2020-12-15
Adds flip/ for Flax improvement processes.

andsteing:
Adds the directory `flip/` that stores "Flax improvement processes".

Note that the `Type: FLIP` issue is currently called described as `Proposals for breaking or major API changes` - this will be changed to `Flax improvement process` later on to make the use of the word consistent.


flip/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+First, create an issue with the [FLIP label]. All pull requests that relate to
+the FLIP (i.e. adding the FLIP itself as well as any implementing pull requests)
+should be linked to this issue.
+
+Then create a pull request that consists of a copy of the `0000-template.md`
+renamed to `0000-your-flip.md` - the number changed should match the pull
+request that adds the FLIP.
+
+[FLIP label]: https://github.com/google/flax/issues?q=label%3A%22Type%3A+FLIP%22
```
avital, line 33: Oh? Is this a cool markdown feature I didn't know about -- is the [Flip label] above linked to this?

flip/README.md
```diff
@@ -0,0 +1,33 @@
+# FLIP : Flax Improvement Process
```
avital, line 1:
```suggestion
# FLIP: Flax Improvement Process
```

avital (commented):
Thanks for doing this, this is great!

One question about the location. We've been moving a lot of markdown files into docs/ -- should this live there?

(I suppose these really aren't user-facing and should not appear on the navigation bar on flax.readthedocs.org, so maybe markdown files in the main repo are indeed the right place)

flip/README.md
andsteing: exactly.

andsteing:
> Thanks for doing this, this is great!
> 
> One question about the location. We've been moving a lot of markdown files into docs/ -- should this live there?
> 
> (I suppose these really aren't user-facing and should not appear on the navigation bar on flax.readthedocs.org, so maybe markdown files in the main repo are indeed the right place)

SGTM - moved.

docs/flip/README.md
```diff
@@ -0,0 +1,33 @@
...
+  discussed themselves in the pull request adding the FLIP.
+
+## How to start a FLIP
+
+First, create an issue with the [FLIP label]. All pull requests that relate to
+the FLIP (i.e. adding the FLIP itself as well as any implementing pull requests)
+should be linked to this issue.
+
+Then create a pull request that consists of a copy of the `0000-template.md`
+renamed to `0000-your-flip.md` - the number changed should match the pull
```
gnecula, line 30: Some features may have multiple pull requests. Some people may want to include the

gnecula, line 30: Is it important to have the PR number part of the name? Can it be some other user-chosen name?

gnecula: In general, the prescriptions on formatting are pretty precise, and I am wondering which are important and which are not?

andsteing: (I used https://github.com/rust-lang/rfcs as a starting point where they also use this scheme)

The `your-flip` was supposed to be the user-chosen name. Clarified.

I thought PR is a useful number, but maybe we would rather want to have shorter IDs only used to differentiate between FLIPs? Updated.

Note that I think it's useful to have a alphabetically increasing list of FLIP filenames. Alternatively we could use the date in YYYYMMDD format - leaving open for further comments.

andsteing:
@avital can you approve once you're happy?
(I'm fine moving it back to the root if you think that's where they should reside...)

avital:
@googlebot I consent.

docs/flip/README.md
avital: I like that the name contains the issue number because it makes it easy to find the discussion thread on GitHub later.

andsteing:
@avital @gnecula  PTAL

avital:
Paging this back in real quick -- let's merge it. I was going to ask "why not just use issue templates" but I realize that discussions on those thread get unwieldy quickly. Maybe after merging this it would be a good exercise to look at the issues under the "Improve Linen" and "Improve Flax outside of Linen" milestones and see which of those could be "ported" to this FLIP process.

avital:
Let's run with this as an experiment and see if @jheek likes it. He's my true beta tester of whether processes serve our needs. :)

avital (approved):

Issue #697, opened 2020-11-30
"Build" workflow too slow


Marvin182:
Our "Build" workflow now takes over 20min. This means that it takes over 20min to merge a pull request into the master. This is too long to iterate fast.
We should aim to get back down to 10-15min. Main contributors seem to be the tests in the linen_examples. I suggest that we change these to run <5min and also make sure that the example tests all run in parallel.

avital:
See also #639

avital:
#737 and #741 seems to have improved things. Looks like tests now take 11m15s (e.g. https://github.com/google/flax/actions/runs/474491706). And we'll soon be removing `examples/` (#801) which will make things even faster. So I think we can close this issue.

Issue #639, opened 2020-11-16
Example tests to run in parallel


avital:
Right now we run each example's tests sequentially in `tests/run_all_tests.sh` (this is due to https://github.com/pytest-dev/pytest/issues/3151). Because of that, we can't run different example tests in parallel (though pytest can do this within a single run).

Consider either finding a way to work around the pytest "can't define two tests with the same filename in different directories", or if that's really not possible, at least try to find a way to rewrite `tests/run_all_tests.sh` to run in parallel but safely (e.g. can we run `pytest` multiple times -- does it keep any state directory that we need to isolate?)

avital:
Marking as "pull requests welcome"

andsteing:
Current timings from test when run on Github:

```
- flax/ in 61.14s (0:01:01)
- examples/pixelcnn in 10.56s
- examples/seq2seq in 20.55s
- examples/mnist/ in 9.11s
- examples/ppo/ in 20.88s
- examples/graph/ in 5.15s
- examples/sst2/ in 22.40s
- examples/lm1b/ in 42.18s
- examples/imagenet/ in 141.12s (0:02:21)
- linen_examples/pixelcnn/ in 30.65s
- linen_examples/wmt/ in 94.05s (0:01:34)
- linen_examples/seq2seq/ in 25.71s
- linen_examples/mnist/ in 10.00s
- linen_examples/nlp_seq/ in 2.28s
- linen_examples/imagenet/ in 330.18s (0:05:30)
```

andsteing:
@avital is there a way to see Github VM CPU usage during the build workflow? How many cores do we have?

When running locally, expensive tests like `wmt` and `imagenet` that make most of the test time max out a single core, so parallelizing would only make sense if we have multiple cores on Github.

avital:
My guess is that the default is one cure, but I'm not sure. Looks like we can use ["self-hosted runners"](https://docs.github.com/en/free-pro-team@latest/actions/hosting-your-own-runners/about-self-hosted-runners) if we want to.

@jheek knows more than me about GitHub actions, maybe he knows how we can find the VM specs for the free GitHub action build machine

jheek:
The specs of the machine seem to be 2 Cores with 7Gb ram.
We could start using `pytest -n auto` to automatically parallelize to the number of CPUs available.

jheek:
I also noticed we are spending most time compiling. I'll try disabling the heavy LLVM optimization in the longer tests

jheek:
[...]

PR #737, opened 2020-12-15
Decreases time needed to run tests.

andsteing:
See #697 and #639 for motivation.

This reduces overall build workflow from 22m to 13m. The pytest & coverage step reduces from 18m to 9m.

Note that `linen_examples/wmt` still takes ~90s to complete - I tried using `chex.fake_pmap()` but that failed with `ValueError: fake pmap does not currently support non-empty static_broadcasted_argnums=(3, 4)`. So maybe that is an idea for a later iteration ...

Note: This PR also includes a fix for TF versions that makes `wmt` example fail on master.

### DIFF
--- a/linen_examples/imagenet/models.py
+++ b/linen_examples/imagenet/models.py
@@ -130,3 +130,7 @@ ResNet152 = partial(ResNet, stage_sizes=[3, 8, 36, 3],
                     block_cls=BottleneckResNetBlock)
 ResNet200 = partial(ResNet, stage_sizes=[3, 24, 36, 3],
                     block_cls=BottleneckResNetBlock)
+
+
+# Used for testing only.
+_ResNet1 = partial(ResNet, stage_sizes=[1], block_cls=ResNetBlock)
--- a/linen_examples/imagenet/train_test.py
+++ b/linen_examples/imagenet/train_test.py
@@ -26,12 +26,9 @@ import tensorflow as tf
 import tensorflow_datasets as tfds
 
 # Local imports.
+import models
 import train
 from configs import default as default_lib
-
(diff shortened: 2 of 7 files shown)


linen_examples/imagenet/train_test.py
```diff
@@ -43,7 +40,7 @@ def setUp(self):
 
   def test_create_model(self):
     """Tests creating model."""
-    model = train.create_model(half_precision=False)
+    model = train.create_model(model_cls=models._ResNet1, half_precision=False)
```
Marvin182, line 43: Do you want to tell the linter that you are OK with the private member access?

setup.py
```diff
@@ -45,8 +45,8 @@
     "pytype",
     "sentencepiece",  # WMT example.
     "svn",
-    "tensorflow",
-    "tensorflow_text",  # WMT example.
+    "tensorflow-cpu==2.4.0",
```
Marvin182, line 48: Would it be better to just require a minimum version?

linen_examples/imagenet/models.py
```diff
@@ -130,3 +130,7 @@ def __call__(self, x, train: bool = True):
                     block_cls=BottleneckResNetBlock)
 ResNet200 = partial(ResNet, stage_sizes=[3, 24, 36, 3],
                     block_cls=BottleneckResNetBlock)
+
+
+# Used for testing only.
+_ResNet1 = partial(ResNet, stage_sizes=[1], block_cls=ResNetBlock)
```
Marvin182, line 136: nice idea :)

Marvin182 (approved):

Issue #739, opened 2020-12-16
Pooling layers still assume batch dimension


jheek:
The standard modules (Conv, Dense, attention...) all work with and without batch dimension to support the single example model pattern. However, pooling layers still enforce a batch dimension and there's no easy way to pool a single input at the moment.

avital:
(See #690 for broader context)

PR #740, opened 2020-12-16
Add support for single example pooling

jheek:
recognizes batch free inputs in pooling layers see. Fixes https://github.com/google/flax/issues/739


levskaya (approved):
lgtm! thanks!

levskaya:
(looks like an unrelated `Could not load dynamic library 'libcudart.so.11.0'` failure is breaking the tests)

jheek:
I think that's a standard warning. The real issue seems to be a symbol mismatch "_ZN10tensorflow8OpKernel11TraceStringEPNS_15OpKernelContextEb". I think the tensorflow and tensorflow-text versions are incompatible. Will investigate further

avital:
Oh, I forgot to ask: Can you please add an item in the CHANGELOG? I think we decided that we'll try to do that on every PR.

PR #741, opened 2020-12-16
disable LLVM optimization on tests with long compilation times

jheek:
timings:
imagenet  226s -> 52s (timed before the model changed)
wmt 63 -> 37s

### DIFF
--- a/linen_examples/imagenet/models_test.py
+++ b/linen_examples/imagenet/models_test.py
@@ -22,6 +22,9 @@ from jax import numpy as jnp
 import models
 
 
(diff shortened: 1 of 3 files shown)


andsteing:
ah, that's a good flag to know!

andsteing (approved):

avital:
Just a quick Q: Should we also at least sometimes be running tests with LLVM optimization? Or do we assume that JAX core sufficiently tests that?

jheek:
For now LLVM passes are only disabled on tests where it is slow. You could argue both ways though. Once LLVM generates buggy code all bets are off really

PR #752, opened 2020-12-17
move DCGAN community example from deprecated to Linen

bkkaggle:
I said i'd update this a while ago but school unfortunately got in the way :) 
Screen Shot 2020-12-17 at 12 56 31 PM

Here's the updated [notebook](https://github.com/bkkaggle/jax-dcgan/blob/main/dcgan.ipynb)

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -66,8 +66,12 @@ Using Linen
 +==================================+=================+========================+==================================+
 | `JAX-RL`_                        | `@henry-prior`_ | Reinforcement learning | N/A                              |
 +----------------------------------+-----------------+------------------------+----------------------------------+
- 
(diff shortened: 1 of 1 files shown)


andsteing:
Thanks for updating this!

bkkaggle:
Hi thanks for merging this in :), I just checked the [docs](https://flax.readthedocs.io/en/latest/examples.html#using-linen) and it looks like I left in an extra space at the beginning of the table row that breaks the table formatting. I pushed up the correct version to my [fork](https://github.com/bkkaggle/flax/commit/defef4cddc09f8a52f5e0ddc233c0af29fde2a79) that should fix this

andsteing:
I think you'll need to open another PR in that case.

bkkaggle:
yep will do that rn

PR #771, opened 2020-12-29
Add "extracting intermediates" pattern doc

### LINKED ISSUE #719: `Module.apply` should accept vanilla dictionaries for variables
Actually the current behavior is a bit strange I agree. We accept a dict of collections but the collections themselves have to be frozen.
It's a little stubborn if the api doesn't just accept a dict here. After all you can just freeze it and it will accept. The api should first freeze the deep-copy the dict before using it thought to avoid potential silent errors!

_Originally posted by @jheek in https://github.com/google/flax/discussions/715#discussioncomment-163338_

avital:
Adds a page under docs/patterns/ for ways to extract intermediate
values from a model. One of the ways suggests using `put_variable`,
which this commit also adds back.

I propose that this format replace our previous HOWTO system, and
we use this to document many of our other patterns, and rewrite
all HOWTOs this way.

Thus, I am looking for a very careful review.

TODO:
- [ ] Add unit tests. (The doc itself is an integration test)
- [ ] Add a docstring to `put_variable`
- [ ] Clone submodules when assigning to `self.layers` in `Sequential` so that we can safely reuse the same `CNN` instance.
- [ ] Add doc tests to run_all_tests.sh
- [ ] Document how to run doc tests locally
- [x] Consider adding RTD branch for every PR
- [ ] Can we highlight certain lines, like in the old HOWTO system?

ADDITIONAL CHANGES
------------------
* Remove the HOWTO system. It was clunky and added maintenance
  burden to all example editors. Instead, this document contains
  `testcode` blocks which means that it is tested directly from
  within the doc, without depending on any examples.

* Add support for tests embedded within docs. This allows for
  writing docs that "feel like" notebooks, but are easier to edit,
  and allow hiding imports and other initialization in `testsetup`

* Allow taking in non-frozen collections in `apply` (resolves #719).

QUESTIONS
---------
1. Is it weird that we need to pass in an empty intermediates collection
   if we want to use `put_variable`? We already specify that we expect
   changes to that collection via `mutable=['intermediates']`.
   We could try making `mutable=['intermediates']` add an additional
   intermediates collection if it's not there but then we'd have to
   theoretically also support `mutable=[True, 'intermediates']`.
   Another alternative is to add another option like mutable with
   an explicit list of collection names that are "output-only"

2. Is the `put_variable` implementation, which allows for either
   receiving intermediate outputs or not based on whether you pass
   in an `intermediates` variable collection good? Otherwise, people
   would have to thread in additional attributes to keep track or
   whether they want to return intermediate outputs or not.

3. Is the pattern in `Sequential` weird? In `setup`, we run
   `self.layers = self.submodules` to bind unbound modules. It's
   hard for users to understand this. I haven't thought too deeply
   about a possible better solution.


jheek:
I think we should have something like `put_variable` but there a few tricky details:

1. It should have the "ignore if not mutable" semantics but this is different from the other methods which fail if something isn't mutable. So ideally the method name already suggests that this is not a standard read/write get/put style api for variables.

2. We probably shouldn't silently override existing values. Because there will be 2 common issues with tracking intermediates: Accidentally keeping them around and feeding them back into apply (memory leak). And we can have multiple intermediates when using shared modules. So a more general version could be:
```
def record(self, name, value, col='intermediate'):
 """Records a value if the target collection is mutable. By default the 'intermediate collection is used.
     Returns True if the collection is mutable and False otherwise."""
  if not self.is_mutable(col):
    return False
  var = self.variable(col, name, lambda: ())
  var += (value,)
  return True
```

jheek:
>> Is it weird that we need to pass in an empty intermediates collection
if we want to use put_variable

Are you sure this is the case? As far as I remember a collection just has to mutable it doesn't need to be passed in

avital:
> > > Is it weird that we need to pass in an empty intermediates collection
> > > if we want to use put_variable
> 
> Are you sure this is the case? As far as I remember a collection just has to mutable it doesn't need to be passed in

marcvanzee:
General question: why did you use the term "pattern" in this PR instead of HOWTO? The PR description states "Adds a page under docs/patterns/ ...", but it actually seems to add it under doct/howtos. Did you first think of switching to patterns and then went back to howtos? It seems clearer to me when we use one of the two consistently, because now I was thinking that patterns weren something different than howtos.

avital:
I'm not sure I thought about it too carefully, I think patterns and HOWTOs are the same thing. This is very much still a WIP PR, happy to follow your lead @marcvanzee on wherever we decide to put things and how we decide to call them.

.github/workflows/build.yml
```diff
@@ -31,15 +31,16 @@ jobs:
         pip install .
         pip install .[testing]
         pip install -r docs/requirements.txt
+    - name: Test testcode blocks in documentation
+      run: |
+        sphinx-build -M doctest docs docs/_build
```
marcvanzee, line 36: This is supercool 😄  Maybe it can replace Colabs all together?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
...
+      x = nn.relu(x)
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = x.reshape((x.shape[0], -1))  # flatten
+      x = nn.Dense(features=256)(x)
+      x = nn.relu(x)
+      x = nn.Dense(features=10)(x)
+      x = nn.log_softmax(x)
+      return x
+
+Because this module uses ``nn.compact``, we don't have direct access to
```
marcvanzee, line 35: I think this deserves a bit more explanation. Why don't we have access to values if we are using `nn.compact`? Aren't all these `x`'s intermediate values? Or do you mean access from the outside?

If we weren't using `compact` but we weren't exposing the intermediate values, we would also not have access to them, right?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
...
+intermediate values. There are a few ways to expose them:
+
+
+Store intermediate values in a new variable collection
+------------------------------------------------------
+
+You can augment any module with calls to ``put_variable``
```
marcvanzee, line 42: Why are these lines so short? Or perhaps it won't show up on RTD?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
...
+    return CNN().apply({"params": params}, x)
+
+  @jax.jit
+  def features(params, x):
+    variables = {
+      "params": params,
+      "intermediates": {},  # collection for intermediate outputs
+    }
+
+    # `mutable=['intermediates']` specified which collections are treated as
```
marcvanzee, line 86:
```suggestion
    # `mutable=['intermediates']` specifies which collections are treated as
```

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
...
+    # mutable during `apply`. The variables aren't actually mutated, instead
+    # `apply` returns a second value, which is a dictionary of the modified
+    # collections.
+    output, modified_variables = CNN().apply(variables, x, mutable=['intermediates'])
+    return modified_variables['intermediates']['features']
+
+  params = init(jax.random.PRNGKey(0), batch)
+  predict(params, batch)
+  features(params, batch)
```
marcvanzee, line 95: It is odd that these two functions return something but we aren't using the returned result. Did you forget it?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
...
+
+  @jax.jit
+  def features(params, x):
+    # Proposal #686 should allow for this alternative:
+    #   return CNN().features.apply({"params": params['features']})
+    return CNN().apply({"params": params}, x,
+      method=lambda module, x: module.features(x))
+
+  params = init(jax.random.PRNGKey(0), batch)
+
```
marcvanzee, line 150: Why aren't we predicting here anymore?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
+Extracting intermediate values
```
marcvanzee, line 1: What do you think of adding something like a TL;DR at the top with the three suggested approaches?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,204 @@
...
+      lambda x: nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2)),
+      lambda x: x.reshape((x.shape[0], -1)),  # flatten
+      nn.Dense(features=256),
+      nn.relu,
+      nn.Dense(features=10),
+      nn.log_softmax,
+    ])
+
+  @jax.jit
+  def init(key, x):
```
marcvanzee, line 194: Nit: If you rename each module differently you only have to define `init` once if you provide the module name as an argument.

marcvanzee (changes requested):
Two high-level questions:

1. Can we see the resulting HOWTO in a readthedocs somewhere? Is it running on the server or should I compile it myself?
2. Can we split off the suggested change to core in a separate PR and get that in first? It seems @jheek has some comments, and this discussion may be helpful to users (and us) later on, so it seems clearer to have that in a separate PR.

avital:
Superceded by #1043

PR #795, opened 2021-01-02
Fix broken document link in italic notation

tokusumi:
a link in italic notation is breaking at [readthedocs documents (see at "Note" section)](https://flax.readthedocs.io/en/latest/notebooks/flax_basics.html#Module-basics) because nested markup (wrap `[comment](url)` by `*...*`) is not supported yet.

This maybe will be resolved in the future, but as a temporary solution, I propose to use normal style instead of italic for links.


andsteing:
Thanks for spotting and fixing this!

PR #798, opened 2021-01-03
Match the standard NumPy convention in JAX (onp -> np)

8bitmp3:
Switching away from using `onp` for ordinary NumPy towards `np`, as discussed with @avital in https://github.com/google/flax/discussions/730.

This is in line with JAX https://github.com/google/jax/pull/3760 (by @jakevdp) and https://github.com/google/jax/pull/2370 (by @hawkinsp) 

(See @mattjj 's comment https://github.com/google/jax/pull/2370#pullrequestreview-370624520: _"...The feedback from internal users was overwhelmingly in favor of this convention (for user code, and we should be consistent internally to avoid confusion)....The confusion can grow in a larger codebase that involves not only JAX code but also, say, TF code; the latter files may naturally have import numpy as np already. Another reason is that this is a convention followed by many other projects that present the NumPy API."_)

```diff
- import numpy as onp
+ import numpy as np
...
```

### DIFF
--- a/docs/annotated_mnist.md
+++ b/docs/annotated_mnist.md
@@ -1,157 +1,245 @@
-# Annotated full end-to-end MNIST example
+# The annotated MNIST image classification example with Flax
 
-```py
+This tutorial demonstrates how to construct a simple convolutional neural network (CNN) using the Flax Linen API and train the network for image classification on the MNIST dataset.
+
+1. Import JAX, [JAX NumPy](https://jax.readthedocs.io/en/latest/jax.numpy.html), Flax, ordinary NumPy, and TensorFlow Datasets (TFDS). Flax can use any data-loading pipeline and this example demonstrates how to utilize TFDS.
+
+```python
 import jax
-import flax
+import jax.numpy as jnp            # JAX NumPy
+
+from flax import linen as nn        # The Linen API
+from flax import optim              # Optimizers
+
+import numpy as np                 # Ordinary NumPy
+import tensorflow_datasets as tfds  # TFDS for MNIST
 ```
 
-Load vanilla NumPy for use on host.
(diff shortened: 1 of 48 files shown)


docs/annotated_mnist.md
```diff
@@ -8,7 +8,7 @@ import flax
 Load vanilla NumPy for use on host.
 
 ```py
-import numpy as onp
+import numpy as np
```
andsteing, line 11: Do we actually use this anywhere in the example? Otherwise let's simply remove the entire paragraph.

8bitmp3: Good point @andsteing

If NumPy isn't used, I'll make sure I remove the import

avital:
This looks great @8bitmp3, thanks! Is it ready to be merged?

8bitmp3:
Done @avital. By the way, `distributed-training.diff` and `ensembling.diff` in `howtos/diff` are the only two files I haven't touched - not sure what to do. LMKWYT 👍

avital (approved):

levskaya:
@8bitmp3 - hey sorry for this not going through yet.  There's a weird bug in our source sync system that makes the import script go nuts when there's over a certain number of SHAs/commits in a PR, and this PR hit it.

Any chance you could rebase on master and submit a squashed version of this PR?  (We also prefer to keep only a few commits per PR for cleaner commit-history's sake.)

8bitmp3:
@levskaya @avital Ok thanks, on it

8bitmp3:
@levskaya @avital I squeezed most of the commits into one large one 😬 Hope this doesn't break anything. I can't figure out what to fix in my merge conflicts though, please advise 👍

levskaya:
This is superseded by a clean-merged version of the PR at #1027

levskaya:
@8bitmp3 - would you mind adding your CLA consent to the updated (merged) version of this PR at #1027 ?

Issue #658, opened 2020-11-20
Carefully rename `linen_examples/` -> `examples/`


avital:
Make sure to catch all references to `linen_examples`, including in publicly linked documents (like in `flax/linen/README.md`).

avital:
This is the last remaining way that people accidentally start using `flax.nn` so we should do this soon. The one issue is that we haven't ported LM1B yet. The easiest thing is to keep an `examples/lm1b/deprecated` directory with the `flax.nn` implementation until we have some way to train LM1B in Linen -- either as a standalone example, or as a configuration option for the WMT example as @levskaya has been suggesting.

jheek:
I think lm1b also uses a deprecated tokenizer? I guess it would stop working anyway?

andsteing:
IIRC @levskaya mentioned recently that he would refactor the `wmt` example so it can also be used as a `lm1b` LM.

so moving to `.../deprecated` and then deleting once `wmt` is updated SGTM

levskaya:
Yeah I'm porting and combining lm1b into wmt example right now to minimize pointless code duplication (it just results in two training loop implementations in the example vs cloning all the modeling code).  It's pretty far along I just need to test a bunch of stuff.

Marvin182:
From my side it would be good to make this change this week. I don't plan to submit any changes to the examples (that could create merge conflicts).

avital:
OK. I will probably just re-make this PR with the most recent changes indeed in the next few days.

avital:
Done.

Issue #567, opened 2020-10-29
Updating examples/lm1b


andsteing:
This issue tracks updates the `lm1b` example to follow practices outlined in #231.

- [x]  Port to linen API - once ported all subsequent changes should be done in `linen_examples/lm1b`
- [x]  Update `README.md`
- [x]  Add `requirements.txt`
- [x]  Update file structure
- [x]  Use `ml_collections.ConfigDict`
- [ ]  Add benchmark test
- [x]  Add unit test for training/eval step
- [ ]  Add Colab
- [x]  Adhere to Google Python style
- [ ]  Add mypy annotations
- [ ]  Shorten/beautify training loop (consider using `clu` for this)

andsteing:
Note that "file structure" in #231 got changed:

> * To make it easier to test, maintain and reuse code structure the code as follows:
>   
>   * `main.py` contains the flags and calls a method from `train.py` to run the training loop. This should be the only file defining flags! This can be almost identical for all examples, please copy from `linen_examples/wmt/main.py`.
>   * `train.py` contains classes and methods for training and evaluating the model.
>   * `train_test.py` for test cases of the training code. At a minimum the test should run a single training step but more fine grained unit tests are a bonus. You can use `tfds.testing.mock_data` to avoid real data from disk.

avital:
@levskaya -- note that the lm1b example now fails tests with:

```
ValueError: Failed to construct dataset lm1b: BuilderConfig subwords32k not found. Available: [] 
```

avital:
I filed that as a separate issue: https://github.com/google/flax/issues/885

levskaya:
Closing - we do have a new linen lm1b example now.

levskaya:
I realize there's more on the above checklist than just getting it to work... so I'll re-open until we decide what fraction of these changes we want to do across examples.

andsteing:
Closing this since most issues from the checklist are resolved.

Adding a Colab would be nice, but I don't think it's a hard requirement. Might also be more interesting to add the Colab to the `wmt` example instead.

PR #801, opened 2021-01-05
Delete old examples/, rename linen_examples/ -> examples/

avital:
Keep the old LM1b example as examples/lm1b_deprecated, until #567
is resolved, per @levskaya's plan.


avital:
This PR resolves #658

tests/run_all_tests.sh
```diff
@@ -53,22 +53,27 @@ else
...
+    # TODO: Remove after github.com/google/flax/issues/567 is resolved
+    if [[ "$egd" == "examples/lm1b_deprecated" ]]; then
+        continue
+    fi
+
     # use cd to make sure pytpe cache lives in example dir and doesn't name clash
     # use *.py to avoid importing configs as a top-level import which leads tot import errors
     # because config files use relative imports (e.g. from config import ...). 
-    (cd $egd ; pytype "*.py")
+  if [[ "$OSTYPE" == "darwin"* ]]; then
```
andsteing, line 72: Maybe extract this into a function if we're using it twice?
Currently it prints the same message for every example.

andsteing:
(Also note there are a couple of conflicts that need to be resolved first)

tests/run_all_tests.sh
avital: I restructured this code to run pytype separately, and will file a new PR because this one became stale.

avital:
Re-did this PR as https://github.com/google/flax/pull/884

PR #813, opened 2021-01-07
Yet another test: remove duplicated copyright notice

avital:


avital:
cc @levskaya I am just testing something.

PR #823, opened 2021-01-11
Linen `Module` instances are now Frozen after `setup` has been called.

### LINKED ISSUE #674: Linen modules should be frozen
Currently we don't enforce linen Modules to be frozen after setup. However, this should be the case because Module instances need to be clone-able to work correctly. `__setattr__` should refuse to set attributes after setup is finished.

Update:

Actually there are more sharp edges that can be fixed by freezing correctly.
Currently we accept lists and dicts of sub modules which are registered on assignment. But we can actually freeze them to avoid this common trap:

```
def setup(self):
  self.sub_modules = [Dense()]
  self.sub_modules.append(Dense())
```

We could avoid this by making sub_modules is stored as a tuple and similarly we can avoid the same issue with dicts by transforming them into a FrozenDict

jheek:
Previously mutations after setup could be dropped silently. Now the stateless requirement
   is enforced by raising a TypeError in `__setattr__` after `setup`.

   Pytrees of dicts and lists are transformed into FrozenDict and tuples during attribute assignment.
   This avoids undetected submodules and inner state.

Closes #674

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ vNext
  - Added OptimizedLSTM: ~33% faster than the original LSTM when using <=1024 units
  - Bug Fix `Scope.variable` mutability check, before a variable could only be initialized
    if the 'params' collection was mutable.
+ - Linen `Module` instances are now Frozen after `setup` has been called.
+   Previously mutations after setup could be dropped silently. Now the stateless requirement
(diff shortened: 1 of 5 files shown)


avital:
Give my kids school-at-home obligations I'd appreciate if @levskaya can take a look at this

levskaya (approved):
This all looks great to me!

avital (approved):

PR #836, opened 2021-01-13
Fix breaking test

jheek:
should not really on instances of Jax array checking dtype and shape instead.


avital:
LGTM but how did the code pass CI before (both GitHub and Google internal?)

avital (approved):

andsteing:
Also I didn't get the error when running tests on my local computer. Is it maybe due to some numerical differences with a new jax/jaxlib version (I didn't upgrade packages locally)?

jheek:
It's not a numerics issue. It's about the class instance of a device buffer with the new jax release. We should not test that anyway because that's testing against Jax internals

andsteing:
So there's a second instance of this? The check is still failing at `tests/linen/linen_test.py:133`

jheek:
You are right. I overlooked it because somehow the second one didn't trigger locally?! I made a seperate PR for the other failure

PR #845, opened 2021-01-15
Add is_mutable method to Variable

### LINKED ISSUE #844: Mutable / Unmutable state when training
Hi Flax team,

I'm working on a model with an internal state which gets updated during training. When calling the model during validation, I do not want to update these variables. I could technically introduce a training variable, but I feel it could be more elegantly (and much simpler!) by simply checking if a state is mutable:

1 ) Is there an easy way to check if a variable is mutable, so that my code only updates when it is? e.g.
`if is_initialized and var.is_mutable(): var.value = f(....)`
2 ) If I set mutable=False, I only get back the output. Is there a way to get back the state regardless? e.g.
`output, updated_state = model.apply(inputs, mutable=False, return_state=True)`
My usecase is that for my validation metrics I call my loss function with the testdata and extract the metrics, so that I can use the same code for both training and validation.

Thanks!

jheek:
Closes #844

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ vNext
    This avoids undetected submodules and inner state. 
  - Add support for dict/FrozenDict when using `ModelParamTraversal`
    As a result `MultiOptimizer` can be used properly with linen modules.
- - Added `is_mutable()` method to `Variable`.
+ - Added `is_mutable()` method to `Variable` and `is_mutable_collection()` to `flax.linen.Module`.
 
 v0.3
 -----
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ vNext
    This avoids undetected submodules and inner state. 
  - Add support for dict/FrozenDict when using `ModelParamTraversal`
    As a result `MultiOptimizer` can be used properly with linen modules.
+ - Added `is_mutable()` method to `Variable`.
 
 v0.3
 -----
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -223,6 +223,10 @@ class Variable(Generic[T]):
     """Updates the value of this Variable."""
     self.scope.put_variable(self.collection, self.name, value)
 
+  def is_mutable(self) -> bool:
+    """Checks if this Variable is mutable."""
+    return self.scope.is_mutable_collection(self.collection)
+
 
 class Scope:
   """A Scope allows easy access to variables and manages RNGS of a neural network layer.
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -654,6 +654,12 @@ class Module:
       raise ValueError("Can't access variables on unbound modules")
     return self.scope.has_variable(col, name)
 
+  def is_mutable_collection(self, col: str) -> bool:
+    """Returns true if the collection `col` is mutable."""
+    if self.scope is None:
+      raise ValueError("Can't check mutability on unbound modules")
+    return self.scope.is_mutable_collection(col)
+
   def make_rng(self, name: str) -> PRNGKey:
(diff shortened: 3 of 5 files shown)


avital:
I'm not 100% sure it's good to expose this to users, but if you are sure this doesn't lead to footguns and a user asked for it then sure. The implementation seems correct to me.

avital (approved):

avital:
For example, what happens if a module really changes behavior based on whether a variable was mutable or not? That could be very surprising to the user of that module, that when they call `apply` it may do one or two totally different things based on whether they passed `mutable=True` or not. I can't think of a concrete case where someone would actually do this, but I mean something like:

```
class Foo(nn.Module):
  def __call__(self, x):
    v = self.variable('bar', ...)
    if v.is_mutable():
      return x*2
    else
      return x
```

avital:
Here's a more realistic example -- what if people use this to implement batch norm without an `is_training` attribute, e.g:

```
class BatchNorm(nn.Module):
  def __call__(self, x):
    moving_mean = self.variable('moving_mean', ...)
    if moving_mean.is_mutable():
      # compute batch norm in "training mode"
    else:
      # compute batch norm in "prediction mode"
```

Then indeed you'd get different results by simply changing `mutable=True`. Seems like it could be a footgun. (Whereas in our current implementation you'd get an error, which is good). So I currently feel like by adding this API we'd be encouraging users to write modules in a way that could lead to footguns. Not that they couldn't do it without this particular API, they'd just be tempted differently.

Per the comment on #844, the user said "I don't want to have to pass another attribute" but I now think maybe it's actually a feature to force users to think about this additional attribute.

WDYT?

jheek:
With BatchNorm in particular I think it is easy to make a mistake. There are other cases where I think it is less of an issue. For example you could have something like:
```
x = self.variable('log', 'x')
if log.is_mutable()
```

where log is a collection used for debugging/collecting some auxiliary information. 

I think it's good that we can control whether something is mutable but I don't think we should enforce a double check pattern.

avital:
For the log example you pointed out, wouldn't it be better to then have something like `put_variable()` which doesn't do anything if the variable isn't mutable, or `append_variable()` similarly? Basically give users controlled mutators?

(Like the patterns discussed in #771)

GJBoth:
> For the log example you pointed out, wouldn't it be better to then have something like `put_variable()` which doesn't do anything if the variable isn't mutable, or `append_variable()` similarly? Basically give users controlled mutators?
> 
> (Like the patterns discussed in #771)

That sounds like a better option! Now the forward pass fails if you don't pass mutable=True...

avital:
So Gert-Jan if we add those primitives (put and append, that are noops if
the variable isn't present or immutable) do you think we should wait before
exposing is_mutable? (That's my personal inclination at the moment)

jheek:
Those primitives have the same problem in reverse though. If you have a mutator that is a silent no-op if it the collection isn't mutable you can silently drop mutations that should have happened.

GJBoth:
If I understood you correctly, I think so. In case I didn't, I'd like this to happen:
Case 1 - state is not mutable: state doesn't get updated, we return the input state.
Case 2 - state is mutable: state gets updated, new state gets returned.

For me giving an error when the state is not mutable strikes me as unexpected behaviour...
As to how this is implemented, I'm more ambivalent. Something like 
` if is_initialized and var.is_mutable` seems reasonable, although something like `update_var(update_fn, *vars)` could be cleaner...

avital:
> Those primitives have the same problem in reverse though. If you have a mutator that is a silent no-op if it the collection isn't mutable you can silently drop mutations that should have happened.
@jheek -- yes but still, you can only drop updates -- you can't actually compute a different result. So the footgun risk is reduced.

PR #864, opened 2021-01-17
Improve submodule handling for transforms and top-level usage

levskaya:
This PR includes two changes:
1. Fixing the module scope collection and setting for use with multi-scope functional-core transforms to be properly recursive to handle complex, nested aliasing edge-cases.
2. Introducing the ability to use top-level instantiated modules as attributes in another top-level module.  This is a very common workflow during interactive development and it has been cited as a pain-point and a badly confusing gotcha a number of times by our users. All top-level instantiated  modules passed in as attributes are 'adopted' or 'reparented' onto the root, top-level module they are passed into.  Effectively this function registers any such top-level instantiated submodule as if they had been defined in the top setup() function, taking care to avoid name collisions.

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -52,7 +52,8 @@ def _dedup_scopes(scopes):
       path.append(scope.name)
       scope = scope.parent
     if max_parent is not leaf:
-      del minimal_set[leaf]
+      if leaf in minimal_set:
+        del minimal_set[leaf]
     paths.append((max_parent, max_parent_path))
   return tuple(minimal_set), tuple(paths)
 
@@ -67,8 +68,10 @@ def _dup_scopes(orig_scopes, scopes, paths):
     scopes.append(scope)
   return scopes
 
+
 def _transpose(xs):
   return tuple(zip(*xs))
+
 
 def pack(fn: Callable[..., Any],
          in_variable_filters: Sequence[CollectionFilter],
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -65,6 +65,7 @@ def _attr_repr(value: Any):
     value_rep = repr(value)
   return value_rep
 
+
 def _module_repr(module: 'Module', num_spaces: int = 4):
   """Returns a pretty printed representation of the module"""
   cls = type(module)
@@ -91,6 +92,7 @@ def _module_repr(module: 'Module', num_spaces: int = 4):
   else:
     return f'{cls_name}()'
 
+
 # Track parent relationship across Modules.
 # -----------------------------------------------------------------------------
 class _DynamicContext:
@@ -103,7 +105,7 @@ class _DynamicContext:
       self._thread_data.module_stack = [None,]
     return self._thread_data.module_stack
 
-# The global context 
+# The global context
 _context = _DynamicContext()
 
 class _Sentinel:
@@ -132,16 +134,29 @@ def _get_suffix_value_pairs(
     tree_or_leaf: Any) -> List[Tuple[str, Type["Module"]]]:
   """Helper for naming pytrees of submodules."""
   dict_or_leaf = serialization.to_state_dict(tree_or_leaf)
-  if dict_or_leaf == {} or not isinstance(dict_or_leaf, dict):
+  if not isinstance(dict_or_leaf, dict) or dict_or_leaf == {}:
     return [('', tree_or_leaf)]
   else:
     flat_dict = traverse_util.flatten_dict(dict_or_leaf)
     return [('_' + '_'.join(k), v) for k, v in flat_dict.items()]
 
 
+def _map_over_modules_in_tree(fn, tree_or_leaf):
+  """Helper for mapping function over submodules."""
+  dict_or_leaf = serialization.to_state_dict(tree_or_leaf)
+  if not isinstance(dict_or_leaf, dict) or dict_or_leaf == {}:
+    return fn('', tree_or_leaf)
+  else:
+    flat_dict = traverse_util.flatten_dict(dict_or_leaf)
+    mapped_flat_dict = {k: fn('_' + '_'.join(k), v)
+                        for k, v in flat_dict.items()}
+    return serialization.from_state_dict(
+        tree_or_leaf, traverse_util.unflatten_dict(mapped_flat_dict))
+
+
 def _all_names_on_object(obj: Any) -> Set[str]:
   """Gets all names of attributes on `obj` and its classes throughout MRO.
-  
+
   Args:
     obj: The object to get names for.
   Returns:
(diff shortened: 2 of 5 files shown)


flax/linen/module.py
```diff
@@ -132,16 +137,29 @@ def _get_suffix_value_pairs(
     tree_or_leaf: Any) -> List[Tuple[str, Type["Module"]]]:
   """Helper for naming pytrees of submodules."""
   dict_or_leaf = serialization.to_state_dict(tree_or_leaf)
-  if dict_or_leaf == {} or not isinstance(dict_or_leaf, dict):
+  if not isinstance(dict_or_leaf, dict) or dict_or_leaf == {}:
```
levskaya, line 140: btw: switching the test order here is important since numpy arrays (but not jax arrays!) will throw an exception if compared to {} then cast to boolean - in the numpy case one ends up with an array of `False`s that can't be converted to a bool.

flax/core/lift.py
```diff
@@ -52,7 +52,8 @@ def _dedup_scopes(scopes):
       path.append(scope.name)
       scope = scope.parent
     if max_parent is not leaf:
-      del minimal_set[leaf]
+      if leaf in minimal_set:
```
levskaya, line 55: This was altered to support duplicate leaf scopes that occur in some edge-cases with top-level module adoption.

jheek: Thanks, duplicate leaves are supposed to be supported but probably forgot to test.

flax/linen/module.py
```diff
@@ -102,8 +104,11 @@ def module_stack(self):
     if not hasattr(self._thread_data, 'module_stack'):
       self._thread_data.module_stack = [None,]
     return self._thread_data.module_stack
+  @property
+  def at_top_level(self):
+    return len(self.module_stack) == 1
```
jheek, line 109: This kind of check is pretty dangerous. It means there is a global "top-level". So the things that depend on at_top_level will never work if you do something like module.apply inside another module. The old flax also special cased the top frame and that resulted in some nasty edge cases for users.

flax/linen/module.py
```diff
@@ -432,11 +451,15 @@ def __setattr__(self, name: str, val: Any):
...
     else:
       val = _freeze_attr(val)
+      already_processed_modules = set()
       for suffix, subvalue in _get_suffix_value_pairs(val):
         if isinstance(subvalue, Module):
           if not self._state.in_setup:
             raise ValueError(
                 "You can only assign submodules to self in setup().")
+          # If a modules appears multiple times in the pytree, attach it only once.
+          if id(subvalue) in already_processed_modules:
```
jheek, line 461: So is it intended that:
```
a = Dense()
self.a = a
self.b = b # raises error
```
but
```
a = Dense()
self.a = [a, a]
```
does not raise an error?

flax/linen/module.py
```diff
@@ -547,12 +576,59 @@ def setup(self):
...
 
-      3. Immediately when a module is constructed inside a method wrapped with 
+      3. Immediately when a module is constructed inside a method wrapped with
          :meth:`compact`.
 
     """
     pass
 
+  @wrap_method
+  def _adopt_toplevel_modules(self):
```
jheek, line 586: An easier solution would be to a "bind-to-first" behaviour. Currently we fail if you bind a module twice but we could just bind the parent on the first attribute assignment and make later bindings a no-op. This would be more consistent because you don't need the top-level exception case to be handled separately. Instead you would always bind modules passed as attributes which were unbound before. This also removes the distinction between constructor attributes and setup attributes. What do you think?

jheek (changes requested):
Second commit looks good to me but I'm not sure about the special casing around top level modules

flax/linen/module.py
levskaya: Sorry, I should have commented on this change, or even split it out for later -- it's actually completely orthogonal to the rest - I just noticed that the latter example you give does in fact fail with an exception now.  I could imagine someone making a list w. replicated layers in setup() to do something like e.g. "universal transformer" which is a stack of shared param layers.

In first example, I'm assuming you meant
```
a = Dense()
self.a = a
self.b = a # raises error
```
That is another interesting example that would still error-out in this case.

Now I'm thinking that perhaps we should stick to the current "unique-names on self in setup()" policy and only try to "fix" this if there's an actual problem that users encounter.  At the very least I think I'll split this out to discuss later/elsewhere, as it's completely separate from the other changes.

flax/linen/module.py
levskaya: The only thing this doesn't address is sharing-by-reference semantics.  Basically if you write code w. shared refs at the top, then we don't respect parameter-sharing relationships in the top-init'd module in a simple bind-to-first implementation.  Maybe we say that's just how it is, but it _is_ a different set of rules for "top-level style" vs "internal style" which seems like something painful to explain or debug later.  A subset of our users coming from pytorch really seem to like defining models in this "top-level style", and I'm sure someone eventually will try setting up sharing and wonder why it's broken.

flax/linen/module.py
levskaya: What exactly are we supposed to do with Modules passed in as attributes to an inner `module.apply` (in which case they would never naturally be parent=None)?  That's actually a very interesting edge-case to consider... I'm guessing we re-parent them in the same way we're talking about here?  Throw an error?  Perhaps we need to keep track of a module's "root" as well as its parent to even detect this kind of thing? (root might be useful for other stuff anyway...)

flax/linen/module.py
levskaya: I've removed this from this PR - it's a complicated issue worth covering later.

flax/linen/module.py
levskaya: I should also add that "bind to first" doesn't save a ton of code, as you can't bind directly in `__setattr__` since the parent module hasn't been instantiated yet, you have to do this at the end of post-init in a helper function in any case.

flax/linen/module.py
levskaya: Actually I'm being dumb - I'm pretty sure we already handle passing in submodules as attributes to internal `module.init/apply` just fine.  Some quick experimentation verifies that transforms, mutability, etc. work naturally when passing into internal `init/apply`.  It's acts like a trivial transform, there's absolutely nothing special here.

Which brings us back to the "top_level" check.  I think in this case there really -is- a unique top-top-level edge-case that we need to handle.  We don't naturally have `parent=None` modules appearing inside a module, this is very much only a top-top-level concern and _should_ be a no-op otherwise.

So I _generally_ agree with you that having top-level checks is sketchy and undesirable, but I think this is a special case where there really is a unique concern that only ever occurs at the "true" top level.

flax/linen/module.py
jheek: I still don't get how top-level is a special case:

I would expect the following to work after with PR:
```
class Foo(nn.Module):
  def setup(self):
    self.foo = Bar(bar=nn.Dense(1))

class Bar(nn.Module):
  bar: Any

  def setup(self):
    pass

  def __call__(self, x):
    return self.bar(x)

x = jnp.ones((2,))
Foo().init(random.PRNGKey(0), x)
```
but it actually raises an error because it will not bind to `bar.bar` as I would expect. It would be weird if this same pattern did work if you pass a Module instance to the top-level constructor right (which to me makes less sense)?

levskaya:
We need to stage these and some other related changes carefully, so closing this PR in favor of a separate series.

PR #876, opened 2021-01-18
Adds a comment on batch sharding

marcvanzee:
Users sometimes encounter OOM if they run WMT on fewer machines with the same batch size, this comment explains why.

### DIFF
--- a/linen_examples/wmt/train.py
+++ b/linen_examples/wmt/train.py
@@ -548,6 +548,8 @@ def train_and_evaluate(config: ml_collections.ConfigDict, workdir: str):
       is_last_step = step == config.num_train_steps - 1
 
       # Shard data to devices and do a training step.
+      # We use a global batch size and shard the batch, so the per-device batch 
+      # size decreases if the number of devices increases.
       with jax.profiler.StepTraceContext("train", step_num=step):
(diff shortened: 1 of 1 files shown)


linen_examples/wmt/train.py
```diff
@@ -548,6 +548,8 @@ def decode_tokens(toks):
       is_last_step = step == config.num_train_steps - 1
 
       # Shard data to devices and do a training step.
+      # We use a global batch size and shard the batch, so the per-device batch
```
Marvin182, line 551: That is not true.
github.com/google/flax/blob/master/linen_examples/wmt/input_pipeline.py#L349
The config contains the per-device batch size. The input pipeline runs on every machine and we multiply the per-device batch size with `jax.local_device_count()`. So if you have `config.per_device_batch_size=64` and you train on 2 machines with 8 TPU devices each you get a global batch size of 64*8*2 but each machine prepares batch of size 64*8. If you reduce the number of devices on each machine the batch size per machines should also go down.

That is how it should work. If this is not the case it's a bug and we should fix the bug.

marcvanzee: Yes, you are correct. I based this comment on a thread on chat, which contained false information, and I did not check the config myself. Thanks for clarifying this, I'll close this PR.

PR #884, opened 2021-01-19
Delete old examples/, rename linen_examples/ -> examples/

avital:
Keep the old LM1b example as examples/lm1b_deprecated, until #567
is resolved, per @levskaya's plan.

Continuation of abandoned #801


tests/run_all_tests.sh
```diff
@@ -43,33 +43,41 @@ handle_errors () {
...
-# validate types
-if [[ "$OSTYPE" == "darwin"* ]]; then
-  echo "Pytype is currently not working on MacOS, see https://github.com/google/pytype/issues/661"
-else
-  pytype flax/
-fi
+pytest -n 4 tests $PYTEST_OPTS
 
 # Per-example tests.
-if [[ $ALL_EXAMPLES == 'true' ]]; then
```
andsteing, line 56: Please also remove `ALL_EXAMPLES=false`, help message and `--all` switch above.

tests/run_all_tests.sh
```diff
@@ -43,33 +43,41 @@ handle_errors () {
...
 done
 
+# validate types
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    echo "Pytype is currently not working on MacOS, see https://github.com/google/pytype/issues/661"
+else
+    pytype flax/
+
+    # Skip pytype on deprecated example
+    # TODO: Remove after github.com/google/flax/issues/567 is resolved
```
andsteing, line 69: Maybe move this comment inside the `if` below to make it clearer?

tests/run_all_tests.sh
```diff
@@ -43,33 +43,41 @@ handle_errors () {
...
+# we apply pytest within each example to avoid pytest's annoying test-filename collision.
+# In pytest foo/bar/baz_test.py and baz/bleep/baz_test.py will collide and error out when
+# /foo/bar and /baz/bleep aren't set up as packages.
+for egd in $(find examples -maxdepth 1 -mindepth 1 -type d); do
     pytest $egd
+
     # use cd to make sure pytpe cache lives in example dir and doesn't name clash
     # use *.py to avoid importing configs as a top-level import which leads tot import errors
     # because config files use relative imports (e.g. from config import ...). 
     (cd $egd ; pytype "*.py")
```
andsteing, line 59: Do you want to test pytype twice for examples ?
(see line 78 below)

andsteing (commented):
Note: the deprecated lm1b is currently failing. It was probably failing for a while but because we deactivated the deprecated example tests in our default build workflow nobody noticed.

tests/run_all_tests.sh
avital: Oops, I was missing a commit that I forgot to push.

tests/run_all_tests.sh
avital: Good idea.

tests/run_all_tests.sh
avital: Good idea.

avital:
> Note: the deprecated lm1b is currently failing. It was probably failing for a while but because we deactivated the deprecated example tests in our default build workflow nobody noticed.

:(

The TFDS tokenized dataset (e.g. lm1b/subwords32k) were already deprecated for a while, I guess they removed it altogether. In the meanwhile, since the example was already broken before, I'm going to have lm1b_deprecated not run tests. @levskaya -- we should make a clear plan about the future of the lm1b example...

andsteing:
About that failing `lm1b` test : maybe you want to disable the test manually and create an issue to fix it?
(Given that we keep the example because people want to use it we should probably make it work)

avital:
Filed as https://github.com/google/flax/issues/885

andsteing (approved):

andsteing (changes requested):
Now failing with another build error:
`ERROR: Cannot apply howto checkpointing! ==> PLEASE FIX HOWTO`

andsteing (approved):

avital:
Something wacky happened with Copybara -- this change went in as https://github.com/google/flax/commit/d068512a932da3e05b822790a591bac391aeab36 but without some of the changes that aren't mirrored by Copybara. So I updated this PR to only include the remaining changes.

PR #907, opened 2021-01-19
Adds imagenet Colab 

andsteing:
As part of #480

**Note**: for testing the Colab, you'll need to modify the setup cell to:

```
repo, branch = 'https://github.com/andsteing/flax', 'imagenet'
```

See Colab at:

https://colab.research.google.com/github/andsteing/flax/blob/imagenet/examples/imagenet/imagenet.ipynb

Features

- editing files using Colab's built-in editor with horizontal split
- optional persistence of changes in user's Drive
- visualizing data & training from `imagenette` dataset
- loading pretrained checkpoint from Cloud repository, inference & evaluation on `imagenette`


examples/imagenet/configs/default.py
```diff
@@ -36,11 +36,14 @@ def get_config():
 
   # As defined in the `models` module.
   config.model = 'ResNet50'
+  # We use 'imagenette' in the Colab.
```
avital, line 39:
```suggestion
  # Allow the dataset to be configurable, since we fine-tune on 'imagenette' in the Colab.
```

examples/imagenet/configs/default.py
```diff
@@ -36,11 +36,14 @@ def get_config():
...
+  config.dataset = 'imagenet2012:5.*.*'
 
   config.learning_rate = 0.1
+  config.warmup_epochs = 5.0
   config.momentum = 0.9
   config.batch_size = 128
-  config.num_epochs = 100
+  config.num_epochs = 100.0
```
avital, line 46: Has this intentionally become a floating point number? Seems like it really should be an int? But if there was some good reasoning behind this then ignore me!

avital (approved):
Thanks, this looks great! See small comments below.

examples/imagenet/configs/default.py
andsteing: In the current Colab we don't fine-tune but rather train from scratch on imagenette, then load a pre-trained model from cloud and evaluate on iamagenette. I renamed the sections ("train from scratch", "load pre-trained") to make this a bit clearer, PTAL.

Maybe we should modify the imagenet example and gear it more towards transfer learning, WDYT?

If we want to do so, then maybe we should add some scripts for this as well and update the README accordingly - in a separate PR.

andsteing: But anyways, I don't think this comment is really useful. I'd rather say what form the string should have.

examples/imagenet/configs/default.py
andsteing: It is intentional - it's fine to specify 2.5 epochs if you want. This is in parallel with the `warmup_epochs` (which in turn needs to be a float for the Colab's 0.5 warmup epochs).

andsteing:
Thanks for the comments, updated accordingly.

avital:
Due to a weird fluke in our internal code mirroring, this ended up being submitted as part of https://github.com/google/flax/commit/293b1f62127ab293356e9347cc680833787c66e1. (cc @marcvanzee )

@andsteing -- can you double check that all of the remaining changes you had made indeed ended up in that commit?

andsteing:
I can confirm changes are in the commit (one image is missing in the Colab but it's not an important one).

avital:
Thanks for checking @andsteing

PR #908, opened 2021-01-20
Updates comment on Models

marcvanzee:
Flax Modules don't carry state, but according to this comment they do. Probably a left-over from pre-Linen.

### DIFF
--- a/flax/serialization.py
+++ b/flax/serialization.py
@@ -28,7 +28,7 @@
 # limitations under the License.
 """Serialization utilities for Jax.
 
-All flax classes that carry state like Model and Optimizer can be turned into a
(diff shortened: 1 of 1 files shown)


jheek (approved):

flax/serialization.py
```diff
@@ -28,7 +28,7 @@
 # limitations under the License.
 """Serialization utilities for Jax.
 
-All flax classes that carry state like Model and Optimizer can be turned into a
+All Flax classes that carry state (e.g. ,Optimizer) can be turned into a
```
avital, line 31: Looks like a mistake -- "(e.g. , Optimizer)?

PR #909, opened 2021-01-20
Updates optimizer docstrings and RTD

marcvanzee:
One of the most asked questions by users is how to optimize different parameters with different optimizers, or variations on this, e.g.

- How to only train a subset of the parameters?
- How to train different parameters with different learning rates?
- How to train different parameters with different optimizers?

In this PR I try to improve the docstring on the optimizers, and make sure that they show up nicely on RTD.

The new version on RTD can be viewed here: https://flax.readthedocs.io/en/multi-optim/flax.optim.html

### DIFF
--- a/docs/flax.optim.rst
+++ b/docs/flax.optim.rst
@@ -2,44 +2,40 @@
 flax.optim package
 ===================
 
-Optimizer core
-------------------------
-
 .. currentmodule:: flax.optim
 
-.. automodule:: flax.optim
-
+Optimizer Base Classes
+------------------------
 
 .. autoclass:: Optimizer
-   :members: apply_gradient, compute_gradients, optimize
+   :members: apply_gradient
 
 .. autoclass:: OptimizerDef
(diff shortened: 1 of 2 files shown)


docs/flax.optim.rst
```diff
@@ -2,44 +2,40 @@
...
 .. currentmodule:: flax.optim
 
-.. automodule:: flax.optim
-
+Optimizer Base Classes
+------------------------
 
 .. autoclass:: Optimizer
-   :members: apply_gradient, compute_gradients, optimize
+   :members: apply_gradient, replicate, unreplicate
```
jheek, line 11: replicate & unreplicate have been deprecated for a long time so maybe they should not be added to the docs now

marcvanzee: I forgot about that, and it seems they also don't have a deprecation warning. I added it to both of them and remove them from this list.

jheek (approved):

PR #910, opened 2021-01-20
Clarifies usage of jit on Module.init

### LINKED ISSUE #879: Improve Documentation: Jitting init()
In some of our examples we `jax.jit` the `init()`, for instance in the [WMT example](https://github.com/google/flax/blob/master/linen_examples/wmt/train.py#L472), and in other examples we don't.

@Marvin182 mentioned in a chat: "Jitting the model.init() of the PixelCNN example takes forever (>5min) but runs without jitting in <20s. " [on TPU]

@jheek replied: "jitting init is a difficult thing. On one hand we save time because we are lazy and potentially avoid lot's of small compiles. on the other hand we have a lot of duplicate HLO's when a model has many paramaters with the same shape & dtype"

It thus seems there are some best practices on how and when to `jit` the `init()` of Flax modules, and it would be useful to document this since it can make a big difference in practice, especially on TPU.

marcvanzee:
Closes #879

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -725,6 +725,12 @@ class Module:
            **kwargs) -> VariableDict:
     """Initializes a module method with variables and returns modified variables.
 
+    Calling `jax.jit` on init will make JAX trace through the model and compile
+    it at once, which means we possibly avoid many small op-by-op compilations.
+    However, it does mean we cannot re-use ops (e.g., for compiling many kernels
+    of the same shape), so it depends on the specifics of the model whether this
+    is a good idea or not.
+
     Args:
       rngs: The rngs for the variable collections.
       method: An optional method. If provided, applies this method. If not
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -725,11 +725,12 @@ class Module:
            **kwargs) -> VariableDict:
     """Initializes a module method with variables and returns modified variables.
 
-    Calling `jax.jit` on init will make JAX trace through the model and compile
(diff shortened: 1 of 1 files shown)


examples/wmt/train.py
```diff
@@ -469,6 +469,12 @@ def decode_tokens(toks):
   target_shape = (config.per_device_batch_size, config.max_target_length)
 
   m = models.Transformer(eval_config)
+
+  # Calling `jit` on init will make JAX trace through the model and compile it
```
jheek, line 473: I'm wondering about the discoverability of this explanation.
Wouldn't it make more sense to add a jit snippet with this explenation to the init method itself?

marcvanzee: Yes I liked that idea. I've moved it there now.

flax/linen/module.py
```diff
@@ -725,6 +725,12 @@ def init(self, rngs: Union[PRNGKey, RNGSequences], *args,
            **kwargs) -> VariableDict:
     """Initializes a module method with variables and returns modified variables.
 
+    Calling `jax.jit` on init will make JAX trace through the model and compile
```
jheek, line 728: I would focus here on the laziness aspect (not computing the actual forward pass) and show an example of how to do that. The info here is correct but only relevant if you also have a lazy_init which we currently don't offer

marcvanzee: Done.

jheek (approved):
LGTM

PR #911, opened 2021-01-20
Adds HOWTO: TF functions in Flax

### LINKED ISSUE #899: What is the Flax equivalent of Tensorflow functions?
For reference, I'm listing some Tensorflow functions below for which users asked what the equivalent is in Flax.

| TF Function  | Flax/jax Function |
| ------------- | ------------- |
| `tf.identity`  | `lambda x: x`  |
| `tf.name_scope()`  |   "Create submodule" [1] |
| `tf.print`  |  [`host_callback.id_print`](https://jax.readthedocs.io/en/latest/jax.experimental.host_callback.html#jax.experimental.host_callback.id_print) |

[1] Putting anything inside a new Module would have that effect. You can also always give submodules a name explicitly with the `name=... kwarg` to the submodule constuctor.

marcvanzee:
Closes #899 

RTD preview: https://flax.readthedocs.io/en/tf-howto/howtos/tf_functions.html

### DIFF
--- a/docs/howtos/tf_functions.rst
+++ b/docs/howtos/tf_functions.rst
@@ -1,0 +1,31 @@
+Tensorflow Functions in Flax
+==============================
+
+Below is a list of Tensorflow functions and their Flax counterpart.
+
+Please feel free to file a Pull Request adding more functions!
+
+
+.. list-table:: TF vs. Flax functions
+   :widths: 25 25
+   :header-rows: 1
+
+   * - TF function
+     - Flax function
+   * - `tf.identity`_
+     - ``lambda x: x``
+   * - `tf.name_scope`_
+     - "Create submodule" (1)
+   * - `tf.print`_
+     - `jax.experimental.host_callback.id_print`_
+
+Notes:
+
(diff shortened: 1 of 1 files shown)


avital:
I don't think many researchers outside of Google use TF these days so this seems like a very low priority HOWTO to add, and in fact may confuse people reading our RTD page to think that there's any relationship between TF and JAX... My vote would be to not add this to our official docs, but maybe there are additional perspectives

marcvanzee:
A few users asked these questions, and I thought it would be useful to have them documented somewhere. Ideally these would have been asked in a Github Discussion where a user asks "how do I do X in TF?", and someone replies. However, the conversation was held somewhere else.

I do agree with the fact that these look a bit odd and potentially confusing as HOWTOs. Perhaps we just ignore these cases and just answer them again on Github Discussions if someone ends up asking them again. WDYT?

avital:
Another option is to simply file a discussion now with the content of this HOWTO and call it a day ("Just posting this here for future reference: some people asked about ...") -- at least that way we can point people to a discussion URL if they ask again.

marcvanzee:
Closed by #916

PR #913, opened 2021-01-20
Keep the behavior consistent when mutable is an empty dictionary

yang-song:
`mutable` can be an empty dictionary when the network has no mutable states. In this case, `apply` should return another state even though an empty state was provided. This is the expected behavior of the API, and enables a unified approach for multiple architectures, with or without mutable latent states. The current application simply equates empty state and False, and return no state in the output. This can be surprising to many new users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ vNext
  -
  -
  -
- -
+ - Bug Fix `flax.core.apply` and `Module.apply`. Now it returns a tuple containing the output and a frozen empty collection when `mutable` is specified as an empty list.
  -
  -
  -
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -602,10 +602,10 @@ def apply(fn: Callable[..., Any],
     new_variables = _unfreeze_variables(variables, mutable)
     with Scope(new_variables, rngs=rngs, mutable=mutable).temporary() as root:
       y = fn(root, *args, **kwargs)
-    if mutable:
-      mutated_variables = {
-          k: v for k, v in new_variables.items() if in_filter(mutable, k)
-      }
+    if mutable is not False:
+      mutated_variables = {k: v
+                           for k, v in new_variables.items()
+                           if in_filter(mutable, k)}
       return y, freeze(mutated_variables)
     else:
       return y


marcvanzee:
Thanks @yang-song for your PR! Out of curiosity: was there a specific use-case you had in mind when writing it?

>  In this case, apply should return another state even though an empty state was provided. This is the expected behavior of the API

Why is this the expected behavior? Intuitively I understand the behavior as: "If there are mutable variables, return them as the second argument, if there are no mutable variables, don't return them". So following that reasoning, it should not matter whether `mutable` is `False` or `{}`, since in both cases we have no mutable variables.

I am not sure I understand why we need to make the distinction between `False` and `{}`, but maybe i am missing something.

> and enables a unified approach for multiple architectures, with or without mutable latent states. 

If we really want a unified approach, then I think we could always return a pair `(immutable_variables, mutable_variables)`, and just let `mutable_variables` be empty if there are none.

yang-song:
Thanks for the fast reply! From my understanding of the documentation and `linen_examples`, we should supply `mutable=False` during test time, and supply a dictionary of mutable variables (if there is any) during training time. There is one common case that I think the PR can help: we have a unified, model-agnostic training and test pipeline that is designed to work without modification for a large family of models, some of which might have batch normalization (and other latent variables) while some do not. The original implementation requires us to manually check whether the model has latent variables when implementing this pipeline.

avital:
I find this a pretty compelling argument:
> we have a unified, model-agnostic training and test pipeline that is designed to work without modification for a large family of models, some of which might have batch normalization (and other latent variables) while some do not. 

If we do make this change, we need to also update the docstring.

@jheek -- as the original author of this API do you agree?

jheek:
Yes this is what I would expect as well. I made the classic mistake of forgetting that Python would cast an empty container into `False`.

Please don't forget a CHANGELOG entry.

yang-song:
Also a reminder to check all code in `linen_examples` in case this PR is incorporated.

marcvanzee:
@yang-song could you please add an entry to the CHANGELOG as well? Then I will approve this PR.

yang-song:
@marcvanzee Sorry for the delay, but I have updated `CHANGELOG.md`. I also checked all the linen examples and didn't find anything that might be broken due to this change.

avital (approved):

avital:
Thanks! I suggested an improvement to the CHANGELOG entry, and looks like you need to rebase because there's a conflict in scope.py now. Then let's merge it! (How much work for 1 LOC change -- an important one! ;))

yang-song:
Just rebased and resolved conflicts. Things look fine on my side. Let me know if anything else needs to be changed.

avital:
Thanks @yang-song! I'll merge this PR

PR #925, opened 2021-01-22
Use doctest for new HOWTOs, instead of our old diff system

avital:
I split this PR from https://github.com/google/flax/pull/771
as it's orthogonal

This allows us to write embedded code snippets inside .rst files,
that are validated during the doc build. This seems like a better
alternative than using Colabs as documentation, except for cases
where we expicitly want people to fork and run a notebook.


avital:
@marcvanzee once this is merged you should be able to start using it to make some HOWTOs -- follow #771 for an example (with my current schedule it'll take me a few days to get it merged)

marcvanzee (approved):
Thanks for splitting this off!

PR #926, opened 2021-01-22
Allow passing in unbound modules into other modules

avital:
This change enables a `nn.Sequential` type pattern.

(though it's still confusing that you have to separately assign
submodules in `setup` -- #864 explores automating this somewhat.)

The original code was designed to allow using unbound modules as
"module templates" which you could bound arbitrarily in your tree,
but if people want that after this PR, they can always pass in lambdas.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -438,6 +438,10 @@ class Module:
             raise ValueError(
                 "You can only assign submodules to self in setup().")
           if subvalue.parent is _unspecified_parent:
+            raise ValueError("Unexpected -- parent should have been assigned during __post_init__")
+          elif subvalue.parent is None:
+            # Attaching unbound modules -- this pattern is used
+            # for "model surgery" on `nn.Sequential` type models.
             subvalue.parent = self
           elif subvalue.parent is not self:
             raise ValueError("Can't attach to remote parent in setup, pass in "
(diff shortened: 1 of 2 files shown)


avital:
(Thinking a bit more about this -- we may need to `clone` the submodule during `__setattr__`, otherwise if you pass in the same `Conv` instance into two Sequentials it will fail). 

This may also cause problems with JAX transformations inside modules, I haven't thought too carefully about that.

jheek:
This is related to https://github.com/google/flax/pull/864

Passing in Modules is brittle at the moment. I think #864 handles a few more edge cases but also doesn't work yet when using the pattern inside setup.

avital:
@jheek Yes but this change is a very small one that I think doesn't make anything worse -- do you mind if we merge this in, in the meanwhile before generalizing? Users have been asking about this recently (e.g. how do you implement Sequential -- without this I think it may not be possible)

jheek:
If we merge something we should probably merge #864 it does seem like an improvement to me and works in more cases.
That PR would also make sequential work.

But I prefer not to merge something half baked. Right now users basically get an error right away when passing modules instances through attributes (except in compact where it all works fine). I think this is better than having half-baked and buggy support for it.

levskaya:
I think we need to merge something a little more complete than this, and realize belatedly that this change would actually interfere w. the larger fixes if users started expecting this behavior.

PR #933, opened 2021-01-23
[.github/workflows/build.yml] Add pylint

SamuelMarks:
As per request in #841.

Maybe before this is merged I'll create another PR to get all the lint errors to pass…

PR #841, opened 2021-01-15
[flax/core/scope.py] Rename "Arguments:" to "Args:"

SamuelMarks:
I've written custom parsers and emitters for everything from docstrings to classes and functions. However, I recently came across an issue with the JAX and flax codebase: inconsistent use of `Args:` and `Arguments:` in its docstrings:

```sh
(flax#cb128e)$ for name in 'Args:' 'Arguments:'; do
    printf '%-10s %04d\n' "$name" "$(rg -IFtpy --count-matches "$name" | paste -s -d+ -- | bc)"; done
Args:      0328
Arguments: 0002
```

It is easy enough to extend my parsers to support both variants, howevever it looks like `Arguments:` is wrong anyway, as per:

  - https://google.github.io/styleguide/pyguide.html#doc-function-args @ [`ddccc0f`](https://github.com/google/styleguide/blob/ddccc0f/pyguide.md)
  - https://chromium.googlesource.com/chromiumos/docs/+/master/styleguide/python.md#describing-arguments-in-docstrings @ [`9fc0fc0`](https://chromium.googlesource.com/chromiumos/docs/+/9fc0fc0/styleguide/python.md)
  - https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html @ [`c0ae8e3`](https://github.com/sphinx-contrib/napoleon/blob/c0ae8e3/docs/source/example_google.rst)

Therefore, only `Args:` is valid. This PR replaces them throughout the JAX codebase.

PS: For similar contributed to the [TensorFlow](https://github.com/tensorflow/tensorflow), [keras-team](https://github.com/keras-team), and [PyTorch](https://github.com/pytorch/pytorch) organisations, see trackbacks at https://github.com/tensorflow/tensorflow/pull/45420.

PR #935, opened 2021-01-24
[*.py] Lint

SamuelMarks:
I'm a little worried that this PR changes too much and thus will never be merged.

The first commit is from:
```
fd -HIepy -x bash -c 'autoflake --remove-all-unused-imports -i "$0" && isort --atomic "$0" && python -m black "$0"' {} \;
```

The goal is to have all these lint checks passing:
```
flake8 . --count --select=$(printf '%s,' {A..Z}) --ignore='W503,E203,E402,E731,E741,C901' --show-source --max-line-length=119 --statistics 
pylint flax -rn --load-plugins=pylint.extensions.check_docs --accept-no-param-doc=n
```

This is related to this PR #933


tests/linen/module_test.py
```diff
@@ -768,18 +768,6 @@ def bar(self):
         # attributes defined during `setup`
         self.assertEqual(empty.bar(), 3)
 
-    def test_call_unbound_noncompact_module_methods(self):
```
avital: Why as this test removed?

avital (commented):
Thank you Samuel for spending the energy investigating this! I have always wanted us to be properly linting our code, this looks like a great direction.

Would the `flake8` and `pylint` commands you described pass on this branch? If so, that's a pretty strong reason for us to merge it! (and looks like it already removed some unused variables)

avital:
We would have to review this carefully as it would break the git blame chain, and that would happen every time we change lint configuration going forward

tests/linen/module_test.py
SamuelMarks: It was duplicated. See how it looked before https://github.com/google/flax/blob/3448764de94461bd08caef2cf90e113a711ffa3c/tests/linen/module_test.py#L759

SamuelMarks: It was duplicated. See how it looked before https://github.com/google/flax/blob/3448764de94461bd08caef2cf90e113a711ffa3c/tests/linen/module_test.py#L759-L769 and https://github.com/google/flax/blob/3448764de94461bd08caef2cf90e113a711ffa3c/tests/linen/module_test.py#L771-L781

SamuelMarks:
@avital Yep, I agree completely. So that's why it's important to get linters setup as early as possible in the project. Introducing these to TensorFlow for example would be completely untenable [without some executive decision: _good luck_], as there are over 200 open PRs.

In terms of what linters pass currently, all the `flake8` tests pass. Just wanted to wait for your feedback before proceeding with updating [almost] all the docstrings.

jheek:
We should probably update the pylint configuration with the style that is used right now.
We need at least 2 space indent instead of 4 space otherwise pretty much every line changes

SamuelMarks:
@jheek One can just add this line to the GitHub Action:
```
flake8 . --count --select=$(printf '%s,' {A..Z}) --ignore='W503,E203,E402,E731,E741,C901' --show-source --max-line-length=119 --statistics 
```

In terms of two spaces versus four, I suppose I can modify that easily, but not sure that's good practice. With the exception of Google repositories, most every repo I see is four spaces (not tabs). Anyway I'm happy if a little surprised that you're actually considering merging this that changes so many lines.

Good!

jheek:
I think it's best to pick the dominant style of coding as source of truth for the linting rules for now. 
That way the issue doesn't become too invasive.
We can always switch some style in the codebase to a different indentation or string quote but that shouldn't be in the linting PR.
Otherwise we are going to mix a discussion about consistency (which we obviously want) with subjective discussions like indentations (which obviously everyone is going to disagree on)

marcvanzee:
Closing this due to inactivity. Also, we are incrementally fixing linting errors in our codebase, but this is quite a difficult process due to how we set up our linting infrastructure, so submitting it all in one PR is unfortunately not feasible.

SamuelMarks:
Darn

marcvanzee:
Hi @SamuelMarks, I was going over all PRs and issues and realize now I probably have closed this PR too soon.

Apologies! I didn't mean to disregard all your hard work. We are trying to minimize the number of open issues so we can keep maintaining our repo effectively.

I have now read both this PR and the discussion in #841 in more detail.

> So that's why it's important to get linters setup as early as possible in the project. Introducing these to TensorFlow for example would be completely untenable [without some executive decision: good luck], as there are over 200 open PRs.

I share this sentiment! One problem is that we are internally using `pylint` (which is a google product), but the external version is slightly different, so if we decide to enforce it we will run into problems.

The pylint team recommended using something like this: https://github.com/google/styleguide/blob/gh-pages/pylintrc

Anyway I am reopening this issue, sorry again!

PR #938, opened 2021-01-25
Call setup lazily.

levskaya:
We need to call `setup` lazily rather than immediately at object construction to avoid a nasty exponential double-recursion via repeated setup calls when methods are under nested transformations  (e.g. named_call for profiling.).

I'm aware that this is the opposite direction of what some of us had chatted about before vis a vis `setup` behavior, but I think this is an unavoidable fix given the way we handle transforms (by recreating the modules as clones inside a pure function).

With this PR setup is called _once_ lazily per instance:
  1. Just before any module method such as `__call__` is invoked.
  2. When as-yet non-existent setup-defined attributes are accessed (via `__getattr__`).

A simple motivating example for why this needs to be done if our users begin writing larger models in "setup-style":
```python
D, N = 4, 3

class Repeat(nn.Module):
  mdl_def: Any
  def setup(self):
    self.lyrs = [self.mdl_def() for _ in range(N)]
  def __call__(self, x):
    for lyr in self.lyrs:
      x = lyr(x)
    return x

def nested_repeat(mdl):
  for _ in range(D):
    mdl = functools.partial(Repeat, mdl)
  return mdl()

dense = functools.partial(nn.Dense, 2, use_bias=False)
_ = nested_repeat(dense).init(random.PRNGKey(0), jnp.ones((2,)))

# with named call enabled, before and after this PR:
# D=4, N=3 : 24sec --> 250ms  speedup: ~100x
# D=4, N=4 : ~infinity? --> 699ms  speedup:  >1000x
# D=3, N=5 : 68sec --> 645ms  speedup: ~100x
# D=3, N=10:  ~infinity? --> 2.3sec  speedup: >1000x
```

This PR also fixes some sloppiness in our linen decorators, guaranteeing that the module state-management decorator always occurs only once at the "bottom" of a decorator stack containing linen transforms.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -65,6 +65,7 @@ def _attr_repr(value: Any):
     value_rep = repr(value)
   return value_rep
 
+
 def _module_repr(module: 'Module', num_spaces: int = 4):
   """Returns a pretty printed representation of the module"""
   cls = type(module)
@@ -90,6 +91,7 @@ def _module_repr(module: 'Module', num_spaces: int = 4):
     return f'{cls_name}(\n{_indent(rep, num_spaces)})'
   else:
     return f'{cls_name}()'
+
 
 # Track parent relationship across Modules.
 # -----------------------------------------------------------------------------
@@ -205,7 +207,7 @@ def _get_local_method_names(cls: Any, exclude: Iterable[str] = ()) -> Tuple[str]
   return tuple(true_methods.difference(set(exclude)))
 
 
-def wrap_method(fun: Callable[..., Any]) -> Callable[..., Any]:
+def wrap_method_once(fun: Callable[..., Any]) -> Callable[..., Any]:
   """Manages Module state for a given user-defined method.
   
   Args:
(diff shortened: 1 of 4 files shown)


flax/linen/module.py
```diff
@@ -461,6 +478,21 @@ def __setattr__(self, name: str, val: Any):
     # Finally, always run default __setattr__ to attach to self.__dict__.
     object.__setattr__(self, name, val)
 
+  def __getattr__(self, name: str) -> Any:
```
jheek, line 481: stupid edge case: but I think you need to override `__dir__` because it should also trigger setup.
We also use `__dir__` ourselves in the module code sometimes so probably worth it.

flax/linen/module.py
```diff
@@ -344,6 +359,7 @@ def __init_subclass__(cls):
     cls._wrap_module_methods()
     # Set empty class defaults.
     cls._state = _uninitialized_module_internal_state
+    cls._setup_called = False
```
jheek, line 362: I think this belongs in ModuleInternalState

jheek:
I think this is the right solution. Without the lazy trick there will either be double recursion or submodule attributes become completely inaccessible.


> I'm aware that this is the opposite direction of what some of us had chatted about before vis a vis setup behavior,

That was my intuition too. But actually I think this solution would allow setup to be called without a Scope while avoiding the double recursion issue.

jheek (changes requested):

flax/linen/module.py
levskaya: Actually this is interesting, and correct me if I'm wrong, but I -think- this actually should be object-local, rather than state that is copied across transform boundaries.  We don't want `_setup_called` True inside a cloned object to copy to the outside object, since it in fact has _not_ had `setup` called. (For instance __call__ has been called under a transform, but now a user triggers `__getattr__`, or in the case of multiple methods - a second method needs to rebuild the object inside its own transform context).  We could still stick it inside `ModuleInternalState` but it would require manually writing these  scoping rules.

flax/linen/module.py
levskaya: Done.

flax/linen/module.py
jheek: I would put it in ModuleInternalState regardless. Just makes it easier to handle when you have less magic attributes especially with all this setattr getattr logic.

jheek (approved):

flax/linen/module.py
levskaya: Yeah you're right, all runtime state should be kept together and the transform-boundary behavior made explicit - fixed.

PR #939, opened 2021-01-25
Improves documentation in various places

marcvanzee:
For the "how to cite Flax" I mirrored [what the JAX team](https://github.com/google/jax#citing-jax) did very closely.


flax/optim/base.py
```diff
@@ -122,8 +122,10 @@ def create(self, target, focus=None):
     See docstring of :class:`Optimizer` for more details.
 
     Args:
-      target: the object to be optimized. This will typically be
-        an instance of `flax.nn.Model`.
+      target: the object to be optimized. This will typically be an instance of
+        `flax.linen.Module`, but it can also be a container with Modules, e.g.
```
levskaya, line 126: this isn't quite correct - we don't optimize w. `Module` as a target, but rather the 
variables" that come from a `Module.init`.

levskaya (changes requested):

flax/optim/base.py
marcvanzee: Oh yes of course, updated it.

levskaya (approved):

PR #946, opened 2021-01-27
Adopt toplevel modules and fix setup deferred handling.

levskaya:
Two things:

1. Introduces the ability to use top-level instantiated modules as attributes in another top-level module.  This is a very common workflow during interactive development and it has been cited as a pain-point and a badly confusing gotcha a number of times by our users.

All top-level instantiated  modules passed in as attributes are 'adopted' or 'reparented' onto the root, top-level module they are passed into.  Effectively this function registers any such top-level instantiated submodule as if they had been defined in the top setup() function, taking care to avoid name collisions.

2. In `setup()` code like this currently breaks:
   ```python
   self.foo = Bar(submod=nn.Dense(2))
   ```
  since our current `__setattr__` binding logic doesn't recurse into module attributes.  This PR as-is does not yet solve some of the weirder edgecases around shared-modules in e.g. a list of modules passed to setattr, as this would require relaxing the name override check.  With this change the above snippet will have the same variable tree as if the same code has been written in "compact" style.

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -52,7 +52,8 @@ def _dedup_scopes(scopes):
       path.append(scope.name)
       scope = scope.parent
     if max_parent is not leaf:
-      del minimal_set[leaf]
+      if leaf in minimal_set:
+        del minimal_set[leaf]
     paths.append((max_parent, max_parent_path))
   return tuple(minimal_set), tuple(paths)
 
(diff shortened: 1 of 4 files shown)


levskaya:
for history's sake: this is a cleaned up copy of part of the old #864 PR

avital:
Is this PR superseded by https://github.com/google/flax/commit/9bf71619dcd9eae0edc7a72ce0864d25de8eb6cf?

levskaya:
This PR has been superseded by the merged #976

PR #947, opened 2021-01-27
Add automatic giant array chunking in msgpack checkpoints.

levskaya:
msgpack can only support total leaf encoded buffers sizes of max length 2^32-1  Some giant embedding arrays exceed this, so add an automatic reversible array chunking pass to msgpack serialization.

This PR should -not- break compatibility with existing checkpoints.

### DIFF
--- a/flax/serialization.py
+++ b/flax/serialization.py
@@ -12,20 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
-# Copyright 2020 The Flax Authors.
-#
-# Licensed under the Apache License, Version 2.0 (the "License");
-# you may not use this file except in compliance with the License.
-# You may obtain a copy of the License at
-#
-#      http://www.apache.org/licenses/LICENSE-2.0
-#
-# Unless required by applicable law or agreed to in writing, software
-# distributed under the License is distributed on an "AS IS" BASIS,
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
 """Serialization utilities for Jax.
 
 All flax classes that carry state like Model and Optimizer can be turned into a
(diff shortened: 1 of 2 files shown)


flax/serialization.py
```diff
@@ -246,6 +246,74 @@ def _msgpack_ext_unpack(code, data):
...
+
+
+_tuple_to_dict = lambda tpl: dict([(str(x), y) for x, y in enumerate(tpl)])
+_dict_to_tuple = lambda dct: tuple(dct[str(i)] for i in range(len(dct)))
+
+
+def _chunk(arr):
+  """Convert array to a canonical dictionary of chunked arrays."""
+  if isinstance(arr, jax.xla.DeviceArray):
+    arr = np.array(arr)
```
jheek, line 266: I'm always a bit worried about the serialisation logic operating on jax arrays. Should we just move things to numpy at the very start of serialisation?

flax/serialization.py
```diff
@@ -246,6 +246,74 @@ def _msgpack_ext_unpack(code, data):
...
+  chunksize = max(1, int(MAX_CHUNK_SIZE / arr.dtype.itemsize))
+  data = {'__msgpack_chunked_array__': True,
+          'shape': _tuple_to_dict(arr.shape),
+          'chunks': {}}
+  flatarr = arr.reshape(-1)
+  idx, pos = 0, 0
```
jheek, line 272: nit: this loop duplicates the tuple_to_dict and range logic. I think it's nicer to have:
```
chunks = [flatarr[i: i + chunksize] for i in range(0, flatarr.size, chunksize)]
data['chunks'] = _tuple_to_dict(chunks)
```

jheek (commented):
Looks good just some minor comments

flax/serialization.py
levskaya: sure, added a pass to do this first.

flax/serialization.py
levskaya: yeah, agreed - replaced it.

PR #949, opened 2021-01-27
Proposal for better error messages

marcvanzee:
This is a proof-of-concept proposal for improving our error messages.  It currently implements a few cases in `flax.core.scope`. If we get consensus on this proposal then I will continue to implement it for our entire code base (~70 cases in total).

I think the proposal improves the user experience a lot, since we can give concrete code examples in the error messages, rather than throwing generic Python exception with a one-line explanation. The error class doc-strings serve as documentation on RTD directly, which makes the proposal quite light-weight.

RTD demo: https://flax.readthedocs.io/en/improve-error/flax.errors.html

Before:

```
>>> from flax.core import scope
>>> scope.in_filter(3, 4)
  File "<stdin>", line 1, in <module>
  File "/Users/marcvanzee/github/flax/flax/core/scope.py", line 105, in in_filter
    raise TypeError(f'Invalid Filter: "{filter_like}"')  
TypeError: Invalid Filter: "3"
```

After:

```
>>> from flax.core import scope
>>> scope.in_filter(3, 4)
  File "<stdin>", line 1, in <module>
  File "/Users/marcvanzee/github/flax/flax/core/scope.py", line 105, in in_filter
    raise InvalidFilterError(f'Invalid Filter: "{filter_like}"')
flax.errors.InvalidFilterError: Invalid Filter: "3" (https://flax.readthedocs.io/en/improve-error/flax.errors.html#flax.errors.InvalidFilterError)
```


jheek:
I don't have a super strong opinion but just for consideration:

We could make the message the responsibility of the error class itself so you have things like:

```
class InvalidFilterError(FlaxError):
  def __init__(self, filter_like):
    super().__init__(f'Invalid Filter: "{filter_like}"')
    self.filter_like = filter_like
```

This way all error messages are in one module which should also make maintenance easier. For example it is very easy to keep messages consistent in style.

docs/index.rst
```diff
@@ -56,11 +56,12 @@ For a quick introduction and short example snippets, see our `README
    :maxdepth: 2
    :caption: API reference
 
+   flax.errors
```
avital, line 59: I don't think we should change the order here -- things should appear in order of importance.

I'd put errors at the bottom, and definitely not move jax_utils

flax/core/scope.py
```diff
@@ -323,9 +324,9 @@ def reserve(self, name: str):
       name: the name to reserve.
     """
     if not isinstance(name, str):
-      raise ValueError('Variable and child scopes should have a string name.')
+      raise ScopeNamingError('{name} is not a string.')
```
avital, line 327: Would the "mini-guide" for this use of `ScopeNamingError` and the one right below be different? (I would think so -- the thing you'd have to do to fix it, or what you need to understand behind it are different) so should they be different class types?

flax/errors.py
```diff
@@ -0,0 +1,67 @@
...
+  """
+  Collections that are immutable cannot be modified. When you are applying a
+  Module, you should specify which variable collections are mutable::
+
+    class MyModule(nn.Module):
+      @nn.compact
+      def __call__(self, x):
+        ...
+        var = self.variable('batch_stats', 'mean', ...)
+        ...
```
avital, line 36: To really make this example tight, add `var.value = ...` (otherwise it's valid to not make `batch_stats` mutable if you pass it in in `apply` and it's not modified)

flax/errors.py
```diff
@@ -0,0 +1,67 @@
...
+    vars = MyModule.init(...)
+    ...
+    logits = MyModule.apply(vars, batch)  # This throws an error.
+    logits = MyModule.apply(vars, batch, mutable=['batch_stats'])  # This works.
+  """
+
+
+class ScopeNamingError(FlaxError):
+  """
+  Scope names should be strings and unique within a subscope::
```
avital, line 47: I think we should not use the term "scope" in user-facing errors (both class names and more concretely in error docstrings) because 90% of users (my random assessment) will only interact with Modules and have never heard of scopes. No need to force them to learn about that just because they got this error.

avital: (If we're raising this error inside scope.py rather than module.py then maybe we should add another check in module.py to raise a better error, or use a vague name like "VariableNameReuseError".)

avital (approved):
I like this! I think I agree with Jonathan's point, but also don't feel strongly.

flax/core/scope.py
marcvanzee: Yes I agree, updated it.

flax/errors.py
marcvanzee: yes good point!

flax/errors.py
marcvanzee: I agree. For now I've named these error to not include "Scope", and added a TODO saying we should raise these errors in module.py and rename them.

marcvanzee:
> We could make the message the responsibility of the error class itself so you have things like:

Yes I like that idea, I've implemented it.

docs/flax.errors.rst
```diff
@@ -0,0 +1,9 @@
+
+flax.error package
+===================
+
+Flax has the following classes of errors.
+
+.. automodule:: flax.errors
+    :members:
+    :exclude-members: FlaxError
```
avital, line 9: BTW I didn't know about `:exclude-members:` -- this is helpful!

flax/errors.py
```diff
@@ -0,0 +1,120 @@
...
+    ...
+    logits = MyModule.apply(vars, batch)  # This throws an error.
+    logits = MyModule.apply(vars, batch, mutable=['batch_stats'])  # This works.
+  """
+  def __init__(self, col, variable_name, scope_path):
+    super().__init__(f'Cannot update variable "{variable_name}" in '
+                     f'"{scope_path}" because collection "{col}" is immutable.')
+
+
+# TODO(marcvanzee): Make sure this error is thrown in Modules and not in Scope
```
avital, line 85: To clarify, my position is a bit less strong: it's fine if errors are thrown in Scope. As long as the *user-facing* message doesn't require understanding scope.

avital (approved):
Thanks Marc, I think we should merge this and iterate onwards.

One additional thing that would be useful (not as part of this PR) is to make sure we have tests for each error type -- that indeed the right error class is raised under each situation.

Are you keeping a list of TODOs for this errors project? Could you add that there?

flax/errors.py
marcvanzee: Yes I agree.

jheek (approved):

marcvanzee:
> Are you keeping a list of TODOs for this errors project? Could you add that there?

I created a tracking issue containing all TODOs: #957

PR #950, opened 2021-01-27
Fix lazy setup bug

jheek:
Fixes cases where setup is called on an unbound module:

- setup could be called on modules that have a parent but no name
- setup could be called recursively within setup if calling other methods


levskaya (approved):
Looks good, thanks!  Checking existence of `scope` is obviously the correct thing to do.  
And thanks for unifying the try_setup checks, I had meant to do that and forgot!

jheek:
Judging the positive sentiment of your comment, I think you forgot to press approve ;)

PR #952, opened 2021-01-27
Log info message when a checkpoint is not found in the directory.

hmph:


levskaya (approved):
Thanks, there was a recent commit that altered this code region so I merged it in with that.

PR #953, opened 2021-01-28
Proposal for a new HOWTO side-by-side view

marcvanzee:
This is a proposal for showing side-by-side diff code for our HOWTOs in Read The Docs.

It turns out that it is possible to show code diff side-by-side by combining Sphinx tables with code blocks. However, this is a pain to write since the tables have to be nicely aligned (see example in the test file of this PR).

To simplify this, this PR proposes a new Sphinx extension `codediff`, that takes two code snippets separated by `---`, and rewrites this to two code blocks that are shown side-by-side in a Sphinx table.

Here's a live demo of the "ensembling HOWTO" (note: this currently only shows the diff, the HOWTO itself still has to be written): https://flax.readthedocs.io/en/howto-ensembling/howtos/ensembling.html


docs/_ext/codediff_test.py
```diff
@@ -0,0 +1,67 @@
...
+  extra_line
+  return initial_params
+---
+@jax.pmap
+def get_initial_params(key):
+  init_val = jnp.ones((1, 28, 28, 1), jnp.float32)
+  initial_params = CNN().init(key, init_val)['params']
+  return initial_params'''
+
+    expected_output= r'''+----------------------------------------------------------+----------------------------------------------------------+
```
avital, line 37: Nice test!

docs/howtos/ensembling.rst
```diff
@@ -0,0 +1,25 @@
...
+=============================
+
+We show you how to rewrite code for training a single mode to training an 
+ensemble of models, one on each device.
+
+Example of the new directive:
+
+.. codediff::
+  :title_right: Wit ensembling
+  :highlight_left: 1
```
avital, line 11: Should we also consider a version where the line numbers aren't specified here, but instead embedded in the code via a comment like `#  <highlight>`? (that would then get stripped in the output view, and instead that line would be highlighted)

The upside is that it's easier to then edit HOWTOs without having to recompute the line numbers...

WDYT?

avital (approved):
Super cool! I left one question but I think we can definitely merge this and start from here...

In your live demo (https://flax.readthedocs.io/en/howto-ensembling/howtos/ensembling.html), though the left and right side line numbers don't seem to align correctly. I didn't understand if this is expected.

One additional question: if you try getting this to work with `testcode` blocks -- what is the right behavior? Would it run both sides? I wonder if that will work well in all examples -- at the very least you'd have to make sure both sides of the diff are "full functions" (not just function snippets)... Just a thought -- we can start trying and see how it works out.

docs/howtos/ensembling.rst
marcvanzee: Yes I think that is actually nicer. I've implemented it and also simplified the code a bit.

marcvanzee:
your live demo (https://flax.readthedocs.io/en/howto-ensembling/howtos/ensembling.html), though the left and right side line numbers don't seem to align correctly. I didn't understand if this is expected.

Hmm yes I also noticed that. It is kind of weird and I will look into it in a follow-up.

> One additional question: if you try getting this to work with `testcode` blocks -- what is the right behavior? Would it run both sides? I wonder if that will work well in all examples -- at the very least you'd have to make sure both sides of the diff are "full functions" (not just function snippets)... Just a thought -- we can start trying and see how it works out.

Yes this is also something we should think about better.

PR #954, opened 2021-01-28
Minor README updates

avital:


Marvin182 (approved):
One could also say that Flax started inside the Brain Team but is now developed by the researchers community. Even if right now the majority of contribution are still coming from Google.

avital:
> One could also say that Flax started inside the Brain Team but is now developed by the researchers community. Even if right now the majority of contribution are still coming from Google.

I gave it a shot here: https://github.com/google/flax/pull/960

PR #976, opened 2021-02-03
Generalize attribute assignment to support passing module instances

jheek:
- Fix attributes not being frozen
- Clone detached modules when passed as attributes
- Harmonize attributes and fields by handling submodules the same way.

Co-authored-by: Anselm Levskaya <levskaya@google.com>


levskaya (approved):
This all looks great, and much cleaner than my initial prototypes!  So sorry that I was so slow to review.

PR #989, opened 2021-02-04
Add v1 logo to RTD.

avital:
If we link this, we can also add this same image into
the GitHub README.

But thought I'd try here first.

See rendered docs with logo here: https://flax--989.org.readthedocs.build/en/989/


andsteing:
I think we should position it differently. In the current form it removes almost 50% of screen estate on a typical screen (landscape orientation).

I checked some other popular readthedocs pages and they often have the logo in the side menu (like [JAX](https://jax.readthedocs.io/en/latest/)), or have a much smaller logo (like [solidity](https://docs.soliditylang.org/)). We could also experiment with having it on the right side with text flowing around it (didn't see any examples doing that, but it would allow to have the logo on top and not stealing screen estate from main page or menu).

I also don't think that we need "FLAX" in the logo if it's written as a title just under it.

8bitmp3:
It's a great-looking logo @avital. 

I'd suggest a few minor things: 1) trim the margins a little (see below); 2) like @andsteing said, have the logo in the sidebar (like JAX's ReadTheDocs); 3) have a high resolution version somewhere; 4) add a low-res favicon 🚀 

Awesome logo 👍 

So the margins make the whole thing take up a bit "too much" space, IMHO:

image

Then, when you zoom in, it looks pixelated, so a high res PNG would fix it:

image

P.S. Looking forward to FLAX stickers.

avital:
Hi @8bitmp3 -- thanks for chiming in :) I see you've thought about this a lot more than me.

Any interest in fixing the padding and figuring out how best to integrate the image into the docs? Here is a vector PDF of the raw logo, in case you're excited about this. (If not that's of course fine, but just wanted to supply anything needed in case you or someone else wants to go ahead and do it)

[flax sketch.pdf](https://github.com/google/flax/files/5931269/flax.sketch.pdf)

avital:
Regarding the favicon, it may not look good with all of the "layers" in such a small resolution, we'd need to check. But perhaps if we remove the green layers and only keep blue and purple then there will be less detail and it would look better as a favicon

avital:
(oops, sorry, that was the old logo version -- I'm deleting my comment, I'll upload the PDF of the new one shortly)

avital:
Here's the latest logo in PDF form: 
[Uploading flax sketch_03.pdf…]()

joaogui1:
The link is broken

8bitmp3:
> Any interest in fixing the padding and figuring out how best to integrate the image into the docs? 

@avital Ok

avital:
> The link is broken
Oops! I updated the comment above with a working link to a PDF file.

joaogui1:
Looks really cool, looking forward to the stickers :D

andsteing (changes requested):
(Same text I put into a comment previously, this time as a "request changes" review for better visibility)

I think we should position it differently. In the current form it removes almost 50% of screen estate on a typical screen (landscape orientation).

I checked some other popular readthedocs pages and they often have the logo in the side menu (like [JAX](https://jax.readthedocs.io/en/latest/)), or have a much smaller logo (like [solidity](https://docs.soliditylang.org/)). We could also experiment with having it on the right side with text flowing around it (didn't see any examples doing that, but it would allow to have the logo on top and not stealing screen estate from main page or menu).

I also don't think that we need "FLAX" in the logo if it's written as a title just under it.

8bitmp3:
@andsteing how about:

image

8bitmp3:
https://github.com/avital/flax/pull/2 @avital

andsteing:
I would vote for having the logo only in the "main" section on the right - i.e. above / besides (float layout?) the "Flax documentation" title.

Having it in the left hand menu wastes a lot of vertical space. For example on my macbook screen I currently see the first part of "API REFERENCE" only. If we waste vertical space for the logo then I won't even see the "API REFERENCE" without scrolling...

Also, I think for the RTD docs it's actually more important having a good favicon. Because that will really help users finding the open tab with the docs, while the big logo on the page itself looks nice but gives little added value.

andsteing:
Implemented as #1392

PR #999, opened 2021-02-05
Updates some error messages in module.py and scope.oy

### LINKED ISSUE #893: Improve Error Message: Calling a method without initializing
Example code:

```
from flax import linen as nn
import jax.numpy as jnp
test_dense = nn.Dense(10)
test_dense(jnp.ones((5,5)))
```

Gives error:

```

    146 
    149 
    150     if is_compact_method:

```

This code does work:

```
```

marcvanzee:
Closes #893

### DIFF
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -25,8 +25,32 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Flax error classes."""
+"""Flax error classes.
 
+=== Suggested error classes naming conventions:
+
+* If the error occurs when doing something, name the error
+  <Verb><Object><TypeOfError>Error
+
+  For instance, if you want to raise an error when applying a module with an 
+  invalid method, the error can be: ApplyModuleInvalidMethodError.
+
+ <TypeOfError> is optional, for instance if there is only one error when 
+  modifying a variable, the error can simply be: ModifyVariableError.
+
+* If there is no concrete action involved the only a description of the error is
+  sufficient. For instance: InvalidFilterError, NameInUseError, etc.
+
+
+=== Copy/pastable template for new error messages:
+
+class Template(FlaxError):
+  "" "
+
+  "" "
+  def __init__(self):
+    super().__init__(f'')
+"""
 
 class FlaxError(Exception):
   def __init__(self, message):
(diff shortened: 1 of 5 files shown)


flax/errors.py
```diff
@@ -117,4 +116,44 @@ def __call__(self, x):
...
+  this, add this to your imports::
+    
+    from jax.config import config
+    config.enable_omnistaging()
+  """
+  def __init(self):
+    super().__init__(f'Flax Linen requires Omnistaging to be enabled')
+
+
+class UnboundCompactModuleError(FlaxError):
```
avital, line 134: This new error alone is so valuable

flax/errors.py
```diff
@@ -117,4 +116,44 @@ def __call__(self, x):
...
+
+    test_dense = nn.Dense(10)
+    test_dense(jnp.ones((5,5)))
+
+  Instead, you should pass the variables (parameters and other state) via 
+  :meth:`Module.apply() <flax.linen.Module.apply>` (or use 
+  :meth:`Module.init() <flax.linen.Module.init>` to get initial variables)::
+
+    # Create the initialized variables with a random key for inits:
+    params = test_dense.init(random.PRNGKey(0), jnp.ones((5,5)))['params']
```
avital, line 151: I think this code doesn't work. It should be `variables = ` and don't do `['params']` at the end

flax/linen/module.py
```diff
@@ -231,7 +228,7 @@ def wrapped_module_method(self, *args, **kwargs):
 
     if is_compact_method:
       if self.scope is None:
-        raise ValueError("Can't call compact methods on unbound modules")
+        raise UnboundCompactModuleError(self.name)
```
avital, line 231: I'm curious whether we also get this error when calling non-compact methods

flax/errors.py
avital: Ah, I get a different error then -- `setup` doesn't run so the submodule I defined isn't there

https://colab.research.google.com/drive/1SsRzkBoGqf1DNYRVseh5QxjA_SwCb7gb

marcvanzee: I don't seem to get a different error when I do as you suggested:

```
from flax import linen as nn
import jax.numpy as jnp

test_dense = nn.Dense(10)

 # Create the initialized variables with a random key for inits:
import jax.random
vars = test_dense.init(random.PRNGKey(0), jnp.ones((5,5)))

# Apply the NN to the variables + input to get output.
y = test_dense.apply(vars, jnp.ones((5,5)))
```

flax/linen/module.py
marcvanzee: It doesn't seem to be possible since it states `if is_compact_pmethod` before, right?

avital: Right, I mean that it would be good to similarly detect that -- this was independently filed as #1014

marcvanzee:
Closing this since I've created another PR #1070 which uses my local branch.

flax/linen/module.py
marcvanzee: Good idea, I've moved your suggestion to the description of the new PR #1070 and I'll implement it there.

PR #1000, opened 2021-02-05
Enable broadcast in lift.scan

jheek:
Fixes a bug that made broadcast impossible to use with `lift.scan`


marcvanzee:
Congrats on filing GH artefact number 1000!! 🎉 🎈 🎂

levskaya (approved):

bastings:
Thanks for this! :1st_place_medal:

Issue #906, opened 2021-01-19
Add explicit deprecation warning for flax.nn


avital:
`flax.nn` is planned to be officially no longer supported during May 2021 (in favor of `flax.linen`). Let's start putting deprecation warnings now, so that people aren't surprised too much later. /cc @marcvanzee

avital:
Also good to add deprecation messages to all docstrings. I just searched for "Flax Batch Norm" and found this: https://flax.readthedocs.io/en/latest/_autosummary/flax.nn.BatchNorm.html

A lot of users won't notice the small "(deprecated)" at the top in the nav bar

joaogui1:
Opened a PR fixing this @avital

avital:
go for it @joaogui1!

joaogui1:
I think this can be closed :D

avital:
Yup! Closed by #1001

PR #1001, opened 2021-02-05
Deprecation

joaogui1:
Adds deprecation warnings per #906


joaogui1:
Hey @avital can you review this please?

flax/nn/attention.py
```diff
@@ -45,7 +45,9 @@ def dot_product_attention(query,
                           dropout_rate=0.,
                           deterministic=False,
                           precision=None):
-  """Computes dot-product attention given query, key, and value.
+  """DEPRECATION WARNING:
```
avital, line 48: Do we not need to also call `warnings.warn(..., DeprecationWarning)`? That's how we've done deprecations, e.g. see `flax.jax_utils.pmean`.

Or maybe we've been doing it wrong the whole time? We should standardize on the right way to do deprecations in Python.

avital: ...ah I see, you added it in flax/nn/init.py. OK maybe that's good.

flax/nn/__init__.py
```diff
@@ -29,3 +29,6 @@
 from .recurrent import LSTMCell, GRUCell, ConvLSTM, OptimizedLSTMCell
 from .stochastic import make_rng, stochastic, dropout, is_stochastic
 # pylint: enable=g-multiple-import
+import warnings
+warnings.filterwarnings("default", category=DeprecationWarning, module=__name__)
+warnings.warn("the nn module is deprecated, use linen instead", DeprecationWarning)
```
avital, line 34:
```suggestion
warnings.warn("The `flax.nn` module is Deprecated, use `flax.linen` instead. Learn more and find an upgrade guide at https://github.com/google/flax/blob/master/flax/linen/README.md", DeprecationWarning)
```

flax/nn/__init__.py
```diff
@@ -29,3 +29,6 @@
 from .recurrent import LSTMCell, GRUCell, ConvLSTM, OptimizedLSTMCell
 from .stochastic import make_rng, stochastic, dropout, is_stochastic
 # pylint: enable=g-multiple-import
+import warnings
+warnings.filterwarnings("default", category=DeprecationWarning, module=__name__)
```
avital, line 33: What does this line do? (Can we add a comment before it?) If we don't include it does the `warning.warn` below not work?

flax/nn/attention.py
```diff
@@ -166,7 +168,9 @@ def scan_in_dim(*args, **kwargs):
 
 
 class Cache(Collection):
-  """Collect intermediate activations for efficient autoregressive decoding."""
+  """DEPRECATION WARNING:
+  `flax.nn` is deprecated, use `flax.linen` instead.
```
avital, line 172: Please add the link to the README, like I suggested above, to these deprecation warnings as well.

flax/nn/__init__.py
avital, line 34: How does this work in Python? If someone does `import flax.nn` it will execute this code. But what if someone does `import flax.nn.base`?

avital (changes requested):
Thanks so much @joaogui1 ! I added a few small comments/requests

flax/nn/__init__.py
joaogui1: Deprecation warning are muted by default, this makes sure the person sees the warning
https://stackoverflow.com/a/51903782

flax/nn/__init__.py
joaogui1: I lost the reference, but yeah if import anything from it it shows a deprecation warning

flax/nn/__init__.py
avital: Can you write this as a comment in the code then, for future people looking at it?

joaogui1: Done!

avital (approved):

avital:
Thanks @joaogui1!

PR #1011, opened 2021-02-09
FLIP #1009: New Optimizer API

### LINKED ISSUE #1009: New Optimizer API
This FLIP discusses a new Optimizer API that is to replace the [current API].

[current API]: https://github.com/google/flax/blob/master/flax/optim/base.py

andsteing:
Adds a FLIP that proposes to replace the existing optimizer API with a new functional one.

Fixes #1009


docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,232 @@
...
+hyperparameters, and target variables. This pattern is relatively complex for
+implementing a simple optimizer, while being quite verbose in the typical Linen
+train step (especially when using mutable state collections).
+
+This FLIP proposes a new purely [functional API] where optimizers are simply
+defined with a `init()` and an `update()` function and a new [Linen helper] that
+reduces boilerplate in a typical Linen train step.
+
+
+# Functional API
```
andsteing, line 27: Could you add a subsection comparing this to Optax API?

avital: Yes, I would like to understand what a version where we adopt Optax looks like -- if you directly use the Optax API, how would this proposal change? Would we still need some utilities or a wrapper due to how we represent variable collections?

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,232 @@
...
+and from an `OptimizerDef` that defines how to update optimizer state,
+hyperparameters, and target variables. This pattern is relatively complex for
+implementing a simple optimizer, while being quite verbose in the typical Linen
+train step (especially when using mutable state collections).
+
+This FLIP proposes a new purely [functional API] where optimizers are simply
+defined with a `init()` and an `update()` function and a new [Linen helper] that
+reduces boilerplate in a typical Linen train step.
+
+
```
avital, line 26: (As more of a comment on the FLIP template: Perhaps we should have a "Motivation" section?)

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,232 @@
...
+- The functions `init_param_state()` and `apply_param_gradient()` are called
+  for every leaf in the params/grads pytree. This makes it possible to write the
+  calculations directly without `jax.tree_multimap()`.
+- The interface was defined in pre-Linen without the distinction of `params` vs.
+  other collections in `variables` in mind. The original API was elegant because
+  one only needed to pass around the optimizer, which included the parameters,
+  optimizer state, optimizer hyperparameters, and a reference to the
+  `OptimizerDef` to perform the param/state update.
+- Note that the current implementation contains some additional functionality
```
avital, line 182: This is another potential gap with Optax (at least at the moment) -- if we consider switching to Optax, we should find what the new pattern for `MultiOptimizer` and passing in `focus`, and whether we can (or should) keep a backwards compatibility layer.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,232 @@
...
+
+Our current API (referred to as [previous API] in this document) uses a pattern
+where an `Optimizer` dataclass is created from a pytree of `target` variables
+and from an `OptimizerDef` that defines how to update optimizer state,
+hyperparameters, and target variables. This pattern is relatively complex for
+implementing a simple optimizer, while being quite verbose in the typical Linen
+train step (especially when using mutable state collections).
+
+This FLIP proposes a new purely [functional API] where optimizers are simply
+defined with a `init()` and an `update()` function and a new [Linen helper] that
```
avital, line 23: Is the Linen helper still under construction? There's no link here.

docs/flip/1009-optimizer-api.md
andsteing: Added.

docs/flip/1009-optimizer-api.md
andsteing: Refactored the entire FLIP. It now starts with Optax in the motivation.

andsteing:
For a demonstration of using Optax (incl. `DynamicScale`) in an example see #1053 (still running - currently at 35.53% test accuracy after 4 epochs, which is in line with [previous runs](https://tensorboard.dev/experiment/iJzNKovmS0q6k5t6k5wvOw/#scalars&_smoothingWeight=0&regexInput=v100_x8_mixed_precision)).

docs/flip/1009-optimizer-api.md
andsteing: Added.

docs/flip/1009-optimizer-api.md
andsteing: Completed.

docs/flip/1009-optimizer-api.md
andsteing: Refactored the entire FLIP. Now Optax is in motivation & has its own subsection. See also #1053.

docs/flip/1009-optimizer-api.md
andsteing: Optax can be used with a tiny wrapper.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+
+We can then augment this dataclass with some code (similar to
+`flax.optim.Optimizer` and `flax.nn.Model`) that takes care of splitting and
+merging the variable collections appropriately:
+
+```python
+  def update_with_loss(self, loss_fn, inputs, labels, *, mutable=False):
+
+    def loss_from_params(params):
+      if mutable:
```
avital, line 330: If you use `mutable=[]` instead of `mutable=False` then you always get back a second `new_model_state` return value and you don't need this conditional. (See #913 )

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+        opt_state=new_opt_state,
+    ), loss
+```
+
+Users can derive from this dataclass and put any additional state into that
+object:
+
+```python
+@flax.struct.dataclass
+class TrainState(LinenHelper):
```
avital, line 356: It's pretty nice that this seems to work well with dataclass inheritence!

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+- Note how `optimizer.apply_gradients()` can take additional arguments to update
+  hyper parameters, like the learning rate from an independent function
+  `get_learning_rate()` in this case.
+
+
+# Update Plan
+[Update Plan]: #update-plan
+
+1. Finalize discussions on this FLIP
+2. Test existing optimizers for numerical equivalence (e.g. `flax.optim.Adam`
```
avital, line 506: I believe @mtthss was interested in helping with this -- it would entail copying over the Flax optimizers into the Optax test directory as the hope is to eventually deprecate flax.optimizers so we shouldn't do this by having optax tests depend on Flax itself.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+   and `optax.adamw`).
+3. Update examples to use Optax and verify that they reach the same final
+   performance with the same computational cost. We probably want some examples
+   to directly use the optimizer while others might use [Linen helper].
+4. Update all documentation to talk exclusively about Optax optimizers.
+5. Mark optimizers in `flax.optim` as deprecated.
+
```
avital, line 513: I think there are more things that need to be done:
* Update README and other documentation with examples (e.g. Flax guided tour)
* Update HOWTOs
* Port missing optimizers to Optax (e.g. Adafactor)

avital (changes requested):
Thanks for getting the ball rolling on this @andsteing! The Linen Helper in particular looks like it'll simplify Flax training loops significantly!

I added a few small comments and a larger question about the need for an optimizer abstraction. PTAL.

Also ccing @mtthss re: Optax, this document describes our work-in-process thinking about what it would take to transition Flax to use Optax and deprecate Flax optimizers. There are a few questions about patterns that Optax users use below, and happy to hear any other feedback!

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+A similar scheme could be implemented with optimizers following the functional
+API:
+
+```python
+# flax/opt/base.py
+
+FilterFn = Callable[[str], bool]
+
+
+def multi_optimizer(
```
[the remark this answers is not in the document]
n2cholas: I'm not sure if this is the idiomatic pattern in `optax`, but there is a approved [PR](https://github.com/deepmind/optax/pull/27) on a mask wrapper to update only a subset of the parameters in a pytree. I think this is a clean way to apply different optimizers to different parameters within an `optax.chain`.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+```
+
+Remarks:
+
+- We'll call `Optimizer` instances from the functional API `optim`, and refer
+  to `flax.optim.Optimizer` instances from the [previous API] as `optimizer`
+  for the remainder of this document.
+- As opposed to the [previous API], this API does not have the option to update
+  hyper parameters. When [using Optax], updating the hyper parameters is instead
+  done by defining a schedule that takes care of updating the hyper parameters
```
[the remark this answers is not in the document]
n2cholas: As far as I know, `optax` currently only supports scheduling the learning rate (using the `optax.scale` transform). Here is an [issue](https://github.com/deepmind/optax/issues/20) about scheduling other hyperparameters with some potential solutions.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+# Motivation
+[motivation]: #motivation
+
+Flax contains some optimizers in the `flax.optim` package, but that list is far
+from exhaustive and ideally we would instead use JAX optimizers from a dedicated
+PyPi package.
+
+Deepmind's [Optax] library is such a dedicated package that already implements a
+choice of interesting optimizers and it also provides a framework to compose new
+optimizers from reusable gradient transformations.
```
8bitmp3, line 48: Small suggestion:

- Nit/minor refactoring

```suggestion
DeepMind already has a dedicated library — [Optax] — that already implements a
wide range of interesting optimizers and provides a framework to compose new
optimizers from reusable gradient transformations.
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+    new_params, new_momentum = inner(params, grads, state.momentum)
+    return new_params, MomentumState(new_momentum)
+```
+
+Remarks:
+
+- We'll call `Optimizer` instances from the functional API `optim`, and refer
+  to `flax.optim.Optimizer` instances from the [previous API] as `optimizer`
+  for the remainder of this document.
+- As opposed to the [previous API], this API does not have the option to update
```
8bitmp3, line 131: Small suggestion: nit

```suggestion
- Compared with the [previous API], this new API does not have the option to update
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+  using the [previous API], but the difference is smaller with Linen. If you
+  care about a boilerplate-free training loop see the proposed [Linen helper]
+  further down.
+
+## Using Optax
+[Using Optax]: #using-optax
+
+Optax is centered on the idea of composable gradient transformations. Since
+above [functional API] is based on Optax's interface it is straightforward to
+wrap an Optax transformation to comply with the `Optimizer` interface:
```
8bitmp3, line 145: Suggestion: nit

```suggestion
Optax is centered on the idea of composable gradient transformations. Since
the above-mentioned [functional API] is based on Optax's interface it is straightforward to
wrap an Optax transformation to comply with the `Optimizer` interface:
```

("above-mentioned" or the "new"/"proposed")

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+opt_state = optim.init(variables['params'])
+train_step = make_train_step(model.apply, optim.update)
+for batch in ds.as_numpy_iterator():
+  opt_state, variables, loss = train_step(
+      opt_state, variables, batch['image'], batch['label'])
+```
+
+Remarks:
+
+- As opposed to the [previous API], we now can keep the entire `variables`
```
8bitmp3, line 289: Suggestion: nit

```suggestion
- Compared with the [previous API], we can now keep entire `variables`, 
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+[Linen helper]: #linen-helper
+
+While the above training loop is perfectly fine, there is quite a number of
+lines that are going to be exactly the same for every Linen training loop,
+namely the splitting of variables, passing in parameters separately because
+we only need the gradients for those, updating the parameters with the opimizer
+and then merging the updated parameters, any other updated collections, and
+the unchanged collections to generate the new variables. We also need to keep
+track of the model's `apply_fn`, the otpimizer's `update_fn`, and the optimizer
+state, in addition to the data.
```
8bitmp3, line 308: Suggestion: nit (some trimming/grammar/reducing the sentence complexity)

```suggestion
While the above training loop is perfectly fine, a number of lines are going to be
exactly the same for every Linen training loop, namely: 1) the splitting of variables,
2) passing in parameters separately because we only need the gradients for those;
3) updating the parameters with the opimizer and then merging the updated
parameters, any other updated collections; 4) and the unchanged collections to
generate the new variables. We also need to keep track of the model's 
`apply_fn`, the otpimizer's `update_fn`, and the optimizer state, in addition to the data.
```

(The first sentence was a bit hard to follow and I may have numbered the items incorrectly @andsteing )

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+Users can derive from this dataclass and put any additional state into that
+object:
+
+```python
+@flax.struct.dataclass
+class TrainState(LinenHelper):
+  step: int
+```
+
+This will reduce above training loop to something much simpler:
```
8bitmp3, line 360: Suggestion: nit

```suggestion
This will reduce the above training loop to something much simpler:
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+  signatures if there is demand for that.
+
+# Previous API
+[previous API]: #previous-api
+
+## Previous Implementation
+[Previous Implementation]: #previous-implementation
+
+The optimizer itself would be implemented by creating a new class deriving
+from `OpimizerDef`:
```
8bitmp3, line 403: Sorry, question: a new class "deriving" or "derived" from `OptimizerDef`? Maybe it's "derived" 🤷

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+    return new_params, _MomentumParamState(new_momentum)
+```
+
+Remarks:
+
+- Note the relationship between `OptimizerDef` and `Optimizer` : When the
+  function `Optimizer.apply_gradient()` is called from the user code, it calls
+  into `OptimizerDef.apply_gradient()` (among other things) which in turn will
+  call `OptimizerDef.apply_param_gradient()` which is implemented by subclasses
+  of `OptimizerDef`.
```
8bitmp3, line 443: Suggestion: nit 

(when tryin to follow, present & future tense mixing may be a bit confusing)
("which.... which..." -> "which... (...)")

```suggestion
- Note the relationship between `OptimizerDef` and `Optimizer`: When the
  function `Optimizer.apply_gradient()` is called from the user code, it calls
  into `OptimizerDef.apply_gradient()` (among other things) which in turn
  calls `OptimizerDef.apply_param_gradient()` (implemented by subclasses
  of `OptimizerDef`).
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+An optimizer would first be constructed from its definition and the pytree of
+target params:
+
+```python
+optimizer_def = flax.optim.Momentum(learning_rate=0.1, beta=0.9)
+optimizer = optimizer_def.create(variables['params'])
+```
+
+The target variables would then optimized in the train step (assuming a single
+non-params collection "batch_stats"):
```
8bitmp3, line 465: Suggestion: nit (grammar)

```suggestion
Then, the target variables would be optimized during the train step (assuming a single
non-params collection "batch_stats"):
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+for step, batch in enumerate(ds)
+  optimizer, batch_stats, loss = train_step(
+      optimizer, batch_stats, batch['image'], batch['label'])
+```
+
+Remarks:
+
+- Note how `optimizer.apply_gradients()` can take additional arguments to update
+  hyper parameters, like the learning rate from an independent function
+  `get_learning_rate()` in this case.
```
8bitmp3, line 499: Suggestion: nit (grammar)

```suggestion
- Notice how `optimizer.apply_gradients()` can take additional arguments to update
  hyperparameters, such as the learning rate from an independent function
  `get_learning_rate()` in this case.
```

("Note that..." or "Notice how", I think)

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+
+# Update Plan
+[Update Plan]: #update-plan
+
+1. Finalize discussions on this FLIP
+2. Test existing optimizers for numerical equivalence (e.g. `flax.optim.Adam`
+   and `optax.adamw`).
+3. Update examples to use Optax and verify that they reach the same final
+   performance with the same computational cost. We probably want some examples
+   to directly use the optimizer while others might use [Linen helper].
```
8bitmp3, line 510: Suggestion: nit

```suggestion
3. Update examples of using Optax and verify that they reach the same final
   performance with the same computational cost. We probably want some examples
   to directly use the optimizer, while other users might use the [Linen helper].
```

8bitmp3 (commented):
A few tiny suggestions, if you can @andsteing @avital . Feel free to disregard them. Awesome step, Optax is 👍 @n2cholas

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+of parameters and optimizer state with a `jax.tree_multimap()`. For more
+complex optimizers this can become a hassle but we can provide a small
+functional helper that takes care of this:
+
+```python
+  def update_fn(
+    grads: Gradients, params: Params, state: OptimizerState
+  ) -> Tuple[Params, OptimizerState]:
+
+    @multimap
```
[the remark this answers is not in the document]
andsteing: No it is a small helper we would add (above "we can provide a small functional helper"). `jax.tree_multimap()` only supports a single return value. Mentioning name in paragraph above to make this clearer.

docs/flip/1009-optimizer-api.md
andsteing: Thanks for the suggestion - updated.

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
andsteing: Thanks for the pointer n2cholas@ !
Added an example of scheduling the learning rate below in [using Optax]

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+
+Optax is centered on the idea of composable gradient transformations. Since
+above [functional API] is based on Optax's interface it is straightforward to
+wrap an Optax transformation to comply with the `Optimizer` interface:
+
+```python
+import optax
+
+
+def optax_optimizer(tx: optax.GradientTransformation) -> Optimizer:
```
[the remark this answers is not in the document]
andsteing: It has the advantage that we adapt our code using this optimizer abstraction and then use any optimizer (including previous `flax.optim.Optimizer`) - but it adds another wrapper...

Adding this to Optax also SGTM, but it litterally saves a single line for the price of an additional indirection - also, Optax is built around the gradient transformation idea, so I can imagine they would prefer to keep this visible in the user code.

Leaving open for discussion...

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,529 @@
...
+```
+
+Remarks:
+
+- As opposed to the [previous API], we now can keep the entire `variables` 
+  including the `params` as an input and output to the `train_step()`.
+- Splitting `params` from `variables` is still necessary inside the train step
+  because we only want to compute gradients with respect to `params` and not the
+  entire `variables`.
+- A further complication is that we cannot provide `apply_fn` and
```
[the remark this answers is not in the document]
andsteing: That would also work. Updated paragraph below.

docs/flip/1009-optimizer-api.md
andsteing: Thanks, updated with your suggestion and modified format to use markdown listed items.

docs/flip/1009-optimizer-api.md
andsteing: Oh I missed that update - thanks!

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
andsteing: Yeah maybe that's a bit clearer. Updated.
(I don't see a semantic difference between the two, merely stylistic)

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
andsteing: I think that's a good idea. If somebody from Optax team can confirm here, then I'll update the description accordingly.

docs/flip/1009-optimizer-api.md
andsteing: I think `update examples to use` is clearer than `update examples of using`, so leaving unchanged.

docs/flip/1009-optimizer-api.md
andsteing: Added.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
avital: Can you explain what you mean by "It has the advantage that we adapt our code using this optimizer abstraction and then use any optimizer (including previous flax.optim.Optimizer)"? I couldn't follow. Are you saying that it makes the transition from our existing examples easier?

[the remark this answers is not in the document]
andsteing: It does not make the transition easier, but it would allow us to refactor our existing code once with the new optimizer abstraction and then we can simply exchange one optimizer implementation with another (e.g. `flax.optim.Adam` vs. `optax.adam`) by changing the initialization.

If we decide to update our existing code with the optax abstraction directly (i.e. passing around a `tx` transformation and `opt_state` and then calling `tx.update()` and `optax.apply_updates()`) then we'll have to update the code in multiple places if we want to switch from one implementation to another.

I can see advantages for both approaches and it's really about how much we value one less redirection vs. some more flaxibility.

docs/flip/1009-optimizer-api.md
8bitmp3: Gotcha!

docs/flip/1009-optimizer-api.md
8bitmp3: No worries, glad to be of service

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+5. We also need to keep track of the model's `apply_fn`, the otpimizer's
+   `update_fn`, and the optimizer state, in addition to the data.
+
+
+If we put all this state into a dataclass then we only have to pass around a
+single object (and the train data):
+
+```python
+@flax.struct.dataclass
+class LinenHelper:
```
jheek, line 330: `class LinenHelper(PyTreeNode)` is better for pytype

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
n2cholas: Now that the PR is merged, here's how a `multi_optimizer` _could_ look in pure Optax:

```python
def dict_map_with_path(fn, d, path=''):  # a crude "ModelParamTraversal"
    if isinstance(d, dict):
        return {k: dict_map_with_path(fn, v, path+'/'+k) for k, v in d.items()}
    return fn(path, d)

# bool masks with same structure as (or a prefix of) the params dict
is_bias_mask = dict_map_with_path(
    lambda path, value: path.endswith('/bias'), params)
not_bias_mask = jax.tree_map(lambda x: not x, is_bias_mask)

multi_optimizer = optax.chain(
    optax.masked(optax.sgd(learning_rate=0.1), mask=is_bias_mask),
    optax.masked(optax.sgd(learning_rate=0.05), mask=not_bias_mask))
```

Here's a [colab gist](https://colab.research.google.com/gist/n2cholas/2c36724f1d740ed9d388681c41f6e479/optax-multioptimizer.ipynb) with a full example. 

@mtthss is this the idiomatic approach, or is there a better way?

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
marcvanzee: @andsteing did you write "flaxibility" on purpose? 😄 

> It does not make the transition easier, but it would allow us to refactor our existing code once with the new optimizer abstraction and then we can simply exchange one optimizer implementation with another (e.g. flax.optim.Adam vs. optax.adam) by changing the initialization.

That is true, but does it mean we will always have the additional abstraction, even if we completely switch to Optax? I think that in the end, if we like Optax, there is no need to have our own optimizers, and then having an additional abstraction may only make things more complicated.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+[Previous Implementation]: #previous-implementation
+
+The optimizer itself would be implemented by creating a new class derived
+from `OpimizerDef`:
+
+```python
+# flax/optim/momentum.py
+
+@flax.struct.dataclass
+class _MomentumHyperParams:
```
marcvanzee, line 419: Why did you remove two of the hyperparameters here? (`nesterov` and `weight_decay`). The reason I am asking is because I was confused since `nesterov` and `decay` are used in the "Using Optax" section, so I expect it to be here as well. Perhaps it is clearer to make both uses cases consistent?

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+    new_params, new_momentum = inner(params, grads, state.momentum)
+    return new_params, MomentumState(new_momentum)
+```
+
+Remarks:
+
+- We'll call `Optimizer` instances from the functional API `optim`, and refer
+  to `flax.optim.Optimizer` instances from the [previous API] as `optimizer`
+  for the remainder of this document.
+- Compared with the [previous API], this new API does not have the option to
```
marcvanzee, line 131: Just for my understanding: I guess this means that with the new functional API, we are using optax exclusively? I was somehow thinking we want to support both optimizers initially, but I think I misunderstood that. Could you please clarify?

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+For example, to implement `flax.optim.Momentum` in this new interface:
+
+```python
+# flax/opt/momentum.py
+
+class MomentumState(OptimizerState):
+  momentum: PyTree
+
+
+def momentum(learning_rate: float, beta: float = 0.9) -> Optimizer:
```
jheek, line 86: I think it's good to discuss the pros/cons of this construction pattern a bit more. I think there are some downsides to the higher-order function pattern here:
- hyper params are not inspectable
- requires static_argnums= to pass into Jax functions
- you cannot implement auxiliary functions like `__repr__` or other utility like methods beside init/update

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+optim = optax_optimizer(tx)
+```
+
+Remarks:
+
+- The only difference is that the Optax udpate function returns a transformed
+  gradient and the updated optimizer state (which makes them composable), so we
+  apply this to the parameters in a final step and return the updated parameters
+  and optimizer state.
+- We can use a `get_learning_rate()` that returns the learning rate depending on
```
jheek, line 178: we often log the learning rate in metrics. How do I retrieve metrics like learning rate and grad norm (when clipping is used) from the update?

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+    return new_params, new_states
+
+  return Optimizer(init_fn, update_fn)
+```
+
+Remarks:
+
+- The `flax.optim.ModelParamTraversal` is reused.
+- The interface is a bit simplified, allowing to filter by path only (filtering
+  by value was never used), and shortening user code a bit by instantiating the
```
jheek, line 254: filtering by shape and/or dtype might still be useful

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
mtthss: I do like the approach supported by @n2cholas's recently contributed `masked` wrapper,
because it makes no assumption on how you may want to specify the `mask`
(e.g. it compatible with flax/stax/haiku/etc users to each generate the mask however is more idiomatic for them)
but would be good to hear from flax's multiopt users whether it works for them

@avital what do you think about this?

docs/flip/1009-optimizer-api.md
mtthss: Yes I am happy to help with this!

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
avital: Seems reasonable to me, and as far as I can tell, this is just as powerful as our https://flax.readthedocs.io/en/latest/flax.optim.html#flax.optim.ModelParamTraversal. (See the docstring [here](https://flax.readthedocs.io/en/latest/flax.optim.html#flax.optim.MultiOptimizer) for an example usage)

But one thing I don't like so much about this optax mask API is that it represents paths as "full strings" (IIUC, or maybe I am misunderstanding?). Can we instead make them lists of (nested) keys?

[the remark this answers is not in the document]
mtthss: I am probably missing something, but I am not sure what you mean by "it represents paths as full strings"?

Right now the mask is a pytree of booleans with the same structure as the params (or a prefix thereof)
So for instance a very simple mask that just looks as shapes would be:
```
mask = jax.tree_util.tree_map(lambda x: x.ndim != 1, params)
custom_weight_decay = optax.masked(optax.add_decayed_weights(0.001), mask)
```

Could you elaborate on what make them "lists of nested keys" would look like?

I am not particularly "married" to the current mask representation,
my main desiderata would be:
1) simple to do simple things
2) flexible enough to do complicated things
3) nn library agnostic

We can certainly choose a different one and switch to it if you think there is a more appropriate one.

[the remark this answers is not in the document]
avital: Oh I see -- this design allows `optax.masked` to bypass the question of "how to represent nested keys" by asking you to tree_map yourself, I now understand and I like it. So there remains the questions of how one would make a mask that depends on keys, e.g. "only weights, not biases".

@n2cholas shared a utility function above, `dict_map_with_path` that allows for that but that implementation encodes nested keys as "key1/key2/key3", that is what I was referring to. I prefer that we make this function exist somewhere general (maybe in optax for now, but should be in some "jax commons" place) but with a slightly different implementation:

```
def dict_map_with_path(fn, d, path=()):
    if isinstance(d, dict):
        return {k: dict_map_with_path(fn, v, path + (k, )) for k, v in d.items()}
    return fn(path, d)
```

this would allow us to suggest a pretty easy transition plan for users of `flax.optim.ModelParamTraversal` (the `filter_fn` argument to `ModelParamTraversal` could be fed directly into `dict_map_with_path`)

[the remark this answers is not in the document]
avital: (deleted my previous comment because I misremembered how `ModelParamTraversal` worked -- it does represent the argument passed to `filter_fn` as string e.g. '/Conv_0/weight', not tuples/lists). 

Updated, correct comment below:

Oh I see -- this design allows optax.masked to bypass the question of "how to represent nested keys" by asking you to tree_map yourself, I now understand and I like it. So there remains the questions of how one would make a mask that depends on keys, e.g. "only weights, not biases".

@n2cholas shared a utility function above, `dict_map_with_path` that allows for that but that implementation encodes nested keys as "key1/key2/key3", that is what I was referring to. To make the smoothest For people who currently use Flax's `ModelParamTraversal`, we'd want to use something very similar to that. The signature of the `filter_fn` argument to the `ModelParamTraversal` constructor is very similar to what you'd pass to `dict_map_with_path` (Flax also adds a leading /)

So your proposal sounds like a good plan.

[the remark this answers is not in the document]
n2cholas: @avital I just want to clarify `dict_map_with_path` was very ad-hoc and just for the sake of example. I like the idea of this function using a tuple for the path, as this matches the format of Flax's [`flatten_dict`](https://flax.readthedocs.io/en/latest/flax.traverse_util.html#flax.traverse_util.flatten_dict)

Haiku also has some nice utilities including a `haiku.data_structures.map` function that provides similar information (docs [here](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.data_structures.map)).

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,541 @@
...
+  `make_train_step()` function like above, or declared as `static_argnums` to
+  `jax.jit()`.
+
+## Linen Helper
+[Linen helper]: #linen-helper
+
+While the above training loop is perfectly fine, a number of lines are going to
+be exactly the same for every Linen training loop, namely:
+
+1. The splitting of variables.
```
avital, line 315: I think many of these things indeed are boilerplate but not all of them. For example the assumption that there's a single `apply_fn` isn't accurate in many cases and I think that would be better off written explicitly.

The "variable collection" munging (your items 1-4 -- taking params out of variables and back in) is boilerplate for 99% of cases though and I agree that it's annoying now and we should improve on it.

So I'm wondering what the "Linen Helper" that only does 1-4 look like?

Another option, perhaps, is simply to stop using variable dictionaries as idiomatic Flax training code and instead keep track of "params" and "other_variables", e.g.

```
params = ...
other_variables = {'batch_stats': ..., ...}

output, other_variables = module.apply({"params": params, **other_variables}, inputs, mutable=['batch_stats'])
```

avital: I wonder if structuring the code like that reduces the boilerplate for optimization steps as well?

n2cholas: I think it's also worth considering a very common use case is transfer learning, so users will want to freeze some subset of parameters and train others.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
andsteing: We wouldn't have to use it in that case. But it would allow us to easily replace an Optax optimizer with another optimizer function that transforms `(params, opt_state) -> (params, opt_state)`.

I can see Avital & you don't like it, so I've removed that abstraction completely and go "full Optax" in the FLIP.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
andsteing: Cool, updated the FLIP accordingly.

Note though that now we need to have `params` ready already when we define the optimizer - previously this was deferred to calling `optim.init()` - I don't think it makes a difference for our examples, though.

docs/flip/1009-optimizer-api.md
andsteing: Ah right!

docs/flip/1009-optimizer-api.md
andsteing: We switch to Optax exclusively - reformulated `Summary` to make this clearer

 The `Update Plan` is discussed in the end of the FLIP and ends with `Mark optimizers in flax.optim as deprecated` as a last step.

If you have specific ideas how to make the plan clearer feel free to put some comments in that section.

docs/flip/1009-optimizer-api.md
andsteing: Thanks for highlighting these points...

- inspect construction params : added comment to discussion under `Gradient Transformations`
- inspect hyper parameters that are scheduled : see comments about `learning_rate` in discussion `Optax Training Step`
- wrt `static_argnums` : solved with the proposed `Train State` helper
- in principle `OptState` could be extended with utility functions shared between all optimizers, and individual `OptState` can be freely extended ...

docs/flip/1009-optimizer-api.md
andsteing: Mentioning this in the discussion under `Optax Training Step`.

docs/flip/1009-optimizer-api.md
andsteing: Re-added.

docs/flip/1009-optimizer-api.md
andsteing: okok maybe that went too far :) 

@avital reduced the pattern to a `TrainState` with some predefined fields and helper `.param` property & `.update()` function, PTAL

@n2cholas yeah that's a good point where you would want to have more control over how variables are split; I think in that case you could either do it manually in the training step or override `.param` & `.update()` accordingly (referring to the updated section `Train State` in the FLIP)

docs/flip/1009-optimizer-api.md
andsteing: Note that the `optax.trace()` gradient transformation implements momentum; for simplicity I opted not to implement the `nesterov` in the `class Momentum(flax.optim.OptimizerDef)` example. Note that `optax` does not have a default argument for `trace(nesterov)` so I need to set it to `False` here.

Added a paragraph to the remarks to make this clearer.

docs/flip/1009-optimizer-api.md
n2cholas: Thanks @andsteing, that makes sense. Perhaps I'm missing something, so please correct me if I'm wrong, but what's the value in providing this `update()` method? Isn't the method equivalent to:

```python
state = state.replace(
    ...
    variables={**variables, 'params': new_params, **new_model_state}
)
```

n2cholas: If providing support for easy transfer learning is out of scope for this proposal, you can disregard this comment. But if it is in scope, another potential idea is to make state, trainable params, and frozen params more explicit. This sort-of mirrors how Keras provides access to `trainable_weights` and `non_trainable_weights` separately.

```python
class TrainState(flax.struct.PyTreeNode):
  apply_fn: Callable = flax.struct.field(pytree_node=False)
  trainable_params: flax.core.FrozenDict[str, Any]
  frozen_params: flax.core.FrozenDict[str, Any]
  model_state: flax.core.FrozenDict[str, Any]
  tx: optax.GradientTransformation = flax.struct.field(pytree_node=False)
  opt_state: optax.OptState
        
  @property
  def params(self):
    return merge(self.trainable_params, self.frozen_params)
    
  @property
  def variables(self):
    return {'params': self.params, **self.model_state}
```

Where `merge` behaves like [`haiku.data_structures.merge`](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.data_structures.merge) (recursively merges sub-dicts as well). A rough implementation for this in pure flax could look like:

```python
from flax.traverse_util import flatten_dict, unflatten_dict

def merge(*dicts):
    flat_merged = {}
    for d in dicts:
        flat_merged.update(flatten_dict(d))
    return unflatten_dict(flat_merged)
```
It's important that the merge is recursive, to give users flexibility to freeze arbitrary sets of parameters (e.g. only training biases of the layers, or perhaps unfreezing only batchnorm).

This would change the training step's loss function to:
```python
def loss_fn(trainable_params):
  variables = {'params': merge(trainable_params, state.frozen_params), 
               **state.model_state}
  outputs, new_model_state = state.apply_fn(
      variables, inputs, mutable=['batch_stats'])
  loss = xent_loss(outputs, labels)
  return loss, new_model_state
```

This makes transfer learning straightforward, while the normal use cases is not impacted significantly (the user can just use `trainable_params` as usual). But there are some caveats:
- `.params` is unexpectedly expensive for a property, since you have to do a merge upon every access. This doesn't matter inside a JIT'd function, does outside of JIT.
- I'm not sure how FrozenDicts work under the hood, but I would assume you need to do add some freezing and unfreezing to the code I provided to make it run, which would make it more expensive.

n2cholas: Another alternative is to keep the `TrainState` as is in the proposal, and let users override and to a `frozen_params`, and also provide a `merge` function through the library (which sounds like what you were already proposing @andsteing).

n2cholas: On second thought, @andsteing your suggestion makes more sense from a library perspective (letting the user override to add a `frozen_params` attribute and change the  `.params` themselves if needed). I still think it would be beneficial to provide a [`haiku.data_structures.merge`](https://dm-haiku.readthedocs.io/en/latest/api.html#module-haiku.data_structures) equivalent to facilitate this.

andsteing: @n2cholas I simplified `TrainState` with your suggestions

wrt optimizing only some parameters : this is currently supported via `MultiOptimizer` (i.e. `optax.masked()`) - if we want to make this usecase more efficient for cases where we only need gradients for a fraction of the `params`, then we could do this via multiple collections; note though that AFAIK this is not a common pattern in Flax and that there is some special logic for the `params` collection that would need to be replicated for other `params`-like collections. I think that's an interesting usecase but maybe for another FLIP. Curious to hear @avital's take on this.

cgarciae:
Hey! Is there a special reason for the optimizer to be part of `TrainState`?  Normally I just pass the optax optimizers via a closure capture or through e.g. `self.optimizer` when using classes.

andsteing:
@cgarciae - it's a  Flax pattern to include the optimizer in `TrainState` so we have fewer objects to pass around. But it's perfectly fine to pass in `tx` and `model.apply` in an outer function or with `static_argnums` - see the section `Optax training step` where I do exactly that.

docs/flip/1009-optimizer-api.md
```diff
@@ -190,11 +190,8 @@ Remarks:
 [Train State]: #train-state
 
 In Flax it is common to hand around a `TrainState` object that can then be
-used for checkpointing. We could extend this pattern slightly for Linen and
-shorten the above [Optax Training Step] significantly. The augmented
-`TrainState` would provide an easy access to `variables['params']` and the added
-function `.update()` would accept dataclass items as well as collections inside
-`.variables`:
+used for checkpointing. This simplifies above [Optax training step] a bit by
```
8bitmp3, line 193: Suggestion: nit

```suggestion
used for checkpointing. This simplifies the above [Optax training step] a bit by
```

docs/flip/1009-optimizer-api.md
n2cholas: @andsteing thanks for addressing my comments, that makes sense! I'm curious to hear @avital's thoughts as well.

docs/flip/1009-optimizer-api.md
n2cholas: The current proposal for scheduling hyperparameters in Optax stores the hyperparameters in the `OptState`, which makes logging them convenient. We'd appreciate feedback on this design on the [PR](https://github.com/deepmind/optax/pull/48)!

docs/flip/1009-optimizer-api.md
andsteing: Accepted.

docs/flip/1009-optimizer-api.md
```diff
@@ -83,9 +83,9 @@ Remarks:
   code illustrates how this can be a drop-in replacement for a function we also
   use in our [previous training step], where this update function already exists
   (notice how we need to invert the sign because we add the gradient update to
-  the parameters). See also [Optax
-  #20](https://github.com/deepmind/optax/issues/20) for an ongoing discussion
-  about scheduling hyper parameters.
+  the parameters). Since
+  [schedule_hyperparams](https://github.com/deepmind/optax/pull/48) Optax
+  supports scheduling arbitrary hyper parameters, not only the learning rate.
```
8bitmp3, line 88: Small suggestion: nit (refactor/grammar)

```suggestion
  the parameters). In addition,
  you can use [`schedule_hyperparams`](https://github.com/deepmind/optax/pull/48) in Optax 
  to schedule the learning rate and other arbitrary hyper parameters.
```

@andsteing LMKWYT

andsteing: Updated similar to your suggestion. Note that the learning rate is schedule differently (using `scale_by_schedule()`) from other hyper parameters.

8bitmp3: Cool @andsteing Thanks for letting us know

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,453 @@
...
+    if isinstance(d, dict):
+        return {k: dict_map_with_path(fn, v, path+'/'+k) for k, v in d.items()}
+    return fn(path, d)
+
+is_bias_mask = dict_map_with_path(
+    lambda path, value: path.endswith('/bias'), variables['params'])
+not_bias_mask = jax.tree_map(lambda x: not x, is_bias_mask)
+
+tx = optax.chain(
+    optax.masked(optax.sgd(learning_rate=0.1), mask=is_bias_mask),
```
jheek, line 173: I wonder if this should be a mask_fn instead or have that option. It seems to break the pattern where you can define the optimizer without knowing the exact variable structure

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,453 @@
...
+Note how we first define a traversal that selects parameters based on their
+path (which is the concatenation of module scopes and variable name), and then
+create a `MultiOptimizer` that binds a different optimizer for each of these
+separate traversals.
+
+Optax has recently implemented `optax.masked()` that can be used for specifying
+gradient transformations that only applied to a subset of the gradients:
+
+```python
+def dict_map_with_path(fn, d, path=''):  # a crude "ModelParamTraversal"
```
jheek, line 163: nit: this breaks our current path repr as a tuple of names

A simplified implementation using flatten_dict could be:
```
def dict_map_with_path(fn, data):
  flat = traverse_util.flatten_dict(data)
  return traverse_util.unflatten_dict({k: fn(k, v) for k, v in flat.items()})
```

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,453 @@
...
+    optax.masked(optax.sgd(learning_rate=0.05), mask=not_bias_mask),
+)
+```
+
+Remarks:
+
+- Above `dict_map_with_path()` emulates `flax.optim.ModelParamTraversal`.
+- Note that now we need `variables['params']` already when the gradient
+  transformation is defined, i.e. before calling `tx.init()`.
+- Note the change in behavior when a parameter matches multiple masks: With
```
jheek, line 183: We never properly defined this case. I would consider this a bug in the current implemenation.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,453 @@
...
+
+## Train State
+[Train State]: #train-state
+
+In Flax it is common to hand around a `TrainState` object that can then be
+used for checkpointing. This simplifies the above [Optax training step] a bit by
+reducing the number of arguments and getting rid of the `static_argnums`.
+
+```python
+class TrainState(flax.struct.PyTreeNode):
```
jheek, line 197: I think composition is better than inheritance here.
This is very similar to the old idea of having an optimizer dataclass:
```
class Optimizer(PyTreeNode):
  params: flax.core.FrozenDict[str, Any]
  model_state: flax.core.FrozenDict[str, Any]
  tx: optax.GradientTransformation = flax.struct.field(pytree_node=False)
  opt_state: optax.OptState

  def apply_gradient(self, grads):  # should apply_gradient take a model_state update?
    updates, new_opt_state = self.tx.update(
      grads, self.opt_state, self.params)
    new_params = optax.apply_updates(self.params, updates)
    return self.replace(params=new_params, opt_state=new_opt_state)

  @classmethod
  def create(cls, variables, tx):
     ...
```

inheritance implies that the TrainState only has a single model & optimizer. Also dataclasses don't behave very well under inheritance in general.

Another question is here whether you should store variables or params and model_state separately. I don't have a strong opinion here. We could even define the other one using `@property`. To me it seems splitting is a little bit easier because you only ever need to merge again inside the loss function and it's easy to accidentally use `variables` instead of `variables.copy(params=params_inside_loss_fn)`.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,453 @@
...
+
+And then [Optax Training Step] becomes:
+
+```python
+@jax.jit
+def train_step(state, inputs, labels):
+
+  def loss_fn(params):
+    outputs, new_model_state = state.apply_fn(
+      state.variables.copy({'params': params}), inputs, mutable=['batch_stats'])
```
jheek, line 220: nit: the downside of storing variables is that it's easy to accidentally use `state.variables` here which would lead to 0 gradients.

docs/flip/1009-optimizer-api.md
andsteing: Yeah, I put a similar comment above:
https://github.com/google/flax/pull/1011#discussion_r588426554

@n2cholas @mtthss what are your thoughts on this?

n2cholas: This is a good point, I think updating `optax.masked` to accept either a `mask` or `mask_fn` (but not both) is a good idea. `mask_fn` would take a pytree as input and return a `mask`.

It may even be sufficient to change `optax.masked` to only accept a `mask_fn` (if the user really wants to pass in a premade `mask`, they can simply do `mask_fn = lambda: mask`).

docs/flip/1009-optimizer-api.md
n2cholas: +1 for storing params and model_state separately.

docs/flip/1009-optimizer-api.md
n2cholas: Opened an [issue](https://github.com/deepmind/optax/issues/57) on the Optax repo for more visibility/discussion.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,453 @@
...
+  use in our [previous training step], where this update function already exists
+  (notice how we need to invert the sign because we add the gradient update to
+  the parameters). In addition, you can use
+  [`schedule_hyperparams()`](https://github.com/deepmind/optax/pull/48) to
+  schedule arbitrary hyper parameters with Optax.
+
+## Optax Training Step
+[Optax Training Step]: #optax-training-step
+
+```python
```
marcvanzee, line 93: Since we aren't using the `variables` in this example, I guess we can simplify it to this, but I guess you put the variables in here to demonstrate how it works in general?

```python
@functools.partial(jax.jit, static_argnums=(4, 5))
def train_step(opt_state, params, inputs, labels, apply_fn, tx_update_fn):

  def loss_fn(params):
    logits = apply_fn({'params': params}, inputs, mutable=['batch_stats'])
    loss = xent_loss(logits, labels)
    return loss

  loss, grads = jax.value_and_grad(loss_fn, has_aux=True)(params)
  updates = tx_update_fn(grads, opt_state, params)
  new_params = optax.apply_updates(params, updates)
  return new_opt_state, new_params, loss

opt_state = tx.init(params)
for batch in ds.as_numpy_iterator():
  opt_state, params, loss = train_step(
      opt_state, params, batch['image'], batch['label'], model.apply,
      tx.update)
  print(loss)
```

docs/flip/1009-optimizer-api.md
n2cholas: The PR has been merged! The `mask` argument for `optax.masked` can either be a pytree of booleans or a callable that returns such a pytree given the parameters/updates.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
andsteing: Note: with https://github.com/deepmind/optax/pull/60 above comment about `optim.init()` does not apply anymore.

docs/flip/1009-optimizer-api.md
andsteing: Updated to use Optax's new `mask=lambda ...` function & to adhere to Flax's "a path is a tuple" convention.

docs/flip/1009-optimizer-api.md
andsteing: Fixed in https://github.com/google/flax/pull/1135

docs/flip/1009-optimizer-api.md
andsteing: Yeah I think composition is much better here. Updated.

I decided to include `model_state` in the signature to make it harder to forget that. Since the signature is different I thought it's also better to use a new name (going for `update()` as in `GradientTransformation.update()`).

docs/flip/1009-optimizer-api.md
andsteing: Agreed - split "params" from variables.
(undoing https://github.com/google/flax/pull/1011#discussion_r588727155)

jheek (approved):
Looks great!

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+  [Optimizer and OptimizerDef] where we define a Momentum optimizer without
+  Nesterov momentum (note that the `beta` parameter corresponds to the `decay`
+  parameter of the `optax.trace()` transformation, and the learning rate is
+  applied in a second chained transformation).
+- Note that hyper parameters like `decay` or `nesterov` only exist in the inner
+  scope of the higher order functions returning the `GradientTransformation`.
+  Such a gradient transformation is currently defined as a `NamedTuple` of the
+  `init()` and the `update()` function. In principle this pattern could be
+  extended to also store parameters, maybe a point to discuss on the [Optax]
+  repo.
```
n2cholas, line 80: Did you mean "...extended to also store **hyper**parameters...", instead of "parameters"?

andsteing: Absolutely! Fixed.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+        tx=tx,
+        opt_state=opt_state,
+    )
+```
+
+Users can then compose their own data class containing this optimizer, the
+model's `.apply()` function and arbitrary additional data:
+
+```python
+class TrainState(flax.struct.PyTreeNode):
```
avital, line 221: What's the benefit of having separate `Optimizer` and `TrainState` classes? (instead of putting everything in `TrainState`?)

One downside of the current proposal, is something I've heard from confused users -- when you load from a checkpoint, why do you need to go through an "optimizer object" to get the parameters if all you're doing is making a prediction?

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+  step: int
+  apply_fn: Callable = flax.struct.field(pytree_node=False)
+  optimizer: Optimizer
+```
+
+With this the [Optax Training Step] becomes:
+
+```python
+@jax.jit
+def train_step(state, inputs, labels):
```
avital, line 231: rename `state` to `train_state` to disambiguate with `opt_state` and `model_state` above?

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+
+  def loss_fn(params):
+    outputs, new_model_state = state.apply_fn(
+        state.optimizer.model_state.copy({'params': params}),
```
avital, line 235: `{'params': params, **state.optimizer.model_state}` does the same thing, I think? And doesn't require using special Flax APIs

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+2. Test existing optimizers for numerical equivalence (e.g. `flax.optim.Adam`
+   and `optax.adamw`).
+3. Update examples to use Optax and verify that they reach the same final
+   performance with the same computational cost. We probably want some examples
+   to directly use the optimizer while others might use [Linen helper].
+4. Port missing optimizers to Optax (e.g. Adafactor) - and verify above points.
+5. Update all documentation (including README, Flax guided tour, HOWTOs, ...) to
+   talk exclusively about Optax optimizers.
+6. Mark optimizers in `flax.optim` as deprecated.
+
```
avital, line 392: I think we also need:
* A transition guide, similar to the [one for Linen](https://docs.google.com/document/d/1hYavTVPaKVVe9Be8pCB7yW7r6dDv3RALVNit8NZca4c/edit), with common patterns in the old and new form
* As part of this transition guide and (2), document how to translate hyperparameters from current flax.optim to Optax (including default values) -- @mtthss is working on this here: https://github.com/deepmind/optax/blob/e186087ddce5ef92d5df936ff341e819b29bc576/optax/_src/equivalence_test.py#L100

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+# Update Plan
+[Update Plan]: #update-plan
+
+1. Finalize discussions on this FLIP
+2. Test existing optimizers for numerical equivalence (e.g. `flax.optim.Adam`
+   and `optax.adamw`).
+3. Update examples to use Optax and verify that they reach the same final
+   performance with the same computational cost. We probably want some examples
+   to directly use the optimizer while others might use [Linen helper].
```
avital, line 387: I guess there's no longer a LinenHelper proposal? Should this be removed?

avital (commented):
This is looking really good! I added a few questions/small reservations.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+  the parameters). In addition, you can use
+  [`inject_hyperparams()`](https://github.com/deepmind/optax/pull/48) to
+  schedule arbitrary hyper parameters with Optax.
+
+## Optax Training Step
+[Optax Training Step]: #optax-training-step
+
+```python
+@functools.partial(jax.jit, static_argnums=(4, 5))
+def train_step(opt_state, variables, inputs, labels, apply_fn, tx_update_fn):
```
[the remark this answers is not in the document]
andsteing: This would remove two lines from `train_step`, but at the cost of having one more input (now 7 args) and one more return value.

Personally, I find it a bit more natural to think of `variables` in general, and then only split it "close to" the optimizer step where we actually need to treat the individual variable collections differently.

Also note that  this is only necessary with models that update the `model_state`; other models would simply return `params` alone. So with your proposal adding a mutable collection would not only change the signature of `module.apply()`, but also the signature of `train_step()`.

(But I don't have a clear preference here, so I'm happy to update as you proposed if you think it's worth it.)

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,471 @@
...
+        grads, self.opt_state, self.params)
+    new_params = optax.apply_updates(self.params, updates)
+    return self.replace(
+        params=new_params,
+        opt_state=new_opt_state,
+        model_state=new_model_state,
+    )
+
+  @classmethod
+  def create(cls, variables, tx):
```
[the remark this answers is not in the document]
andsteing: (I think here it's easier to use `variables` because that's what is returned by `module.init()` - so we keep the calling code simpler and push this detail into the helper.)

docs/flip/1009-optimizer-api.md
andsteing: In an earlier version I had an implementation that moved everything at the top level in the dataclass. But jheek and n2cholas convinced me in [above comment](https://github.com/google/flax/pull/1011#discussion_r593151131) to use composition instead.

docs/flip/1009-optimizer-api.md
andsteing: Renamed.

docs/flip/1009-optimizer-api.md
andsteing: Updated.

docs/flip/1009-optimizer-api.md
andsteing: Good idea - added.

docs/flip/1009-optimizer-api.md
andsteing: Good catch - removed.

docs/flip/1009-optimizer-api.md
n2cholas: Sorry if I was unclear before, I prefer having everything in `TrainState` as well. I was just in favour of splitting up `params` and `model_state` _within_ the `TrainState` class.

andsteing: Got it, sorry for mischaracterising your comment!

That still leaves jheek's suggestion to use composition. Both his arguments about dataclasses not behaving well under inheritance, and the possibility to have multiple optimizers/models, speak in favour of composition.

While I see Avital's point that users might not expect to load the model weights "from the optimizer", I think this is just a new pattern to get used to, and I would weight the arguments about inheritance & having multiple optimizers/models more heavily.

Pinging @jheek if he wants to add something .

avital: I am proposing neither inheritance nor composition -- a single `TrainState` dataclass.

jheek: I think there are 2 viable options:
- Let users roll their own "flat" TrainState where they manually track params, model_state, opt_state, and the optimizer (as a meta). 
- Create an Optimizer dataclass that users can use for their own TrainState (actually usually it replaces TrainState all together because these 4 fields are the only state).

I prefer the second option considering how much discussion we had already about the optimizer pattern. The bookkeeping itself is non-trivial and it might not be obvious to users that params & model_state need to be split or that the Optax function needs to be a meta field. The abstraction itself doesn't get in the way and users could also opt-out of it easily by "inlining" these fields into their own TrainState object. Eventually we might have more utilities that make it easier to use like an optimize method that takes a loss_fn and a gradient reducer or a property that gives the variables by combining params & modle_state.

avital: Yes indeed I misunderstood the intent that the proposal is that `TrainState` is user-defined, while `Optimizer` is part of Flax. I now better understand the rationale.

Perhaps I am stating a small preference for not calling it `Optimizer` then... We can't call it `OptimizerState` because that's an Optax concept. I still think this confuses many users as it's neither an optimizer nor the optimizer state. It's more like "model and optimizer metadata"... 

It's like the good old saying about only two hard things in software engineering...

I am thinking out loud here: Why wouldn't we call what's proposed as `Optimizer` here, instead, `TrainState`? Or `BaseTrainState`? Or `CommonTrainState`? It's not really the optimizer state, as it also holds parameters -- I don't think most folks would intuitively expect an "optimizer instance" to hold parameters.

Then it goes back to being confusing that users will then go ahead and define their own `TrainState` that holds a `CommonTrainState`... I'd like to better understand the concrete downsides of using inheritence, because this seems pretty similar to how we define `Module` and users extend it, which they can then call `MyModule` or whatever they want... To me this seems quite similar. I'll chat with @jheek offline for a few minutes to understand his objection.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
n2cholas: I don't have a strong preference, but I _slightly_ prefer @andsteing's suggestion here. The function args/return list is already quite long for `train_step`, and I think it makes more sense for the signature of `train_step` to stay the same regardless of whether the model has state or not.

I agree with @avital that using fewer Flax specific features would make the code more approachable. Perhaps we can use normal python dict merging here: `{**variables, 'params': params}` instead of `variables.copy(dict(params=params))`?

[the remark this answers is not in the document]
andsteing: Followed n2cholas's suggestion and replaced calls to `FrozenDict.copy()` with a simple `{**frozen_dict, 'name': value}`. Leaving `train_step()` arguments as-is.

docs/flip/1009-optimizer-api.md
andsteing: Reworked this part:

- As per Avital's suggestions, switched back to inheritance (which should work just fine with dataclasses as long as they don't have optional fields), and keeping everything at the root level. This will allow users to access parameters simply via `state.params`, not going through the `state.optimizer` attribute.
- Concerning the naming, going simply for `TrainState`. Since this will live in a different module, users can later subclass via `class TrainState(flax.training.TrainState)`
- Also added `apply_fn` and handling of `step` to the default implementation because `apply_fn` will be required in most cases and I think it's often useful to have the step number in the train state (anyways it shouldn't harm when it's not needed).
- The current solution works without modification for the simple state-less case.
- It's easy and safe to subclass `TrainState` if the user wants to store/update mutable model state or other data.

docs/flip/1009-optimizer-api.md
```diff
@@ -0,0 +1,475 @@
...
+        variables.copy(dict(params=params)), inputs, mutable=['batch_stats'])
+    loss = xent_loss(logits, labels)
+    return loss, new_model_state
+
+  variables, params = variables.pop('params')
+  (loss, new_model_state), grads = jax.value_and_grad(loss_fn, has_aux=True)(
+      params)
+  updates, new_opt_state = tx_update_fn(grads, opt_state, params)
+  new_params = optax.apply_updates(params, updates)
+  new_variables = variables.copy(dict(params=new_params, **new_model_state))
```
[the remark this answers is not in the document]
andsteing: Safety first!

Changed `model_state` to be explicit so it's clearly visible what state is being copied.

docs/flip/1009-optimizer-api.md
avital: Looks like with the latest `TrainState` proposal, you could achieve the transfer learning use-case by either using the new `masked` in Optax, or re-implementing the `update` method in your "user-space" TrainState subclass to mix parts of the old and new parameter dict (either using hk.data_structure or something similar, or directly). So for my original request and what I think @n2cholas was aluding to, I think everything has a simple solution now.

I prefer this current approach than explicitly creating a distinction between "optimized parameters" and "unoptimizer parameters". What if you want to sometimes optimize different subsets, like after training step 100? Make the user be explicit rather than adding a small number of insufficient hooks.

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
avital: LGTM!

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
avital: Nice!

docs/flip/1009-optimizer-api.md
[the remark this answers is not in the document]
avital: Yes I now agree

docs/flip/1009-optimizer-api.md
avital: I really like this latest version. It's short, understandable, really easy to extend (just add more dataclass attributes for additional state collections), you can probably easily override `update` if you want to do some modifications to parameters on updating if for some reason it's not supported in Optax, and you can also simply fork the whole TrainState class or not use it at all.

avital (approved):
Thanks @andsteing for the long haul review process! It looks great to me now.

docs/flip/1009-optimizer-api.md
n2cholas: I agree, the latest version looks great!

andsteing:
Thanks everybody for all the helpful comments!

And special thanks to Nicholas for pushing changes to Optax in parallel to make this FLIP a lot cleaner.

Going to merge and will start with the implementation.

PR #1020, opened 2021-02-12
Fix init/apply return annotation

jheek:


bastings (approved):
Thanks for the quick fix!

PR #1023, opened 2021-02-15
Fix incorrect attribute wrapping

jheek:
Module attributes that are callable but not methods were wrapped incorrectly.
The wrapper had self as a
required arg so things like inner
classes would fail to construct.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -19,6 +19,7 @@ import functools
 import inspect
 import os
 import threading
+import types
 import weakref
 
 from typing import (Any, Callable, Sequence, Iterable, List, Optional, Tuple,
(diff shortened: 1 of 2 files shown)


flax/linen/module.py
```diff
@@ -241,7 +241,11 @@ def wrap_method_once(fun: Callable[..., Any]) -> Callable[..., Any]:
     return fun
 
   @functools.wraps(fun)
-  def wrapped_module_method(self, *args, **kwargs):
+  def wrapped_module_method(*args, **kwargs):
+    if args and isinstance(args[0], Module):
```
avital, line 245: Can you add a comment explaining what this check does? I can't quite figure it out immediately. Is this to distinguish class methods from instance methods?

jheek: Done

avital: Wouldn't it be clearer to only wrap the instance methods in `_wrap_module_methods` instead of within `wrapped_module_method`? (since it's a bit misleading -- sometimes the function that's returned from `wrapped_module_method` won't be wrapped after this change)

avital (approved):
Thanks!

Issue #1034, opened 2021-02-19
Linen upgrade guide


andsteing:
It would be great to have a short document that explains in some detail how to
upgrade a model from the old API to the new Linen API.

avital:
Is that not covered in the "upgrade guide" linked to from https://github.com/google/flax/tree/master/flax/linen? Is there more we're missing? Or is it hard to find this link?

andsteing:
Yes htat is what I was looking for!

I would say it's not exactly easy to discover.

I think we want to avoid having a main RTD documentation and then a parallel Github-README documentation.

Maybe we should add a new main menu point "Linen" to RTD and add subpages there (brief intro, upgrade guide, design principles)?

Note also that we have both a Google doc [Linen Design Principles](https://docs.google.com/document/d/1ZlL_4bXCw5Xl0WstQw1GpnZqfb9JFOeUGAPcBVk-kn8) and an RTD page [Linen Design Principles](https://flax.readthedocs.io/en/latest/design_notes/linen_design_principles.html)

The bulk text about Linen we could move from the README into `docs/flax.linen.rst` so it would show up in RTD's [flax.linen package](https://flax.readthedocs.io/en/latest/flax.linen.html) description.

avital:
I agree that we should move all docs into RTD. There are a few other stranglers, like https://github.com/google/flax/blob/master/docs/design_notes/arguments.md

I think your proposal to add a "Linen" section in RTD is good, and unify the design principles under there.

marcvanzee:
This can be closed. Update guide is here: https://flax.readthedocs.io/en/latest/howtos/linen_upgrade_guide.html

PR #1036, opened 2021-02-19
Adds checkpoints.convert_pre_linen()

andsteing:
See #1034

### DIFF
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -235,6 +235,16 @@ def convert_pre_linen(params):
   With Linen the resulting params would instead have had the structure:
     {'Conv_0': { ... }, 'Dense_0': { ... } }
 
+  Note that you can also use this utility to convert pre-Linen collections
+  because they're following the same module naming. Note though that collections
+  were "flat" in pre-Linen and first need to be unflattened before they can be
+  used with this function:
+
+    batch_stats = convert_pre_linen(flax.traverse_util.unflatten_dict({
+        tuple(k.split('/')[1:]): v
+        for k, v in model_state.as_dict().items()
+    }))
+
   Args:
     params: Parameter pytree in pre-Linen format. If the pytree is already in
       Linen format, then the returned pytree is unchanged (i.e. this function
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -26,6 +26,7 @@ import re
 from absl import logging
 
 from flax import serialization
+from flax import core
 from tensorflow.io import gfile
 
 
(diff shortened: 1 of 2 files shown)


andsteing:
Please also review updated upgrade documentation at

https://docs.google.com/document/d/1hYavTVPaKVVe9Be8pCB7yW7r6dDv3RALVNit8NZca4c/edit?ts=6033d24d#heading=h.nvxz7b6p6tce

flax/training/checkpoints.py
```diff
@@ -208,3 +211,51 @@ def read_chunk(i):
       return serialization.msgpack_restore(checkpoint_contents)
     else:
       return serialization.from_bytes(target, checkpoint_contents)
+
+
+def convert_pre_linen(params):
```
jheek, line 216: Did you intent to only convert pre-linen param trees? Because I think this code will not handle the old Collections

tests/checkpoints_test.py
```diff
@@ -148,6 +205,69 @@ def test_save_restore_checkpoints_target_none(self):
...
+    state_dict = flax.serialization.to_state_dict(train_state)
+    # Linen.
+    model = Model()
+    variables = model.init(rng, inputs)
+    optimizer = flax.optim.GradientDescent(0.1).create(variables['params'])
+    optimizer = optimizer.restore_state(
+        flax.core.unfreeze(
+            checkpoints.convert_pre_linen(state_dict['optimizer'])))
+    optimizer = optimizer.apply_gradient(variables['params'])
+    batch_stats= flax.core.freeze(flax.traverse_util.unflatten_dict({
```
jheek, line 261: This `Collection` unflatten trick should probably be in the docstring of convert_pre_linnen or be natively supported by convert_pre_linnen.

jheek (changes requested):
Looks good, I have one small request about Collection handling

flax/training/checkpoints.py
andsteing: Added comment to the pydoc.

tests/checkpoints_test.py
andsteing: Added comment to the pydoc.

I prefer to keep the function simple and do a single thing. Copying the unflattening from the pydoc seems easier and safer to use than to make the function contract more complicated.

tests/checkpoints_test.py
```diff
@@ -258,10 +258,11 @@ def test_convert_checkpoint(self):
         flax.core.unfreeze(
             checkpoints.convert_pre_linen(state_dict['optimizer'])))
     optimizer = optimizer.apply_gradient(variables['params'])
-    batch_stats= flax.core.freeze(flax.traverse_util.unflatten_dict({
-        tuple(k.split('/')[1:]): v
-        for k, v in checkpoints.convert_pre_linen(model_state.as_dict()).items()
-    }))
+    batch_stats = checkpoints.convert_pre_linen(
```
andsteing, line 261: While writing the doc I noticed that the order of conversion was wrong!

Is it expected that Linen doesn't complain when the mutable collections (`batch_stats` in this case) don't match?

avital: ping @jheek

PR #1040, opened 2021-02-23
Updated Module docstring

joelgarde:
change the list default parameter of the example in the docstring to a tuple, since a dataclass default parameter cannot be of type list.


marcvanzee (approved):
Thanks for fixing this!

Issue #984, opened 2021-02-04
Make sure codediff code is tested


marcvanzee:
Currently the code in our [codediff directive](https://github.com/google/flax/blob/master/docs/_ext/codediff.py) is using Sphinx `codeblock`, however we would like it to use Sphinx `testcode` so we can make sure the code compiles.

Unfortunately the testing logic of Sphinx [doctest.py](https://github.com/sphinx-doc/sphinx/blob/284d9e703bf9919c34414353afd5992218518745/sphinx/ext/doctest.py#L436) retrieves the code to test from the original content in the file, and since our original code only contains the `codeblock` logic, this is not working and it throws a syntax error.

We can resolve this by creating a `TestNode` object in our directive, and by making sure it points to the lines in our codeblock directive that should be tested. As can be seen in the function [get_line_number](https://github.com/sphinx-doc/sphinx/blob/284d9e703bf9919c34414353afd5992218518745/sphinx/ext/doctest.py#L380), if we set the line number to the right value the doc test logic should be able to find the code to test correctly.

marcvanzee:
So to give an example, if we have the following code in our RTD:

```
.. codediff::
  :title_left: Single-model
  :title_right: Ensemble

  @jax.jit #!
  def get_initial_params(key):
    init_val = jnp.ones((1, 28, 28, 1), jnp.float32)
    initial_params = CNN().init(key, init_val)['params']
    return initial_params

  ---
  @jax.pmap #!
  def get_initial_params(key):
    init_val = jnp.ones((1, 28, 28, 1), jnp.float32)
    initial_params = CNN().init(key, init_val)['params']
    return initial_params
```

If you then simply transform this code so it creates two tables and the code is written using the `testcode` directive, then the test will run on all the content in the `codediff` block, and it breaks. 

A better solution is to create a TestCode node programmatically in `codediff.py`, and make sure the lines to test are the ones from the second code block.

PR #1043, opened 2021-02-24
Add functionality for capturing intermediates

jheek:
- `Module.sow` tracks values in a collection if it's mutable and acts as a no-op otherwise
- `Module.apply(capture_intermediates=...)` can be used to track return values of submodules (by default from the __call__ method) but a custom filter can be passed if necessary.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,6 +36,7 @@ jobs:
         sphinx-build -M doctest docs docs/_build
     - name: Build documentation
       run: |
+        sphinx-build -M doctest docs docs/_build
         sphinx-build -M html docs docs/_build
     - name: Test with pytest and generate coverage report
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,6 @@ jobs:
         sphinx-build -M doctest docs docs/_build
     - name: Build documentation
       run: |
-        sphinx-build -M doctest docs docs/_build
         sphinx-build -M html docs docs/_build
     - name: Test with pytest and generate coverage report
       run: |
--- a/docs/howtos/extracting_intermediates.rst
+++ b/docs/howtos/extracting_intermediates.rst
@@ -146,6 +146,51 @@ can define all submodules in ``setup`` and avoid using ``nn.compact`` altogether
   features(params, batch)
 
 
+Use `capture_intermediates`
+---------------------------
+
+Linen supports the capture of intermediate return values from submodules automatically without any code changes.
+This pattern should be considered the "sledge hammer" approach to capturing intermediates.
+As a debugging and inspection tool it is very useful but using the other patterns described in this howto.
+
+In the following code example we check if any intermediate activations are non-finite (NaN or infinite):
+
+.. testcode::
+
+  class CNN(nn.Module):
+    @nn.compact
+    def __call__(self, x):
+      x = nn.Conv(features=32, kernel_size=(3, 3))(x)
+      x = nn.relu(x)
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = nn.Conv(features=64, kernel_size=(3, 3))(x)
+      x = nn.relu(x)
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = x.reshape((x.shape[0], -1))  # flatten
+      x = nn.Dense(features=256)(x)
+      x = nn.relu(x)
+      x = nn.Dense(features=10)(x)
+      x = nn.log_softmax(x)
+      return x
+
+  @jax.jit
+  def init(key, x):
+    variables = CNN().init(key, x)
+    return variables
+
+  @jax.jit
+  def predict(variables, x):
+    y, state = CNN().apply(variables, x, capture_intermediates=True, mutable=["intermediates"])
+    intermediates = state['intermediates']
+    fin = jax.tree_map(lambda xs: jnp.all(jnp.isfinite(xs)), intermediates)
+    return y, fin
+
+  variables = init(jax.random.PRNGKey(0), batch)
+  y, is_finite = predict(variables, batch)
+  all_finite = all(jax.tree_leaves(is_finite))
+  assert all_finite, "non finite intermediate detected!"
+
+
 Use ``nn.Sequential``
 ---------------------
 
--- a/docs/howtos/extracting_intermediates.rst
+++ b/docs/howtos/extracting_intermediates.rst
@@ -219,7 +219,7 @@ In the following code example we check if any intermediate activations are non-f
   all_finite = all(jax.tree_leaves(is_finite))
   assert all_finite, "non finite intermediate detected!"
 
-By default only the intermediates of `__call__` methods are collected.
+By default only the intermediates of ``__call__`` methods are collected.
 Alternatively, you can pass a custom filter based on the ``Module`` instance and the method name.
 
 .. testcode::
@@ -231,13 +231,13 @@ Alternatively, you can pass a custom filter based on the ``Module`` instance and
   dense_intermediates = state['intermediates']
 
 
-Use ``nn.Sequential``
+Use ``Sequential``
 ---------------------
 
 You could also define ``CNN`` using a simple implementation of a ``Sequential`` combinator (this is quite common in more stateful approaches). This may be useful
 for very simple models and gives you arbitrary model
 surgery. But it can be very limiting -- if you even want to add one conditional, you are 
-forced to refactor away from ``nn.Sequential`` and structure
+forced to refactor away from ``Sequential`` and structure
 your model more explicitly.
 
 .. testcode::
--- a/docs/howtos/extracting_intermediates.rst
+++ b/docs/howtos/extracting_intermediates.rst
@@ -14,6 +14,27 @@ Let's start with this simple CNN that uses :code:`nn.compact`.
 
   batch = jnp.ones((4, 32, 32, 3))
 
+  class SowCNN(nn.Module):
+    @nn.compact
+    def __call__(self, x):
+      x = nn.Conv(features=32, kernel_size=(3, 3))(x)
+      self.sow('intermediates', 'conv1', x)
+      x = nn.relu(x)
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = nn.Conv(features=64, kernel_size=(3, 3))(x)
+      self.sow('intermediates', 'conv2', x)
+      x = nn.relu(x)
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = x.reshape((x.shape[0], -1))  # flatten
+      self.sow('intermediates', 'features', x)
+      x = nn.Dense(features=256)(x)
+      self.sow('intermediates', 'conv3', x)
+      x = nn.relu(x)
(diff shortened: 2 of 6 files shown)


marcvanzee:
Nice! @jheek what do you think of using the `codediff` directive for the HOWTO, since it seems to be based on many side-by-side diff views?

jheek:
> Nice! @jheek what do you think of using the `codediff` directive for the HOWTO, since it seems to be based on many side-by-side diff views? Here is how it is used in other HOWTOs: https://github.com/google/flax/blob/master/docs/howtos/lr_schedule.rst

I do lose the ability to test the code inline I guess?

marcvanzee:
> I do lose the ability to test the code inline when using `codediff` I guess?

The idea is that only the right-hand side of the code diff is using a `testcode` directive, so that is tested inline. However, there's a bug I haven't fixed yet so right now it isn't working, but I am planning to do this soon (see #984). If you use the directive here, then you can assume the RHS will be tested once that issue is fixed.

marcvanzee:
> > I do lose the ability to test the code inline when using `codediff` I guess?
> 
> The idea is that only the right-hand side of the code diff is using a `testcode` directive, so that is tested inline. However, there's a bug I haven't fixed yet so right now it isn't working, but I am planning to do this soon (see #984). If you use the directive here, then you can assume the RHS will be tested once that issue is fixed.

To clarify: I thought it would be better to only test the RHS, because it is based on a before/after diff view so you may get errors if you want to test both version. Technically I think it is feasible to also make the LHS tested. if you have thought / suggestions here, please let me know.

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,272 @@
...
+
+  import flax.linen as nn
+  import jax
+  import jax.numpy as jnp
+  from flax.core import FrozenDict
+  from typing import Sequence
+
+  batch = jnp.ones((4, 32, 32, 3))
+
+  class SowCNN(nn.Module):
```
avital, line 17: I think this shouldn't be in the testsetup block as it's defined later again?

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,272 @@
...
+  params = init(jax.random.PRNGKey(0), batch)
+
+  features(params, batch)
+
+
+Use `capture_intermediates`
+---------------------------
+
+Linen supports the capture of intermediate return values from submodules automatically without any code changes.
+This pattern should be considered the "sledge hammer" approach to capturing intermediates.
```
avital, line 198: I like the metaphor :)

docs/howtos/extracting_intermediates.rst
```diff
@@ -0,0 +1,272 @@
...
+surgery. But it can be very limiting -- if you even want to add one conditional, you are 
+forced to refactor away from ``nn.Sequential`` and structure
+your model more explicitly.
+
+.. testcode::
+
+  class Sequential(nn.Module):
+    submodules: Sequence[nn.Module]
+
+    def setup(self):
```
avital, line 237: I think this `setup` block is no longer necessary, after a recent change by @levskaya

jheek:
I think it's fine to test the RHS only. I just duplicated the RHS into testsetup for now to make sure the code is still tested. I added a codediff for the sow example where a line by line comparision makes most sense for the other code snippets I think it doesn't make it easier to understand. Let me know what you think

flax/linen/module.py
```diff
@@ -847,9 +885,79 @@ def variables(self) -> VariableDict:
...
+      col: the variable collection.
+      name: the name of the variable.
+      reduce_fn: The function used to combine the existing value with
+        the new value the default is to append the value to a tuple.
+      init_fn: For the first value stored reduce_fn will be passed
+        the result of `init_fn` together with the value to be stored.
+        The default is an empty tuple.
+
+    Returns:
+      `True` if the value has been stored succesfully, `False` otherwise.
```
avital, line 943: This is a nice return value. And thanks for the clear docstring!

flax/linen/module.py
```diff
@@ -847,9 +885,79 @@ def variables(self) -> VariableDict:
...
+          self.sow('intermediates', 'h', h)
+          return nn.Dense(2)(h)
+
+      y, state = Foo.apply(params, x, mutable=['intermediates'])
+      print(state['intermediates'])  # {'h': (...,)}
+
+    Args:
+      col: the variable collection.
+      name: the name of the variable.
+      reduce_fn: The function used to combine the existing value with
```
avital, line 936: This is nice and general

tests/linen/module_test.py
```diff
@@ -1024,6 +1024,40 @@ def __call__(self, x):
         return x+1
     self.assertTrue(isinstance(X.Hyper(a=1), X.Hyper))
 
+  def test_sow(self):
+    class Foo(nn.Module):
+      @nn.compact
+      def __call__(self, x):
+        self.sow('intermediates', 'h', x)
+        self.sow('intermediates', 'h', 2 * x)
```
avital, line 1032: Should you add examples with other reducers? Like adding numbers maybe?

flax/linen/module.py
```diff
@@ -344,10 +356,20 @@ def reimport(self, other):
...
+
+
+capture_call_intermediates = lambda _, method_name: method_name == '__call__'
+
 
 # Base Module definition.
 # -----------------------------------------------------------------------------
+
+
+_caches = weakref.WeakKeyDictionary()
```
avital, line 370: Can you add a comment explaining why we need weak references here?

flax/linen/module.py
```diff
@@ -107,6 +108,12 @@ def module_stack(self):
     if not hasattr(self._thread_data, 'module_stack'):
       self._thread_data.module_stack = [None,]
     return self._thread_data.module_stack
+  
+  @property
+  def capture_stack(self):
+    if not hasattr(self._thread_data, 'capture_stack'):
```
avital, line 114: Can you add a (maybe very short) docstring to this property method?

flax/linen/module.py
```diff
@@ -788,6 +811,12 @@ def apply(self, variables: VariableDict, *args, rngs: RNGSequences = None,
                treated as mutable: ``bool``: all/no collections are mutable.
                ``str``: The name of a single mutable collection. ``list``: A
                list of names of mutable collections.
+      capture_intermediates: If `True`, captures intermediate return values
+        of all Modules inside the "intermediates" collection. By default only
+        the return value of the `__call__` method is stored. A function can
```
avital, line 816:
```suggestion
        the return values of all `__call__` methods are stored. A function can
```

flax/linen/module.py
```diff
@@ -788,6 +811,12 @@ def apply(self, variables: VariableDict, *args, rngs: RNGSequences = None,
...
+        be passed to change the filter behavior. The filter function takes
+        the Module instance and method name and returns a bool indicating
+        whether the output of that method invocation should be stored.
```
avital, line 819: Can we should an example of how you'd write such a custom filter function? (Maybe show how the default one is implemented?)

flax/linen/module.py
```diff
@@ -798,7 +827,16 @@ def apply(self, variables: VariableDict, *args, rngs: RNGSequences = None,
     else:
       method = _get_unbound_fn(method)
     fn = lambda scope: method(self.clone(parent=scope), *args, **kwargs)
-    return apply(fn, mutable=mutable)(variables, rngs=rngs)
+    if capture_intermediates is True:
+      capture_intermediates = capture_call_intermediates
+    if capture_intermediates:
+      mutable = union_filters(mutable, 'intermediates')
+    _context.capture_stack.append(capture_intermediates)
```
avital, line 834: Why do we need a capture stack? Can there be multiple "capture intermediate" scopes nested into each other?

avital (changes requested):
Looks really good! Added some comments/questions.

avital:
> I think it's fine to test the RHS only. I just duplicated the RHS into testsetup for now to make sure the code is still tested. 

I believe RHS is already tested, so you don't need the same code in testsetup. But maybe I'm misunderstanding

marcvanzee:
> > I think it's fine to test the RHS only. I just duplicated the RHS into testsetup for now to make sure the code is still tested.
> 
> I believe RHS is already tested, so you don't need the same code in testsetup. But maybe I'm misunderstanding

Right now it isn't tested, but I plan to fix this soon (see #984). So putting it in testsetup is fine for now and then I'll remove it once it is fixed.

flax/linen/module.py
jheek: If you use apply inside apply yes

flax/linen/module.py
jheek: Oops I accidentally switched this code around. Doesn't actually belong to the PR

tests/linen/module_test.py
jheek: I don't see how that adds anything number addition is strictly simpler because the order of a + b no longer matters.

flax/linen/module.py
jheek: Adding to howto

.github/workflows/build.yml
```diff
@@ -36,6 +36,7 @@ jobs:
         sphinx-build -M doctest docs docs/_build
     - name: Build documentation
       run: |
+        sphinx-build -M doctest docs docs/_build
```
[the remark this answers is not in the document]
jheek: Oops

docs/howtos/extracting_intermediates.rst
jheek: The code in codediff blocks isn't currently part of the tests

docs/howtos/extracting_intermediates.rst
jheek: True

avital:
@googlebot I consent.

flax/linen/module.py
avital: Ah I see. (Though I'm curious how that actually works [at least within nn.compact] because once you instantiate a module it's a child of the current module...)

docs/howtos/extracting_intermediates.rst
avital: I see. Yeah @marcvanzee said he'll remove this once he brings in the change that tests the RHS of codediff blocks

tests/linen/module_test.py
avital: You're just not currnetly testing the `reduce_fn` and `init_fn` functions -- the unit tests isn't complete.

flax/linen/module.py
avital: It's probably best as part of the docstring, as people should be able to understand an API from its docstring. Or maybe add a link to the HOWTO from the docstring otherwise?

jheek: done

tests/linen/module_test.py
```diff
@@ -1024,6 +1024,48 @@ def __call__(self, x):
...
+        return 3 * x
+
+    _, state = Foo().apply({}, 1, mutable=['intermediates'])
+    self.assertEqual(state, {
+      'intermediates': {'h': (1, 2)}
+    })
+    _, state = Foo().apply(
+        {}, 1,
+        init_fn=lambda: 0,
+        reduce_fn=lambda a, b: a + b,
```
avital, line 1042: thanks for this (I'd have put it in a separate test entirely, but NBD)

avital (approved):

PR #1051, opened 2021-02-27
Updated the version number to match the release number

mplemay:
Bumped the projects version from `0.3.0` to `0.3.1`.


mplemay:
@googlebot I signed it!

avital:
I don't know how I missed this, thank you!

avital:
Looks like this change was made in the original #1046 but was reverted accidentally due to a mistake of mine when importing the change into Google via copybara. I'm release 0.3.2 that also includes the version update and another small change -- see #1055

PR #1055, opened 2021-03-01
Release 0.3.2

avital:
1. In the 0.3.1 release we (I) forgot to update the README for the new version
2. In the 0.3.1 release, `import flax` led to a `flax.nn` deprecation warning. In order to eliminate this, we now only raise a deprecation warning when instantiating a `flax.nn.Module` instance, not simply when importing `flax.nn`.
3. Un-revert accidental changes to CHANGELOG that erased the 0.3.1 changes (f63d210)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,35 +8,85 @@ vNext
  -
  -
  -
- - Bug Fix `flax.core.apply` and `Module.apply`. Now it returns a tuple containing the output and a frozen empty collection when `mutable` is specified as an empty list.
- -
- -
- - Add `sow` method to `Module` and `capture_intermediates` argument to `Module.apply`.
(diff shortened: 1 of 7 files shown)


flax/optim/base.py
```diff
@@ -534,7 +532,10 @@ def __init__(self, filter_fn):
 
   @staticmethod
   def _get_params_dict(inputs):
-    if isinstance(inputs, base.Model):
+    if inputs.__class__.__name__ == 'Model':
```
jheek, line 535: Wouldn't it be easier to move the deprecation warning to some piece of the old Module code that is always triggered when you actually use it? Like the Module constructor? Other codebases might also important the old nn api for simple instance checks and such or other backward compatibility tasks.

avital: Good point. That's much simpler indeed. /cc @levskaya.

PR #1070, opened 2021-03-02
Moves all errors in scope.py to errors.py

### LINKED ISSUE #893: Improve Error Message: Calling a method without initializing

marcvanzee:
After this PR, all error raised in `scope.py` have corresponding docstrings in `errors.py`


flax/core/scope.py
```diff
@@ -320,9 +322,9 @@ def reserve(self, name: str):
       name: the name to reserve.
     """
     if not isinstance(name, str):
-      raise NameTypeError(name)
+      raise errors.ScopeNameTypeError(name)
```
jheek, line 325: I think wrong types should not have a separate error. Otherwise you could have many of this type of errors. I would prefer the standard `TypeError`

flax/core/scope.py
```diff
@@ -596,13 +595,9 @@ def wrapper(variables: VariableDict,
...
     if not _is_valid_variables(variables):
-      raise ValueError('The first argument passed to an apply function '
-                       'should be a dictionary of collections. '
-                       'Each collection should be a dictionary '
-                       'with string keys.')
+      raise errors.ApplyScopeInvalidVariablesError()
     if rngs is not None and not _is_valid_rngs(rngs):
-      raise ValueError('rngs should be a dictionary mapping strings to '
-                       '`jax.PRNGKey`.')
+      raise errors.ApplyScopeInvalidRngsError()
```
jheek, line 600: Here the input type validation should also be TypeError or ValueError I think

flax/core/scope.py
```diff
@@ -596,13 +595,9 @@ def wrapper(variables: VariableDict,
               **kwargs) -> Union[Any, Tuple[Any, VariableDict]]:
 
     if not _is_valid_variables(variables):
-      raise ValueError('The first argument passed to an apply function '
-                       'should be a dictionary of collections. '
-                       'Each collection should be a dictionary '
-                       'with string keys.')
+      raise errors.ApplyScopeInvalidVariablesError()
```
jheek, line 598: dito

flax/core/scope.py
```diff
@@ -632,9 +627,7 @@ def init(fn: Callable[..., Any],
   @functools.wraps(fn)
   def wrapper(rngs, *args, **kwargs) -> Tuple[Any, VariableDict]:
     if not _is_valid_rng(rngs) and not _is_valid_rngs(rngs):
-      raise ValueError(
-          'First argument passed to an init function should be a `jax.PRNGKey` '
-          'or a dictionary mapping strings to `jax.PRNGKey`.')
+      raise errors.InitScopeInvalidRngsError()
```
jheek, line 630: Also type validation

jheek (changes requested):
Looks good, I just think we shouldn't name type errors because you might end up with an error class for almost every argument.

flax/core/scope.py
marcvanzee: I agree this one is probably not useful.

marcvanzee:
> Looks good, I just think we shouldn't name type errors because you might end up with an error class for almost every argument.

I agree that some of these errors aren't useful (e.g, the name error as you suggested), but in other cases it seems very useful to have more documentation than just a single line. For instance, you suggest to change `error.InitScopeInvalidRngsError()` into a type error, but then we lose the docstring I wrote for it, which provides quite a bit of useful information. What do you think of this?

jheek:
I see, I think the documentation for the RNGs is useful but not sure if that belongs in errors. It seems like it's a missing piece of documentation for the init & apply docstring?

marcvanzee:
I think there is quite a bit overlap with these error messages and other documentation, but that doesn't seem to be harmful to me.

I'm not entirely sure about the heuristic you are proposing for custom error messages. Shouldn't the heuristic be: "If this error message requires more explanation than a one-liner, it is useful to add it as a separate error class"? I think the main goal of these error classes is to provide more information to users, so that specific Scope errors can be explained with more concrete examples that users often run into.

jheek:
I agree, some duplication in docstrings cannot be avoided and I like this heuristic. Perhaps it's good to add a short guide on when to make a new error in the errors.py docstring but that doesn't need to happen in this PR

jheek (approved):

marcvanzee:
> I agree, some duplication in docstrings cannot be avoided and I like this heuristic. Perhaps it's good to add a short guide on when to make a new error in the errors.py docstring but that doesn't need to happen in this PR

I added some explanation to the top of `errors.py`.

Issue #1014, opened 2021-02-10
Raise clearer Exception when calling method of unbound module


ingmarschuster:
Using this minimal example
```
import jax.numpy as np
from jax.numpy import log, exp
import jax.random as rand
import flax.linen as ln

class MultipleForw(ln.Module):
    def setup(self):
        self.s1 = self.param("s1", ln.initializers.ones, (1,))

    def __call__(self, X, ):
        return X * log(1 + exp(self.s1 - 1))



mf = MultipleForw()
X = np.arange(5)
mf.init(rand.PRNGKey(0), X)
mf(X)
```

### Problem you have encountered:
The last line raised the rather opaque error message `AttributeError: 'MultipleForw' object has no attribute 's1'`

### What you expected to happen:
The raised Exception should contain a hint that makes clear that calling a linen-Module correctly is by using `mf.apply(parameters, input)`. See Discussion #1013

marcvanzee:
In #1072, I tried fixing this by creating a custom error class for Module AtttributeError, but after a discussion with @avital we found that this is not a very natural solution because users expect a normal `AttributeError` when they are trying to access an unknown attribute in a Module.

Solving this issue is a bit more work, and probably not our highest priority. For now I'm lowering the priority of this issue because it seems we won't fix it soon, and we can higher it when it turns out that more users run into this problem.

marcvanzee:
Also unassigning myself since I don't plan to work on this soon.

cgarciae:
I'd like to take this issue. A simple solution would be to customize the current error message to suggest calling `apply` if `self.scope is None`. I'll create a PR as this is a rather simple fix, if we want to tackle it a different way we can discuss there.

PR #1072, opened 2021-03-02
Moves errors in module.py to errors.py [1/2]

### LINKED ISSUE #847: Improve Error Message: Naming a module in setup
The error `TypeError: Module instance is frozen outside of setup method.` is thrown  at any time you assign module attributes somewhere other than within setup. It is not always clear to users how to resolve this.

-- Example 1

This code
```
def setup(self):
  self.layer = nn.Dense(...)
  self.layer.name = 'dense'
```

Throws the error `TypeError: Module instance is frozen outside of setup method.`, which confuses users.

-- Example 2
See: #936

### LINKED ISSUE #893: Improve Error Message: Calling a method without initializing

marcvanzee:
Closes #893
Closes #847

I'm separating this change in 2 PRs to keep the reviewing manageable (and hopefully high quality since these docstrings are super important for our users).

RTD preview: https://flax--1072.org.readthedocs.build/en/1072/flax.errors.html


flax/core/scope.py
```diff
@@ -329,9 +329,10 @@ def reserve(self, name: str):
       name: the name to reserve.
     """
     if not isinstance(name, str):
-      raise errors.ScopeNameTypeError(name)
+      raise TypeError('The type of scope "{name}" should be string but '
+                     f'it is {type(name)}')
     if name in self.reservations:
-      raise errors.ScopeNameInUseError(name)
+      raise ValueError(f'Duplicate use of scope name: "{name}"')
```
avital, line 335: Don't we lose a valuable docstring that existed on `ScopeNameInUseError`?

flax/errors.py
```diff
@@ -69,67 +69,82 @@ def __init__(self, message):
...
+    class Bar(nn.Module):
+      @nn.compact
+      def __call__(self, x):
+        some_param = self.param('some_param', nn.initializers.zeros, (1, ))
+        dropout_rng = self.make_rng('dropout')
+        x = nn.Dense(features=4)(x)
+        ...
 
-       SomeModule(...).init({'params': jax.random.PRNGKey(0)}, ...)
+    class Foo(nn.Module):
```
avital, line 102: You could just get rid of `class Foo`, and use `Bar` directly?

flax/errors.py
```diff
@@ -298,4 +298,297 @@ def __call__(self, x):
...
+class ModuleAttributeNotFoundError(FlaxError):
+  """
+  This error occurs when trying to access a module attribute that does not
+  exist. For instance, the following code will throw the error because attribute
+  ``unknown`` is not defined::
+
+    class Foo(nn.Module):
+      @nn.compact
+      def __call__(self, x):
+        y = self.unknown  # <--- ERROR
```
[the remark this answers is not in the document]
avital: (But maybe there's no way to catch this error otherwise?)

flax/errors.py
```diff
@@ -297,5 +323,263 @@ def __call__(self, x):
...
+        def __call__(self, x):
+          return self.dense(x)
+  """
+  def __init__(self, module_cls, attr_name, attr_val):
+    super().__init__(f'Can\'t set {attr_name}={attr_val} for Module of type '
+                    f'{module_cls}: Module instance is frozen outside of '
+                     'setup method.')
+
+
+class MultipleMethodsCompactError(FlaxError):
```
avital, line 485: Also here would be nice as a TODO to add link to a short design note explaining this as it's key to having no need for an equivalent to hk.transparent, and makes submodule naming much more sane (no need to prefix the method name)

flax/errors.py
```diff
@@ -297,5 +323,263 @@ def __call__(self, x):
...
+
+    class Foo(nn.Module):
+      features: int = 6
+
+      @nn.compact
+      def __call__(self, x):
+        return nn.Dense(self.features)(x)
+
+    variables = SomeModule(features=3).init(random.PRNGKey(0), jnp.ones((1, )))
+
```
avital, line 443: Maybe a TODO to link to a design note explaining why it's necessary for modules to stay frozen (otherwise we can't safely clone them, which we use for lifted transformations)

flax/errors.py
```diff
@@ -69,67 +69,82 @@ def __init__(self, message):
...
-  :meth:`Module.init() <flax.linen.Module.init>`, the first argument can be of
-  two forms:
+  def __init__(self, scope_name, rng_name):
+    super().__init__(f'{scope_name} needs PRNG for "{rng_name}"')
 
-  1. A single PRNGKey. This is in case only one PRNGKey is needed to initialize
-     the ``params`` collection. Note that this::
 
-       SomeModule(...).init(jax.random.PRNGKey(0), ...)
+class InvalidRngError(FlaxError):
```
avital, line 87: I don't think this error is ever actually raised? Can/should we combine all of the different RNG errors into this one?

flax/errors.py
```diff
@@ -69,67 +69,82 @@ def __init__(self, message):
...
+  If a Module only requires an rng for ``params``, you don't have to provide
+  rngs for apply at all::
+
+       SomeModule().apply(variables, inputs)  # rngs=None
+  """
+  def __init__(self, scope_name, col):
+    super().__init__(f'Scope {scope_name} needs PRNG for collection "{col}"')
+
+
+class BindInvalidRngsError(FlaxError):
```
avital, line 144: (Like this doesn't give the user much more than `InvalidRngError` -- what is a scope?)

flax/errors.py
```diff
@@ -297,5 +323,263 @@ def __call__(self, x):
...
+  
+  * ``parent``: The parent Module of this Module.
+  * ``name``: The name of this Module.
+  """
+  def __init__(self, annotations):
+    super().__init__(f'properties `parent` and `name` are reserved: '
+                     f'{annotations}')
+
+
+class InitModuleInvalidRngsError(FlaxError):
```
avital, line 510: Same here, the user knows they are in init by looking at the stack trace. So I think instead of this class you can just raise `InvalidRngError`?

avital: (Is there also `ApplyRngError`?)

flax/errors.py
```diff
@@ -297,5 +323,263 @@ def __call__(self, x):
...
+
+
+class ApplyModuleInvalidMethodError(FlaxError):
+  """
+  When calling :meth:`Module.apply() <flax.linen.Module.apply>`, you can specify
+  the method to apply using parameter ``method``. This error is thrown if the
+  provided parameter is not a method in the Module and not a function with at
+  least one argument.
+
+  A typical call looks as follows::
```
avital, line 528: I think from here on should be part of the `apply` docstring instead, and then replace this part with, e.g. "learn more on the on the reference docs for apply".

tests/linen/module_test.py
```diff
@@ -615,9 +648,37 @@ def __call__(self, x):
...
+      
+      def test(self):
+        pass
+
+    # We can use both instance and class methods in apply.
+    Foo().apply({}, method=Foo.test)
+    Foo().apply({}, method=Foo().test)
+
+    # We can even use a function that is not in the provided Module, although
+    # it requires an argument for the class (Foo in this case).
```
avital, line 665: I don't understand this -- the lambda would get passed `cls` as the first argument? Is this a real use-case? What would someone do with `cls`? (I guess I don't understand how you'd use this pattern)

avital (changes requested):
Real nice, I added some discussions

flax/errors.py
```diff
@@ -298,4 +298,297 @@ def __call__(self, x):
...
+
+
+#################################################
+# module.py errors                              #
+#################################################
+
+
+class  ModuleNameExistsError(FlaxError):
+  """
+  This error is raised when trying to create a submodule with an exsiting name.
```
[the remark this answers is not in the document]
marcvanzee: Yes great point! I've added it.

flax/errors.py
```diff
@@ -298,4 +298,297 @@ def __call__(self, x):
...
+    def setup(self):
+      self.conv = functools.partial(nn.Conv, features=3)
+
+    def __call__(self, x):
+      x = self.conv(kernel_size=4)(x)
+      return x
+
+    Foo().init(random.PRNGKey(0), jnp.zeros((1,)))
+
+  In this case, ``self.conv(...)`` inside ``__call__`` constructs a
```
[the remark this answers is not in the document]
marcvanzee: Added it.

flax/errors.py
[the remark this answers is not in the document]
marcvanzee: I'm not sure what you mean, this is normal behavior right? We are simply accessing an object attribute that doesn't exist. Consider the following Python code:

```
class Test:
  def hoi(self):
    print(self.unknown)

Test().hoi()
```

That also throws an error: `AttributeError: 'Test' object has no attribute 'unknown'`

Or do you mean it is surprising that we raise a custom error?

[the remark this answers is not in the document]
avital: Yes that it is surprising that we raise a custom error here.

flax/linen/module.py
```diff
@@ -294,8 +292,7 @@ def wrapped_module_method(*args, **kwargs):
 def _wrap_hash(hash_fn: Callable[..., Any]) -> Callable[..., Any]:
   @functools.wraps(hash_fn)
   def wrapped(self):
-    if self.scope is not None:
-      raise ValueError('Can\'t call __hash__ on modules that hold variables.')
+    assert self.scope == None, 'Can\'t call __hash__ on modules that hold variables.'
```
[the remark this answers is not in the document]
marcvanzee: Yes I agree, I am actually not sure why I changed it to an assertion but I've changed it back now.

flax/core/scope.py
marcvanzee: Actually the docstring on `ScopeNameInUseError` describes the problem using Modules: https://flax.readthedocs.io/en/latest/flax.errors.html#flax.errors.ScopeNameInUseError

Therefore, I've now moved that description to `errors.NameInUseError` and made sure it is thrown in `module.py`. The error here is only thrown if a user is using Scopes directly, which isn't something the average user will do (I think).

flax/errors.py
marcvanzee: I've done that to clarify that even if a Module has submodules, there is no nesting in the provided RNGS (the provided RNGs is always a flat dict). But if you think this is a bit artificial I can also simplify it.

flax/errors.py
marcvanzee: Good point. I'm changing it so all RNG errors point to here.

tests/linen/module_test.py
marcvanzee: This example is a bit contrived, but I wanted to test that the first argument to the custom apply function is actually the `Foo` class. I did that by directly returning it with a lambda and then checking the types of the returned value is indeed `Foo`.

I am not sure what you question is, but I think it is quite useful that someone can define a custom apply function without having to change the original Module:.

```python
class Foo(nn.Module):
  features: int = 4

  def setup(self):
    self.dense = nn.Dense(self.features)

  @nn.compact
  def __call__(self, x):
    x = self.dense(x)
    # do some stuff
    return x

def custom_apply(cls, x):
  x = cls.dense(x)
  # do some custom stuff
  return x

Foo().apply(..., ) # this calls Foo's apply
Foo().apply(..., method=custom_apply)  # this uses the custom apply function
```

flax/errors.py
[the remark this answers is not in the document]
marcvanzee: I agree, but the main reason I added is so that we can explain this problem better: #1014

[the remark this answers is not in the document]
avital: I understand, and perhaps there's no good solution here. The problem is that it catches the error too broadly. Most likely someone just had a typo and it's unrelated to whether they used `apply` or not...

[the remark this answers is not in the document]
marcvanzee: Yes I agree, the current solution is a bit unnatural. It would be better if we can catch the error in #1014 in another way and not abuse AttributeError for it. I've reverted it.

flax/errors.py
avital: Sure, good point.

tests/linen/module_test.py
avital: How does the line `x = cls.dense(x)` work?

Wouldn't it need to be an instance of Foo (with `self.features`?).

flax/core/scope.py
avital: Ah I understand now. Makes sense.

tests/linen/module_test.py
marcvanzee: No, I don't think that is how it works (but please correct me if I am wrong).  I think you don't need any instances in the apply method, you only need the classes, and apply will then instantiate everything automatically.

In `apply`, we first "unbound" the `__call__` function, which means it is no longer attached to `self`: https://github.com/google/flax/blob/fa826152a5c092c8f4e8a50d01f0d560ea7a6af3/flax/linen/module.py#L303

So in that sense `__call__` is exactly the same as any function that has a first argument representing the class. After creating an unbound instance of the function, the function is called with an instantiated clone of the module (`Foo` in this case), where the parent scope is initialized correctly.

flax/linen/module.py
```diff
@@ -533,8 +533,8 @@ def __getattr__(self, name: str) -> Any:
     if name in self.__dict__:
       return self.__dict__[name]
     else:
-      raise AttributeError(
-          f"'{self.__class__.__name__}' object has no attribute '{name}'")
+       raise AttributeError(
```
avital, line 536: incorrect whitespace?

flax/linen/module.py
```diff
@@ -871,21 +864,36 @@ def apply(self, variables: VariableDict, *args, rngs: RNGSequences = None,
...
+      encoded = model.apply({'params': params}, x, method=Transformer.encode)
+  
+    If a function instance is provided, the unbound function is used. For 
+    instance, the example below is equivalent to the one above::
+
+      encoded = model.apply({'params': params}, x, method=model.encode)
+
+    Note ``method`` can also be a function that is not defined in
+    ``Transformer``. In that case, the function should have at least one
+    argument representing the Module class::
```
avital, line 881: Similarly this isn't right. the argument is the module instance, not class

flax/linen/module.py
```diff
@@ -871,21 +864,36 @@ def apply(self, variables: VariableDict, *args, rngs: RNGSequences = None,
...
+
+      def other_fn(cls, ...):
```
avital, line 883: and the argument shouldn't be named cls

PR #1075, opened 2021-03-03
Fixes a bug in Module and adds test

### LINKED ISSUE #1074: Bug in error catching
The following code throws an error at the indicated line:

```
class SomeModule(nn.Module):
  @nn.compact
  def __call__(self, x):
    dense = nn.Dense(10)
    dense.features = 20
    dense.new_attr = 20  # <--- ERROR!
    return dense(x)

SomeModule().init(random.PRNGKey(0), jnp.ones((5, 5)))
```

The error is: `Module instance is frozen outside of setup method.` This seems odd: if `dense` were frozen, why are we allowed to modify the existing attribute `features`, but we aren't allowed to add a new one called `new_attr`? It seems we should not be allowed to modify `dense.features` are all and we should already throw an error the line before.

marcvanzee:
Closes #1074

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -329,11 +329,16 @@ class _ModuleInternalState:
   in_compact_method: bool = False
   in_setup: bool = False
   setup_called: bool = False
+  is_initialized: bool = False
   autoname_cursor: Optional[dict] = dataclasses.field(default_factory=dict)
   children: Dict[str, Union[str, 'Module']] = dataclasses.field(default_factory=dict)
 
   def reset(self):
-    """Resets transient state."""
+    """Resets transient state. 
+    
+    This function is called after each module method, so only attributes that
+    are method-dependent are reset.
+    """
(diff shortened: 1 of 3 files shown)


flax/linen/module.py
```diff
@@ -504,8 +504,9 @@ def __setattr__(self, name: str, val: Any):
       val: Value of the attribute.
     """
     is_dataclass_attr = name in self.__dataclass_fields__ and self.__dataclass_fields__[name].init  # pytype: disable=attribute-error
-    
-    if not self._state.in_setup and not is_dataclass_attr:
+    is_initialized = self.scope != None
```
jheek, line 507: this will probably fail in setup and top level because a constructed module will not immediately get a scope.

marcvanzee: I've now added `is_initialized` to `Module` and added some more tests.

jheek (approved):
Looks good, Thanks!

PR #1076, opened 2021-03-03
Squeeze and Excite Block, take 2

joaogui1:
Hey @marcvanzee is this version correct?


marcvanzee:
Thanks for the second take!

I think the code is correct, but we will truly know it when we use the class 😄 .  Did you use this class in your own code? In that case it would be nice if we could somehow show how this class can be used. It doesn't have to be very complex, it can even be very simple, but it would both be useful for users and good since we then test whether the class actually works.

What do you think?

PR #1078, opened 2021-03-03
implement functional init/apply and bind.

jheek:
- adds init/apply as functional APIs so you can use `apply_fn = nn.apply(fn, module)` making it easier to combine apply with jax transforms and making it more obvious that you can make arbitrary functions that operate on a bound module
- adds `Module.bind` which gives you an interactive instance. It fully supports RNGs and mutability but the docstrings explicitly warns users not to use stateful training loops outside of notebook/debug code.

### DIFF
--- a/examples/vae/train.py
+++ b/examples/vae/train.py
@@ -143,16 +143,17 @@ def train_step(optimizer, batch, z_rng):
 
 @jax.jit
 def eval(params, images, z, z_rng):
-  recon_images, mean, logvar = model().apply({'params': params}, images, z_rng)
+  def eval_model(vae):
+    recon_images, mean, logvar = vae(images, z_rng)
+    comparison = jnp.concatenate([images[:8].reshape(-1, 28, 28, 1),
+                                  recon_images[:8].reshape(-1, 28, 28, 1)])
 
-  comparison = jnp.concatenate([images[:8].reshape(-1, 28, 28, 1),
-                                recon_images[:8].reshape(-1, 28, 28, 1)])
+    generate_images = vae.generate(z)
+    generate_images = generate_images.reshape(-1, 28, 28, 1)
+    metrics = compute_metrics(recon_images, images, mean, logvar)
+    return metrics, comparison, generate_images
 
-  generate_images = model().apply({'params': params}, z, method=VAE.generate)
-  generate_images = generate_images.reshape(-1, 28, 28, 1)
-  metrics = compute_metrics(recon_images, images, mean, logvar)
-
-  return metrics, comparison, generate_images
+  return nn.apply(eval_model, model())({'params': params})
 
 
 def prepare_image(x):
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -586,7 +586,7 @@ def _unfreeze_variables(variables, mutable):
 def bind(variables: VariableDict,
          rngs: Optional[RNGSequences] = None,
          mutable: CollectionFilter = False):
-  """Bind variables and rngs to a new ```Scope```.
+  """Bind variables and rngs to a new ``Scope``.
   
   bind provides a ``Scope`` instance without transforming a function
   with ``apply``. This is particulary useful for debugging and
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -290,7 +290,14 @@ class Scope:
     """Invalidates the Scope."""
     self._invalid = True
 
-  def variables(self) -> Collection:
+  def mutable_variables(self) -> VariableDict:
+    """Returns an immutable copy of the mutable variables belonging to this Scope."""
+    self._populate_collections()
+    xs = {k: v for k, v in self._variables.items()
+          if in_filter(self.mutable, k)}
+    return freeze(xs)
+
+  def variables(self) -> VariableDict:
     """Returns an immutable copy of the variables belonging to this Scope."""
     self._populate_collections()
     return freeze(self._variables)
@@ -576,6 +583,29 @@ def _unfreeze_variables(variables, mutable):
   return new_variables
 
 
+def bind(variables: VariableDict,
+         rngs: Optional[RNGSequences] = None,
+         mutable: CollectionFilter = False):
+  """Bind variables and rngs to a new ```Scope```.
+  
+  bind provides a ``Scope`` instance without transforming a function
+  with ``apply``. This is particulary useful for debugging and
+  interactive use cases like notebooks where a function would limit
+  the ability split up code into different cells.
+
+  a ``Scope`` instance is a stateful object. Note that idiomatic JAX is functional
+  and therefore a ``Scope` does not mix well well with vanilla JAX APIs. Therefore,
+  we recommend using ``apply`` when code should be reusable and compatible
+  across the JAX software ecosystem.
+  """
+  if not _is_valid_variables(variables):
+    raise errors.ApplyScopeInvalidVariablesError()
+  if rngs is not None and not _is_valid_rngs(rngs):
+    raise errors.ApplyScopeInvalidRngsError()
+  new_variables = _unfreeze_variables(variables, mutable)
+  return Scope(new_variables, rngs=rngs, mutable=mutable)
+
+
 def apply(fn: Callable[..., Any],
           mutable: CollectionFilter = False) -> Callable[..., Any]:
   """Functionalize a `Scope` function.
@@ -593,19 +623,10 @@ def apply(fn: Callable[..., Any],
               *args,
               rngs: Optional[RNGSequences] = None,
               **kwargs) -> Union[Any, Tuple[Any, VariableDict]]:
-
-    if not _is_valid_variables(variables):
-      raise errors.ApplyScopeInvalidVariablesError()
-    if rngs is not None and not _is_valid_rngs(rngs):
-      raise errors.ApplyScopeInvalidRngsError()
-    new_variables = _unfreeze_variables(variables, mutable)
-    with Scope(new_variables, rngs=rngs, mutable=mutable).temporary() as root:
+    with bind(variables, rngs=rngs, mutable=mutable).temporary() as root:
       y = fn(root, *args, **kwargs)
     if mutable is not False:
-      mutated_variables = {k: v
-                           for k, v in new_variables.items()
-                           if in_filter(mutable, k)}
-      return y, freeze(mutated_variables)
+      return y, root.mutable_variables()
     else:
       return y
 
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -32,7 +32,8 @@ import numpy as np
 import flax
 from flax import traverse_util
 from flax import serialization
-from flax.core import Scope, apply
+from flax import core
+from flax.core import Scope
 from flax.core.scope import CollectionFilter, Variable, VariableDict, FrozenVariableDict, union_filters
 from flax.core.frozen_dict import FrozenDict, freeze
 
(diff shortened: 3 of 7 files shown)


flax/linen/module.py
```diff
@@ -1011,61 +1050,145 @@ def __call__(self, train: Optional[bool] = None):
...
+  JAX transformations like ``jax.jit``::
+
+    def f(foo, x):
+      z = foo.encode(x)
+      y = foo.decode(z)
+      # ...
+      return y
+    
+    foo = Foo()
+    f_jitted = nn.jit(nn.apply(f, foo))
```
avital, line 1073: Is this really `nn.jit` or `jax.jit`? And how does one use `f_jitted` afterwards? Can you pass `foo` as an argument? (I didn't think we registered Modules against pytree so how does this work?)

flax/linen/module.py
```diff
@@ -790,9 +791,52 @@ def make_rng(self, name: str) -> PRNGKey:
...
+           mutable: CollectionFilter = False):
+    """Creates an interactive Module instance by binding variables and RNGs.
+
+    bind provides an "interactive" instance of a Module directly without
+    transforming a function with ``apply``. This is particulary useful for debugging
+    and interactive use cases like notebooks where a function would limit the ability
+    split up code into different cells.
+
+    Once the variables (and optionally RNGs) are bound to a ``Module`` it becomes a
+    stateful object. Note that idiomatic JAX is functional and therefore an interactive
```
avital, line 804: I think this statement isn't true if `mutable=False` -- you're not able to shoot yourself in the foot (correct me if I am wrong)

If you agree, would be good to clarify in the comment.

avital:
I vote for renaming the `mutable` argument to bind to `unsafe_mutable`, and adding an actual Python warning when you use it. Then this warning guides you to a docs/HOWTO page showing how to take code that uses an "unsafe mutable" module instance and put it in a function that you can safely pass to `nn.apply`.

flax/linen/module.py
jheek: In practise, I think it doesn't matter that much. Normally you can train a model or compile a train step with `mutable=False` but you cannot combine this api with optimizers or `jax.jit`. The pitfall I describe here is about composability with Jax and other apis that expect functional objects which still holds for `mutable=False`.

avital: Isn't this fine?
```
@jax.jit
def foo(params, inputs):
  return CNN().bind({"params": variables})(inputs)
  
jax.grad(foo)
```
And then `foo` works fine with `jax.grad` etc?

jheek: Yes but here you are basically inlining apply. The bound Module lives inside a function. The problem is that you can keep a bound module around and just pass the bound module around until you start using jax transforms.

jheek: So of course we could tell users to bind only inside functions but it seems better to me to stick to a single pattern `apply` to avoid having multiple ways to do the same thing.

avital: Ah. I guess there are two different things:

1. Binding a module within a function. This works fine also with mutable, e.g.:

```
@jax.jit
def prediction(params, batch_stats, inputs):
  cnn = CNNWithBatchNorm().bind({"params": variables, "batch_stats", batch_stats}, mutable=['batch_stats'])
  outputs = cnn(inputs)
  return outputs, cnn.variables['batch_stats']
```
  
2. Colab style tinkering, e.g.:
 
```
cnn = CNNWithBatchNorm().bind(...)
  
@jax.jit
def predict(inputs):
  return cnn(inputs)
```

IIUC this would technically work even if you use `mutate=False`, but it would compile the variables in `cnn` as constants when jitting `predict` which would make the compiled HLO much larger, and would be unsafe if anyone modifies `cnn` or `cnn.variables` later on. So maybe this is what your comment was referring to.

avital: I guess the real thing is that to use JAX safely/effectively you never want to:
1. Leak a bound module "out" of a function, e.g. in example 1 above, if you would return `cnn` from the `prediction` function; or
2. Leak a bound module "in" to a function (e.g. closing over a bound instance e.g. example 2 above)

examples/vae/train.py
```diff
@@ -143,16 +143,17 @@ def loss_fn(params):
 
 @jax.jit
 def eval(params, images, z, z_rng):
-  recon_images, mean, logvar = model().apply({'params': params}, images, z_rng)
+  def eval_model(vae):
```
avital, line 146: Curious why you ported this example to use `nn.apply`?

flax/core/scope.py
```diff
@@ -577,6 +584,33 @@ def _unfreeze_variables(variables, mutable):
   return new_variables
 
 
+def bind(variables: VariableDict,
+         rngs: Optional[RNGSequences] = None,
+         mutable: CollectionFilter = False):
+  """Bind variables and rngs to a new ```Scope```.
```
avital, line 590:
```suggestion
  """Bind variables and rngs to a new ``Scope``.
```

flax/linen/module.py
```diff
@@ -790,9 +791,52 @@ def make_rng(self, name: str) -> PRNGKey:
...
+    instance does not mix well well with vanilla JAX APIs. Therefore, we recommend using
+    ``apply`` when code should be reusable and compatible across the JAX software ecosystem.
+
```
avital, line 807: It would really complement this docstring nicely, to have a very short HOWTO showing how to convert code that uses `bind` to code that uses `nn.apply` (by simply putting code inside a function).

Basically showing the autoencoder example that you have in the docstring and then on the RHS of the diff showing:
```
def f(model, x):
  z = model.encode(x)
  return model.decode(z)

ae = AutoEncoder()
x_reconstructed = ae.apply(f)(variables, x)
```

wdyt?

flax/linen/module.py
avital: Thanks, I understand now. I'll resolve the conversation.

flax/linen/module.py
```diff
@@ -790,9 +791,52 @@ def make_rng(self, name: str) -> PRNGKey:
...
+    bind provides an "interactive" instance of a Module directly without
+    transforming a function with ``apply``. This is particulary useful for debugging
+    and interactive use cases like notebooks where a function would limit the ability
+    split up code into different cells.
+
+    Once the variables (and optionally RNGs) are bound to a ``Module`` it becomes a
+    stateful object. Note that idiomatic JAX is functional and therefore an interactive
+    instance does not mix well well with vanilla JAX APIs. Therefore, we recommend using
+    ``apply`` when code should be reusable and compatible across the JAX software ecosystem.
```
avital, line 806: Not something that should block merging this PR, but it could be nice to have a short note showcasing exactly how this fails, as I think this may be a recurring question for people new to JAX who will see `bind` and want to use it more broadly. But we can also wait to see if people ask.

flax/linen/module.py
avital: Basically, then the docstring guides you out of your hole, once you want to use JAX transformations after you've tinkered in Colab for a while.

flax/linen/module.py
```diff
@@ -790,9 +791,52 @@ def make_rng(self, name: str) -> PRNGKey:
...
+        about variables.
+      rngs: a dict of PRNGKeys to initialize the PRNG sequences.
+        The "params" PRNG sequence is used to initialize parameters.
+      mutable: Can be bool, str, or list. Specifies which collections should be
+               treated as mutable: ``bool``: all/no collections are mutable.
+               ``str``: The name of a single mutable collection. ``list``: A
+               list of names of mutable collections.
+    Returns:
+      A copy of this instance with bound variables and RNGs.
+    """
```
avital, line 833: Thank you for the clear docstring!

flax/linen/module.py
```diff
@@ -1011,61 +1050,146 @@ def __call__(self, train: Optional[bool] = None):
...
+    f_jitted(variables, x)
+
+  Args:
+    fn: The function that should be applied. The first argument passed will
+      be an module instance of the ``module`` with variables and RNGs bound
+      to it.
+    module: The ``Module`` that will be used to bind variables and RNGs to.
+      The ``Module`` passed as the first argument to ``fn`` will be a clone
+      of module.
+    method: The literal name of a method in this class. If provided, applies
```
avital, line 1083: There's no `method` argument here, right?

avital (approved):
This looks really good, it all seems to fit together very nicely. I left a few comments/suggestions but nothing that needs to block merging.

examples/vae/train.py
jheek: It's not essential but I think it's nice to have at least one usage in the examples. Here we avoid passing custom methods and a double apply

flax/linen/module.py
jheek: Yes I agree, I would like to think a bit more about what we would put into an apply/bind howto so I'll make an issue for it

flax/linen/module.py
jheek: This is also something that could be part of the HOWTO

flax/linen/module.py
jheek: Fixed

examples/vae/train.py
avital: Ah yes I now fully agree that this is a good idea. I always disliked the `method` argument to `apply`... (And if we really want to we /could/ get rid of it now... But not sure)

flax/core/scope.py
```diff
@@ -576,6 +583,29 @@ def _unfreeze_variables(variables, mutable):
   return new_variables
 
 
+def bind(variables: VariableDict,
+         rngs: Optional[RNGSequences] = None,
+         mutable: CollectionFilter = False):
+  """Bind variables and rngs to a new ``Scope``.
+  
+  bind provides a ``Scope`` instance without transforming a function
```
andsteing, line 591: nit: ''bind'' (i.e. backticks)

flax/linen/module.py
andsteing: Would it be possible/desirable to avoid above pitfalls by throwing an error when a `module.bind()` is used inside a JAX transformation and/or refusing to be called from a Python module (as opposed to an interactive environment)?

flax/linen/module.py
```diff
@@ -790,9 +791,52 @@ def make_rng(self, name: str) -> PRNGKey:
...
+      z = model.encode(x)
+      x_reconstructed = model.decode(z)
+
+
+    Args:
+       variables: A dictionary containing variables keyed by variable
+        collections. See :mod:`flax.core.variables` for more details
+        about variables.
+      rngs: a dict of PRNGKeys to initialize the PRNG sequences.
+        The "params" PRNG sequence is used to initialize parameters.
```
andsteing, line 826: Doesn't `variables` require initialized `"params"`?

(Maybe mention that these `rngs` will be mutated in the attribute `.scope.rngs`)

flax/linen/module.py
```diff
@@ -1011,60 +1050,144 @@ def __call__(self, train: Optional[bool] = None):
...
-  # # TODO: Consider whether this is a helpful abstraction, and think about naming.
-  # # See its use in design_test/linen/weight_std.py
-  # def materialized(self, variables={}, rngs={}):
-  #   assert self.scope is None, ("Can't attach a module twice."
-  #                               " Maybe you want to clone first?")
-  #   return self.clone(parent=Scope(variables, rngs))
+def apply(fn: Callable[..., Any], module: Module,
+          mutable: CollectionFilter = False,
+          capture_intermediates: Union[bool, Callable[[Module, str], bool]] = False) -> Callable[..., Any]:
+  """Creates an apply function for the given function and ``Module`` instance.
```
andsteing, line 1056: What about

```
  """Creates an apply function to call ``fn`` with a bound module.
```

I feel mentioning `bound module` makes it clearer what this function is for...
(same could be argued for other 1st lines pydoc strings)

And/or a reference to `Module.bind()` would help understanding this function.

flax/linen/module.py
```diff
@@ -790,9 +791,52 @@ def make_rng(self, name: str) -> PRNGKey:
...
+    """Creates an interactive Module instance by binding variables and RNGs.
+
+    bind provides an "interactive" instance of a Module directly without
+    transforming a function with ``apply``. This is particulary useful for debugging
+    and interactive use cases like notebooks where a function would limit the ability
+    split up code into different cells.
+
+    Once the variables (and optionally RNGs) are bound to a ``Module`` it becomes a
+    stateful object. Note that idiomatic JAX is functional and therefore an interactive
+    instance does not mix well well with vanilla JAX APIs. Therefore, we recommend using
```
andsteing, line 805: Maybe `we recommend using apply when the code should be reusable` could be formulated a bit more strongly, something along the lines `bind() should only be used for interactive experimentation, and in all other cases we strongly encourage to use apply()` ?

(that's how I understood its purpose anyways)

PR #1088, opened 2021-03-05
Improve Flax Linen API __call__ and __init__ formatting in docstrings

8bitmp3:
Noticed that the formatting varies and the "more correct" way of formatting `__call__` and `__init__` methods in the Linen API's `module.py` is to write is as with two back ticks on each side:

```
``__call__``
```

which returns:

image

on the site.

Hope this helps @avital

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -421,7 +421,8 @@ class Module:
     # automatically transform Modules into dataclasses at subclass creation
     # time, and we set the last dataclass arguments to `parent` and `name`.
     cls._customized_dataclass_transform()
-    # We wrap user-defined methods including setup and __call__ to enforce
+    # We wrap user-defined methods including setup and _
+    to enforce
     # a number of different checks and to provide clear error messages.
     cls._verify_single_or_no_compact()
     cls._wrap_module_methods()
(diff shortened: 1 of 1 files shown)


flax/linen/module.py
```diff
@@ -421,7 +421,8 @@ def __init_subclass__(cls):
     # automatically transform Modules into dataclasses at subclass creation
     # time, and we set the last dataclass arguments to `parent` and `name`.
     cls._customized_dataclass_transform()
-    # We wrap user-defined methods including setup and __call__ to enforce
+    # We wrap user-defined methods including setup and _
```
avital, line 424: ?

avital (changes requested):
Thanks! (I added a comment)

flax/linen/module.py
8bitmp3: Good catch — undoing whatever that was. Thanks @avital !

```suggestion
    # We wrap user-defined methods including setup and __call__ to enforce
```

flax/linen/module.py
```diff
@@ -421,7 +421,8 @@ def __init_subclass__(cls):
...
+    to enforce
```
8bitmp3, line 425: Not sure what happened. Good catch @avital ! Undoing:

```suggestion
    # We wrap user-defined methods including setup and __call__ to enforce
```

avital (approved):
Thanks!

PR #1092, opened 2021-03-08
use config file, add note about how to use bucket for data

brettkoonce:


levskaya (approved):

examples/imagenet/README.md
```diff
@@ -66,6 +66,11 @@ python main.py \
 #### Running on Cloud
 
 See commands in [../cloud/README.md](../cloud/README.md)
+
+To use a dataset stored in a bucket, set the following shell parameter:
+
+    export TFDS_DATA_DIR="gs://$GCS_TFDS_BUCKET/tensorflow_datasets"
```
andsteing, line 72: This is described in more detail under `## Training the imagenet example` in ../cloud/README.md - why do you think it is helpful to add this comment?

Maybe it would be better to note that there is a longer description about how to prepare the `imagenet2012` dataset in that other README.

examples/imagenet/README.md
```diff
@@ -46,7 +46,7 @@ Configuration flag is defined using
 follows:
 
 ```shell
-python main.py --workdir=./imagenet_default --config.num_epochs=100
+python main.py --workdir=./imagenet_default --config.num_epochs=100 --config=configs/default.py
```
andsteing, line 49: Nit: we usually switch the order : `--config=config/default.py --config.num_epochs=100`

andsteing, line 49: This should also be added to line 37, no?

andsteing (changes requested):
Hi brettkonce !

Thanks for your PR - I'm sorry for the delay in replying, but the email slipped between my inbox filters and I just became aware of this PR now.

Best, Andreas

examples/imagenet/README.md
brettkoonce: tweaked!

brettkoonce: tweaked!

examples/imagenet/README.md
brettkoonce: As far as I can tell, it's not documented in the cloud/README.md file, am I missing something?  The only reference I see is https://github.com/google/flax/blob/4fd597c1d4c987d561abdb19f34eca58f15c7743/examples/cloud/startup_script.sh#L58 where the python script is setting said shell variable for the user.  My reference point into this is running flax on a TPU directly and so by extension having to manually set these variables to get the demo to work.

andsteing: Got it, LGTM in that case.

andsteing (approved):
Thanks Brett!

PR #1116, opened 2021-03-12
Remove FrozenDict checks to determine mutability.

jheek:
Previously lift used isinstance(FrozenDict) to determine if a Collection is mutable,
because CollectionFilter did not support representing "everything but x".

- Adds DenyList as a filter
- add subtract_filter(a, b)
- Remove FrozenDict check instead consistenly use mutable filters
- Fix lift.transform by using subtraction.


levskaya (approved):
a nice improvement!

PR #1117, opened 2021-03-12
Updates docstrings on module.py

marcvanzee:
Clarifying a difference between `self.param` and `self.variable`.


jheek (approved):
Looks great!

PR #1119, opened 2021-03-12
Remove deprecated LM1b example

avital:
We never ported the Colab but I think it's purpose was mainly to have /some/ example for TPUs and we now have https://github.com/google-research/google-research/tree/master/flax_models/t5x.


levskaya (approved):
Thanks!  We could resurrect LM1B notebook if desired once 1VM works on colab if people want it...

PR #1120, opened 2021-03-12
Fix checkpoint overwrite

### LINKED ISSUE #1091: flax.training.checkpoint.save_checkpoint with keep=1 leads to "file already exists" error
I'm using Jax latest, Tensorflow latest, Jaxlib 0.1.59.

Let say I run a python code which does a save_checkpoint with keep=1 once. Then, I rerun the python code again, it will give me a "tensorflow.python.framework.errors_impl.AlreadyExistsError: file already exists". This is really annoying because this means that I have to manually delete the checkpoints if I want to reuse the same python code which happens a lot when debugging. 

I think that this happens because the extra files are only deleted at the end of save_checkpoint.

>  File "/localscratch/jolicoea.63359842.0/1/ScoreSDEMore/run_lib.py", line 469, in evaluate
>     checkpoints.save_checkpoint(
>   File "/localscratch/jolicoea.63359842.0/1/env/lib/python3.8/site-packages/flax/training/checkpoints.py", line 99, in save_checkpoint
>     gfile.rename(ckpt_tmp_path, ckpt_path)
>   File "/localscratch/jolicoea.63359842.0/1/env/lib/python3.8/site-packages/tensorflow/python/lib/io/file_io.py", line 548, in rename_v2
>     _pywrap_file_io.RenameFile(
> tensorflow.python.framework.errors_impl.AlreadyExistsError: file already exists

jheek:
save_checkpoint behaviour was ill-defined when checkpoint with the same or larger step already exists.
- Add overwrite option to save_checkpoint (defaults to False)
- Raise an error  when overriding a checkpoint accidentally
- Raise an error when storing a checkpoint that has a smaller step than the latest.

Fixes #1091

### DIFF
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -25,6 +25,7 @@ import re
 
 from absl import logging
 from flax import core
+from flax import errors
 from flax import serialization
 from tensorflow.io import gfile
 
@@ -73,7 +74,8 @@ def save_checkpoint(ckpt_dir,
                     target,
                     step,
                     prefix='checkpoint_',
-                    keep=1):
+                    keep=1,
+                    overwrite=False):
   """Save a checkpoint of the model.
 
   Attempts to be pre-emption safe by writing to temporary before
@@ -85,7 +87,8 @@ def save_checkpoint(ckpt_dir,
     step: int or float: training step number or other metric number.
     prefix: str: checkpoint file name prefix.
     keep: number of past checkpoint files to keep.
-
+    overwrite: overwrite existing checkpoint files if a checkpoint
+      at the current or a later step already exits (default: False).
   Returns:
     Filename of saved checkpoint.
   """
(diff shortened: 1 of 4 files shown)


flax/training/checkpoints.py
```diff
@@ -94,16 +97,38 @@ def save_checkpoint(ckpt_dir,
...
+  print(ckpt_path)
+
+  # Remove newer checkpoints
+  if overwrite:
+    ind = checkpoint_files.index(ckpt_path) + 1
+    newer_ckpts = checkpoint_files[ind:]
+    checkpoint_files = checkpoint_files[:ind]
+    for path in newer_ckpts:
+      logging.info('Removing checkpoint at %s', path)
+      gfile.remove(path)
```
levskaya, line 129: probably fine since the default is not overwrite - but if people have this on a lot I'm sure someone will accidentally erase their checkpoints by accidentally re-using the same working directory. :)

levskaya (approved):

flax/training/checkpoints.py
jheek: Actually you have to set overwrite=True and disable checkpoint restore.

PR #1121, opened 2021-03-12
lm1b: tweak readme command

brettkoonce:


examples/lm1b/README.md
```diff
@@ -17,7 +17,7 @@ rate schedule. Based off of Machine Translation `wmt` example.
 
 ### How to run
 
-`python main.py --workdir=./lm1b_64 --config.per_device_batch_size=64`
+`python main.py --workdir=./lm1b_64 --config=configs/default.py`
```
avital, line 20: This changes the per device batch size (which is 128 in configs/default.py) Was that intentional?

(Unfortunately this README doesn't contain information on which precise accelerator this was tested, and whether 64 was the maximal batch size that fit into memory there)

brettkoonce: i had to cut it down to get things to run on my system (a lowly m40), but can tweak the config file to be 64 if you like, let me know!

brettkoonce: figure 2 from https://dl.acm.org/doi/pdf/10.1145/3360307 claims <= 256 is the 'perfect' setting, fwiw

avital: I wouldn't change the config file, but perhaps just add a comment here saying that you ran it on a m40. If you feel really excited, you could add a link to http://tensorboard.dev/ like we have in some of our other examples (e.g. the ImageNet or PixelCNN READMEs).

brettkoonce: @avital restored the override, let me know what you think!

avital (approved):

PR #1135, opened 2021-03-16
Checks no param is matched multiple times.

andsteing:


flax/optim/base.py
```diff
@@ -463,7 +465,20 @@ def __init__(self, *traversals_and_optimizers):
...
+
+      keykeys = traverse_util.unflatten_dict({
+        key_tuple: '/'.join(('',) + key_tuple)
+        for key_tuple in traverse_util.flatten_dict(params)
+      })
+      for key in traversal.iterate(keykeys):
+        if key in seen:
+          raise ValueError(f'Key "{key}" processed by multiple optimizers: '
+                           f'{seen[key]}, {idx}.')
+        seen[key] = idx
```
marcvanzee, line 480: Can't we just unflatten the params and do a raw string comparison?

```suggestion
    seen = set()
    for traversal, opt in zip(self.traversals, self.sub_optimizers):
      for key_tuple in traverse_util.flatten_dict(params):
        if key_tuple in seen:
          path = '/'.join(key_tuple)
          raise ValueError(f'Key "{path}" processed by multiple optimizers')
        seen.add(key_tuple)
```

marcvanzee (changes requested):

flax/optim/base.py
andsteing: We only want to bail if the same key is matched multiple times by different `traversal.iterate()` - so line 476 is crucial with this test. Note that `traverse_util.Traversal` only provides an iterator over the *values* hence the creation of `keykeys`.

marcvanzee: Ah ok that makes sense, so since `traversal.iterate` can be different dependning on which traversal we use, we want to call it here explicitly.

marcvanzee: Ahh ok, that makes sense. So in that case it seems the flatted/unflatten logic is there to be able to determine whether two `ModelParamTraversal` instances yield values for the same paths in the params, is that right? (I think the API of `traverse_util.Traversal` doesn't require using a dict-like structure, this seems to be specific to `ModelParamTraversal`.)

So since that logic seems quite special for `ModelParamTraversal`, and we could get other traversal classes in the future, I wonder if it would make sense to instead add something like `traverse_util.Traversal.iterate_with_path`, which also yields the path. Then you could simply do something like this here:

```python
for path, _ in traversal.iterate_with_path(params):
  if path in seen:
    raise ValueError(f'Key "{path}" processed by multiple optimizers: '
                     f'{seen[path]}, {idx}.')
  seen[path] = idx
```

I think then you also wouldn't have to do the extra flattening/unflattening you are doing here. 

WDYT?

marcvanzee (changes requested):

andsteing:
I agree that it should accept any traversal, not only `ModelParamTraversal`.

I wish `traverse_util.Traversal.iterate()` returned the path as well. If we want to change that interface it will be backwards-incompatible though and we would need to update all existing implementations.

(Note that adding a new function `traverse_util.Traversal.iterate_with_path()` doesn't really solve the problem because it will be absent in existing implementations, so we're back to only be able to use e.g. `ModelParamTraversal` if we implement the function there.)

Having an additional flattening/unflattening is suboptimal (note though it's only called during initialization) - but updating the interface seems like a big step for this one usage.

Given above, do you think we should replace the API?

marcvanzee:
Yes I see what you mean, if we were to use `iterate_with_path` and someone would implement a new `Traversal` class but only use `iterate`, then this might break.

It would indeed be best if `iterate` would just return the path and value, and I agree it is a bit involved for this change. But maybe we could add a TODO or file an issue with "PR welcome"? It would be nice if someone did this at some point.

marcvanzee (approved):

andsteing:
I like that idea. Filed #1170

Issue #1155, opened 2021-03-17
Implement a Sequential Module


marcvanzee:
Users often ask for this so it would be good to just add it. I can be as simple as this:

```python
class Sequential(nn.Module):
  layers: Sequence[nn.Module]

  def __call__(self, x):
    for layer in self.layers:
      x = layer(x)
    return x
```
Example usage:

```
class Foo(nn.Module):
  feature_sizes: List[int]
  @nn.compact
  def __call__(self, x):
    return Sequential([nn.Dense(sz, name=f'layers_{idx}')
                       for idx,sz in enumerate(self.feature_sizes)])(x)
```

baheytharwat:
Hi @marcvanzee,
I am interested in working on this issue. is that possible?

wookayin:
`flax.nn.Sequential`: https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Sequential.html

PR #1156, opened 2021-03-18
Implemented Sequential Module #1155

baheytharwat:
I solve #1155, I've made a new file in the linen directory for the sequential class and add it to the init file. 
I've made a colab [notebook](https://colab.research.google.com/drive/1PdZI7fvnTZM279TjdpI1OxYDMm_tbmfT?usp=sharing) to test the sequential module
Let me know your thoughts!


marcvanzee:
Looks good! Two comments:

- The build fails due to a pytype error, could you please fix this?
- How about we move your Colab to the test file?

baheytharwat:
Hi @marcvanzee, 
Thanks for your feedback! It helped me a lot. 
I have made as you mentioned, Let me know your thoughts about the changes

tests/linen/linen_sequential_test.py
```diff
@@ -0,0 +1,67 @@
...
+		key1, key2 = random.split(random.PRNGKey(0), 2)
+		x = random.uniform(key1, (4,4))
+
+		init_variables1 = model1.init(key2, x)
+		init_variables2 = model2.init(key2, x)
+
+		y1 = model1.apply(init_variables1, x)
+		y2 = model2.apply(init_variables2, x)
+
+		self.assertTrue(np.all(y1 == y2))
```
marcvanzee, line 63: Please use 2 space for a tab, and remove redundant newlines.

flax/linen/sequential.py
```diff
@@ -0,0 +1,13 @@
+from typing import Sequence
+from flax.linen.module import Module
+from typing import Type
+
```
marcvanzee, line 4:
```suggestion
```

flax/linen/sequential.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Sequential(Module):
```
marcvanzee, line 7: Please add documentation to both the Module and the `__call__` function. For instance as we have for the `Dense` layer: https://github.com/google/flax/blob/master/flax/linen/linear.py#L140

tests/linen/linen_sequential_test.py
```diff
@@ -0,0 +1,67 @@
...
+
+from absl.testing import absltest
+
+
+from flax import linen as nn
+import jax
+from jax import  random
+import numpy as np
+
+from typing import Sequence,List
```
marcvanzee, line 23:
```suggestion
from typing import Sequence, List
```

marcvanzee (changes requested):
Thanks! Few small additional comments.

flax/linen/sequential.py
```diff
@@ -0,0 +1,40 @@
...
+from typing import Type
+
+
+class Sequential(Module):
+
+  """ A sequential module for stacking layers sequentially. Each layer has only one input and one output. Layers will be added to it in the order they are passed in.
+    Attributes:
+      layers: A sequence of the layers that needed to be stacked sequentially.
+  """
+  layers: Sequence[Type[Module]]
```
relational, line 26: It seems `layers` is a sequence of Module instances, so shouldn't this just be `Sequence[Module]`?

baheytharwat: `Sequence[Module]` fails in building due to a pytype error. as 'Module' object is not callable. but it works well if you pass in Module instances. so, `Sequence[Type[Module]]` is used to pass instances of Module object.

relational: If you look at the documentation for typing.Type:
https://docs.python.org/3/library/typing.html#typing.Type
you'll see that it is used to represent class objects, not instances.

If you set layers as
```
  layers: Sequence[Type[Module]]
```

I think the issue is that you are doing `layer(x)`, which assumes that `__call__` is defined as a method.
While this is common for most modules (like Dense), the base class `Module` does not define `__call__` and you can define modules which do not have a `_call__` method.

When you declare layers as `Sequence[Type[Module]]` you are tricking pytype to think `layer(x)` is OK the same way that `nn.Dense(3)` is OK -- i.e. because classes define constructors.

flax/linen/sequential.py
```diff
@@ -0,0 +1,13 @@
...
+from typing import Type
+
+
+
+class Sequential(Module):
+  layers: Sequence[Type[Module]]
+
+  def __call__(self, x):
+    for layer in self.layers:
+      x = layer(x)
```
relational: Here you could perhaps first do:
```
if not callable(layer):
  raise ValueError(f"Each layer should implement `__call__`, but layer `{layer}` is not callable.")
```
This should make pytype stop complaining when you declare `layers: Sequence[Module]`.

flax/linen/sequential.py
n2cholas: Is there any reason `layers` has to be a `Sequence[Module]` instead of `Sequence[Callable]`? For example, users could use `lambda x: flax.linen.max_pool(x, (2, 2))` as a layer as well.

anirbanl:
Hi, Adding relu activation fails (not sure if I added it correctly): Here is the [Colab](https://colab.research.google.com/drive/1zcBDt_uNQhVatsuTBa6VESgUjSld2eqg?usp=sharing)

flax/linen/sequential.py
levskaya: mark this as Sequence[Callable] - that should work now w. typechecking and Modules

flax/linen/sequential.py
```diff
@@ -0,0 +1,40 @@
+# Copyright 2021 The Flax Authors.
```
levskaya, line 1: Can we call this file combinators.py instead in case we add other things like this.

levskaya:
Sorry this languished - summer has been a bit crazy for the maintainers - understood if you've moved on from this PR and focused on other things.  If not, please rebase and squash the PR into a single commit - we like keeping repo history neat.

kamalkraj:
@baheytharwat  @marcvanzee 
I am really interested in this Sequential Module feature. It 
Is it okay if I create a new PR from this one with the proposed changes from the reviews?

marcvanzee:
> @baheytharwat @marcvanzee I am really interested in this Sequential Module feature. Is it okay if I create a new PR from this one with the proposed changes from the reviews?

Yes, that would be awesome, thanks a lot for picking this up!

kamalkraj:
> Hi, Adding relu activation fails (not sure if I added it correctly): Here is the [Colab](https://colab.research.google.com/drive/1zcBDt_uNQhVatsuTBa6VESgUjSld2eqg?usp=sharing)

Why does adding `relu` activation function fails? Any idea? @marcvanzee

marcvanzee:
> > Hi, Adding relu activation fails (not sure if I added it correctly): Here is the [Colab](https://colab.research.google.com/drive/1zcBDt_uNQhVatsuTBa6VESgUjSld2eqg?usp=sharing)
> 
> Why does adding `relu` activation function fails? Any idea? @marcvanzee

I suspect this is due to RNG splitting. The `Dense` layers are called in different ways in the two modules, and they use an RNG for their kernel initializer, so if the RNG is split differently their kernels will have different initial values.

You can avoid it by passing `kernel_init=linen.initializers.ones` to all `nn.Dense` constructors, and then they will have the same values.

kamalkraj:
Thanks @marcvanzee 
passing `kernel_init=linen.initializers.ones` worked. 

For handling layers like `flax.linen.Dropout` and `flax.linen.BatchNorm`, which have additional input flags to control the behaviour in training and inference, I am thinking of a similar approach used in TF sequential module. 

https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/engine/sequential.py#L376-L387

Let me know if you have any other suggestions

marcvanzee:
> Thanks @marcvanzee passing `kernel_init=linen.initializers.ones` worked.
> 
> For handling layers like `flax.linen.Dropout` and `flax.linen.BatchNorm`, which have additional input flags to control the behaviour in training and inference, I am thinking of a similar approach used in TF sequential module.
> 
> https://github.com/keras-team/keras/blob/2c48a3b38b6b6139be2da501982fd2f61d7d48fe/keras/engine/sequential.py#L376-L387
> 
> Let me know if you have any other suggestions

Hmm I am not sure if that is the right approach. It will just plumb through any argument provided to `Sequential`'s `__call__` method if any of the provided layers have it as well, and we have many layers that require arguments to `__call__` besides the standard `input`, e.g.:

- `deterministic` (Dropout)
- `inputs_q`, `inputs_kv`, `mask`, `deterministic` (attention)
- `use_running_average` (BatchNorm)
- `carry` (LSTMCell)

So I feel that if we just plumb through everything then things may get a bit unwieldy, and if we get more layers with additional arguments this can lead to bugs (parameters could have the same name but different meaning).

Given that this issue is open for quite a long time now, I think it would be good to have an initial version of `Sequential` in Flax, so how about we start with not providing support for providing additional arguments to `__call__`, and we can always improve this later? Then I'll think a bit more about how to support this as well.

melissatan:
Hi @baheytharwat, are you still working on this PR? If not, I'd like to pick up this issue. 
If I don't hear from you in a week I'll assume you are not working on it anymore :) Thanks!

marcvanzee:
Closing this PR since @melissatan implemented the Sequential module in #1955 .

PR #1174, opened 2021-03-22
Add a comment to MultiOptimizer

marcvanzee:
This is following a question from a user who wasn't sure how to freeze part of a network for the first K steps.


marcvanzee:
> Maybe just merge the where into the block where we change the hyper params? In that case we have a full example and the previous block becomes a bit more realistic

Yes nice idea, I changed it.

jheek (approved):
Great!

PR #1175, opened 2021-03-22
Fixes ModelParamTraversal & value-dependent filter

andsteing:
Follow up to #1135

### DIFF
--- a/flax/optim/base.py
+++ b/flax/optim/base.py
@@ -15,7 +15,7 @@
 """Flax Optimizer api."""
 
 import dataclasses
-from typing import Any, Tuple
+from typing import Any, List, Tuple
 import warnings
 
 from .. import jax_utils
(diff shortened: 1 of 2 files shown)


andsteing:
The fix proposed in #1135 does not work with `ModelParamTraversal` that have a `filter_fn` that filters based on the `value` of a leaf. This PR expands the solution to fake the `.dtype` and `.shape` attributes, while maintaining the `key` and `sub_optimizers` indexing for improved error messaging

flax/optim/base.py
```diff
@@ -479,18 +493,18 @@ def __init__(
   def init_state(self, params):
     sub_states = []
     seen = {}
+    dummies = traverse_util.unflatten_dict({
```
jheek, line 496: Is it intentional to still use flatten_dict here? It's nice for the error message but it still breaks non-dict optimizer targets this way.

andsteing: Oh - I was only thinking of `ModelParamTraversal`...

PTAL, should work with arbitrary traversals now (incl. useful error message - assuming containers know how to `__str__`).

andsteing:
Note that`MultiOptimizer` might change some more soon : #1177

jheek (approved):
Looks good!

PR #1194, opened 2021-04-01
Remove explicit omnistaging enabling

jheek:
Jax will soon remove `jax.config.enable_omnistaging()`. We should no longer call this from examples and tests


levskaya (approved):
Ha!  I had totally forgotten that we manually set this, thanks for catching this!

PR #1200, opened 2021-04-05
Checkpoint fix

hmph:
# What does this PR do?

Fix bug in the case where job was preempted after the temporary checkpoint was written, but before it was renamed to the final checkpoint name.

Currently, this will result in an "InvalidCheckpointError" being thrown.

Fixes # (issue)

## Checklist
- [ ] This PR fixes a minor typo or improves the docs (you can dismiss the other
      checks if that's the case).
      link).


jheek (approved):
I'm surprised this didn't pop up before. Thanks for the fix!

flax/training/checkpoints.py
```diff
@@ -111,6 +111,8 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
     checkpoint_files.append(ckpt_path)
 
   checkpoint_files = natural_sort(checkpoint_files)
+  if checkpoint_files[-1] == ckpt_tmp_path:
```
avital, line 114: Can we add a short comment here explaining what this check does? (Basically the contents of this PR)? So that future collaborators on this code don't have to `git blame` to find this PR?

hmph: I have done this now

avital: Thanks Anurag!

PR #1203, opened 2021-04-06
Fix super calls in compact & setup

### LINKED ISSUE #1192: super().__call__ not generally safe to call in subclass __call__
```
class Foo(nn.Module):
  @nn.compact
  def __call__(self, x):
    return nn.Dense(4)(x)

class Bar(Foo):
  @nn.compact
  def __call__(self, x):
    y = super().__call__(x)
    return nn.Dense(4)(y)

k = random.PRNGKey(0)
x = random.randint(k, (4, 7), 0, 256)

variables = Bar().init(k, x)
y = Bar().apply(variables, x)
```
returns 
```
AssignSubModuleError: Submodule Dense must be defined in `setup()` or in a method wrapped in `@compact` (https://flax.readthedocs.io/en/improve-error/flax.errors.html#flax.errors.AssignSubModuleError)
```

This happens because the parent `super().__call__` is wrapped to mark itself as "compact" upon entry, clearing the "compact" state upon exit, leading to any remaining variable/submodule instantiations in the subclass `__call__` causing an error because the function no longs "looks compact".

I think we could fix this by adding a bit of logic to the `wrap_method_once` function and by passing in the class to `wrap_method_once` during subclass initialization to detect when `self` isn't an instance of the class it's "supposed to be", and having the wrapper acting as a passthrough.

jheek:
Fixes #1192

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -259,12 +259,16 @@ def wrap_method_once(fun: Callable[..., Any]) -> Callable[..., Any]:
     is_compact_method = hasattr(fun, 'compact')
     is_setup_method = fun.__name__ == 'setup'
     # We lazily call setup() only when needed.
-    if not is_setup_method:
+    if is_setup_method:
+      is_recurrent = self._state.in_setup
+      self._state.in_setup = True
+    else:
       self._try_setup()
 
     if is_compact_method:
       if self.scope is None:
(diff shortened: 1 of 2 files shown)


levskaya (approved):
Oh thanks for getting to this, I was about to look into this - looks great!

PR #1237, opened 2021-04-12
Fix DenseGeneral bug

jheek:
- Correctly expand bias to account
for batch & non-contracting dims.
- Sort axes for consistent output under permutations

### DIFF
--- a/tests/linen/linen_linear_test.py
+++ b/tests/linen/linen_linear_test.py
@@ -279,10 +279,24 @@ class LinearTest(parameterized.TestCase):
         return nn.DenseGeneral(features=6, axis=1, name='dense')(x)
 
     x = jnp.ones((2, 4, 8))
-    params = Foo().init(random.PRNGKey(0), x)['params']
-    self.assertEqual(jax.tree_map(jnp.shape, params), {
+    y, variables = Foo().init_with_output(random.PRNGKey(0), x)
+    self.assertEqual(jax.tree_map(jnp.shape, variables['params']), {
       'dense': {'kernel': (4, 6), 'bias': (6,)}
     })
+    self.assertEqual(y.shape, (2, 8, 6))
+  
+  def test_non_final_axes(self):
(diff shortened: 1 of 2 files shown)


tests/linen/linen_linear_test.py
```diff
@@ -144,7 +144,7 @@ def _counter_init(rng, shape, dtype, state):
     np.testing.assert_allclose(y, target)
 
   @parameterized.parameters([((-2, 3), (), 'bijk,jklm->bilm'),
-                             ((3, -2), (), 'bijk,kjlm->bilm'),
+                             ((3, -2), (), 'bijk,jklm->bilm'),
```
marcvanzee, line 147: So our testcase was actually testing wrong behavior? That's odd 😄

tests/linen/linen_linear_test.py
```diff
@@ -271,6 +271,18 @@ def test_embed(self):
     np.testing.assert_allclose(y, dummy_embedding[None])
     z = embed_module.apply(initial_params, jnp.ones((3,)), method=embed_module.attend)
     np.testing.assert_allclose(z, 3. * jnp.arange(4))
+  
+  def test_non_final_axis(self):
```
marcvanzee, line 275: For completeness we could also add a testcase where `axis` is a tuple, so we make sure it works on multiple axes as well.

marcvanzee (approved):

levskaya (approved):

tests/linen/linen_linear_test.py
jheek: No this was correct. We are free to pick the order of the kernel dimensions. But I made it invariant wrt to the order in which you pass the axes. I think this is more consistent and it makes the implementation easier.

tests/linen/linen_linear_test.py
jheek: done

PR #1240, opened 2021-04-12
Add "max 5 commits in PR" to GH actions

marcvanzee:
# What does this PR do?

Adds workflow item that counts the number of commits in the current branch, if it is > 5 compared to master, the build will fail.

This is to ensure users squash their commits and we won't break our source syncs accidentally.

I verified this by adding 6 commits to this PR and it indeed fails.

Currently this PR contains 3 commits, and the output is:

```
+ diff=4
+ ((  4 > 6 ))
```

As I add in a comment in the build file, the diff is one too big, so this check is correct.


avital (approved):
I wonder if there's a nice guide online we can link to, that showcases git rebase -i

avital (changes requested):
Actually I'm not sure this is correct -- if commit-count was merged off of a older version of master (which is quite likely) then $((eval-base)) will be smaller than it should be.

Maybe something like `git rev-list origin/master...commit-count` works?

marcvanzee:
> I wonder if there's a nice guide online we can link to, that showcases git rebase -i

Yes I was thinking the same, I've added a subsection to `contributing.md` and linked to it in the error message.

marcvanzee:
> Actually I'm not sure this is correct -- if commit-count was merged off of a older version of master (which is quite likely) then $((eval-base)) will be smaller than it should be.
> 
> Maybe something like `git rev-list origin/master...commit-count` works?

Fixed it.

avital:
I'd also like a review from @froystig and @jakevdp as they may want to incorporate this into JAX as well. (and then curious to see if we should share a single rebase tutorial, and if so, where)

marcvanzee:
> I'd also like a review from @froystig and @jakevdp as they may want to incorporate this into JAX as well. (and then curious to see if we should share a single rebase tutorial, and if so, where)

Yes I was planning to discuss this during our next sync, but doing the discussion here makes sense as well!

.github/workflows/build.yml
```diff
@@ -26,10 +26,29 @@ jobs:
       run: |
         sudo apt-get -y install pandoc
     - uses: actions/checkout@v2
+      with:
+        fetch-depth: 10
```
avital, line 30: what is this? can you add a comment explaining it

avital: and why is 10 sufficient?

.github/workflows/build.yml
```diff
@@ -26,10 +26,29 @@ jobs:
...
+    # We allow at most 5 commits in a branch to ensure our CI doesn't break.
+    - name: Check commit count in PR
+      if: always()
+      shell: bash
+      run: |
+        set -x 
+        git fetch origin $GITHUB_REF:commit-count
+        git fetch origin master
+        diff=$(git rev-list --count origin/master...commit-count)
+        # $GITHUB_REF adds an additional commit to the commit tree, so $diff is
```
avital, line 44: what is $GITHUB_REF?

avital: (maybe put a general comment explaining what GITHUB_REF is right after the `set -x` line?)

avital:
I added some comments. The doc looks GTM.

docs/contributing.md
```diff
@@ -103,6 +103,44 @@ Follow these steps to contribute code:
...
+
+If you encounter this error message, you should squash your commits. You can do
+this in various ways, and we demonstrate two options here.
+
+* If you simply want to squash the last 8 commits and write a new commit
+  message from scratch, make sure you checkout your branch and run:
+
+  ```bash
+  git reset --soft HEAD~8 && git commit
+  ```
```
jakevdp, line 126: The workflow I personally use is `$git rebase master && git reset --soft master && git commit`. Seems safer than counting commits... what do you think?

jakevdp:
I like this idea – we'll probably adapt the same check and language for JAX. On the question of combining developer docs: that seems like it would be a win, although there are some repo-specific differences in our workflows (e.g. JAX doesn't use codecov because we need multiple runs to cover X32 & X64 code paths, so the coverage number from a single run is not particularly accurate).

levskaya:
@marcvanzee @avital - just curious what's up with this PR - is there a reason we never merged it?

marcvanzee:
Oops, totally forgot about this one, I'll take a look at it this week. Thanks for reminding me @levskaya!

docs/contributing.md
marcvanzee: Good idea, updated it, and also updated the explanation!

.github/workflows/build.yml
marcvanzee: Added a comment.

.github/workflows/build.yml
marcvanzee: Added a comment.

marcvanzee:
@avital I finally took a look at this PR again, PTAL!

avital:
Assuming you've tested this, it LGTM! (But just to make sure, did you test the following: 

1. Check out an old version of main locally
2. branch off of that local (old) version of main
3. Add 4 commits, make a PR, see if it passes
4. Add 2 more commits, make a PR, see that it fails.

That is -- I see that you force pushed this change: "[marcvanzee](https://github.com/marcvanzee) [force-pushed](https://github.com/google/flax/compare/9b950c43c843b51fac175abb190687cde6c58b05..e5a27a581d15597d299cb24c0b49333077b4859c) the max-commits branch from [9b950c4 ]", so I'm not sure you tested the common case where people fork from an old version. (which is what my [comment above](https://github.com/google/flax/pull/1240#pullrequestreview-633898965) was about)

marcvanzee:
> Assuming you've tested this, it LGTM! (But just to make sure, did you test the following:
> 
> 1. Check out an old version of main locally
> 2. branch off of that local (old) version of main
> 3. Add 4 commits, make a PR, see if it passes
> 4. Add 2 more commits, make a PR, see that it fails.
> 
> That is -- I see that you force pushed this change: "[marcvanzee](https://github.com/marcvanzee) [force-pushed](https://github.com/google/flax/compare/9b950c43c843b51fac175abb190687cde6c58b05..e5a27a581d15597d299cb24c0b49333077b4859c) the max-commits branch from [[9b950c4](https://github.com/google/flax/commit/9b950c43c843b51fac175abb190687cde6c58b05) ]", so I'm not sure you tested the common case where people fork from an old version. (which is what my [comment above](https://github.com/google/flax/pull/1240#pullrequestreview-633898965) was about)

I'm not sure if you noticed, but I did actually implement the change you suggested in your comment!

I tested it as you suggested here https://github.com/google/flax/pull/1973, and indeed it works. Up to 5 commits passes but after that it starts failing: https://github.com/google/flax/runs/5468078293?check_suite_focus=true

PR #1243, opened 2021-04-13
Adds FLIPs link to main RST nav.

andsteing:
I also tried creating a manual index `.rst` file and inlining the markdown files in RST, but it did not look great. Simply adding the link makes FLIPs easily discoverable and it's also easy to search & browse FLIPs directly on Github.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -48,6 +48,7 @@ For a quick introduction and short example snippets, see our `README
    :titlesonly:
 
    design_notes/*
+   FLIPs <https://github.com/google/flax/tree/master/docs/flip>
 
 .. toctree::
    :maxdepth: 1


marcvanzee (commented):
Thanks for adding this. I do wonder whether it is easy for users to access flips now. Maybe it is nicer to have a `flips.rst` doc that points to all individual FLIPs with maybe a one line explanation for each FLIP?

andsteing:
I think we should keep everything as markdown on Github instead of mixing "two views".

But if you think it would be useful to have one-line summaries I could add tose to the beginning of the `README.md` - WDYT?

marcvanzee:
Yes I agree with your point that we should't duplicate information. 

However, I think we can also output .md in RTD directly (which is probably nicer so users just stay on RTD). The problem is just that FLIPs are right now using "#" for all headings, but RTD assumes this is only used for titles.

So one solution would be to add a "#" to all headings in all FLIPs documents and simply add this to `index.rst`:

```
.. toctree::
   :maxdepth: 1
   :caption: FLIPS
   :glob:
   :titlesonly:

   flip/README
   flip/*
```

And then if you then add `flip/0000-template.md` to `exclude_patterns` in `docs/conf.py`, it should show you a table of content in the side bar with README first, and then all flips in the `flip/` directory. WDYT?

andsteing:
While I do think that FLIPs are important, they're more about technical details, and in the future we might have more FLIPs that are less relevant to the user than the Optimizer FLIP.

So I'm not sure we want to automatically list all of them in the LHS navigation (we already fill a complete screen in landscape orientation, and all of the current items are more important than the average FLIP).

As for showing the FLIPs inside RST that's what I originally wanted to do but I found that the rendering is rather horrible:

flax_rst_flip_md

I think the main purpose is the FLIP process and in Github MD the documents look nice (i.e. as opposed to move to RST to make the FLIPs show nicely in RTD).

marcvanzee:
Ok -- I agree with your point about FLIPs being mostly internal and not part of the documentation. So let's keep it as it is.

I do think it should be possible to format the md files nicer than shown in your screenshot, but let's not spend too much time on it then.

marcvanzee (approved):

Issue #1053, opened 2021-02-28
Replace flax.optim with Optax in examples


andsteing:
See https://github.com/google/flax/blob/master/docs/flip/1009-optimizer-api.md#update-plan

The following examples need to be updated

- [x] imagenet #1251 
- [x] mnist #1302 
- [x] nlp_seq #1916
- [x] pixelcnn #1413
- [x] ppo #1404
- [x] seq2seq #1396
- [x] vae #1361 
- [x] wmt #1476
- [x] lm1b #1479

andsteing:
Run finished:
https://tensorboard.dev/experiment/w4PcKXloQMG7gXJhjskVjw/

- final test accuracy 0.7640
- total run time 5h4m

Compared to previous run using `flax.optim.Adam` (from `examples/imagenet/README.md`):
https://tensorboard.dev/experiment/iJzNKovmS0q6k5t6k5wvOw/#scalars&_smoothingWeight=0&regexInput=v100_x8_mixed_precision

- final test accuracy 0.7647
- total run time 4h51m

Note that the final test accuracy of the imagenet example fluctuates between 0.7625 and 0.7650, so the result is compatible.

avital:
Nice! Where is the branch of the code that you ran that used Optax?

andsteing:
Another finished run with the Optax code:
https://tensorboard.dev/experiment/xOpycRYnT7m3inYbEuNMxw/

- final test accuracy 0.766
- total run time 5h1m

marcvanzee:
Updated the checklist since PixelCNN++ doesn't exist anymore.

PR #1251, opened 2021-04-15
Updates examples/imagenet to use Optax.

andsteing:
#1053

### DIFF
--- a/examples/imagenet/train.py
+++ b/examples/imagenet/train.py
@@ -35,6 +35,7 @@ import models
 from flax.metrics import tensorboard
 from flax.training import checkpoints
 from flax.training import common_utils
+from flax.training import train_state
 
 import jax
 from jax import lax
@@ -68,8 +69,7 @@ def initialized(key, image_size, model):
   def init(*args):
     return model.init(*args)
   variables = init({'params': key}, jnp.ones(input_shape, model.dtype))
-  model_state, params = variables.pop('params')
-  return params, model_state
+  return variables['params'], variables['batch_stats']
 
 
 def cross_entropy_loss(logits, labels):
@@ -141,12 +141,10 @@ def train_step(state, batch, learning_rate_fn):
     # Re-use same axis_name as in the call to `pmap(...train_step...)` below.
     grads = lax.pmean(grads, axis_name='batch')
   new_model_state, logits = aux[1]
-  new_state = state.update(
-      grads=grads, batch_stats=new_model_state['batch_stats'])
   metrics = compute_metrics(logits, batch['label'])
   metrics['learning_rate'] = lr
 
-  new_state = state.update(
+  new_state = state.apply_gradients(
       grads=grads, batch_stats=new_model_state['batch_stats'])
   if dynamic_scale:
     # if is_fin == False the gradients contain Inf/NaNs and optimizer state and
@@ -196,7 +194,7 @@ def create_input_iter(dataset_builder, batch_size, image_size, dtype, train,
   return it
 
 
-class TrainState(checkpoints.TrainState):
+class TrainState(train_state.TrainState):
   batch_stats: Any
   dynamic_scale: flax.optim.DynamicScale
 
@@ -220,6 +218,8 @@ cross_replica_mean = jax.pmap(lambda x: lax.pmean(x, 'x'), 'x')
 
 def sync_batch_stats(state):
   """Sync the batch statistics across replicas."""
+  # Each device has its own version of the running average batch statistics and
+  # we sync them before evaluation.
   return state.replace(batch_stats=cross_replica_mean(state.batch_stats))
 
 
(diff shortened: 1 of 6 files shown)


andsteing:
Results from testrun:

- `v100_x8_mixed_precision` 76.27% 4h16m (compared to run from README with 76.47% 7h39m)
- `v100_x8` 76.72% 13h9m (compared to run from README with 76.72% 13h32m)

flax/training/checkpoints.py
```diff
@@ -313,3 +315,62 @@ def __call__(self, x):
     params_renamed[name] = convert_pre_linen(value)
 
   return core.freeze(params_renamed)
+
+
+class TrainState(struct.PyTreeNode):
```
avital, line 320: I propose putting this in a new file as it can be used independently of checkpointing.

flax/training/checkpoints.py
```diff
@@ -313,3 +315,62 @@ def __call__(self, x):
...
+    opt_state: The state for `tx`.
+  """
+  step: int
+  apply_fn: Callable = struct.field(pytree_node=False)
+  params: core.FrozenDict[str, Any]
+  tx: optax.GradientTransformation = struct.field(pytree_node=False)
+  opt_state: optax.OptState
+
+  def update(self, *, grads, **kwargs):
+    """Updates `step`, `params`, `opt_state` and `**kwargs` in return value."""
```
avital, line 354: Can we clarify in the docstring that actually takes a gradient step, and in addition updates other kwards? The current docstring doesn't really say much. And/or maybe even the name? Like apply_gradients?

flax/training/checkpoints.py
```diff
@@ -313,3 +315,62 @@ def __call__(self, x):
...
+    return self.replace(
+        step=self.step + 1,
+        params=new_params,
+        opt_state=new_opt_state,
+        **kwargs,
+    )
+
+  @classmethod
+  def create(cls, *, apply_fn, params, tx, **kwargs):
+    """Creates a new instance with `step=1` and initialized `opt_state`."""
```
avital, line 367:
```suggestion
    """Creates a new instance with `step=0` and initialized `opt_state`."""
```
```suggestion
    """Creates a new instance with `step=1` and initialized `opt_state`."""
```

examples/imagenet/train.py
```diff
@@ -236,10 +234,15 @@ def create_train_state(rng, config: ml_collections.ConfigDict,
     dynamic_scale = None
 
   params, model_state = initialized(rng, image_size, model)
-  optimizer = optim.Momentum(
-      beta=config.momentum, nesterov=True).create(params)
-  state = TrainState(
-      step=0, optimizer=optimizer, model_state=model_state,
+  tx = optax.chain(
```
avital, line 237: Can we use optax's `alias.sgd` instead?

https://github.com/deepmind/optax/blob/master/optax/_src/equivalence_test.py#L103

examples/imagenet/train.py
```diff
@@ -188,14 +196,9 @@ def create_input_iter(dataset_builder, batch_size, image_size, dtype, train,
...
 
-# flax.struct.dataclass enables instances of this class to be passed into jax
-# transformations like tree_map and pmap.
-@flax.struct.dataclass
-class TrainState:
-  step: int
-  optimizer: optim.Optimizer
-  model_state: Any
-  dynamic_scale: optim.DynamicScale
+class TrainState(checkpoints.TrainState):
```
avital, line 199: This worked out quite nicely!

examples/imagenet/train.py
```diff
@@ -217,16 +220,11 @@ def save_checkpoint(state, workdir):
 
 def sync_batch_stats(state):
   """Sync the batch statistics across replicas."""
-
-  # Each device has its own version of the running average batch statistics and
```
avital, line 221: Do we no longer need this code?

avital (approved):
Looks really good! And I think we now no longer use `variables.pop` in the imagenet example? Which would help as we'll probably make apply and init no longer return FrozenDicts: #1223 

I added some comments.

examples/imagenet/train.py
jheek: the new code looks good. I would not delete the comment because users often don't understand why this sync happens

flax/training/checkpoints.py
andsteing: Moved to `flax.training.training_state`. Don't hesitate to inspire better naming if you can...

flax/training/checkpoints.py
andsteing: Renamed & expanded documentation to make this clearer.

Note that we used to write `.apply_gradient()` (note the missing `s`) with `flax.optim.Opimizer` - but I think `.apply_gradients()` is better because we usually say `params` and `grads` when talking about entire pytrees of parameters/gradients. Also, we we're not trying to be compatible with `flax.optim.Optimizer` so I think it's fine to add the `s` here...

flax/training/checkpoints.py
andsteing: Updated.

examples/imagenet/train.py
andsteing: Ah cool, didn't see the recent update to directly use schedules in the aliases yet.

andsteing: This is quite a bit cuter, but less instructional. I can see advantages with both approaches.

flax/training/train_state.py
```diff
@@ -0,0 +1,78 @@
...
+  opt_state: optax.OptState
+
+  def apply_gradients(self, *, grads, **kwargs):
+    """Updates `step`, `params`, `opt_state` and `**kwargs` in return value.
+    
+    Note that internally this function calls `.tx.update()` followed by a call
+    to `optax.apply_updates()` to update `params` and `opt_state`.
+
+    Args:
+      grads: Gradients that match `.params`.
```
avital, line 49:
```suggestion
      grads: Gradients that have the same pytree structure as `.params`.
```

flax/training/checkpoints.py
avital: Agreed.

avital (approved):

PR #1254, opened 2021-04-16
Don't update batch statistics during initialization.

### LINKED ISSUE #1250: Re-used BatchNorm layer with named axis can't be initialised in train mode

When trying to initialise a model with a re-used `BatchNorm` layer a failure occurs when `use_running_average=False` and I've set a named axis (e.g. `axis_name="batch"`). Here is a minimal example which will fail:

```
class TestNet(nn.Module):
    @nn.compact
    def __call__(self, x, train: bool = True):
        norm = nn.BatchNorm(
            name="norm",
            use_running_average=not train,
            momentum=0.9,
            epsilon=1e-5,
            axis_name="batch"
        )

        for _ in range(2):
            x = norm(x) 

        return x

key = random.PRNGKey(0)
model = TestNet()
variables = model.init(key, jnp.ones((10,)))
```

I'd expect the initialization to be successful since this works if any of these three conditions are not met:
1. `use_running_average=False`,
2. there is a named axis, and
3. the `BatchNorm` layer is not reused. 

Instead, I get the following error...

```
NameError: unbound axis name: batch. The following axis names (e.g. defined by pmap) are available to collective operations: []
```

### Steps to reproduce:
Here is a Colab to reproduce the failure as well as successful cases when any of the conditions above are not met: https://colab.research.google.com/drive/1N7Wk6eUdW4UO6Ckj_tUscDhOlO0DtY0P?usp=sharing&forceEdit=true&sandboxMode=true

Please let me know if I can clarify anything!

jheek:
- This PR switches to checking if params is mutable to determine whether the running average of batch stats should be updated.
- Previously init behavior failed / was inconsistent when a batchnorm layer was used multiple times. 

Fixes #1250

### DIFF
--- a/tests/linen/linen_test.py
+++ b/tests/linen/linen_test.py
@@ -156,11 +156,13 @@ class NormalizationTest(absltest.TestCase):
             axis_name="batch",
         )
         x = norm(x)
-        return norm(x)
+        return x, norm(x)
 
     key = random.PRNGKey(0)
     model = Foo()
-    variables = model.init(key, jnp.ones((10,)))
+    x = random.normal(random.PRNGKey(1), (2, 4))
+    (y1, y2), variables = model.init_with_output(key, x)
+    np.testing.assert_allclose(y1, y2, rtol=1e-4)
 
 class StochasticTest(absltest.TestCase):
 
--- a/tests/linen/linen_test.py
+++ b/tests/linen/linen_test.py
@@ -146,6 +146,21 @@ class NormalizationTest(absltest.TestCase):
     with self.assertRaises(ValueError):
       model_cls.init_with_output(key2, x)
 
+  def test_batch_norm_multi_init(self):
+    class Foo(nn.Module):
+      @nn.compact
+      def __call__(self, x):
+        norm = nn.BatchNorm(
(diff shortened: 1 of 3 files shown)


tests/linen/linen_test.py
```diff
@@ -146,6 +146,21 @@ def test_group_norm_raises(self):
...
+            name="norm",
+            use_running_average=False,
+            axis_name="batch",
+        )
+        x = norm(x)
+        return norm(x)
+
+    key = random.PRNGKey(0)
+    model = Foo()
+    variables = model.init(key, jnp.ones((10,)))
```
avital, line 163: This test only checks that no error is thrown. Can we also test that the return values are correct? It sounds like a simple reasonable check is that the two calls to norm() return the same thing (I think?) so you could return them both from Foo.__call__ (`return norm(x), norm(x)`) and then just assert that you get same arrays out

avital (changes requested):

tests/linen/linen_test.py
jheek: Done

avital (approved):

PR #1256, opened 2021-04-19
Expose deny list in flax.core and flax.linen

jheek:


levskaya (approved):
lgtm - just a drive-by comment - wouldn't calling it something like "nn.Complement" or "nn.Not" be a little more self-explanatory if we're making this public?

jheek:
>> wouldn't calling it something like "nn.Complement" or "nn.Not" be a little more self-explanatory if we're making this public?

Yeah I guess DenyList is still a bit of a new term let's see if it sticks. We could always make an alias if it turns out problematic.

Issue #1261, opened 2021-04-21
Saving/Loading parameters changes weight format and can lead to silent errors


patrickvonplaten:
Provide as much information as possible. At least, this should include a description of your issue and steps to reproduce the problem. If possible also provide a summary of what steps or workarounds you have already tried.

### Problem you have encountered:

If one saves & reloads a `state` as follows:

```python
  model_file = os.path.join(temp_dir, "flax_model")
  with open(model_file, "wb") as f:
      model_bytes = to_bytes(state)
      f.write(model_bytes)

  with open(model_file, "rb") as state_f:
      state = from_bytes(AwesomeModel, state_f.read())
```

The `state` previously contained weights of type `jaxlib.xla_extension.DeviceArray`, but is then changed to simply numpy arrays: `np.ndarray`. 

This can lead to silent bugs when, *e.g.*, an embedding layer `flax.linen.Embed` is used since it behaves differently depending on the format of its weights.

### What you expected to happen:

I would expect saved weights to be loaded in the exact same format. Or maybe am I using the wrong save/load functions?

### Steps to reproduce:
If you run the following code, you can see how saving/loading can lead to unexpected errors:

```python
#!/usr/bin/env python3
import flax.linen as nn
from flax.serialization import to_bytes, from_bytes
import numpy as np
import jax.numpy as jnp
from jax.random import PRNGKey
import tempfile
import os


class AwesomeModel(nn.Module):
    dtype: jnp.dtype = jnp.float32
    num_embeddings = 20
    embedding_size = 16

    def setup(self):
        self.word_emb = nn.Embed(self.num_embeddings, self.embedding_size, dtype=self.dtype)

    def __call__(self, input_ids):
        return self.word_emb(input_ids)


input_ids = jnp.broadcast_to(np.arange(10)[None, :], (4, 10))  # => shape [4, 10]

model = AwesomeModel()
state = model.init(PRNGKey(0), input_ids)

# No problem here!
embeddings = model.apply(state, input_ids)

# Now let's save & reload the weights
with tempfile.TemporaryDirectory() as temp_dir:
    model_file = os.path.join(temp_dir, "flax_model")
    with open(model_file, "wb") as f:
        model_bytes = to_bytes(state)
        f.write(model_bytes)

    with open(model_file, "rb") as state_f:
        state = from_bytes(AwesomeModel, state_f.read())

# ...and rerun the forward pass
embeddings = model.apply(state, input_ids) # Error - this is a bit unexpected
```

[...]

PR #1262, opened 2021-04-22
Fix behavior of Embed when embedding is a numpy array

### LINKED ISSUE #1157: np.array parameters may lead to a silent failure
Passing np.array parameters (instead of jnp.array) to a linen module may lead to a silent failure, see the following example:

```
import flax.linen as nn

import jax
import jax.numpy as jnp

import numpy as np

t = jnp.zeros([2, 196], jnp.int32)
print(f'Input shape: {t.shape}')

m = nn.Embed(32, 10)
rng = jax.random.PRNGKey(0)
vars = m.init(rng, t)
o1 = m.apply(vars, t)
print(f'Expected output shape: {o1.shape}')

o2 = m.apply(jax.tree_map(np.array, vars), t)
print(f'Numpy output shape: {o2.shape}')
```

Output:

```
Input shape: (2, 196)
Expected output shape: (2, 196, 10)
"Numpy params" output shape: (196,)  <-- Different output shape
```

jheek:
Use take in Embed because fancy indexing numpy arrays with JAX indices does not work correctly.

Fixes #1157, #1261

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -407,7 +407,8 @@ class Embed(Module):
     """
     if not jnp.issubdtype(inputs.dtype, jnp.integer):
       raise ValueError('Input type must be an integer or unsigned integer.')
-    return self.embedding[inputs]
+    # Use take because fancy indexing numpy arrays with JAX indices does not work correctly.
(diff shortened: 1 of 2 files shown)


marcvanzee (approved):
Nice, thanks!

PR #1267, opened 2021-04-23
Add design note on lifted transformations

jheek:


docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+    out_variables = repack_fn(scope)
+    return y, out_variables
+
+  return lift.pack(wrapper, (variables,), (variables,), (rngs,))
+
+x = jnp.ones((3, 2))
+y, params = init(lift_id(core_nn.dense))(random.PRNGKey(0), x, 4)
+```
+
+Normallly users shouldn't have to
```
levskaya, line 258: shouldn't have to what?

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+  return lift.pack(wrapper, (variables,), (variables,), (rngs,))
+
+x = jnp.ones((3, 2))
+y, params = init(lift_id(core_nn.dense))(random.PRNGKey(0), x, 4)
+```
+
+Normallly users shouldn't have to 
+
+### Supported transformatiosn
```
levskaya, line 260: transformations

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+## Alternatives
+
+Other numerical computation frameworks consider variables a first-class citizen.
+An alternative to functionalization would be to use a variable system either integrated or on top of JAX.
+An advantage of this is that per-variable lifting becomes easier.
+If variables are part of the JAX IR (JAXPR), we could inspect which variables have to be lifted in a certain computation.
+Optionally, they could be annotated with a collection tag to decide on various lifting options.
+
+The downside of this approach is that a variable system is more complicated.
+Variables are related references and break a core assumption of Functional Programming (see [referentia transparency](https://en.wikipedia.org/wiki/Referential_transparency))
```
levskaya, line 337: referential

levskaya (approved):
This is great!  just spelling nits, etc.

jheek:
> This is great! just spelling nits, etc.

You will be happy to learn I just found myself a VSCode spell checker plugin :)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
+# Lifted transformation
+
+⚠️ Advanced topic ⚠️
+
+This design note explains the underlying implementation of `flax.linen.transform` which enable JAX transformations inside `Module`.
```
8bitmp3, line 5: Suggestion:

(unless `transform` in this context is plural)

```suggestion
This design note explains the underlying implementation of `flax.linen.transform`, which enables JAX transformations inside `Module`.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+## Introduction
+
+JAX uses a functional API meaning that it only garantuees correct behavior when using functions without side effects.
+Typically these side effects are the result of mutating an object that lives outside of the function.
```
8bitmp3, line 11: Suggestion: "outside the function" w/o "of"

```suggestion
Typically, these side effects are the result of mutating an object that lives outside the function.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+The functional paradigm has some advantages like the ability to reason about state and stochasticity explicitly.
```
8bitmp3, line 13: Suggestion: "the ability to reason explicitly about state and stochasticity"

```suggestion
The functional paradigm has some advantages like the ability to explicitly reason about state and stochasticity.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+A function output can only change when the input arguments change.
+Therefore, a function is garantueed to behave deterministicly.
+
+But pure functions offer another big advantage to JAX specifically: They enable functional transformations.
```
8bitmp3, line 17: Maybe:

```suggestion
But pure functions offer another big advantage to JAX: specifically, they enable functional transformations.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+For example `jax.vmap(f)` will vectorize a function `f`.
+Because `f` cannot have side-effects the vectorized/parallel version of `f` is well-defined. To see why we need this restriction, consider what happens if `f` would increment a counter or draw a random number.
```
8bitmp3, line 19: Suggestion: "side effects" without the dash (for consistency)

```suggestion
Because `f` cannot have side effects the vectorized/parallel version of `f` is well-defined. To see why we need this restriction, consider what happens if `f` would increment a counter or draw a random number.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+Would `f` draw the same or a different random number for each item in the vector?
+Would each item in the batch have its own counter or is the counter shared among the items?
+And in what order is the counter incremented if `f` is computed in parallel?
+The anser to all these questions is "it depends".
+The behavior is ambigious and the functional constraint elegantly avoids this problem.
+
+Flax introduces a safe way to have limited randomness and stateful variables in a JAX compatible form.
+The reason state in Flax is not problematic is because it is local. Inside a flax `Module` there are variables and PRNG sequences,
```
8bitmp3, line 27: Suggestion (correct me if I'm wrong): "The reason why the  state .. is not problematic is because..." (also "Flax" with a capital "F")


```suggestion
The reason why the state in Flax is not problematic is because it is local: inside a Flax `Module` there are variables and PRNG sequences,
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+but on the outside there are only JAX Arrays and PRNG keys.
+
+For most use cases Flax be used to define models in a stateful way.
```
8bitmp3, line 30: Suggestion: "can" may be missing ("...Flax can be used...")

```suggestion
For most use cases, Flax can be used to define models in a stateful way.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+Becuase a `Module` behaves pure functional externally  we can utilize the full power of JAX with all of its transformations.
```
8bitmp3, line 31: Suggestion: I think this should be reworded:

```
Because a `Module` behaves pure functional externally...
```
Also, "Because" (small typo).

Maybe:

```suggestion
Because a `Module` behaves like a pure function externally, we can utilize the full power of JAX with all of its transformations.
```

(Forgive me if I got this wrong)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+There are however cases where we want to have the best of both worlds by using transformations and Module together.
```
8bitmp3, line 32: Suggestion: use backticks (formatting) for `Module` for consistency

```suggestion
There are, however, cases when we want to have the best of both worlds by using transformations and `Module` together.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+This design note explains how we extend JAX functional transformation to work on modules that have internal state and randomness.
```
8bitmp3, line 33: Suggestion (although maybe I'm incorrect):

```suggestion
This design note explains how we extend JAX's functional transformation to work on `Module`s that have internal state and randomness.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+## Functionalization
+
+Before we jump into the details let's consider a simple example where we would like to use `vmap` inside a `Module`.
+
+First we define a simple MLP without any transformations:
```
8bitmp3, line 40: Suggestion: nit (note the blank line)

```suggestion
First, we define a simple MLP without any transformations:

```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+```
```
8bitmp3, line 41: Suggestion: "```python" for improved rendering of Python code

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+class MLP(nn.Module):
+  @nn.compact
+  def __call__(self, xs):
+    h = nn.Dense(4, name='hidden')(xs)
+    h = nn.relu(h)
+    return nn.Dense(1, name='out')(h)
+```
+
+Now what if I want to have seperate MLP parameters for each item in `xs`?
```
8bitmp3, line 54: Suggestion: 
- for consistecy, maybe "we" > "I". 
- "separate" (typo fix)

```suggestion
Now, what if we want to have separate MLP parameters for each item in `xs`?
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+If this where "vanilla JAX" we could imagene writing something like `jax.vmap(apply_mlp)(mlp_params, xs)`.
```
8bitmp3, line 55: Suggestion:
- "imagine" (typo fix)
- proposal: "If we used "vanilla JAX" > "If this where "vanilla JAX" (or "this was" > "this where")

e.g.

```suggestion
If this was "vanilla JAX", we could imagine writing something like `jax.vmap(apply_mlp)(mlp_params, xs)`.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+But doing something like this in linen will actually fail:
+```
```
8bitmp3, line 57: Suggestion (nit)

```suggestion

```python
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+class NaiveVmapMLP(nn.Module):
+  @nn.compact
+  def __call__(self, xs):
+    mlps = MLP()
+    return jax.vmap(lambda mlp, x: mlp(x))(mlps, xs)  # fails
+```
+
+JAX will refuse to transform the `mlps` because its not a JAX Array or a simple container of arrays.
```
8bitmp3, line 65: Suggestion:

```suggestion
JAX will refuse to transform the `mlps` because it's not a JAX array or a simple container of arrays.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+We can't really blame JAX for refusing this dirty, underspecified task.
```
8bitmp3, line 66: Suggestion (to avoid repetition):

```suggestion
We can't really blame JAX for not allowing this "dirty", underspecified task.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+After all it's not even clear what should happen here.
```
8bitmp3, line 67: Suggestion (it):

```suggestion
After all, it's not even clear what should happen here.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+The parameters inside MLP are not even initialized yet and we will need a seperate PRNG key for each group of parameters.
```
8bitmp3: Suggestion:
- "the MLP" (definite article)
- "separate" (typo fix)

```suggestion
The parameters inside the MLP are not even initialized yet and we will need a separate PRNG key for each group of parameters.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+We can fix this problem by first turning `MLP` into a pure init and apply function.
+Afterwards, we use the `param` method to store the parameters:
+```
```
8bitmp3, line 72: Suggestion (nit):

```suggestion

```python
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+        out: {
+            bias: (3, 1),
+            kernel: (3, 4, 1),
+        },
+    },
+}
+"""
+```
+
+Here `MLP(parent=None)` creates a detached instance of `MLP`.
```
8bitmp3, line 101: Nit:

```suggestion
Here, `MLP(parent=None)` creates a detached instance of `MLP`.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+    mlps = MLP()
+    return jax.vmap(lambda mlp, x: mlp(x))(mlps, xs)  # fails
+```
+
+JAX will refuse to transform the `mlps` because its not a JAX Array or a simple container of arrays.
+We can't really blame JAX for refusing this dirty, underspecified task.
+After all it's not even clear what should happen here.
+The parameters inside MLP are not even initialized yet and we will need a seperate PRNG key for each group of parameters.
+
+We can fix this problem by first turning `MLP` into a pure init and apply function.
```
8bitmp3, line 70: Suggestion (similar to how you used the formatting below):

```suggestion
We can fix this problem by first turning `MLP` into a pure `init` and `apply` function.
```

(or `.init`, `.apply`)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+"""
+```
+
+Here `MLP(parent=None)` creates a detached instance of `MLP`.
+This avoids reserving a name for the submodule inside the current module.
+Although not strictly necessary, this also ensures we cannot accidentaly use the MLP instance in a stateful way and we are forced to use it through either `.init` or `.apply`.
+
+This example is still relatively concise but it already takes a few extra "bookkeeping" statements to make it work.
+However, this implementation has a number of limitations:
+1. During initialization we call the submodule twice through `init_fn` & `apply_fn`. If the submodule used the same trick to do
```
8bitmp3, line 107: Nit:

```suggestion
1. During th initialization we call the submodule twice through `init_fn` and `apply_fn`. If the submodule used the same trick to do
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+   functional tranformation we will end up executing a lot of code as the number of module calls grows like 2^d where d is the number of
+   nested function transformations.
+2. The implemenation assumes the submodule only requires the parameter RNG sequence.
+3. The implementation assumes we init parameters once and we don't create any state variables either during init or apply.
```
8bitmp3, line 114: Suggestion:

```suggestion
3. The implementation assumes we `init` the parameters once and don't create any state variables either during `init` or `apply`.
```

or "initialize the parameters"

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+Point 3 in particular makes manual functionalization cumbersome. Feel free to try and extend the above example with a `nn.BatchNorm`
+layer in the `MLP` module. This should cause a number of complexites further growing the implementation. Like storing the updated
```
8bitmp3: Suggestion (and correct me if I misunderstood this)

```suggestion
layer in the `MLP` module. This should increase the complexity of the implementation. Like storing the updated
```

(the original didn't look right (grammar) IMO)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+batch stats after apply and making sure state isn't mutable inside vmap when it's immutable outside (eval mode).
```
8bitmp3, line 118: Suggestion:

```suggestion
batch stats after `apply` and making sure the state isn't mutable inside `vmap` when it's immutable outside (the `eval` mode).
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+We call the process of transforming a stateful Module into a pure function "functionalization".
```
8bitmp3, line 121:
```suggestion
We call the process of transforming a stateful `Module` into a pure function "functionalization".
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+By temporarily turning a stateful Module into a function we make it compatible with JAX's functional transformations.
```
8bitmp3, line 122: Suggestion (for consistency):

```suggestion
By temporarily turning a stateful `Module` into a function we make it compatible with JAX's functional transformations.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+## Lifting
+
+
+Flax provides an alternative for manual functionalization which we call lifted transformation.
```
8bitmp3, line 126: Suggestion:

```suggestion
Flax provides an alternative for manual functionalization which we call a lifted transformation.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+Lifted transformations are defined in `flax.core.lift`.
+All the lifted JAX transformations are defined with a single generic lifting API called `pack`.
+
+A number of decision decisions had to be made in order to define `pack`. The implementation
```
8bitmp3, line 130: Suggestion (repetition fix)

```suggestion
A number of decisions had to be made in order to define `pack`. The implementation
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+of `pack` controls how variables and rngs are lifted and how fine-grained the user control is.
```
8bitmp3, line 131: Suggestion (RNGs - capitalized acronym)

```suggestion
of `pack` controls how variables and RNGs are lifted and how fine-grained the user control is.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Lifting granularity
+
+
+With the linen API, users can define arbitrary variable collections and PRNG sequences.
+Each variable in a collection is lifted in the same way.
+
+Collections are typically given a semantically meaningful name like "params" or "batch_stats" rather than a general purpose name like "state".
+Because collections carry semantic meaning we can decide at the transformation level how each collection should be lifted.
+For example, we might decide to share parameters if we want to vectorize add a batch dimension to a model.
```
8bitmp3, line 143: Is it "vectorize a batch dimension" without "add"? Or, is it "if we want to add a vectorized batch dimension to a model"

Another suggestion:

```suggestion
For example, we may decide to share parameters if we want to vectorize add a batch dimension to a model. 
```

(might < may/could < will)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+Each variable in a collection is lifted in the same way.
+
+Collections are typically given a semantically meaningful name like "params" or "batch_stats" rather than a general purpose name like "state".
+Because collections carry semantic meaning we can decide at the transformation level how each collection should be lifted.
+For example, we might decide to share parameters if we want to vectorize add a batch dimension to a model. 
+
+At the same time we can write generic code that uses transformations without knowing exactly what kind of variables the submodules will create.
+Collections thus strike a balance between fine-grained control and generality.
+We also avoid brittle string matching code that loops over all variables and tries to split up collections in an ad-hoc way based on
+naming conventions like: does this variable name have a prefix "kernel".
```
8bitmp3, line 148: Maybe there should be a "?" at the end of the sentence. 

E.g.
```suggestion
naming conventions like: does this variable name have a prefix "kernel"? 
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+Collections are typically given a semantically meaningful name like "params" or "batch_stats" rather than a general purpose name like "state".
+Because collections carry semantic meaning we can decide at the transformation level how each collection should be lifted.
+For example, we might decide to share parameters if we want to vectorize add a batch dimension to a model. 
+
+At the same time we can write generic code that uses transformations without knowing exactly what kind of variables the submodules will create.
+Collections thus strike a balance between fine-grained control and generality.
+We also avoid brittle string matching code that loops over all variables and tries to split up collections in an ad-hoc way based on
+naming conventions like: does this variable name have a prefix "kernel". 
+If more fine-grained control is necessary a user can simply split up a set of variables over multiple collections that should be handled differently.
```
8bitmp3, line 149: Suggestion (add a coma which can help read the complex sentence)

```suggestion
If more fine-grained control is necessary, a user can simply split up a set of variables over multiple collections that should be handled differently.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+### Transformation vs variable control
+
+
+Lifting behavior could be defined either at the transformation level or during variable definition.
+We use transformation level definitions of lifting behavior.
+The reason for this choice is that there are many different transformations with various behaviors.
+For example: `vmap` has broadcasted and vectorized arguments, while `scan` has scan, carry, and broadcast arguments.
+A variable would have to define its behavior for all these transformations otherwise a `Module` would not be compatible with
+these tranformations. Or alternatively, we would have to make default decisions for how transformations are handled.
```
8bitmp3: Suggestion (nit):
- "alternatively" may be sufficient without "or"
- "transformations" (fix typo)

```suggestion
these transformations. Alternatively, we would have to make default decisions for how the transformations are handled.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+However, this could lead to silent bugs becuase the behavior might not actually be valid given the users intent.
```
8bitmp3: Suggestion:

- small typo fix
- "the users intent" or "a user's intent" 🤷 

```suggestion
However, this could lead to silent bugs because the behavior might not actually be valid given a user's intent. 
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+Lifting behavior could be defined either at the transformation level or during variable definition.
+We use transformation level definitions of lifting behavior.
+The reason for this choice is that there are many different transformations with various behaviors.
+For example: `vmap` has broadcasted and vectorized arguments, while `scan` has scan, carry, and broadcast arguments.
+A variable would have to define its behavior for all these transformations otherwise a `Module` would not be compatible with
+these tranformations. Or alternatively, we would have to make default decisions for how transformations are handled.
+However, this could lead to silent bugs becuase the behavior might not actually be valid given the users intent. 
+
+The lift package also provides a general purpose `transform` which allows an arbitrary function to transforms a variable collection.
+This can be used to for example tie the weights in a tied autoencoder by transposing the weights.
```
8bitmp3: Suggestion (refactor):

```suggestion
Fox example, this can be used to tie the weights in a tied autoencoder by transposing the weights.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+It is unclear whether a similair general purpose transform could be defined if lifting decisions were made at variable definition.
+
+
+### Linen
+
+The lifting module does not know about linen Modules. Instead it operates directly on instances of `flax.core.Scope`.
+A `Scope` contains the variables and PRNG sequences of a `Module`. Each module has a `Scope` instance linked to it as long
+as its bound to a parent or it was created using a `init` or `apply`.
```
8bitmp3, line 172: Nit:

```suggestion
as it's bound to a parent or created using an `init` or `apply`.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+When a `Module` is transformed we use the `flax.core.lift` APIs to lift the scope and use `Module.clone()` to create a new `Module` instance with the lifted `Scope` bound to it.
```
8bitmp3, line 174: Nit:

```suggestion
When a `Module` is transformed, we use the `flax.core.lift` APIs to lift the scope and use `Module.clone()` to create a new `Module` instance with the lifted `scope` bound to it.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+`flax.linen.transforms` exposes wrappers for the transformations in `flax.core.lift`. The core lifting APIs operate on functions while
+the linen wrappers can transform either a `Module` class or a `Module` method.
+
+Lifting is thus implemented independently from the linen API. This seperation of concern simplifies the implementation while also allowing other Module APIs that chooce a different syntatic sugar for defining Modules to reuse the very generic and flexible `Scope `+ lift APIs for handling state and RNGs.
```
8bitmp3: Suggestion:

- Nit picking
- Typo fix ("separation", "syntactic", "choose")
- `Module`, `lift` and `scope` (of `flax.core`) for consistency

```suggestion
Thus, lifting is implemented independently from the Linen API. This separation of concern simplifies the implementation, while also allowing other `Module` APIs that choose a different syntactic sugar for defining `Module`s to reuse the very generic and flexible `scope `and `lift` APIs for handling state and RNGs. 
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+A number of decision decisions had to be made in order to define `pack`. The implementation
+of `pack` controls how variables and rngs are lifted and how fine-grained the user control is.
+It must also decide whether lifting decisions are made during variable definition or at the transformation.
+
+
+### Lifting granularity
+
+
+With the linen API, users can define arbitrary variable collections and PRNG sequences.
```
8bitmp3, line 138: Suggestion ("Linen"):

```suggestion
With the Linen API, users can define arbitrary variable collections and PRNG sequences.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+## Lifting
+
+
+Flax provides an alternative for manual functionalization which we call lifted transformation.
+Lifted transformations are defined in `flax.core.lift`.
+All the lifted JAX transformations are defined with a single generic lifting API called `pack`.
+
+A number of decision decisions had to be made in order to define `pack`. The implementation
+of `pack` controls how variables and rngs are lifted and how fine-grained the user control is.
+It must also decide whether lifting decisions are made during variable definition or at the transformation.
```
8bitmp3, line 132: Maybe: "at the transformation" < "during the transformation"

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+We use transformation level definitions of lifting behavior.
+The reason for this choice is that there are many different transformations with various behaviors.
+For example: `vmap` has broadcasted and vectorized arguments, while `scan` has scan, carry, and broadcast arguments.
+A variable would have to define its behavior for all these transformations otherwise a `Module` would not be compatible with
+these tranformations. Or alternatively, we would have to make default decisions for how transformations are handled.
+However, this could lead to silent bugs becuase the behavior might not actually be valid given the users intent. 
+
+The lift package also provides a general purpose `transform` which allows an arbitrary function to transforms a variable collection.
+This can be used to for example tie the weights in a tied autoencoder by transposing the weights.
+It is unclear whether a similair general purpose transform could be defined if lifting decisions were made at variable definition.
```
8bitmp3: Suggestion:

- (forgive me if I'm wrong) "during the ..definition" > "at ... definition"
- "similar" (typo fix)

```suggestion
It is unclear whether a similar general purpose transform could be defined if lifting decisions were made during the variable definition.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+### Linen
+
+The lifting module does not know about linen Modules. Instead it operates directly on instances of `flax.core.Scope`.
```
8bitmp3, line 170: Suggestion:

- "Linen"
- `Module` for consistency

```suggestion
The lifting module does not know about Linen `Module`s. Instead, it operates directly on instances of `flax.core.Scope`.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+    In this case we must first find the set of root scopes.
+    A scope is a root if none of its ancestors is in the set of scopes that is too be lifted.
+
+    By only lifting roots we avoid lifting the same variables twice.
+
+    For non-root scopes we store a reference to its ancestor scope and a path such that we can later reconstruct it (stage 4).
+
+2. *Filter stage*
+
+    Variables & PRNG sequences are split up into groups. This way `fn` can lift each group into the seperation seperatly.
```
8bitmp3: Nit:

```suggestion
    Variables and PRNG sequences are split up into groups. This way `fn` can lift each group into the seperation seperatly.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+### Implementation
+
+The `pack(fn, in_vars, out_vars, rngs)` API goes through the following stages:
+
+
+1. *Scope deduplication*
+
+    This stage is only relevant if multiple Scopes are lifted together.
+    In this case we must first find the set of root scopes.
+    A scope is a root if none of its ancestors is in the set of scopes that is too be lifted.
```
8bitmp3, line 191: Suggestion: some grammar stuff (correct me if I'm wrong)

e.g.

```suggestion
    A scope is a root if none of its ancestors are in the set of scopes that need to be lifted.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+    By only lifting roots we avoid lifting the same variables twice.
+
+    For non-root scopes we store a reference to its ancestor scope and a path such that we can later reconstruct it (stage 4).
+
+2. *Filter stage*
+
+    Variables & PRNG sequences are split up into groups. This way `fn` can lift each group into the seperation seperatly.
+    A group is defined by a filter which can be specified as:
```
8bitmp3, line 200: Suggestion (nit):

```suggestion
    A group is defined by a filter, which can be specified as:
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+    - a list of collections/prng names
+    - `True` (match everything)
+    - `False` (match nothing)
+    - `DenyList(filter)` (match everything but the specified filter).
+
+    A collection or PRNG sequence can only be put into a single group. If a collection matches multiple filters it will be put into
```
8bitmp3, line 206: Nit:

```suggestion
    A collection or a PRNG sequence can only be put into a single group. If a collection matches multiple filters, it will be put into
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+    the first group with a matching filter.
+    For example, `in_vars = (["params"], True)` will cause the "params" collection to be put in the first group and all other collection to be put in the second group.
```
8bitmp3, line 208: Suggestion:

```suggestion
    For example, `in_vars = (["params"], True)` will cause the `"params"` collection to be put in the first group and all other collections — in the second group.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+    For each PRNG sequence that is matched we seed a new PRNG sequence by calling `make_rng`.
+    This avoids the need to update the PRNG state after the lifted tranformation finishes.
```
8bitmp3: Suggestion:

- "the ... transformation is complete" instead of "finishes"
- Fix typo ("transformation")

```suggestion
    This avoids the need to update the PRNG state after the lifted transformation is complete.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+3. *Transform specific lifting*
+
+    `fn` is called with the variable and PRNG groups.
+    JAX transform have varying signatures and lifting options. Arguably the cleanest example is `vmap`.
```
8bitmp3, line 216: Nit ("JAX transforms" should be plural here IMO):

```suggestion
    JAX transforms have varying signatures and lifting options. Arguably, the cleanest example is `vmap`.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+    In the case of vmap the function arguments, rngs and variable collections are passed into a `jax.vmap` wrappped function.
+
+4. *Scope reconstruction*
+
+    Now that the variables and rngs are lifted inside the transformation, we want to recreate the lifted scopes. Pack calls
+    `fn` with a `scope_fn` that takes the lifted variables and rngs and returns the reconstructed scopes with the lifted variables and rng sequences.
```
8bitmp3, line 222: Suggestion ("RNGs")

```suggestion
    `fn` with a `scope_fn` that takes the lifted variables and RNGs and returns the reconstructed scopes with the lifted variables and RNG sequences.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+| pmap | ❌ |  |
+| grad | ❌ |  |
+| xmap | ❌ |  |
+| custom_vjp | ❌ |  |
+| custom_jvp | ❌ |  |
+
+
+### Linen examples
+
+Going back to our original example we can now use the lifted vmap:
```
8bitmp3, line 280: Suggestion (nit):

```suggestion
Going back to our original example, we can now use the lifted `vmap`:
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+```
```
8bitmp3, line 282: Nit:

```suggestion
```python
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+        Dense_1: {
+            bias: (3, 1),
+            kernel: (3, 4, 1),
+        },
+    },
+}
+"""
+```
+
+Here we use `variable_axes={'params': 0}` to indicate that paramaters are vectorized rather than shared and `split_rngs={'params': True}` means each set of parameters is initialized independently.
```
8bitmp3: Nit (+ typo fix):

```suggestion
Here, we use `variable_axes={'params': 0}` to indicate that parameters are vectorized rather than shared, and `split_rngs={'params': True}` means each set of parameters is initialized independently.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+class LinenStatefulVmap(nn.Module):
+  @nn.compact
+  def __call__(self, xs):
+    VmapMLP = nn.vmap(StatefulMLP, variable_axes={'params': 0, 'batch_stats': 0}, split_rngs={'params': True}, in_axes=0)
+    return VmapMLP(name='mlp')(xs)
+variables = LinenStatefulVmap().init(random.PRNGKey(0), xs)
+```
+
+All we had to add to `nn.vmap` is `'batch_stats': 0` indicating the batch stats are vectorized rather than shared along the first axis.
```
8bitmp3, line 328: Nit:

```suggestion
All we had to add to `nn.vmap` is `'batch_stats': 0`, indicating that the batch stats are vectorized rather than shared along the first axis.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+    After we have used the lifted scopes we have to retrieve the updated variables (PRNG sequences can simply be discarded).
+    pack passes the `repack_fn` to support this.
+    This stage is similair to stage 2 except that we only lift variables and immutable variables are ignored.
+    Immutable variables cannot be updated so there is no reason to return them from the lifted transformation. 
+
+6. *Commit stage*
+
+    `pack` expects `fn` to return a pair where the first item will simply be returned from pack and the second item should be the repacked variables.
+    The updated variables are stored in the original/unlifted scopes such that the mutation that happened inside the transformation survive beyond the transformation boundary.
```
8bitmp3: Suggestion:
- It "survives" or they "survive"
- I think the present tense would be OK here ("happens")

e.g.

```suggestion
    The updated variables are stored in the original/unlifted scopes, such that the mutation that happens inside the transformation survives beyond the transformation boundary.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+
+### Using pack example
+
+
+A minimal example of using pack:
+
+```
```
8bitmp3, line 242: Nit:

```suggestion
```python
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+    Now that the variables and rngs are lifted inside the transformation, we want to recreate the lifted scopes. Pack calls
+    `fn` with a `scope_fn` that takes the lifted variables and rngs and returns the reconstructed scopes with the lifted variables and rng sequences.
+
+5. *Repack stage*
+
+    After we have used the lifted scopes we have to retrieve the updated variables (PRNG sequences can simply be discarded).
+    pack passes the `repack_fn` to support this.
+    This stage is similair to stage 2 except that we only lift variables and immutable variables are ignored.
+    Immutable variables cannot be updated so there is no reason to return them from the lifted transformation.
```
8bitmp3, line 229: Nit:

```suggestion
    Immutable variables cannot be updated, so there is no reason to return them from the lifted transformation. 
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+3. *Transform specific lifting*
+
+    `fn` is called with the variable and PRNG groups.
+    JAX transform have varying signatures and lifting options. Arguably the cleanest example is `vmap`.
+    In the case of vmap the function arguments, rngs and variable collections are passed into a `jax.vmap` wrappped function.
+
+4. *Scope reconstruction*
+
+    Now that the variables and rngs are lifted inside the transformation, we want to recreate the lifted scopes. Pack calls
```
8bitmp3, line 221: Suggestion ("RNGs"):

```suggestion
    Now that the variables and RNGs are lifted inside the transformation, we want to recreate the lifted scopes. Pack calls
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+    By only lifting roots we avoid lifting the same variables twice.
+
+    For non-root scopes we store a reference to its ancestor scope and a path such that we can later reconstruct it (stage 4).
+
+2. *Filter stage*
+
+    Variables & PRNG sequences are split up into groups. This way `fn` can lift each group into the seperation seperatly.
+    A group is defined by a filter which can be specified as:
+    - a list of collections/prng names
```
8bitmp3, line 201: Suggestion (capitalize the acronym (PRNG))

```suggestion
    - a list of collections/PRNG names
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+
+Lifting behavior could be defined either at the transformation level or during variable definition.
+We use transformation level definitions of lifting behavior.
+The reason for this choice is that there are many different transformations with various behaviors.
+For example: `vmap` has broadcasted and vectorized arguments, while `scan` has scan, carry, and broadcast arguments.
+A variable would have to define its behavior for all these transformations otherwise a `Module` would not be compatible with
+these tranformations. Or alternatively, we would have to make default decisions for how transformations are handled.
+However, this could lead to silent bugs becuase the behavior might not actually be valid given the users intent. 
+
+The lift package also provides a general purpose `transform` which allows an arbitrary function to transforms a variable collection.
```
8bitmp3, line 163: Suggestion:

- "...function to transform a variable collection" (instead of "transforms")

```suggestion
The lift package also provides a general purpose `transform`, which allows an arbitrary function to transform a variable collection.
```

docs/design_notes/lift.md
```diff
@@ -0,0 +1,338 @@
...
+6. *Commit stage*
+
+    `pack` expects `fn` to return a pair where the first item will simply be returned from pack and the second item should be the repacked variables.
+    The updated variables are stored in the original/unlifted scopes such that the mutation that happened inside the transformation survive beyond the transformation boundary.
+
+
+### Using pack example
+
+
+A minimal example of using pack:
```
8bitmp3, line 240: Suggestion: maybe this should be `pack` (with backticks):

```suggestion
A minimal example of using `pack`:
```

docs/design_notes/lift.md
jheek: What is a fox example ;)

why can humans only detect spelling errors in text that they didn't generate? Clearly we are overfitting on our self-generated data

8bitmp3:
@jheek Sorry for too many comments, hope some of them help! This is a great doc, I'm learning so much 👍 Thank you

docs/design_notes/lift.md
```diff
@@ -0,0 +1,341 @@
...
+  @nn.compact
+  def __call__(self, xs):
+    h = nn.Dense(4, name='hidden')(xs)
+    h = nn.relu(h)
+    return nn.Dense(1, name='out')(h)
+```
+
+Now what if I want to have separate MLP parameters for each item in `xs`?
+If this where "vanilla JAX" we could imagine writing something like `jax.vmap(apply_mlp)(mlp_params, xs)`.
+But doing something like this in linen will actually fail:
```
8bitmp3, line 57: Suggestion (nit):

I'm not 100% sure whether we capitalize the first letter.

```suggestion
But doing something like this in Linen will actually fail:
```

jheek:
> @jheek Sorry for too many comments, hope some of them help! This is a great doc, I'm learning so much 👍 Thank you

Your help is much appreciated! GitHubs review UI is a bit verbose for reviewing text but that's what it is.

docs/design_notes/lift.md
jheek: Here we refer to an init/apply function not a particular one or the method that creates them.  So I think this should be text not code

jheek:
> All (the pull request submitter and all commit authors) CLAs are signed, **but** one or more commits were authored or co-authored by someone other than the pull request submitter.
> 
> We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that by leaving a comment that contains only `@googlebot I consent.` in this pull request.

@8bitmp3 could you add your consent? This is for the applied suggestions

8bitmp3:
@googlebot I consent.

docs/design_notes/lift.md
8bitmp3: Understood, thanks for clarifying this @jheek

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+But pure functions offer another big advantage to JAX: specifically, they enable functional transformations.
+For example `jax.vmap(f)` will vectorize a function `f`.
+Because `f` cannot have side effects the vectorized/parallel version of `f` is well-defined. To see why we need this restriction, consider what happens if `f` would increment a counter or draw a random number.
+Would `f` draw the same or a different random number for each item in the vector?
+Would each item in the batch have its own counter or is the counter shared among the items?
+And in what order is the counter incremented if `f` is computed in parallel?
+The answer to all these questions is "it depends".
+The behavior is ambiguous and the functional constraint elegantly avoids this problem.
+
+Flax introduces a safe way to have limited randomness and stateful variables in a JAX compatible form.
```
andsteing, line 26: minor nit: "JAX-compatible"

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+
+## Functionalization
+
+Before we jump into the details let's consider a simple example where we would like to use `vmap` inside a `Module`.
+
+First, we define a simple MLP without any transformations:
+
+```python
+import jax
+from jax import random, NumPy as jnp
```
andsteing, line 44: it's `numpy` (lowercase)
anyways, usually we would say `import jax.numpy as jnp` on a separate line...

(also if you use `jax.random` instead of `random` then I don't think any import is needed in the example since they're pretty obvious)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+
+Now what if we want to have separate MLP parameters for each item in `xs`?
+If this were "vanilla JAX" we could imagine writing something like `jax.vmap(apply_mlp)(mlp_params, xs)`.
+But doing something like this in Linen will actually fail:
+
+```python
+class NaiveVmapMLP(nn.Module):
+  @nn.compact
+  def __call__(self, xs):
+    mlps = MLP()
```
andsteing, line 63: maybe `mlps = [MLP() for _ in xs.shape[0]]` ?
(depends what you want to highlight - but it looks weird having plural in the var name with a singular class name)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+This avoids reserving a name for the submodule inside the current module.
+Although not strictly necessary, this also ensures we cannot accidentally use the MLP instance in a stateful way and we are forced to use it through either `.init` or `.apply`.
+
+This example is still relatively concise but it already takes a few extra "bookkeeping" statements to make it work.
+However, this implementation has a number of limitations:
+1. During initialization, we call the submodule twice through `init_fn` and `apply_fn`. If the submodule used the same trick to do
+   functional transformation we will end up executing a lot of code as the number of module calls grows like 2^d where d is the number of
+   nested function transformations.
+2. The implementation assumes the submodule only requires the parameter RNG sequence.
+3. The implementation assumes we init parameters once and we don't create any state variables either during init or apply.
```
andsteing, line 114: I'm not sure I understand what you mean exactly by "assumes we init parameters once" ... when/why would we init parameters multiple times?

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+The lift package also provides a general purpose `transform`, which allows an arbitrary function to transform a variable collection.
+For example, this can be used to tie the weights in a tied auto-encoder by transposing the weights.
+It is unclear whether a similar general purpose transform could be defined if lifting decisions were made at variable definition.
+
+
+### Linen
+
+The lifting module does not know about the Linen `Module` API. Instead it operates directly on instances of `flax.core.Scope`.
+A `Scope` instance contains the variables and PRNG sequences of a `Module`.
+Each module has a `Scope` instance linked if it is bound to a parent or it was created using `init` or `apply`.
```
andsteing, line 172: This sentence is a bit hard to parse:
- What does "has a scope instance linked" mean exactly? (e.g. what's the alternative)
- It's not clear if it's "(has a scope instance linked if it is bound to a parent) or (it was created using init/apply)" vs. "has a scope instance linked (if it is bound to a parent) or (it was created using init/apply)"

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+2. *Filter stage*
+
+    Variables and PRNG sequences are split up into groups. This way `fn` can lift each group into the transformation separately.
+    A group is defined by a filter specified as:
+    - a list of collections/prng names
+    - `True` (match everything)
+    - `False` (match nothing)
+    - `DenyList(filter)` (match everything but the specified filter).
+
+    A collection or PRNG sequence can only be put into a single group. If a collection matches multiple filters, it will be put into the first group with a matching filter.
```
andsteing, line 206: (I think at this point it is not very clear what it means for a prng/variable to be lifted vs. not to be lifted.)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+    For example, `in_vars = (["params"], True)` will cause the "params" collection to be put in the first group and all other collection to be put in the second group.
+
+    For each PRNG sequence that is matched we seed a new PRNG sequence by calling `make_rng`.
+    This avoids the need to update the PRNG state after the lifted transformation is complete.
+
+3. *Transform specific lifting*
```
andsteing, line 212: nit: transform-specific

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+### Using pack example
+
+
+A minimal example of using `pack`:
+
+```python
+from flax.core import lift
+from flax.core import Scope, init, apply, nn as core_nn
+
+def lift_id(fn, variables=True, rngs=True)
```
andsteing, line 245: While this example is very concise, it might be interesting to expand it a little bit to an example where variables/rngs are selected by filter (i.e. some are lifted and others not).

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+x = jnp.ones((3, 2))
+y, params = init(lift_id(core_nn.dense))(random.PRNGKey(0), x, 4)
+```
+
+NOTE that most users should not need to interact with `pack` directly.
+Please open a GitHub issue when you find a use case that is not supported yet by the existing lifted transformations.
+
+### Supported transformations
+
+| Jax Transform | Supported in Linen? | Comments |
```
andsteing, line 264: Maybe a link to the source would be practical here (or below).

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
...
+```
+
+Here we use `variable_axes={'params': 0}` to indicate that parameters are vectorized rather than shared and `split_rngs={'params': True}` means each set of parameters is initialized independently.
+
+We can also extend the example with some inner state by adding a `BatchNorm` layer:
+
+```python
+class StatefulMLP(nn.Module):
+  @nn.compact
+  def __call__(self, x):
```
andsteing, line 313: Wouldn't we always to pass in `train` when using `BatchNorm`?

andsteing (commented):
This is a great design note!

Thanks for writing this up in such a nice text...

docs/design_notes/lift.md
```diff
@@ -0,0 +1,340 @@
+# Lifted transformation
+
+⚠️ Advanced topic ⚠️
+
+This design note explains the underlying implementation of `flax.linen.transform`, which enables JAX transformations inside `Module`.
```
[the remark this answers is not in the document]
jheek: I'm always feeling a bit uncomfortable with plural of `ClassName`. I guess this is the least evil though

docs/design_notes/lift.md
jheek: I was thinking mlps because you vmap the MLP which is equivalent to applying a bunch of mlps in parallel. But maybe that's a bit confusing. I'll remove the plural here

docs/design_notes/lift.md
jheek: clarified

docs/design_notes/lift.md
jheek: done

docs/design_notes/lift.md
jheek: done

docs/design_notes/lift.md
jheek: add clarification on unlifted prng/variables

docs/design_notes/lift.md
jheek: rephrased

docs/design_notes/lift.md
jheek: oops

docs/design_notes/lift.md
```diff
@@ -167,9 +167,11 @@ It is unclear whether a similar general purpose transform could be defined if li
 
 ### Linen
 
-The lifting module does not know about the Linen `Module` API. Instead it operates directly on instances of `flax.core.Scope`.
+The lifting module does not know about the Linen `Module` API.
+Instead it operates directly on instances of `flax.core.Scope`.
 A `Scope` instance contains the variables and PRNG sequences of a `Module`.
-Each module has a `Scope` instance linked if it is bound to a parent or it was created using `init` or `apply`.
+Each `Module` instance has a `Scope` instance in the `.scope` field if it has a parent or it was created using `init` or `apply`.
+Typically, the top-level `Module` instance on which you call `init` or `apply` is the only `Module` instance that does not have a `Scope` bound to it.
```
8bitmp3, line 174: Suggestion (nit):

- Because there are several moving parts and new material for new users, I suggest the following to help the audience digest this:

```suggestion
Typically, the top-level `Module` instance—on which you call `init` or `apply`—is the only `Module` instance that does not have a `Scope` bound to it.
```

LMKWYT 👍

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+| pmap | ❌ |  |
+| grad | ❌ |  |
+| xmap | ❌ |  |
+| custom_vjp | ❌ |  |
+| custom_jvp | ❌ |  |
+
+References:
+- [Linen transforms documentation](https://flax.readthedocs.io/en/latest/flax.linen.html#module-flax.linen.transforms).
+- [Linen transforms source code](https://github.com/google/flax/blob/master/flax/linen/transforms.py)
+- [Core lifting source code](https://github.com/google/flax/blob/master/flax/core/lift.py)
```
andsteing, line 290: Should we link back from the source to RTD?

(`lift.pack` has very terse documentation currently :-)

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+    target = jax.tree_map(trans, target)
+    variable_groups = (target, rest)
+    scope = scope_fn(variable_groups, rng_groups)
+    y = fn(scope, *args)
+    out_variables = repack_fn(scope)
+    return y, out_variables
+  return lift.pack(wrapper, (target, variables), (variables,), (rngs,))
+
+x = jnp.ones((3, 2))
+y, params = init(lift_id(core_nn.dense))(random.PRNGKey(0), x, 4)
```
andsteing, line 267: `lift_transpose`

andsteing (approved):

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
+# Lifted Transformations
+
+⚠️ Advanced topic ⚠️
+
+This design note explains the underlying implementation of `flax.linen.transform`, which enables JAX transformations inside `Module`s.
+
+
+## Introduction
+
+JAX uses a functional API meaning that it only guarantees correct behavior when using functions without side effects.
```
gnecula, line 10: You may want to cite https://jax.readthedocs.io/en/latest/jax-101/01-jax-basics.html#differences-from-numpy

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+  @nn.compact
+  def __call__(self, xs):
+    mlp = MLP()
+    return jax.vmap(lambda mlp, x: mlp(x))(mlp, xs)  # fails
+```
+
+JAX will raise an error when `vmap` is used on `mlp` because it's not a JAX array or a simple container of arrays.
+We can not really blame JAX for refusing to perform this under-specified job.
+After all, it's not even clear what should happen here.
+The parameters inside the MLP are not even initialized yet and we will need a separate PRNG key for each group of parameters.
```
gnecula, line 70: I think it may be confusing how the PRNG getting into the picture here. I think that it is doing so indirectly, because when this model will later be initialized it will be passed a PRNG, and we would like to specify that we want to pass a different key to each element of the input vector.

docs/design_notes/lift.md
jheek: Probably. I think the core docs need some more thought and love. I'll do that in a seperate PR

docs/design_notes/lift.md
jheek: Agree

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+
+We can fix this problem by first turning `MLP` into a pure init and apply function.
+Afterwards, we use the `param` method to store the parameters:
+
+```python
+class ManualVmap(nn.Module):
```
gnecula, line 76: I suggest adding `MLP` as a suffix to all models, as done until this point.

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+    For non-root scopes we store a reference to its ancestor scope and a path such that we can later reconstruct it (stage 4).
+
+2. *Filter stage*
+
+    Variables and PRNG sequences are split up into groups. This way `fn` can lift each group into the transformation separately.
+    A group is defined by a filter specified as:
+    - a list of collections/prng names
+    - `True` (match everything)
+    - `False` (match nothing)
+    - `DenyList(filter)` (match everything but the specified filter).
```
gnecula, line 206: Do you meant to say `DenyList(list_of_collections)` instead of `DenyList(filter)`?

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+
+    A collection or PRNG sequence can only be put into a single group. If a collection matches multiple filters, it will be put into the first group with a matching filter.
+    If a collection or PRNG sequence does not match any filter it will not be lifted.
+    This means that it cannot be used inside the transformation and and attempting to do this will cause an error to be raised.
```
gnecula, line 210: "and and"

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+      if x.ndim == 2:
+        return x.T
+      return x
+    target = jax.tree_map(trans, target)
+    variable_groups = (target, rest)
+    scope = scope_fn(variable_groups, rng_groups)
+    y = fn(scope, *args)
+    out_variables = repack_fn(scope)
+    return y, out_variables
+  return lift.pack(wrapper, (target, variables), (variables,), (rngs,))
```
gnecula, line 264: I would suggest a comment here with the signature of `lift.pack(fn, in_vars, out_vars, rngs)`

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+
+A minimal example of using `pack` to transpose each matrix in a variable collection:
+
+```python
+from flax.core import lift
+from flax.core import Scope, init, apply, nn as core_nn
+
+def lift_transpose(fn, target='params', variables=True, rngs=True)
+  # by default we transpose 'params' and simply pass through all other variables.
+  def wrapper(scope_fn, repack_fn, variable_groups, rng_groups, *args):
```
gnecula, line 251: I would suggest adding some typing annotations here

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+Here we use `variable_axes={'params': 0}` to indicate that parameters are vectorized rather than shared and `split_rngs={'params': True}` means each set of parameters is initialized independently.
+
+We can also extend the example with some inner state by adding a `BatchNorm` layer:
+
+```python
+class StatefulMLP(nn.Module):
+  @nn.compact
+  def __call__(self, x, *, train):
+    h = nn.Dense(4, name='hidden')(x)
+    h = nn.BatchNorm(axis_name="batch")(h, use_running_average=not train)
```
gnecula, line 330: Single quotes to match the rest of the strings?

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+    h = nn.relu(h)
+    return nn.Dense(1, name='out')(h)
+
+class LinenStatefulVmap(nn.Module):
+  @nn.compact
+  def __call__(self, xs, *, train):
+    VmapMLP = nn.vmap(StatefulMLP, variable_axes={'params': 0, 'batch_stats': 0}, split_rngs={'params': True}, in_axes=0)
```
gnecula, line 337: I am guessing that `batch_stats` is the name of the collection that `nn.BatchNorm` is using?

jheek: yes

docs/design_notes/lift.md
```diff
@@ -0,0 +1,355 @@
...
+    return VmapMLP(name='mlp')(xs, train=train)
+variables = LinenStatefulVmap().init(random.PRNGKey(0), xs)
+```
+
+All we had to add to `nn.vmap` is `'batch_stats': 0`, indicating that the batch stats are vectorized rather than shared along the first axis.
+
+
+## Alternatives
```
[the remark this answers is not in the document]
jheek: Let's consider this for a follow up PR

docs/design_notes/lift.md
jheek: DenyList can compose with an arbitrary filter (we need this internally).

gnecula: The only compositions that are possible are: DenyList(list), DenyList(True), DenyList(False), DenyList(DenyList(...)), right? Of these only the first one seems to make sense. Also, since the name was `DenyList` and `DenyFilter`, I assume you meant to use it with a list.

jheek: Added some clarification

docs/design_notes/lift.md
jheek: Added some more clarificaiton here

PR #1268, opened 2021-04-23
Increase shuffle buffer size

juliakreutzer:
# What does this PR do?

A shuffle size smaller than the input data might have unexpected effects on the training process and thereby the final model quality, especially for heterogeneous and sharded datasets. Since most of the WMT datasets contain several millions of examples, it should be safer to set it to a higher number like this.

Fixes # (issue)

## Checklist
- [ x] This PR fixes a minor issue (e.g.: typo or small bug) or improves the docs (you can dismiss the other
      checks if that's the case).
      link).


jheek:
Shuffling isn't free in `tf.data`. It would be nice if we can see the effects this has on the eval metrics & training speed before merging this.

juliakreutzer:
I'm aware shuffling isn't free, but setting this default so low is risky for users who might not be aware this could hurt model quality. If not changing it globally, I'd suggest making it configurable at least so it can be documented with other hyperparameters.

Files are currently not shuffled when read in, see [here](https://github.com/google/flax/blob/master/examples/wmt/input_pipeline.py#L64). So if there is any ordering in the original data, there's a risk of maintaining that during training.

I'll run a comparison with the default settings and report back.

jheek:
>> Files are currently not shuffled when read in, see here. So if there is any ordering in the original data, there's a risk of maintaining that during training.

@levskaya this seems like a bug to me? I think file shuffling is supposed to be enabled?

levskaya:
> @levskaya this seems like a bug to me? I think file shuffling is supposed to be enabled?
ha - yeah, at a minimum file shuffling should be on by default  -- even w. a fixed seed in case there's a natural bias

jheek (changes requested):
Could you try enabling file shuffling (for training) instead of increasing the buffer size?

jheek:
Closing due to lack of activity

PR #1275, opened 2021-04-26
Add proper cache behavior for lift.jit

jheek:
Fixes cache misses in lift.jit

NOTE: how to cache jit in linen transforms still needs additional changes

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -731,7 +731,8 @@ def jit(fn: Callable[..., Any],
   # Close over scope_fn & repack_fn to avoid recompilation
   # this is impure but we use the fingerprint arg to differentiate between cases
   # where scope_fn or repack_fn actually produce non-identical results.
-  scope_fn, repack_fn = None, None
+  scope_fn = None  # type: Callable
+  repack_fn = None  # type: Callable
   @functools.partial(jax.jit,
                      static_argnums=static_argnums,
                      donate_argnums=donate_argnums,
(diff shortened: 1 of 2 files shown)


flax/core/lift.py
```diff
@@ -715,21 +723,39 @@ def jit(fn: Callable[..., Any],
...
+  # where scope_fn or repack_fn actually produce non-identical results.
+  scope_fn = None  # type: Callable
+  repack_fn = None  # type: Callable
+  @functools.partial(jax.jit,
+                     static_argnums=static_argnums,
+                     donate_argnums=donate_argnums,
+                     device=device, backend=backend)
+  @functools.wraps(fn)
+  def jitted(fingerprint, variable_groups, rng_groups, *args):
+    # fingerprint is only used to differentiate the cache signature for cases
```
levskaya, line 742: would it be useful to mark scope_fn, repack_fn 'nonlocal' here as well just to emphasize what we're doing for future readers?

levskaya (approved):
A clever hack!  I guess we could do this to make a useful version of my old pmap hack as well.

flax/core/lift.py
jheek: > A clever hack! what exactly does your note in the comment above mean? is this unsound when nested inside other transforms?

Just like jit the cache doesn't trigger if you call lift.jit multiple times which is what transforms.py ends up doing.

jheek:
> A clever hack! what exactly does your note in the comment above mean? is this unsound when nested inside other transforms?

Just like jit the cache doesn't trigger if you call lift.jit multiple times which is what transforms.py ends up doing.

PR #1276, opened 2021-04-26
Uses clu.metric_writer & adds config.log_every_step=100

andsteing:
- `clu.metric_writer` writes transparently both to TF summary files and to logging
- It's often useful to get training metrics before an epoch is fully finished; then it's also fine to report average values since the last emission of these metrics instead of logging the results of every batch (which also produces quite large TF summary files)

For comparison of metrics (eval metrics are identical, train metrics are much sparser):

- Before: https://tensorboard.dev/experiment/iJzNKovmS0q6k5t6k5wvOw/#scalars&_smoothingWeight=0&regexInput=v100_x8$
- This PR: https://tensorboard.dev/experiment/OTRYahNGS1m27exR3ryZvA/


jheek (approved):
LGTM! Sorry for the delay!

examples/imagenet/configs/default.py
```diff
@@ -11,20 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-# Copyright 2020 The Flax Authors.
```
jheek, line 15: Was this removed accidentally?

andsteing: The `# Copyright 2021` is enough, no?

andsteing: (I just noted this passing by and thought I would include it in the same CL.)

jheek: Oh I missed the duplicate. You are right

PR #1284, opened 2021-04-29
Fix intermediates during init

jheek:
- Add  mutable arg to Module.init
- Use DenyList("intermediates") as default mutable filter during init.
When calling `init` the 'intermediates' collection is no longer mutable
Therefore, intermediates will no longer be returned from init by default.

### DIFF
--- a/tests/linen/module_test.py
+++ b/tests/linen/module_test.py
@@ -1197,7 +1197,9 @@ class ModuleTest(absltest.TestCase):
         self.sow('intermediates', 'h', x, **sow_args)
         self.sow('intermediates', 'h', 2 * x, **sow_args)
         return 3 * x
-
+    variables = Foo().init(random.PRNGKey(0), 1)
+    # during init we should not collect intermediates by default
+    self.assertTrue('intermediates' not in variables)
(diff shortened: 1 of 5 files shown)


tests/linen/module_test.py
```diff
@@ -1197,7 +1197,9 @@ def __call__(self, x, **sow_args):
         self.sow('intermediates', 'h', x, **sow_args)
         self.sow('intermediates', 'h', 2 * x, **sow_args)
         return 3 * x
-
+    variables = Foo().init(random.PRNGKey(0), 1)
+    # during init we should not collect intermediates by default
+    self.assertTrue('intermediates' not in variables)
```
avital, line 1202: probably we should also have a test that if you pass `mutable=True` to `init` then indeed you do have access to the intermediates?

avital (commented):
Code LGTM. I'm a bit uneasy about hard-coding "intermediate", though it's just a default so it isn't too bad. But I propose we get a broader review of this.

tests/linen/module_test.py
jheek: done

avital (approved):

PR #1288, opened 2021-04-29
Deprecate jax utils.replicate

### LINKED ISSUE #1161: Deprecate flax.jax_utils.replicate
`flax.jax_utils.replicate` should no longer be necessary and superceded by `jax.device_put_replicated`. We need to update our tests and use-cases accordingly, and add a clear and useful deprecation warning (or, perhaps even an error simply asking users to switch to the JAX API -- since as far as I can tell the function signature is identical)

yashs97:
# What does this PR do?

Deprecates `jax_utils.replicate` as mentioned in #1161 

Fixes #1161 

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/docs/howtos/ensembling.rst
+++ b/docs/howtos/ensembling.rst
@@ -230,7 +230,7 @@ Next we transform the ``train_epoch`` function.
     batch_metrics = []
     for perm in perms:
       batch = {k: v[perm, ...] for k, v in train_ds.items()}
-      batch = jax_utils.replicate(batch) #!
+      batch = jax.device_put_replicated(batch) #!
       optimizer, metrics = train_step(optimizer, batch)
       batch_metrics.append(metrics)
 
@@ -281,7 +281,7 @@ than the train dataset so we can do this for the entire dataset directly.
                 epoch, loss, accuracy * 100)
   ---
   train_ds, test_ds = get_datasets()
-  test_ds = jax_utils.replicate(test_ds) #!
+  test_ds = jax.device_put_replicated(test_ds) #!
   
   rng, init_rng = random.split(random.PRNGKey(0))
   params = get_initial_params(random.split(rng, #!
--- a/docs/howtos/ensembling.rst
+++ b/docs/howtos/ensembling.rst
@@ -230,7 +230,7 @@ Next we transform the ``train_epoch`` function.
     batch_metrics = []
     for perm in perms:
       batch = {k: v[perm, ...] for k, v in train_ds.items()}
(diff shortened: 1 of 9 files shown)


yashs97:
Am I supposed to add tests for this PR(only adds a deprecation warning)?

flax/jax_utils.py
```diff
@@ -72,6 +72,8 @@ def replicate(tree, devices=None):
   Returns:
     A new pytree containing the replicated arrays.
   """
+  warnings.warn('use jax.device_put_replicated',
+                DeprecationWarning)
```
marcvanzee, line 76:
```suggestion
  warnings.warn('replicate() will be removed soon.'
                ' Use jax.device_put_replicated instead.',
                DeprecationWarning)
```

flax/jax_utils.py
```diff
@@ -72,6 +72,8 @@ def replicate(tree, devices=None):
   Returns:
     A new pytree containing the replicated arrays.
   """
+  warnings.warn('use jax.device_put_replicated',
```
marcvanzee, line 75: Could you add a deprecation warning to the docstring as well? Similar to e.g. here: https://github.com/google/flax/blob/master/flax/nn/attention.py#L48

docs/howtos/ensembling.rst
```diff
@@ -230,7 +230,7 @@ Next we transform the ``train_epoch`` function.
     batch_metrics = []
     for perm in perms:
       batch = {k: v[perm, ...] for k, v in train_ds.items()}
-      batch = jax_utils.replicate(batch) #!
+      batch = jax.device_put_replicated(batch, jax.local_devices()) #!
```
marcvanzee, line 233: It is a bit annoying that we now have to add `jax.local_devices()` everywhere... I will check with the JAX team whether it makes sense to set `devices` to `jax.local_devices()` by default we so we can simply omit it everywhere.

marcvanzee (changes requested):

docs/howtos/ensembling.rst
yashs97: Even if the JAX team does agree to set `devices` to `jax.local_devices()` by default, wouldn't that be a different PR because from what I can tell `jax.local_devices()` is used everywhere?

marcvanzee:
Note: I have unassigned myself from this PR since it will take a bit to land this. We need to update our internal users as well, and we already have a few ongoing efforts for this, so I don't think we will be able to finish this PR in the short term.

PR #1291, opened 2021-04-30
Expose `dot_product_attention_weights`

avital:
This allows users to access the attention weights if they want to
explicitly observe them, and use einsum directly to then compute
the full result of the attention operation.

### DIFF
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -69,7 +69,11 @@ def dot_product_attention_weights(query: Array,
   Returns:
     Output of shape `[batch..., num_heads, q_length, kv_length]`.
   """
-  assert key.ndim == query.ndim, 'q, k must have same rank.'
+  assert query.ndim == key.ndim, 'q, k must have same rank.'
+  assert query.shape[:-3] == key.shape[:-3], (
+      'q, k batch dims must match.')
+  assert query.shape[-2] == key.shape[-2], (
+      'q, k num_heads must match.')
   assert query.shape[-1] == key.shape[-1], 'q, k depths must match.'
 
   # calculate attention matrix
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -69,7 +69,7 @@ def dot_product_attention_weights(query: Array,
   Returns:
     Output of shape `[batch..., num_heads, q_length, kv_length]`.
   """
-  assert key.ndim == query.ndim, 'q, k, v must have same rank.'
+  assert key.ndim == query.ndim, 'q, k must have same rank.'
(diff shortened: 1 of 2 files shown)


CoderPat:
Hey this is great!
One question that I had is if there is sufficient use cases for having a version where it returns the _unnormalized_ attention weights?
Perhaps an even more general modification is allowing the user to specify the normalization function (such as [Entmax](https://github.com/deep-spin/entmax-jax)), and then the previous case is just a question of passing the identity function and doing the normalization outside
I'm happy to help with this if you think it is a useful thing to have

avital:
Interesting question @CoderPat -- maybe if we can split up `dot_product_attention_weights` into even smaller functions that simply compose that could make sense. But I wouldn't want to add special options to functions, because then you always want more. And we always say "if it's not hard to fork it, that's better than adding an option". If you have an idea in mind for how you'd approach it I'm happy to take a look.

CoderPat:
Ye I get your side. 
One potential compromise is to have the signature of the `dot_product_attention` (and other components that need it) as `dot_product_attention(..., normalization_fn: Callable = nn,softmax, ...)`. 
That way it allows normalization function other than softmax and if users just want the unnormalized weights they can pass the identity function. But ye not sure if you guys want to cross this line

avital:
How about this -- maybe you could open a new GitHub discussion thread with your proposal and let's collect more perspectives from the community. WDYT?

flax/linen/attention.py
```diff
@@ -72,14 +67,9 @@ def dot_product_attention(query: Array,
...
+    Output of shape `[batch..., num_heads, q_length, kv_length]`.
   """
-  assert key.ndim == query.ndim == value.ndim, 'q, k, v must have same rank.'
-  assert query.shape[:-3] == key.shape[:-3] == value.shape[:-3], (
-      'q, k, v batch dims must match.')
-  assert query.shape[-2] == key.shape[-2] == value.shape[-2], (
-      'q, k, v num_heads must match.')
-  assert key.shape[-3] == value.shape[-3], 'k, v lengths must match.'
+  assert key.ndim == query.ndim, 'q, k, v must have same rank.'
   assert query.shape[-1] == key.shape[-1], 'q, k depths must match.'
```
marcvanzee, line 73: Shouldn't you keep the asserts comparing `query` and `key` for the other indices?

marcvanzee (approved):

CoderPat:
@avital sounds good, ill open a discussion

patrickvonplaten:
API looks great to be integrated with transformers :-)

PR #1302, opened 2021-05-04
Updates examples/mnist to use Optax.

andsteing:
See #1053


8bitmp3:
Thank you @andsteing

avital (approved):
Thanks!

PR #1306, opened 2021-05-06
sst2 example: Switch to Optax optimizer and use TrainState.

### LINKED ISSUE #1053: Replace flax.optim with Optax in examples

bastings:
This switches the sst2 example to use Optax for gradient updates and introduces a TrainState to make that work.

Fixes #1053 (partially: for sst2).

## Checklist
      checks if that's the case).
- [X ] This change is discussed in a Github issue/
      link). #1053


examples/sst2/train.py
```diff
@@ -18,20 +18,28 @@
...
 from flax.metrics import tensorboard
+from flax.training import train_state
 import input_pipeline
 import jax
 import jax.numpy as jnp
 import ml_collections
 import models
 import numpy as np
 import tensorflow as tf
+import optax
```
andsteing, line 28: Alphabetically this should go before tensorflow.

examples/sst2/train.py
```diff
@@ -18,20 +18,28 @@
...
 
 Array = jnp.ndarray
 Example = Dict[str, Array]
 
 
+class TrainState(train_state.TrainState):
+  model_state: Any
+  dynamic_scale: Optional[flax.optim.DynamicScale]
```
andsteing, line 37: Why do you need a `dynamic_scale`? Check out the `imagenet` example for an application of that class.

examples/sst2/train.py
```diff
@@ -42,23 +50,35 @@ def sigmoid_cross_entropy_with_logits(*, labels: Array, logits: Array) -> Array:
...
-  variables = model.init(key, token_ids, lengths)
-  state, params = variables.pop('params')
-  return params, state
+  variables = model.init(rng, token_ids, lengths, deterministic=True)
+  model_state, params = variables.pop('params')
+
+  # Remove intermediates for training. Otherwise our model state will fill up
+  # with intermediate outputs (exported using self.sow() commands). This will
+  # cause model_state to have a new shape on each step, triggering a new trace.
+  if 'intermediates' in model_state:
```
andsteing, line 60: Why would `"intermediates"` be in `model_state`, but only sometimes?

The documentation says it's not created during `init`:
https://flax.readthedocs.io/en/latest/howtos/extracting_intermediates.html#store-intermediate-values-in-a-new-variable-collection

examples/sst2/train.py
```diff
@@ -42,23 +50,35 @@ def sigmoid_cross_entropy_with_logits(*, labels: Array, logits: Array) -> Array:
...
-def get_initial_params_and_state(key, model):
+def get_initial_params_and_state(rng, model):
   """Returns randomly initialized parameters and a fresh model state."""
   token_ids = jnp.ones((2, 3), jnp.int32)
   lengths = jnp.ones((2,), dtype=jnp.int32)
-  variables = model.init(key, token_ids, lengths)
-  state, params = variables.pop('params')
-  return params, state
+  variables = model.init(rng, token_ids, lengths, deterministic=True)
+  model_state, params = variables.pop('params')
```
andsteing, line 55: What about keeping individual variable collections in separate variables?
I find that more readable, especially when there are `intermediates` lurking around.

Like we do here:
https://github.com/google/flax/blob/0082b90c36514b4947834cafc99172ec91ed16b5/examples/imagenet/train.py#L72

bastings:
Looks like some changes made in the main branch are causing build to fail.

examples/sst2/configs/default.py
```diff
@@ -42,6 +42,5 @@ def get_config():
   config.num_epochs = 10
 
   config.seed = 0
-  config.deterministic = False
```
andsteing, line 45: How is this change connected to using Optax?

examples/sst2/train.py
```diff
@@ -85,29 +102,26 @@ def model_from_config(config: ml_collections.ConfigDict):
...
   # Make sure to get a new RNG at every step.
-  model = model_from_config(config)
-  step = optimizer.state.step
+  step = state.step
   rngs = {name: jax.random.fold_in(rng, step) for name, rng in rngs.items()}
 
   def loss_fn(params):
-    variables = {'params': params, **model_state}
-    logits, new_model_state = model.apply(
+    variables = {'params': params, **state.model_state}
```
andsteing, line 120: (then we would also be explicit here)

examples/sst2/train.py
```diff
@@ -85,29 +102,26 @@ def model_from_config(config: ml_collections.ConfigDict):
...
+    logits, new_model_state = state.apply_fn(
         variables, batch['token_ids'], batch['length'],
-        rngs=rngs, mutable=list(model_state.keys()))
+        deterministic=False,
+        rngs=rngs, mutable=list(state.model_state.keys()))
```
andsteing, line 124: (and here)

examples/sst2/train_test.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+import train
+
+# Parse absl flags test_srcdir and test_tmpdir.
+jax.config.parse_flags_with_absl()
+
+
+class TrainTest(parameterized.TestCase):
+
+  def test_train_step_returns_correct_output_shape(self):
```
andsteing, line 30: where do you check the output shape?

examples/sst2/train_test.py
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for sst2.train."""
+from absl.testing import absltest
+from absl.testing import parameterized
+from configs import default as default_config
```
andsteing, line 18: nit: I really like ordered imports like this:

```
from absl.testing import absltest
from absl.testing import parameterized
import jax
import jax.test_util
import numpy as np

from configs import default as default_config
import train
```

andsteing (changes requested):

examples/sst2/train_test.py
bastings: I agree with you, but the linter doesn't agree with us :) I think i'd like to stick with keeping the linter happy.

bastings:
> Why would "intermediates" be in model_state, but only sometimes?

That change is very recent! But great, looks like they are no longer there when initiliazing, so  i could remove that code.

examples/sst2/train.py
bastings: This was a recent change, thanks :) We can remove that part now. (And yeah they are really always there, unless the user removed the attention mechanism from the model).

examples/sst2/train.py
bastings: Done

examples/sst2/configs/default.py
andsteing: Ah, got it.

Personally, I prefer being always explicit about that flag, so the change totally LGTM.

examples/sst2/train.py
andsteing: Did you push? I'm still seeing `model_state`...

examples/sst2/train_test.py
bastings: I didn't ;) Updated the test so that it tests that the parameters are updated in the train step.

examples/sst2/train.py
```diff
@@ -13,23 +13,32 @@
...
 import numpy as np
+import optax
 import tensorflow as tf
 
 Array = jnp.ndarray
 Example = Dict[str, Array]
+TrainState = train_state.TrainState
+
+
+class Metrics(struct.PyTreeNode):
```
bastings, line 37: Note, this is new in the last update.

examples/sst2/train.py
andsteing: Ah it was only about `intermediates` in the first place... I tend more and more to always spelling them out because having that `variables` dict is just very opaque.

andsteing (approved):

PR #1323, opened 2021-05-17
Renames `train_state` to `state`.

andsteing:
As @8bitmp3 has mentioned in [a comment](https://github.com/google/flax/commit/c4b3464427ae1d7e12ef3e114dd8ed4022709ddf#commitcomment-50892606) currently it's hard to find documentation about the new `state.apply_gradients()` because we call the variable `train_state` instead of `state` as in the examples. Note that we decided to go for simply `state` in the examples because we stored the new `TrainState` in a module called `flax.training.train_state`.

### DIFF
--- a/docs/flip/1009-optimizer-api.md
+++ b/docs/flip/1009-optimizer-api.md
@@ -228,19 +228,19 @@ With this the [Optax Training Step] becomes:
 
 ```python
 @jax.jit
-def train_step(train_state, inputs, labels):
+def train_step(state, inputs, labels):
 
   def loss_fn(params):
-    outputs, new_model_state = train_state.apply_fn(
-        {'params': params, 'batch_stats': train_state.batch_stats},
+    outputs, new_model_state = state.apply_fn(
+        {'params': params, 'batch_stats': state.batch_stats},
         inputs,
         mutable=['batch_stats'])
     loss = xent_loss(outputs, labels)
     return loss, new_model_state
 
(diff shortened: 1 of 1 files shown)


andsteing:
Note that this PR partially reverts [this commit](https://github.com/google/flax/commit/e45914f1b0c3950681ede50e2163ddb4a735e26b) but at that point we did not yet have decided to call the module `train_state` so there was no naming conflict.

avital:
I get that it's confusing to have `train_state` be both the module name as well as the variable name.

But a global variable named `state` seems like an anti-pattern...

What lives in train_state.py? Looks like it's a single class, just TrainState. Then why not simply re-export it from flax.training? So you could do `from flax.training import TrainState`? Maybe this was discussed somewhere else, apologies if I missed it.

andsteing:
It's not a global variable, but rather instantiated inside the `train()` function and then passed around as a function argument. Personally I don't think that `train_step(train_state, ...)` adds anything to `train_step(state, ...)`. Also, we already call it `state` (and not `train_state`) in our examples, where this never was raised as an issue.

We don't use `flax/training/__init__.py` to import symbols from modules in the same package. I'm not sure how that decision was taken originally, but it would break the existing pattern `from flax.training import <module>` if we suddenly switched to `from flax.training import <class>`.

(Note that you advised me to put that single class into a separate module on #1251 ...)

avital:
OK you've persuaded me!

avital (approved):

Issue #1326, opened 2021-05-18
pip dependency conflict makes our CI build fail


marcvanzee:
Our CI builds are currently failing, example: https://github.com/google/flax/runs/2582750675?check_suite_focus=true

It takes ~6h to resolve pip dependencies and download all the versions.

It seems the problem is with the latest version of Tensorflow, which is released 2 days ago. `pip install --upgrade tensorflow tensorflow-text` causes pip to keep installing different versions. When we use the previous version things work again, so I suggest we temporary add `tensorflow==2.4.1` to our requirements and file an issue to TF.

marcvanzee:
#1327 provides a temporary fix, but I've also filed https://github.com/tensorflow/tensorflow/issues/49253. Once that is fixed we should remove the TF version in `setup.py` again.

PR #1327, opened 2021-05-18
Don't pip install newest TF (see #1326)

marcvanzee:
# What does this PR do?

See #1326


avital:
Thanks for investigating this!

avital (approved):

PR #1351, opened 2021-06-03
Fix citations

lucasb-eyer:
The citation for ResNet50 is wrong, it's currently pointing to Kaiming's init paper.

(and yes, changing to 2016 is correct, it was published CVPR 2016, and one commonly uses the publication date, not arxiv date.)


jheek (approved):
Thanks!

PR #1354, opened 2021-06-04
Fix NamedTuples being converted to tuples in Module

jheek:
NamedTuples are constructed differently so they
have to be special cased by the freezing logic
in Module (*args instead of Iterable)


avital (approved):
Thanks!

Issue #659, opened 2020-11-20
Port Annotated MNIST to Linen


avital:
Port the [annotated MNIST example](https://github.com/google/flax/blob/master/docs/annotated_mnist.md) to the new Linen API.

Then add it to Read The Docs, and fix the currently broken link from the MNIST Colab

8bitmp3:
Hi @avital 👋 Are you looking for assistance with porting flax.nn annotated examples to Linen?

avital:
Yes @8bitmp3! Would you like to port the annotated MNIST example?

8bitmp3:
Sure @avital Do you prefer the Markdown-only format? I think the JAX team uses a Jupyter notebook-to-html extension that yields stuff like https://jax.readthedocs.io/en/latest/notebooks/quickstart.html

avital:
We also have that plugin installed but I prefer the markdown only version here because that allows you to interleave text /within/ a single function, not just between functions.

8bitmp3:
Ok 👍 @avital I can create a "Jupytext" Markdown version which we can edit during the review

avital:
Awesome! Please take a look at the former annotated MNIST linked to from the top comment on this issue for inspiration (people really liked that; perhaps it could be improved but to first order probably the right thing)

8bitmp3:
Sounds good @avital, that's the plan. I'm combining the original annotated MNIST [tutorial](https://flax.readthedocs.io/en/latest/annotated_mnist.html) with the Flax Linen [example](https://github.com/google/flax/blob/master/examples/mnist/mnist_lib.py) (under `/flax/examples/mnist/`). And for one-hot encoding, we could do what the Haiku team did in its MNIST [example](https://github.com/deepmind/dm-haiku/blob/master/examples/mnist.py) and use the ready-made [`jax.nn.one_hot`](https://jax.readthedocs.io/en/latest/_modules/jax/_src/nn/functions.html#one_hot) function.

avital:
Yes jax.nn.one_hot sounds like a good idea!

avital:
Hi @8bitmp3 -- just checking if you are looking at this. I'd love to see this in, one of us could do it instead if you're too busy.

8bitmp3:
Hey @avital 👋 Thanks for checking in. It's been a week of NeurIPS2020 😆. I'll show you what I got real soon, can open as a WIP PR 👍

avital:
#764 closes this. Thanks again @8bitmp3!

[...]

Issue #1357, opened 2021-06-08
Update documentation to `optax`


andsteing:
Examples are currently being refactored to use `optax` #1356

This issue tracks updating the documentation (readthedocs, pydoc) to mark `flax.optim` as deprecated, and to introduce new documentation for using `optax`.

PR #1358, opened 2021-06-08
Updates docs for Optax.

andsteing:
- Marks `flax.optim` as "effectively deprecated" in module pydoc (which is also shown on RTD).
- Marks `flax.training.lr_schedule` as "effectively deprecated" in module pydoc (which is also shown on RTD).
- Improves `flax.training.train_state` docs & includes them in RTD.
- Updates "Flax basics" Colab.
- Reinstates "Annotated MNIST" Colab, based on new MNIST source.

See #1357


flax/optim/__init__.py
```diff
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Flax Optimizer api."""
+"""Flax Optimizer api.
+
+Note that with `FLIP #1009`_ the optimizers in ``flax.optim`` were **effectively
+deprecated** in favor of Optax_ optimizers. There is not feature parity yet (e.g.
+``AdaFactor`` is missing in Optax), but the large majority of usecases is well
```
marcvanzee, line 19:
```suggestion
``AdaFactor`` is missing in Optax), but the large majority of use cases is well
```

flax/optim/__init__.py
```diff
@@ -12,7 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Flax Optimizer api."""
+"""Flax Optimizer api.
+
+Note that with `FLIP #1009`_ the optimizers in ``flax.optim`` were **effectively
+deprecated** in favor of Optax_ optimizers. There is not feature parity yet (e.g.
```
marcvanzee, line 18:
```suggestion
deprecated** in favor of Optax_ optimizers. There is no feature parity yet (e.g.
```

flax/optim/__init__.py
```diff
@@ -12,7 +12,17 @@
...
+``AdaFactor`` is missing in Optax), but the large majority of usecases is well
+supported, and Optax is actively being developed and continuously adds new
+features and optimizers.
```
marcvanzee, line 21:
```suggestion
supported, and Optax is actively being developed and new features and
optimizers are being added continuously.
```

flax/training/train_state.py
```diff
@@ -22,29 +22,30 @@
...
-      grads = grad_fn(state.params, batch)
-      state = state.apply_gradients(grads=grads)
+      state = TrainState.create(
+          apply_fn=model.apply,
+          params=variables['params'],
+          tx=tx)
+      grad_fn = jax.grad(make_loss_fn(state.apply_fn))
+      for batch in data:
+        grads = grad_fn(state.params, batch)
+        state = state.apply_gradients(grads=grads)
```
marcvanzee, line 34: I don't think you need 4 indents, 2 indents and the double colon with an empty newline should be sufficient to get code view.

marcvanzee (changes requested):
Why do we remove `docs/annotated_mnist.md` and we add `docs/notebooks/annotated_mnist.ipynb`?

flax/training/train_state.py
andsteing: Correct. Updated

andsteing:
> Why do we remove `docs/annotated_mnist.md` and we add `docs/notebooks/annotated_mnist.ipynb`?

Because this allows users to open the code in a Colab, inspect & modify it. It's also a lot easier to keep it working by "running all" after changes. In a markdown file it would be laborious copy'n'paste to make sure the code is correct.

Note that I tried to use `jupytext` (as [does JAX](https://jax.readthedocs.io/en/latest/developer.html#syncing-notebooks)) but I couldn't get it to work, so I went for a good old Colab.

The content is exactly the same.

What would be the advantage of a markdown file over a Colab?

andsteing:
Should also fix #659

marcvanzee:
> What would be the advantage of a markdown file over a Colab?

It is a bit easier to read from source and update directly. I just was confused since the change doesn't really seem to be related with the description in the PR, but it makes sense.

marcvanzee (approved):

andsteing:
> It is a bit easier to read from source and update directly. I just was confused since the change doesn't really seem to be related with the description in the PR, but it is fine.

Maybe I'll try again to update our docs to use `jupytext` then we would have the best from both worlds (i.e. simple text edits in markdown, but using Jupyter for code edits to make sure everything works as expected).

8bitmp3:
+1 jupytext @andsteing

Issue #1364, opened 2021-06-14
Add a dummy __call__ method to nn.Module base class for typechecking


levskaya:
for pytype and other checkers we should have something like:

```
if typing.TYPE_CHECKING:
  def __call__(self, *args, **kwargs) -> Any: ...
```

at the very least to remove a class of existing false-negative type-check issues in pytype (when passing modules / module factories as attributes).

levskaya:
A minimal example of one current issue w. the pytype checker is:
```python
from flax import linen as nn
class MyModule(nn.Module):
  sub_module: nn.Module
  def __call__(self):
    return self.sub_module()
```
results in `... in __call__: 'Module' object is not callable [not-callable]`

levskaya:
Closing - this was resolved by #1365

PR #1365, opened 2021-06-15
Add typing stub for Module __call__

levskaya:
Typecheckers like pytype currently don't recognize Module as a Callable
type when "nn.Module" is used as a dataclass attribute type, leading to
spurious typing errors, this adds a type-checking-only __call__ stub,
and moves the previous __init__ stub under the type-checking guard as
well. 

As discussed in #1364

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -20,6 +20,7 @@ import inspect
 import os
 import threading
 import types
+import typing
 import weakref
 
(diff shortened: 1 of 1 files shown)


jheek (approved):
Strictly speaking Module is not callable (only when a subclass implements it).
But otherwise type hints about modules would get very complicated so I think this makes sense

PR #1368, opened 2021-06-18
Change unhashable module exception type to TypeError.

levskaya:
Other libraries when checking whether or not an object
is hashable do so by guarding against a TypeError rather than a
ValueError, so this change brings our unhashable exception in line with
python conventions.


levskaya:
For instance the current code will cause things like webpdb to crash because it looks for a `TypeError` from a failed `hash(x)` call rather than a `ValueError` when it tries to build representations of the stack locals containing a module w. variables.

jheek (approved):
Good catch!

levskaya:
I'm seeing one baffling pytype error on github tests that I can't imagine comes from this PR.  I can't repro it locally, still trying to figure out what's going on with that.

PR #1372, opened 2021-06-18
Updates imagenet Colab & checkpoints for Optax.

andsteing:
As part of #1053 
Note that other imagenet code was ported as part of #1251


avital:
LGTM! I am a bit surprised that this is faster (but that's nice) -- though I was just comparing the tensorflow.dev runs so probably a lot of stuff has changed in JAX and XLA in the meanwhile, that could also be the cause.

And this new result matches the accuracy in our README but is 0.15% lower than the previous tensorflow.dev run. Perhaps this is just noise? Or maybe the Optax optimizer has different initialization? I don't think we necessary need to sweat it but it's one of those things that lights a question mark. For what it's worth, this PR is clearly an improvement because we already ported our actual code so no need to block merging.

cc @jheek

avital (approved):

andsteing:
From internal regression tests we expect the final accuracy to fluctuate somewhat between 76.1% and 76.5%, so I think the value reported here (76.27%) is still within range. Also looking at that internal regression test I don't see any difference after #1251 was submitted on 4/21/21. As for the walltime, I can see a continuous decrease over the last months in the internal test as well, so I think that's in line with expectations (maybe due to many small improvements to JAX?)

Issue #571, opened 2020-10-29
Updating linen_examples/seq2seq


andsteing:
This issue tracks updates the `seq2seq` example to follow practices outlined in #231.

- [x]  Port to linen API - once ported all subsequent changes should be done in `linen_examples/seq2seq`
- [ ]  Update `README.md`
- [x]  Add `requirements.txt` #1396
- [ ]  Update file structure
- [ ]  Use `ml_collections.ConfigDict` ?
- [ ]  Add benchmark test ?
- [x]  Add unit test for training/eval step ?
- [x]  Add Colab #1396
- [ ]  Adhere to Google Python style
- [ ]  Add mypy annotations
- [x]  Shorten/beautify training loop (consider using `clu` for this)

andsteing:
Note that "file structure" in #231 got changed (as part of #634):

> * To make it easier to test, maintain and reuse code structure the code as follows:
>   
>   * `main.py` contains the flags and calls a method from `train.py` to run the training loop. This should be the only file defining flags! This can be almost identical for all examples, please copy from `linen_examples/wmt/main.py`.
>   * `train.py` contains classes and methods for training and evaluating the model.
>   * `train_test.py` for test cases of the training code. At a minimum the test should run a single training step but more fine grained unit tests are a bonus. You can use `tfds.testing.mock_data` to avoid real data from disk.

andsteing:
as discussed in last week's meeting, we don't really need a benchmark/colab in every example, we don't require mypy annotations, and very simple examples don't need configs.

PR #1396, opened 2021-06-29
Improves seq2seq example

andsteing:
This PR does a couple of things (see individual commit messages for details)

1. Removes deprecated @jax.mask transformation.
2. Makes seq2seq use Optax.
3. Adds CLU metric logging.
4. Adds Colab.

Note that some - but not all - of the issues in #571 are resolved with this PR.

Also resolves example for #1053


levskaya (approved):
Thank you!!!

Issue #1402, opened 2021-07-01
Error when specifying type parameter of a nn.Module that inherits from typing.Generic


dhgarrette:
### Problem you have encountered:

When a `nn.Module` inherits from `typing.Generic`, a Flax error is triggered if the type parameter is specified when declaring an attribute or return type.

### Logs, error messages, etc:

```
flax.errors.ReservedModuleAttributeError: properties `parent` and `name` are reserved:
    {'parent': typing.Union[typing.Type[_ForwardRef('Module')],
                            typing.Type[_ForwardRef('Scope')],
                            typing.Type[_ForwardRef('_Sentinel')],
                            NoneType],
     'name': <class 'str'>}
    (https://flax.readthedocs.io/en/latest/flax.errors.html#flax.errors.ReservedModuleAttributeError)
```

### Steps to reproduce:

Generic Module as a return type:

```python
from typing import Generic, TypeVar
from absl import app
from flax import linen as nn
import jax.numpy as jnp

T = TypeVar('T')

class MyComponent(nn.Module, Generic[T]):
  pass

def make_component() -> MyComponent[jnp.ndarray]:  # Flax error here.
  raise NotImplementedError()

def main(_):
  pass

if __name__ == '__main__':
  app.run(main)
```

Generic Module as an attribute:

```python
from typing import Generic, TypeVar
from absl import app
from flax import linen as nn
import jax.numpy as jnp

T = TypeVar('T')

class MyComponent(nn.Module, Generic[T]):
  pass

class MyModule(nn.Module):
  submodule: MyComponent[jnp.ndarray]  # Flax error here.

def main(_):
  pass

if __name__ == '__main__':
  app.run(main)
```

levskaya:
The #1403 PR fixes the proximal issue, but there's something deeper going on...

It seems multiple inheritance from a `Generic[T]` and another parent with any `__init_subclass__` can cause issues, likely due to some clash with the metaclass machinery of `Generic`.  e.g.

```python
class ClassWithInitSubclass:
  @classmethod
  def __init_subclass__(cls):
    pass  # no-op

class MyComponent(ClassWithInitSubclass, Generic[T]):
  pass

MyComponent[jnp.ndarray]  # AttributeError: type object 'MyComponent' has no attribute '__parameters__'

class MyComponent2(Generic[T], ClassWithInitSubclass):
  pass

MyComponent2[jnp.ndarray]  # OK
```

levskaya:
[...]

PR #1403, opened 2021-07-02
Make nn.Module compatible with multiple inheritance with Generic[T].

levskaya:
In some cases of multiple inheritance with `Generic` types, the class is reinstantiated with default annotations and we shouldn't error there.  We also fix here a general hygiene mistake in our `__init_subclass__` of not calling the `super()` `__init_subclass__` (which for instance breaks `Generic`'s metaclass machinery).

Fixes the issue in #1402.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -435,8 +435,9 @@ class Module:
       pass
 
   @classmethod
-  def __init_subclass__(cls):
+  def __init_subclass__(cls, **kwargs):
     """Automatically initializes all subclasses as custom dataclasses."""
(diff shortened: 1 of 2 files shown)


jhclark-google:
My only feedback would be to add a few more `# comments` inline with the fix so the reader knows what oddity these conditionals are getting at.

Also, thanks for the amazingly fast fix! 1 hour after the bug was filed, wow!

levskaya:
@jheek - this was accidentally merged by copybara-service due to an internal lgtm by someone else.  Let me know if anything looks sketchy.

jheek:
LGTM

PR #1404, opened 2021-07-02
Use Optax optimizer for ppo example

jheek:
Use Optax optimizer for ppo example:
- Use TrainState
- use apply_fn inline with train_state pattern

Part of #1053

### DIFF
--- a/examples/ppo/ppo_lib.py
+++ b/examples/ppo/ppo_lib.py
@@ -124,14 +124,15 @@ def loss_fn(
 
   return ppo_loss + vf_coeff*value_loss - entropy_coeff*entropy
 
-@functools.partial(jax.jit, static_argnums=(5,))
+@functools.partial(jax.jit, static_argnums=(2,))
 def train_step(
     state: train_state.TrainState,
     trajectories: Tuple,
+    batch_size: int,
+    *,
     clip_param: float,
     vf_coeff: float,
-    entropy_coeff: float,
-    batch_size: int):
+    entropy_coeff: float):
   """Compilable train step.
 
   Runs an entire epoch of training (i.e. the loop over minibatches within
@@ -145,10 +146,10 @@ def train_step(
                   old_log_probs: shape (steps_per_agent*num_agents, )
                   returns: shape (steps_per_agent*num_agents, )
                   advantages: (steps_per_agent*num_agents, )
+    batch_size: the minibatch size, static argument
     clip_param: the PPO clipping parameter used to clamp ratios in loss function
     vf_coeff: weighs value function loss in total loss
     entropy_coeff: weighs entropy bonus in the total loss
-    batch_size: the minibatch size, static argument
 
   Returns:
     optimizer: new optimizer after the parameters update
@@ -160,10 +161,10 @@ def train_step(
   loss = 0.
   for batch in zip(*trajectories):
     grad_fn = jax.value_and_grad(loss_fn)
-    l, grad = grad_fn(state.params, state.apply_fn, batch, clip_param, vf_coeff,
+    l, grads = grad_fn(state.params, state.apply_fn, batch, clip_param, vf_coeff,
                       entropy_coeff)
     loss += l
-    state = state.apply_gradients(grads=grad)
+    state = state.apply_gradients(grads=grads)
   return state, loss
 
 def get_experience(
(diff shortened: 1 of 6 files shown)


examples/ppo/ppo_lib.py
```diff
@@ -113,31 +117,28 @@ def loss_fn(
...
+  return ppo_loss + vf_coeff*value_loss - entropy_coeff*entropy
 
-@functools.partial(jax.jit, static_argnums=(0,7))
+@functools.partial(jax.jit, static_argnums=(5,))
 def train_step(
-    module: models.ActorCritic,
-    optimizer: flax.optim.base.Optimizer,
+    state: train_state.TrainState,
     trajectories: Tuple,
     clip_param: float,
```
andsteing, line 131: Nit: Could you make these floats kw only args?

examples/ppo/ppo_lib.py
```diff
@@ -161,15 +160,14 @@ def train_step(
   loss = 0.
   for batch in zip(*trajectories):
     grad_fn = jax.value_and_grad(loss_fn)
-    l, grad = grad_fn(optimizer.target, module, batch, clip_param, vf_coeff,
+    l, grad = grad_fn(state.params, state.apply_fn, batch, clip_param, vf_coeff,
```
andsteing, line 163: Nit: Could you name this `grads`?

examples/ppo/ppo_lib.py
```diff
@@ -249,22 +247,46 @@ def process_experience(
...
+      apply_fn=model.apply,
+      params=params,
+      tx=tx)
+  return state
+
+
 def train(
-    module: models.ActorCritic,
-    optimizer: flax.optim.base.Optimizer,
+    model: models.ActorCritic,
```
andsteing, line 276: Nit: could you rename it to `model` everywhere?

examples/ppo/ppo_lib.py
```diff
@@ -274,31 +296,39 @@ def train(
   log_frequency = 40
   checkpoint_frequency = 500
 
-
-  for s in range(loop_steps):
+  initial_params = get_initial_params(jax.random.PRNGKey(0), model)
+  state = create_train_state(initial_params, model, config,
+                             loop_steps * config.num_epochs)
```
andsteing, line 301: I'm not sure I follow how steps are counted:
I think below you say that `state.step = s * num_epochs * num_agents * actor_steps / batch_size`.
So `max(state.step) = loop_steps * num_epochs * num_agents * actor_steps / batch_size`.
But here you say `train_steps = loop_steps * num_epochs` - so that seems to be missing the `num_agents * actor_steps / batch_size` part.

examples/ppo/ppo_lib.py
```diff
@@ -274,31 +296,39 @@ def train(
...
+  del initial_params
+  state = checkpoints.restore_checkpoint(model_dir, state)
+  # number of train iterations done by each train_step
+  iterations = config.num_agents * config.actor_steps // config.batch_size
+  start_step = int(state.step) // config.num_epochs // iterations
```
andsteing, line 306: (It's a bit confusing having `start_step` mean something quite different from `state.step`.)

examples/ppo/ppo_lib.py
```diff
@@ -274,31 +296,39 @@ def train(
...
+  logging.info(f'Start training from step: %s', start_step)
+
+  for s in range(start_step, loop_steps):
```
andsteing, line 309: Nit: Could we call this `step`?
(Or rename it to something more specific; see comment above about `state.step`)

andsteing, line 309: Maybe `range(start_step + 1, loop_steps + 1)`?
Then we would e.g. not checkpoint at the very first step but after `checkpoint_frequency` steps for the first time.

andsteing (changes requested):
Thanks for updating this!

Did you rerun the code to verify we get the same results with Optax?
(We don't expect any changes; but it would still be nice to verify by re-running; thanks for adding the `tensorboard.dev` trace in that case...)

examples/ppo/ppo_lib.py
jheek: Yeah I agree this is confusing but somewhat hard to solve. For performance reasons we need to do multiple optimization steps in a single train_step. This is because RL has really tiny models so the overhead of a single train_step call would be non-trivial.

examples/ppo/ppo_lib.py
jheek: You're right, fixed

examples/ppo/ppo_lib.py
andsteing: And it doesn't help that we have an inner loop over `config.num_epochs` for every "step" :-)

jheek: Yeah I have to dig into that again at some point. I think it made sense from an RL point of view but I need to think about it and put it in a docstring

jheek:
> Did you rerun the code to verify we get the same results with Optax?

Yes I did a re-run and it is consistent with the original run

examples/ppo/ppo_lib.py
andsteing: SGTM

andsteing: SGTM

andsteing (approved):
Thanks for updating the example to Optax!

PR #1413, opened 2021-07-07
Use optax for PixelCNN example

joaogui1:
Part of #1053


marcvanzee:
Nice! I suppose you didn't run this to verify it trains correctly? If not, I'm happy to kick off a run on a v3-8 myself!

joaogui1:
I run for very few steps, so yeah please kick off a run :D

marcvanzee:
Kicked off a run! Let's see in ~4d 😄 . We've had some problems with this example before so I'd like to make sure it reaches the right accuracy.

marcvanzee (changes requested):
Hi @joaogui1, I ran the command from the README: 

```bash
python main.py --workdir=/tmp/pixelcnn --config=configs/default.py --config.batch_size=320
```

And this gave me the following TB: https://tensorboard.dev/experiment/fBxcNKQyRxOULLykAz40RQ/

The loss looks really bad, so I think there is some kind of problem. I took a look at your code but I couldn't easily spot what is going wrong. I'll try running the original code to see if that does behave correctly (it could be we introduced a bug before your PR).

marcvanzee:
I just did a short run on the original example and there everything works fine. The problem is that in your PR, after 43 epochs the loss starts diverging. I ran on 8TPU, but I'll run again on 8GPU to see whether it happens there as well (we have problems running this example on TPU before).

marcvanzee:
@joaogui1 and me discussed offline. The problem is that on TPU, after about 43 epochs the loss starts diverging. Here is an example log: https://pastebin.com/FHTihf12

Joao told me that he made a change to the learning rate schedule, and since he has access to a TPU, he will try running the experiment and see whether the loss behaves better now.

Handing this PR over to @avital , who is the other owner of this example, since I will be OOO for 7 weeks starting tomorrow.

joaogui1:
@marcvanzee also got the divergence here, no idea why it's happening :/

joaogui1:
Any help here @marcvanzee @avital ?

jheek:
Both @marcvanzee and @avital are out of office this week. I'm not sure if they have figured out a possible reason for the divergence. I think @avital will get back to you in a week or so.

marcvanzee:
Hi @joaogui1 I'm back from parental leave! 👋 

I vaguely recall this example was always somewhat problematic on TPU, so I think we should first make sure the divergence only occurs after your change and not before. Therefore, I think we should run the PixelCNN++ example a few times without your change and make sure the loss behave correctly after 43 epochs. If so, then there is definitely something wrong and I think we should bring in someone from the Optax team. Do you still have access to your v3-8?

joaogui1:
Oh, congrats on the child :D @marcvanzee 
I still have access to them, I can test the previous version over the week and see how stable it is, will report back after that's done

joaogui1:
@marcvanzee 
End loss for the optax version is a couple orders of magnitude higher than for normal flax version, we'll probably need help from the optax team
image

marcvanzee:
Closing this since we remove the PixelCNN++ example.

PR #1416, opened 2021-07-09
Improves Flax Modules for RTD

### LINKED ISSUE #1049: Improve Modules __call__ and method arguments on RTD
(from @bastings):

The docstring of [Dense](https://github.com/google/flax/blob/master/flax/linen/linear.py#L140) is pretty good but it doesn't not show up nicely on [RTD](https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Dense.html). The `__call__` method seems to be missing.

Also the arguments to some modules are represented badly, e.g.: https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Conv.html

marcvanzee:
Adds a template for Flax Modules on RTD and applies this template to all Modules. Most importantly, this ensure `__call__` methods are shown, but there are a few more improvements.

Consider the following example for `flax.linen.Conv` (but it applies to all Flax Modules):

- BEFORE: https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Conv.html
- AFTER: https://flax--1416.org.readthedocs.build/en/1416/_autosummary/flax.linen.Conv.html

You see the following things are improved:

- Do not show a docstring for `__init__`
- Do show a docstring for `__call__`
- In the "methods" section in the end, only show the members that aren't inherited, which makes the list much shorter.
- Don't show an `attributes` section.

All these things are done by creating a custom template for our Modules.

I think it would be nice if the parameters showed up a bit more concise, but that seems good for a followup.


flax/linen/linear.py
```diff
@@ -50,7 +50,7 @@ def _canonicalize_tuple(x):
 class DenseGeneral(Module):
   """A linear transformation with flexible axes.
 
-    Attributes:
+    Args:
```
avital, line 53: I don't know about this -- why are we changing the docstring to fix RTD? The docstring was correct before -- these are attributes, not arguments (though they are arguments to the constructor but that's not what you're documenting here)

docs/_templates/autosummary/flax_module.rst
```diff
@@ -0,0 +1,23 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
```
avital, line 3: Is this really necessary? Is it because RTD doesn't have built-in support for dataclasses? Or that it doesn't know that this is a dataclass?

marcvanzee: What do you mean by "this"? This file or this line?

flax/linen/linear.py
marcvanzee: Yes this was a way to fix it so that it appears correctly on RTD, and I did it like this because this is how JAX does it as well. I agree Attributes is more correct, but then they show up so weirdly on RTD. Looking at the example on RTD, we aren't supposed to use Args or Attributes at all! https://sphinx-rtd-theme.readthedocs.io/en/stable/_modules/test_py_module/test.html#Foo

avital: > By the way, RTD actually uses a totally different syntax for attributes and arguments that I think we not using at all. We could consider switching to that at some point, if we think it will work better. https://sphinx-rtd-theme.readthedocs.io/en/stable/_modules/test_py_module/test.html#Foo

Oh, just now saw this! Then yes! Let's move to this -- why not? It's better also because it co-locates the comment with the attribute. Much more useful than scrolling up and down to see the type and the name and the docstring for each attribute?

PhilipVinc: That's also what we are using in NetKet ( see for example https://www.netket.org/docs/_generated/models/netket.models.RBM.html ).
I really suggest you using that syntax.

However it does not seem to work with attributes that are functions, such as the init_kernel arts, because RTD/Sphinx thinks they are methods instead of attributes.

If you find a workaround I'd love to hear it...

levskaya:
I don't think we should mark `Attributes` as `Args` for the sake of sphinx formatting - every style-guide and linter on earth would curse us.

marcvanzee:
> I don't think we should mark `Attributes` as `Args` for the sake of sphinx formatting - every style-guide and linter on earth would curse us.

Yes, good point, I agree. Reverted it and looking for another solution.

docs/_templates/autosummary/flax_module.rst
marcvanzee: This file is there so we can customize how we want our Flax modules to look, because right now they look really bad.

Consider the following example for `flax.linen.Conv` (but it applied to all Flax Modules):

- BEFORE: https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Conv.html
- AFTER: 

You see the following things are improved:

- Nicer formatting of parameters summary view. Less whitespace
- Nicer formatting of parameters details. Types are displayed more concise.
- Do not show a docstring for `__init__`
- Do show a docstring for `__call__`
- In the "methods" section in the end, only show the members that aren't inherited, which makes the list much shorter.
- Don't show an `attributes` section.

All these things are done by creating a custom template for our Modules.

marcvanzee: This is there to make sure we output the Module docstring correctly, see PR description.

flax/linen/linear.py
andsteing: +1 for testing on a couple of editors before changing the pydoc format.
personally, I use editor + codesearch, and rarely look at RTD, so it would be a pity if we changed something that made editor support less good for improving RTD.

andsteing:
As I see it, we now have an `Parameters` section with only names/types:
image

And then the attributes after the return type:
image

And then only methods defined in the module (which in this case actually are "functional attributes"):
image

So I think that's a great first step, and I'm happy about using `Attributes:` and not `Args:` for what really are attributes.

In the future this **could be further improved** with

1. Removing the `Parameters` section
2. Showing all attributes concisely with their type + doc (i.e. making better use of horizontal screen estate)
3. Showing "functional attributes" as attributes and not as methods without docstring.

andsteing (approved):
SGTM as an initial step for improving RTD pydoc rendering.
I have added some more comments to the PR thread (partly in reply to Anselm's note on using `Attributes:` and your reply to that).

PR #1451, opened 2021-07-22
Forbid passing an int as kernel_size

### LINKED ISSUE #1234: Surprising behaviour for integer kernel_size in linen.Conv
I was quite surprised with how `linen.Conv` treats `int` as an argument for `kernel_size`

```
key1, key2 = jax.random.split(jax.random.PRNGKey(0), 2)
image = jax.random.normal(key1, (8, 256, 256, 3))
conv = flax.linen.Conv(features=48, kernel_size=5)
params = conv.init(key2, image)
```

This errors on the last line with
```
AttributeError                            Traceback (most recent call last)
google3/third_party/py/numpy/core/fromnumeric.py in _wrapfunc(obj, method, *args, **kwds)
     55     try:
---> 56         return getattr(obj, method)(*args, **kwds)
     57 

AttributeError: 'tuple' object has no attribute 'take'

During handling of the above exception, another exception occurred:

IndexError                                Traceback (most recent call last)
11 frames
google3/third_party/py/numpy/core/fromnumeric.py in _wrapit(obj, method, *args, **kwds)
     44     except AttributeError:
     45         wrap = None
---> 46     result = getattr(asarray(obj), method)(*args, **kwds)
     47     if wrap:
     48         if not isinstance(result, mu.ndarray):

IndexError: index 3 is out of bounds for size 3
```

And I can see it creates a kernel for only 1 spatial dimension.

So it seems it says `5 == (5,)`

Changing to `conv = flax.linen.Conv(features=48, kernel_size=(5,5))` fixes the error.

Overall I find:
a) the error is cryptic
b) if this is the desired behaviour, perhaps it should only accept kernel_size as a sequence of ints?

jheek:
Users expect that the int will be broadcasted to the appropriate rank.
But this is ambiguous because we don't know if
or how many batch dimensions there are.

In contract, for stride and dilation we can broadcast
because the kernel rank is already known. This PR
therefore implements int broadcasting for these args.

Fixes #1234

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -199,19 +199,19 @@ class Conv(Module):
     kernel_size: shape of the convolutional kernel. For 1D convolution,
       the kernel size can be passed as an integer. For all other cases, it must
       be a sequence of integers.
-    strides: a sequence of `n` integers, representing the inter-window
-      strides.
+    strides: an integer or a sequence of `n` integers, representing the
+      inter-window strides (default: 1).
     padding: either the string `'SAME'`, the string `'VALID'`, or a sequence
       of `n` `(low, high)` integer pairs that give the padding to apply before
       and after each spatial dimension.
-    input_dilation: `None`, or a sequence of `n` integers, giving the
-      dilation factor to apply in each spatial dimension of `inputs`.
+    input_dilation: an integer or a sequence of `n` integers, giving the
+      dilation factor to apply in each spatial dimension of `inputs` (default: 1).
       Convolution with input dilation `d` is equivalent to transposed
       convolution with stride `d`.
-    kernel_dilation: `None`, or a sequence of `n` integers, giving the
+    kernel_dilation: an integer or a sequence of `n` integers, giving the
       dilation factor to apply in each spatial dimension of the convolution
-      kernel. Convolution with kernel dilation is also known as 'atrous
-      convolution'.
+      kernel (default: 1). Convolution with kernel dilation
+      is also known as 'atrous convolution'.
     feature_group_count: integer, default 1. If specified divides the input
       features into groups.
     use_bias: whether to add a bias to the output (default: True).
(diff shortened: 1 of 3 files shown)


flax/linen/linear.py
```diff
@@ -222,11 +222,11 @@ class Conv(Module):
...
   """
   features: int
-  kernel_size: Union[int, Iterable[int]]
-  strides: Optional[Iterable[int]] = None
+  kernel_size: Iterable[int]
+  strides: Union[None, int, Iterable[int]] = 1
   padding: Union[str, Iterable[Tuple[int, int]]] = 'SAME'
-  input_dilation: Optional[Iterable[int]] = None
-  kernel_dilation: Optional[Iterable[int]] = None
+  input_dilation: Union[None, int, Iterable[int]] = 1
```
andsteing, line 228: Please update pydoc (here and `strides` and `kernel_dilation`) to say that the default is 1 (not `None` anymore).

andsteing (changes requested):
Looks good other than failing tests.

flax/linen/linear.py
jheek: Done

flax/linen/linear.py
```diff
@@ -199,19 +199,19 @@ class Conv(Module):
...
-    strides: a sequence of `n` integers, representing the inter-window
-      strides.
+    strides: an integer or a sequence of `n` integers, representing the
+      inter-window strides (default: 1).
     padding: either the string `'SAME'`, the string `'VALID'`, or a sequence
       of `n` `(low, high)` integer pairs that give the padding to apply before
       and after each spatial dimension.
-    input_dilation: `None`, or a sequence of `n` integers, giving the
-      dilation factor to apply in each spatial dimension of `inputs`.
+    input_dilation: `an integer or a sequence of `n` integers, giving the
```
andsteing, line 207: nit: leading backtick should be removed.

andsteing (approved):

PR #1457, opened 2021-07-27
kwargs in remat and pack

### LINKED ISSUE #1455: remat: wrapper() got an unexpected keyword argument 'use_running_average'

The transformed module returned by `remat` does not expect same keyword arguments as original when used as a submodule in `@compact` decorated `__call__` method.

The transformed module expects the same keyword arguments.

```
TypeError: wrapper() got an unexpected keyword argument 'use_running_average'
```

### Steps to reproduce:

```
import jax
import flax.linen as linen 

from jax import numpy as jnp
from typing import Optional

class MyModule(linen.Module):
    
    @linen.compact
    def __call__(self, x, train: Optional[bool]=True):
        return linen.remat(linen.BatchNorm)()(x, use_running_average=not train)

model = MyModule()
key = jax.random.PRNGKey(0)
variables = model.init(key, jnp.ones((10,), jnp.float32))
```
https://colab.research.google.com/drive/1JsmxSn4Msor5D6G5XHpokfHOFlUcsXIJ?usp=sharing

ZYHowell:
# What does this PR do?

Fixes #1455

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -683,7 +683,7 @@ def checkpoint(fn: Callable[..., Any],
       return y, repack_fn(scope)
 
     return rematted(variable_groups, rng_groups, *args, **kwargs)
-  return pack(inner, (variables,), (variables,), (rngs,), name='remat', enable_kwargs=False)
+  return pack(inner, (variables,), (variables,), (rngs,), name='remat', enable_kwargs=True)
 
 
 def _hashable_filter(x):
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -80,7 +80,7 @@ def pack(fn: Callable[..., Any],
   The pack function is the building block for all other lifted transformations.
   """
   @functools.wraps(fn)
-  def wrapper(scope_tree: Scope, *args):
+  def wrapper(scope_tree: Scope, *args, **kwargs):
     # pylint: disable=protected-access
     scopes, treedef = jax.tree_flatten(scope_tree)
     scopes, paths = _dedup_scopes(scopes)
@@ -177,7 +177,7 @@ def pack(fn: Callable[..., Any],
       y, out_variable_groups_xs_t = fn(
           scope_fn, repack,
           variable_groups_xs_t, rng_groups_xs_t,
-          *args)
+          *args, **kwargs)
     finally:
       for inner_scope in inner_scopes:
         inner_scope.invalidate()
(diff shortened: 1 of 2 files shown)


ZYHowell:
@googlebot I signed it!

ZYHowell:
@googlebot I signed it!

flax/core/lift.py
```diff
@@ -177,7 +177,7 @@ def repack(inner_scope_tree):
       y, out_variable_groups_xs_t = fn(
           scope_fn, repack,
           variable_groups_xs_t, rng_groups_xs_t,
-          *args)
+          *args, **kwargs)
```
jheek, line 180: We would need to handle kwargs in other transformations as well. This might be tricky as JAX transforms don't always handle kwargs either.

I think it's also better to pass kwargs as a dict internally to avoid collisions with arg names.

ZYHowell: Does this sounds like a parameter named kwargs but only unpacked later at [here](https://github.com/google/flax/blob/3cea8ad2ad49ec53ae7ad3eefded1ba5b1f557d5/flax/core/lift.py#L671)? I'm not quite clear about the JAX part.

jheek: I think for remat JAX supports it and for jit support was added last week I think. This needs to be checked per transform though. For example with jit you also now need static_arg_names to pass kwargs statically (which is usually the case)

ZYHowell: In this case, what about adding a judgement of the `name` parameter at the beginning of `pack`? It's like when `name` is jit and the function's co_kwonlyargcount is not zero, an error(unimplemented) will be raised.

jheek: I think it would be cleaner to add a kwarg to pack indicating whether kwargs is supported instead of using string matching to figure out who called the function.

ZYHowell: What about adding a parameter like enable_kwargs with default value False, and then set it True in the `lift.checkpoint`?

jheek: Yeah I agree the default should be False

flax/core/lift.py
```diff
@@ -663,16 +674,16 @@ def checkpoint(fn: Callable[..., Any],
...
+    def rematted(variable_groups, rng_groups, *args, **kwargs):
       scope = scope_fn(variable_groups, rng_groups)
-      y = fn(scope, *args)
+      y = fn(scope, *args, **kwargs)
       return y, repack_fn(scope)
 
-    return rematted(variable_groups, rng_groups, *args)
-  return pack(inner, (variables,), (variables,), (rngs,), name='remat')
+    return rematted(variable_groups, rng_groups, *args, **kwargs)
+  return pack(inner, (variables,), (variables,), (rngs,), name='remat', enable_kwargs=False)
```
jheek, line 686: enable_kwargs=True?

flax/core/lift.py
```diff
@@ -663,16 +674,16 @@ def checkpoint(fn: Callable[..., Any],
     A wrapped version of ``fn``. When computing gradients intermediate
     computations will be re-computed when computing gradients.
   """
-  def inner(scope_fn, repack_fn, variable_groups, rng_groups, *args):
+  def inner(scope_fn, repack_fn, variable_groups, rng_groups, *args, **kwargs):
```
jheek, line 677: Please add a test to linen_transform_test.py to test the new behavior

jheek (changes requested):

jheek:
Thanks, looks all good now. Could you rebase the PR to fix the merge conflict?

jheek (approved):

PR #1466, opened 2021-07-30
Update imagenet_fake_data_benchmark to include warmup phase

cheshire:
Update imagenet_fake_data_benchmark to include warmup phase


cheshire:
@hawkinsp

PR #1472, opened 2021-08-03
Bumps examples/cloud script versions.

andsteing:
And some more master->main renaming.

Verified with imagenet example that everything works.


andsteing:
Replaced by #1477

PR #1475, opened 2021-08-04
Fix cloudpickle serialization of wrapped Module methods.

### LINKED ISSUE #1467: Cannot pickle linen Modules
I am using `0.3.4` and I am getting an error when trying to pickle flax modules, specifically `Dense` seems to be the problem but other might have similar issues.

```python
from flax import linen
import pickle

with open("model.pkl", "wb") as f:
    model = pickle.dump(linen.Dense(10), f)
```

> Traceback (most recent call last):
  File "test.py", line 8, in <module>
    model = pickle.dump(linen.Dense(10), f)
AttributeError: Can't pickle local object 'variance_scaling.<locals>.init'

While the previous is solved with `cloudpickle`, this other code doesn't work:
```python
import cloudpickle
from flax import linen
import pickle

class IndentityFlax(linen.Module):
    def __call__(self, x):
        return x

with open("mlp.pkl", "wb") as f:
    cloudpickle.dump(IndentityFlax(), f)
```
> Traceback (most recent call last):
  File "test.py", line 25, in <module>
    cloudpickle.dump(IndentityFlax(), f)
  File "/data/cristian/elegy/.venv/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 55, in dump
    CloudPickler(
  File "/data/cristian/elegy/.venv/lib/python3.8/site-packages/cloudpickle/cloudpickle_fast.py", line 563, in dump
    return Pickler.dump(self, obj)
TypeError: cannot pickle '_thread._local' object

jheek:
Calling _context inside a closure will make cloudpickle attempt to serialize.
It will fail because it cannot serialize ThreadLocals.

Factoring out the body of method_wrapper into a
separate method hides the _context for cloudpickle
because it can store library methods by name.

This PR does not come with a test because it requires us to depend on cloudpickle.
More importantly some tests indirectly uses dill which causes a clash in the pickle registry between dill & cloudpickle.
If we decide to make (cloud)pickle a first class citizen in the future we will need some clever test setup and elaborate to check for leaks of library internals through closures. 

Fixes #1467

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -262,39 +262,9 @@ def wrap_method_once(fun: Callable[..., Any]) -> Callable[..., Any]:
     # otherwise call the wrapped function as is.
     if args and isinstance(args[0], Module):
       self, args = args[0], args[1:]
+      return self._call_wrapped_method(fun, args, kwargs)
     else:
       return fun(*args, **kwargs)
-    is_compact_method = hasattr(fun, 'compact')
-    is_setup_method = fun.__name__ == 'setup'
-    # We lazily call setup() only when needed.
-    if is_setup_method:
-      is_recurrent = self._state.in_setup
-      self._state.in_setup = True
-    else:
-      self._try_setup()
-
-    if is_compact_method:
-      if self.scope is None:
-        raise errors.CallCompactUnboundModuleError()
-      is_recurrent = self._state.in_compact_method
-      self._state.in_compact_method = True
-    _context.module_stack.append(self)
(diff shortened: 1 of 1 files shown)


avital:
The change LGTM. Can you please file a new issue to keep track of the challenge of testing this? So that we could in the future find a solution?

avital (approved):
Thanks!

levskaya:
LGTM - thanks for figuring out what cloudpickle was doing here!

PR #1476, opened 2021-08-05
Transitions `examples/wmt` to use Optax

andsteing:
As part of #1053

This PR also updates the `README.md` with new commands for running on the new Cloud 1VM setup and updates `requirements.txt`.

Note that the learning rate was scheduled with `constant * linear_warmup * rsqrt_decay` in the original code, and this has been replaced with a joint linear + "shifted `rsqrt_schedule`" that also required to update the `config.learning_rate` value (it now is the maximum learning rate reached during training, which seems more aligned with expectations).

Verified on Cloud that results before and after 4b1baddfb93fd7c7f957ce4cc10e5a0844e78030 are identical (incl. learning rate):

https://tensorboard.dev/experiment/sJH8okaqRYyXXf6YSNWg5A

### DIFF
--- a/examples/wmt/README.md
+++ b/examples/wmt/README.md
@@ -14,90 +14,70 @@ rate schedule.
 *   This example additionally depends on the `sentencepiece` and
     `tensorflow-text` packages.
 
-### Supported setups
 
-The model should run with other configurations and hardware, but was explicitly
-tested on the following.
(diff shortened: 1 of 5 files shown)


levskaya (approved):
so sorry I thought that I had lgtm'd this days ago -- looks great!

PR #1478, opened 2021-08-05
Add autocomplete support for Module and struct.dataclass.

jheek:
image

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -38,6 +38,7 @@ from flax import core
 from flax.core import Scope
 from flax.core.scope import CollectionFilter, DenyList, Variable, VariableDict, FrozenVariableDict, union_filters
 from flax.core.frozen_dict import FrozenDict, freeze
+from flax.struct import __dataclass_transform__
 
 # from .dotgetter import DotGetter
 
@@ -396,7 +397,14 @@ capture_call_intermediates = lambda _, method_name: method_name == '__call__'
 # -----------------------------------------------------------------------------
 
 
-class Module:
+# This metaclass + decorator is used by static analysis tools recognize that
+# Module behaves as a dataclass (attributes are constructor args).
+@__dataclass_transform__()
+class ModuleMeta(type):
+  pass
+
+
+class Module(metaclass=ModuleMeta):
   """Base class for all neural network modules. Layers and models should subclass this class.
 
   All Flax Modules are Python 3.7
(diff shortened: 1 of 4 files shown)


flax/linen/module.py
```diff
@@ -51,6 +52,20 @@
 _CallableT = TypeVar('_CallableT', bound=Callable)
 
 
+_T = TypeVar("_T")
+
+def __dataclass_transform__(
```
andsteing, line 57: You already import this?

flax/struct.py
```diff
@@ -39,7 +39,24 @@
 import jax
 
 
-def dataclass(clz: type):
+
+# This decorator is interpreted by static analysis tools as a hint
+# that a decorator or metaclass causes dataclass-like behavior.
+_T = TypeVar("_T")
+def __dataclass_transform__(
```
andsteing, line 46: Should we really add dunders to this?

I don't know that from other libraries, maybe you found some guidance on when to use them?

andsteing (approved):
It's really great that this works, but I don't understand from your comments *why* it's working. If you could extend a bit in the code or the PR that would be interesting...

avital:
Yes adding a comment in the code (not the PR) is ideal so that future
readers immediately can understand

flax/linen/module.py
jheek: fixed

flax/struct.py
jheek: Yes, I've added a link to the proposal that explains this magic

tests/linen/module_test.py
```diff
@@ -1371,15 +1370,18 @@ class Foo(NamedTuple):
...
-    T = TypeVar('T')
-    class MyComponent(nn.Module, Generic[T]):
-      pass
-    class MyModule(nn.Module):
-      submodule: MyComponent[jnp.ndarray]
-    class MyComponent2(Generic[T], nn.Module):
-      pass
-    class MyModule2(nn.Module):
-      submodule: MyComponent2[jnp.ndarray]
+    if sys.version_info.major == 3 and sys.version_info.minor >= 7:
```
andsteing, line 1373: So this bug only appeared with the added `__dataclass_transform__`?

andsteing (approved):

tests/linen/module_test.py
jheek: Yes because the __dataclass_transform__ only works on metaclasses so we have an empty metaclass. In the exceptional edge case where somebody wants do multiple inheritance with Generic on a Module they could probably use `six`

cgarciae:
This is going to be a huge quality-of-life improvement for Flax!

tests/linen/module_test.py
jheek: Refactored the code to only add the metaclass during type checking. This avoids the problem with metaclass inheritance

PR #1479, opened 2021-08-05
Updates `examples/lm1b` to use Optax

jaehlee:
Use Optax optimizer for lm1b example

Part of #1053

### DIFF
--- a/examples/lm1b/train.py
+++ b/examples/lm1b/train.py
@@ -424,7 +424,7 @@ def train_and_evaluate(config: ml_collections.ConfigDict, workdir: str):
                                       jnp.ones(input_shape, jnp.float32))
 
   learning_rate_fn = create_learning_rate_schedule(
-      base_learning_rate=config.learning_rate, warmup_steps=config.warmup_steps)
+      learning_rate=config.learning_rate, warmup_steps=config.warmup_steps)
 
   optimizer = optax.adamw(
       learning_rate_fn, b1=0.9, b2=0.98, eps=1e-9,
--- a/examples/lm1b/train.py
+++ b/examples/lm1b/train.py
@@ -29,17 +29,20 @@ from clu import metric_writers
 from clu import periodic_actions
 from flax import jax_utils
 from flax import linen as nn
-from flax import optim
+
 import input_pipeline
 import models
 import temperature_sampler
 from flax.training import checkpoints
 from flax.training import common_utils
+from flax.training import train_state
+
 import jax
 from jax import random
 import jax.numpy as jnp
 import ml_collections
 import numpy as np
+import optax
 import tensorflow as tf
 
 
@@ -181,7 +184,7 @@ def compute_metrics(logits, labels, weights, label_smoothing=0.0):
 # -----------------------------------------------------------------------------
 
 
-def train_step(optimizer,
+def train_step(opt_state,
                batch,
                config,
                learning_rate_fn,
@@ -199,7 +202,7 @@ def train_step(optimizer,
 
   weights = jnp.where(inputs > 0, 1, 0).astype(jnp.float32)
 
-  dropout_rng = jax.random.fold_in(dropout_rng, optimizer.state.step)
+  dropout_rng = jax.random.fold_in(dropout_rng, opt_state.step)
 
   def loss_fn(params):
     """loss function used for training."""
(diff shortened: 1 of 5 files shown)


examples/lm1b/train.py
```diff
@@ -181,7 +184,7 @@ def compute_metrics(logits, labels, weights, label_smoothing=0.0):
 # -----------------------------------------------------------------------------
 
 
-def train_step(optimizer,
+def train_step(opt_state,
```
andsteing, line 187: We usually refer to `TrainingState` as `state`, not as `opt_state`.

examples/lm1b/train.py
```diff
@@ -450,34 +453,34 @@ def encode_strings(strs, max_len):
...
 
-  # apply an optimizer to this tree
-  optimizer_def = optim.Adam(
-      config.learning_rate,
-      beta1=0.9,
-      beta2=0.98,
-      eps=1e-9,
-      weight_decay=config.weight_decay)
-  optimizer = optimizer_def.create(initial_variables["params"])
+  learning_rate_fn = create_learning_rate_scheduler(
```
andsteing, line 456: I think we should update the `create_learning_rate_scheduler` like here:

https://github.com/google/flax/blob/4b1baddfb93fd7c7f957ce4cc10e5a0844e78030/examples/wmt/train.py#L47-L74

examples/lm1b/train.py
```diff
@@ -450,34 +453,34 @@ def encode_strings(strs, max_len):
...
+      base_learning_rate=config.learning_rate, warmup_steps=config.warmup_steps)
 
-  # We access model params only from optimizer below via optimizer.target.
+  optimizer = optax.adamw(
+      learning_rate_fn, b1=0.9, b2=0.98, eps=1e-9,
+      weight_decay=config.weight_decay
+      )
+  opt_state = train_state.TrainState.create(
```
andsteing, line 463: We usually call this simply `state`, and not `opt_state`.

andsteing (changes requested):
Hi Jaehoon

The change looks good!

I'm currently running your code on Cloud and will update here once I get the results so you can update the PR accordingly. I don't think that we need to do a before/after comparison since the change is identical with the `wmt` example and we already have a regression analysis in #1476.

examples/lm1b/train.py
jaehlee: Done.

examples/lm1b/train.py
jaehlee: Done.

examples/lm1b/train.py
jaehlee: Done.

jaehlee:
Thank you! Will update the README with the results once your run is done. 

> I'm currently running your code on Cloud and will update here once I get the results so you can update the PR accordingly.

andsteing:
You can find the updated results here:

- workdir: https://pantheon.corp.google.com/storage/browser/flax_public/examples/lm1b/default
- metrics: https://tensorboard.dev/experiment/n30WkNOZTJq3RHWD7wNslg/

Thanks for verifying the results are as expected and adding a table with links to the README.

jaehlee:
To me the run results looks consistent with using batch_size=256 (per_device_batch_size=32 & TPU v3-8). Updated the README.md with provided links.

examples/lm1b/train.py
```diff
@@ -29,76 +29,49 @@
...
-        raise ValueError("Unknown factor %s." % name)
-    return jnp.asarray(ret, dtype=jnp.float32)
-
-  return step_fn
+def create_learning_rate_schedule(learning_rate: float, warmup_steps: int):
+  """Creates a rsqrt schedule with linear warmup."""
+  return optax.join_schedules([
+      optax.linear_schedule(
+          init_value=0, end_value=learning_rate, transition_steps=warmup_steps),
+      rsqrt_schedule(init_value=learning_rate, shift=warmup_steps),
```
andsteing, line 72: When using the `rsqrt_schedule` you probably also need to update the learning rate in the config. Please check with #1476 and see [this Colab](https://colab.corp.google.com/drive/1CngW6pCIHH4yppyqsqnTfKXvkvqUzmIt)

andsteing (changes requested):

examples/lm1b/train.py
jaehlee: Thanks! I see that you've changed default `config.learning_rate` from `0.0625` to `0.002` while decreasing the default batch size as well. I requested access for the colab. To be consistent in the end, what's the learning_rate and batch_size used for results above?

andsteing: (sorry gave you access)

right so in your case it would be

```
learning_rate = 0.05
warmup_steps = 1000
num_train_steps = 500_000

lrmax = learning_rate / (warmup_steps ** .5)
lrmax
```

So 0.0015811388300841897 or the 0.0016 that you propose LGTM

andsteing (approved):

PR #1480, opened 2021-08-06
[WIP] Transformer Layers

cgarciae:
Hey! 

# What does this PR do?
Adds the `TransformerEncoderLayer`, `TransformerEncoder`, `TransformerDecoderLayer`, `TransformerDecoder`, and `Transformer` modules, their APIs and implementations are ported as faithful as possible from PyTorch (with obvious deviations when needed).

## Tasks
- [ ] Implementation
- [ ] Tests
- [ ] Docs
- [ ] Example

## Comments
Posting PR early to get some feedback, starting with if this addition is actually of interest to Flax, naming conventions, style, implementation, etc.

### DIFF
--- a/flax/nn/transformer.py
+++ b/flax/nn/transformer.py
@@ -1,0 +1,272 @@
+"""
+"""
+from dataclasses import dataclass
+from typing import Any, Callable, Optional
+
+import jax
+from jax._src.lax.lax import DType
+import jax.numpy as jnp
+import numpy as np
+from flax.linen.attention import MultiHeadDotProductAttention
+from flax.linen.linear import Dense
+from flax.linen.module import Module, compact
+from flax.linen.normalization import LayerNorm
+from flax.linen.stochastic import Dropout
+
+Dtype = Any
+
+
+def sequential(*layers) -> Callable[[np.ndarray], np.ndarray]:
+
+  def _lambda(x):
+    for layer in layers:
+      x = layer(x)
+    return x
+
+  return _lambda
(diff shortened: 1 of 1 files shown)


jheek:
Hi, we have so far refrained from adding Transformer to the main nn library.
The main purpose of the nn library is to provide the primitive building blocks of neural nets but not specific models.
Of course this is somewhat ambiguous and (Self)Attention for example was considered a "grey area" which eventually was added (but later simplified).
For new additions to the nn library it's best to first open a feature request issue so we can discuss whether it's in scope.

cgarciae:
I see. That makes sense. I'll leave it here for now in case someone wants to provide free feedback and will find it a home later.

avital:
Hi @cgarciae -- just doing some GitHub bookkeeping now. Sounds like we won't merge this, so I'll close the PR but hopefully others can find it if they search (if you want to edit the issue with the relevant keywords for what this enables that would help people find it)

DBraun:
Thanks for this code. I'm starting to use it. I noticed that these two lines should be swapped:

https://github.com/cgarciae/flax/blob/a4dc414385d1ef8365933f0ca4414c728ef25fcd/flax/nn/transformer.py#L248-L249

I also had to do some things with passing deterministic to Dropout layers. Could @cgarciae update it since you know it much better? It would be helpful to see it in https://github.com/google/flax/tree/main/flax/nnx/examples (although I haven't transitioned to nnx yet either)

PR #1485, opened 2021-08-12
Add jraph dependency for upcoming ogbg_molpcba example.

ameya98:
Redoing #1470

# What does this PR do?

Fixes build for ogbg_molpcba example.

## Checklist
      checks if that's the case).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ tests_require = [
     "clu",  # All examples.
     "gym",
     "jaxlib",
-    "jraph==0.0.1.dev0",
+    "jraph",
     "ml-collections",
     "opencv-python",
     "pytest",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -38,6 +38,7 @@
     "clu",  # All examples.
     "gym",
     "jaxlib",
+    "jraph==0.0.1.dev0",
```
andsteing, line 41: We try not to have explicit versions for the main `setup.py` but instead simply specify `jraph` for dependencies whenever possible. This simplifies maintenance.
(Note that existing explicit versions were added because without we would have had version conflicts; but I think `jraph` should work, please give it a try and update here if that's not the case)

andsteing (changes requested):

andsteing (approved):

PR #1516, opened 2021-09-01
Do not call `jnp` on shapes in AutoEncoder example

romanngg:
Otherwise can't jit/abstact-eval `model.apply`


README.md
```diff
@@ -137,13 +137,15 @@ output = model.apply(variables, batch)
 ```
 
 ```py
+import numpy as np
```
andsteing, line 140: Could you move that up to the other imports?

andsteing:
Thanks for the change!

PR #1519, opened 2021-09-04
Add nowrap decorator for module methods.

levskaya:
User-defined module methods are automatically wrapped with a state
management wrapper and, if profiling is enabled, with named_call.

This can interfere with constructor helper methods when they are
wrapped by named_call automatically, since "functionalized" methods
cannot return Modules or Variables.

Additionally for advanced cases of overriding e.g. the base param
method, one does not want to wrap such a method with the state
management wrapper.

So a @nn.nowrap method is introduced to prevent automatic wrapping
in those advanced use cases where it is detrimental.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -224,6 +224,42 @@ def compact(fun: _CallableT) -> _CallableT:
   return fun
 
 
+def nowrap(fun: _CallableT) -> _CallableT:
+  """Marks the given module method as a helper method that needn't be wrapped.
+
+  Methods wrapped in @nowrap are private helper methods that needn't be wrapped
+  with the state handler or a separate named_call transform.
+
+  This is needed in several concrete instances:
+   - if you have a helper method that returns Modules or Variables to prevent
+     it from being functionalized by named_call. (Functionalized methods
+     can't return Modules/Variables.)
+   - if you're subclassing a method like Module.param and don't want this
+     overriden core function decorated with the state management wrapper.
+   - If you want a method to be callable from an unbound Module (e.g.: a
+     function of construction of arguments that doesn't depend on params/RNGs)
+
+  For instance::
+
+    @nowrap
+    def _make_dense(self, num_features):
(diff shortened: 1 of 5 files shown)


flax/linen/module.py
```diff
@@ -224,6 +224,35 @@ def compact(fun: _CallableT) -> _CallableT:
...
+  """Marks the given module method as a helper method that needn't be wrapped.
+
+  Methods wrapped in @nowrap are private helper methods that needn't be wrapped
+  with the state handler or a separate named_call transform.
+
+  This is needed in two concrete instances:
+   - if you have a helper method that returns Modules or Variables to prevent
+     it from being functionalized by named_call. (Functionalized methods
+     can't return Modules/Variables.)
+   - if you're subclassing a method like param and don't want this overriden
```
jheek, line 237: Another case:
- If you want a method to be callable from an unbound Module (e.g.: a function of construction of arguments that doesn't depend on params/RNGs)

flax/linen/module.py
```diff
@@ -224,6 +224,35 @@ def compact(fun: _CallableT) -> _CallableT:
...
+     function decorated with the state management wrapper.
+
+  For instance::
+
+    @compact
```
jheek, line 242: Forgot to update example?

flax/linen/transforms.py
```diff
@@ -188,6 +189,14 @@ def set_scopes_inner(x):
   return new_module, new_args, new_kwargs
 
 
+def _test_transformed_return_values(tree, method_name):
```
jheek, line 192: nice!

jheek (changes requested):

flax/linen/module.py
levskaya: good addition, added it.

flax/linen/module.py
levskaya: fixed.

jheek (approved):

PR #265, opened 2020-05-15
Moves ModelParamTraversal to traverse_util and updates references

marcvanzee:

PR #1525, opened 2021-09-07
Separate ModelParamTraversal into traverse_util.

### LINKED ISSUE #62: Make `ModelParamTraversal` more public?
`ModelParamTraversal` is currently somewhat hidden within `optim`. But it is much more generally useful, for example for implementing weight-decay (not as a loss) or weight standardization or spectral norm (I think).
So it seems like putting it in `traverse_util.py` (where I'd look for it) would make sense.

levskaya:
flax.optim is on a long-term deprecation plan, move the more
generally useful ModelParamTraversal out of optim and into
traverse_util where it naturally belongs.

Resolves #62  and finishes #265

### DIFF
--- a/flax/optim/base.py
+++ b/flax/optim/base.py
@@ -30,6 +30,9 @@ from ..nn import base
 
 from ..core import FrozenDict, unfreeze
 
+# Backwards compatibility symbol import.
+ModelParamTraversal = traverse_util.ModelParamTraversal
+
 
 @struct.dataclass
 class OptimizerState:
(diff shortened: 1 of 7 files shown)


flax/optim/base.py
```diff
@@ -546,63 +539,3 @@ def update_hyper_params(self, **hyper_param_overrides):
...
       hps = [hp.replace(**hyper_param_overrides) for hp in hps]
     return hps
-
-
-def _sorted_items(x):
-  """Returns items of a dict ordered by keys."""
-  return sorted(x.items(), key=lambda x: x[0])
-
-
-class ModelParamTraversal(traverse_util.Traversal):
```
jheek, line 556: Maybe re-export ModelParamTraversal from flax.optim for backward compat?

jheek (approved):
Looks great! But consider adding a backward compatible re-export

levskaya:
@jheek - I did add a backwards compat re-export - it was just at the top of base.py

PR #1526, opened 2021-09-07
Clean up pytest.ini

### LINKED ISSUE #210: Investigate and fix cause of warnings defined in pytest.ini
With pytest.ini defined, all the warnings captured during a test execution are reported as exceptions. 

Although there are a few warnings which are independent of flax's implementation (for e.g: DeprecationWarning and UserWarning captured due to importing tensorflow.compat.v2.io import gfile). 

Ideally we should have ZERO entries as exemptions in pytest.ini. Investigate the root cause and fix them.

levskaya:
Lots of old cruft in here, trying to remove obsolete warnings.
This much-reduced set worked on my local machine.

Resolves #210


jheek (approved):
Thanks!

PR #1531, opened 2021-09-09
Fix lifting rng splits

### LINKED ISSUE #1530: `FLAX_PROFILE=1` changes the computation

```
run.py:
print(nn.Dense(1, use_bias=False).init({'params': jax.random.PRNGKey(0)}, np.ones((1, 1), np.float32)))

FLAX_PROFILE= python run.py -> -1.4588
FLAX_PROFILE=1 python run.py -> 1.3333
```

I would expect a profiling option like `FLAX_PROFILE` to not change any computation.

jheek:
Fixes #1530 

- Dynamically check whether named call is enabled.
   Previously enable/disable named call only affected Module construction. This is unexpected and less flexible than checking 
   the global during calls to named_call.

- This commit also introduces override_named_call which locally overrides named_call behavior.
   This is particullary useful for tests that depend on a specific setting.

- Remove unnecessary splitting of rngs in lifting logic
  Previously we split every rng that was lifted. Now the rng_counters are shared between the lifted and unlifted scope.
  This is valid because make_rng can only be called a static number of times and lifted loops/vmap split the rngs manually 
  already

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -109,12 +109,16 @@ def pack(fn: Callable[..., Any],
           if not col_in_out:
             variable_group[col_name] = freeze(collection)
     rng_groups_xs = []
+    inner_rng_counters = []
     for scope in scopes:
+      rng_counters = {}
       rng_groups = group_collections(scope.rngs, rng_filters)
       for rng_group in rng_groups:
         for kind in rng_group:
-          rng_group[kind] = scope.make_rng(kind)
+          rng_group[kind] = scope.rngs[kind]
+          rng_counters[kind] = scope.rng_counters[kind]
       rng_groups_xs.append(rng_groups)
+      inner_rng_counters.append(rng_counters)
     rng_groups_xs_t = _transpose(rng_groups_xs)
 
     inner_scopes = []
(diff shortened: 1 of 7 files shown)


flax/linen/module.py
```diff
@@ -541,10 +553,10 @@ def _wrap_module_methods(cls):
       if hasattr(method, 'nowrap'):
         continue
       wrapped_method = wrap_method_once(method)
-      if _use_named_call and key != 'setup':
+      if key != 'setup':
```
levskaya, line 556: only downside here is increased stack depth - but we can add stack trace filtering soon.

levskaya (approved):
Thanks!

flax/core/lift.py
```diff
@@ -109,12 +109,16 @@ def wrapper(scope_tree: Scope, *args, **kwargs):
...
             variable_group[col_name] = freeze(collection)
     rng_groups_xs = []
+    inner_rng_counters = []
     for scope in scopes:
+      rng_counters = {}
       rng_groups = group_collections(scope.rngs, rng_filters)
       for rng_group in rng_groups:
         for kind in rng_group:
-          rng_group[kind] = scope.make_rng(kind)
+          rng_group[kind] = scope.rngs[kind]
```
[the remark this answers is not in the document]
jheek: yes the issue is only about named_call but the fix is general

flax/linen/module.py
jheek: Named call might actually be a context manager soon. So hopefully that will also clean up (and speed up!) tracing

PR #1535, opened 2021-09-10
Remove Implicit rank promotion

### LINKED ISSUE #1534: Avoid rank promotion in Flax codebase
Silent rank promotion can conceal errors that would otherwise throw exception, so I always prefer to run my code with
```
from jax.config import config
config.update("jax_numpy_rank_promotion", "raise")
```
This way an exception is thrown if rank promotion happens anywhere (https://jax.readthedocs.io/en/latest/rank_promotion_warning.html).

Unfortunately this appears incompatible with flax, since it uses rank promotion in biases, normalization, some examples etc, so I can't use the above setting to catch unintentional rank promotion in my code.

It would be very nice if flax could be rank promotion-free. Thanks a lot for the great library!

jheek:
- users can now disable implicit rank promotion in JAX without Flax throwing errors.
- Tests are no run with JAX_NUMPY_RANK_PROMOTION=raise
- This change uncovered 2 faulty LSTM tests.

Fixes #1534

### DIFF
--- a/tests/run_all_tests.sh
+++ b/tests/run_all_tests.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+export JAX_NUMPY_RANK_PROMOTION=raise
 export FLAX_PROFILE=1
 
 PYTEST_OPTS=
@@ -42,7 +43,7 @@ handle_errors () {
 sphinx-build -M doctest docs docs/_build
 
 # Run battery of core FLAX API tests.
-pytest -n 4 tests $PYTEST_OPTS
+pytest -n auto tests $PYTEST_OPTS
 
 # Per-example tests.
 #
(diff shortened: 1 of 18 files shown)


tests/run_all_tests.sh
```diff
@@ -42,7 +43,7 @@ handle_errors () {
 sphinx-build -M doctest docs docs/_build
 
 # Run battery of core FLAX API tests.
-pytest -n 4 tests $PYTEST_OPTS
+pytest -n auto tests $PYTEST_OPTS
```
levskaya, line 46: didn't know about `auto`

levskaya (approved):
This all looks great!  Thanks so much for isolating all of these.

tests/run_all_tests.sh
jheek: We had it before, It was accidentally removed in the linen refactor

PR #1536, opened 2021-09-10
Fix PixelCNN example

juliuskunze:
Fixes `TypeError: eval_step() got multiple values for argument 'config'` on running the PixelCNN example with default config.


examples/pixelcnn/train.py
```diff
@@ -231,7 +231,7 @@ def train_and_evaluate(config: ml_collections.ConfigDict, workdir: str):
         # Load and shard the TF batch
         eval_batch = load_and_shard_tf_batch(eval_batch)
         # Step
-        metrics = p_eval_step(ema, eval_batch)
+        metrics = p_eval_step(params=ema, batch=eval_batch)
```
jheek, line 234: can we just use partial(eval_step, config) instead of partial(eval_step, config=config) to be consistent with train_step?

PR #1537, opened 2021-09-12
Enable traceback filtering of flax internal frames.

levskaya:
Add an option to remove Flax internal frames from tracebacks
for more parseable exception stack traces.  Allow users to
enable/disable flax filtering by environment variable or dynamically
with simple function calls.

### DIFF
--- a/flax/traceback_util.py
+++ b/flax/traceback_util.py
@@ -1,0 +1,57 @@
+# Copyright 2021 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


levskaya:
One main question here is whether to filter flax internal frames by default.  Probably?

flax/traceback_util.py
```diff
@@ -0,0 +1,78 @@
...
+  elif val in ('n', 'no', 'f', 'false', 'off', '0'):
+    return False
+  else:
+    raise ValueError(
+        "invalid truth value %r for environment %r" % (val, varname))
+
+
+# Globals:
+# Whether to filter flax frames from traceback.
+_flax_filter_tracebacks = bool_env('FLAX_FILTER_FRAMES', True)
```
jheek, line 45: Maybe put the configs filter tracebacks & named_call in a dedicated flax/config.py

jheek (commented):
> One main question here is whether to filter flax internal frames by default. Probably?

Yes, let's enable by default

PR #1545, opened 2021-09-16
Unify normalization layers

jheek:
- Refactor normalization helpers to share code and remove subtle differences in implemenation.
- Use clipping to avoid negative variances due to fusion optimization.


levskaya (approved):
This is great, thanks for unifying all this.

PR #1548, opened 2021-09-16
make `save_checkpoint` support non-blocking mode

### LINKED ISSUE #1543: `flax.training.checkpoints.save_checkpoint` is blocking
[flax.training.checkpoints.save_checkpoint](https://github.com/google/flax/blob/1dfc64b6203ab4e69d55bd9e9365c5410c5fdad6/flax/training/checkpoints.py#L93) default behavior is blocking, this becomes a problem as we scale up, specially if we save the checkpoints to GCS or any high latency storage. if we scale up by "napkins math" for a v3-2048 it can be up to 40% of the total wall time (my reference point is ~25 mins on a v3-128)

**EDIT: The read is from a US bucket to a TPUv3-8 in eu4  and write to bucket is in eu4 as well.**

Consider the below code that simulates the imagenet MLPerf run on a v3-2048:

### Steps to reproduce:
```python

from flax.training import checkpoints
from threading import Thread
import time
from tqdm import tqdm

state=None
save_times=80 # save every ~ 10 epochs
compute=5 # ~ 10 epochs of compute on v3-2048 it should be "4.6875" but ~ should be good enough to size the problem. 
readable_bucket_path="gs://flax_public/examples/imagenet/tpu_v3_32"

writeable_bucket="YOUR_BUCKET"
writable_path=f"gs://{writeable_bucket}/perf"

print("Reading parallel GCS","="*30)
state=checkpoints.restore_checkpoint(readable_bucket_path,state,parallel=True)


print("Starting blocking writes to bucket")

total_write=0
# Writing in blocking mode
tic_total = time.time()

for step in tqdm(range(save_times)):
    tic_write = time.time()
    checkpoints.save_checkpoint(writable_path,state,step=step,keep=3)
    toc_write = time.time()
    total_write += (toc_write-tic_write)
    # simulating a compute load of 5 secs because otherwise the writes will be get caught in racing condition. 
    # and sometimes they do anyway because of the non deterministic nature of the GCS access time (which in its own a tell on how slow GCS can be!)
    time.sleep(compute)
toc_total = time.tic = time.time()
blocking_total = toc_total - tic_total
print("Total time on blocked GCS write:",blocking_total)
print("Total write time on blocked GCS write:",total_write)

# =======================================

print("Starting non-blocking writes to bucket")

blocked_write=total_write


total_write=0
ts=[]

# Writing in non-blocking mode

tic_total = time.time()
for step in tqdm(range(save_times,save_times*2)):
    tic_write = time.time()
    thread = Thread(target = checkpoints.save_checkpoint, kwargs= {'ckpt_dir':writable_path,'target':state,'step':step,'keep':3})
    thread.start()
    toc_write = time.time()
    total_write += (toc_write-tic_write)
    # simulating a compute load of 5 secs same as above.
    time.sleep(compute)
    ts.append(thread)
    
    


print("Threaded on GCS before join:",total_write)

for t in tqdm(ts):
    
    t.join()

toc_total= time.time()
non_blocking_total = toc_total - tic_total
non_bocked_write=total_write
print("Time on non-blocked GCS write",non_blocking_total)
print(f"Total speedup ={blocking_total/non_blocking_total}X")
print(f"Total write speedup ={blocked_write/non_bocked_write}X")
```
yields
```
Reading parallel GCS ==============================
Starting blocking writes to bucket
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 80/80 [09:31<00:00,  7.14s/it]
Total time on blocked GCS write: 571.1965103149414
Total write time on blocked GCS write: 170.71105670928955
Starting non-blocking writes to bucket
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 80/80 [06:40<00:00,  5.01s/it]
Threaded on GCS before join: 0.09925699234008789
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 80/80 [00:00<00:00, 49947.06it/s]
Time on non-blocked GCS write 400.5715479850769
Total speedup =1.4259537732725367X
Total write speedup =1719.8894776538862X
```
I'll try to pull on this tonight or tomorrow evening to unblock `flax.training.checkpoints.save_checkpoint`

mjsML:
closes #1543

### DIFF
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -19,6 +19,7 @@ other numerical metric in filename.  Cleans up older / worse-performing
 checkpoint files.
 """
 
+from concurrent.futures._base import Future
 from concurrent.futures import thread
 import os
 import re
@@ -95,12 +96,15 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
                       prefix: str = 'checkpoint_',
                       keep: int = 1,
                       overwrite: bool = False,
-                      blocking:bool = True) -> str:
+                      keep_every_n_steps: Optional[int] = None,
+                      blocking:bool = True) -> Union[str,Future]:
     """Save a checkpoint of the model.
 
     Attempts to be pre-emption safe by writing to temporary before
     a final rename and cleanup of past files.
-
+    A note about parallism: This function is blocking by default (see the `blocking` argument). 
+    It's generally a good idea to try to unblock the compute thread as soon as possible to avoid blocking the training.
+    Examples of blocking operations: logging, saving model, etc.
     Args:
       ckpt_dir: str or pathlib-like path to store checkpoint files in.
       target: serializable flax object, usually a flax optimizer.
@@ -109,26 +113,31 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
       keep: number of past checkpoint files to keep.
       overwrite: overwrite existing checkpoint files if a checkpoint
         at the current or a later step already exits (default: False).
-      blocking: bool: if True, wait for the file write to be completed before returning (default: True).
+      keep_every_n_steps: if defined, keep every checkpoints every n steps (in
+      addition to keeping the last 'keep' checkpoints).
+      blocking: bool: `blocking=True` will block until the checkpoint is written (default behaviour).
+      `blocking=False` will return a future that one needs to call `result()` on to make sure the checkpoint is written or not.
+        This is helpful to unblock the compute thread while the checkpoint is being written.
+      if `blocking=False`, you can't use `overwrite=True` as it will raise a `OverwriteWithNonBlockingError`.
     Returns:
-      Filename of saved checkpoint or concurrent.futures._base.Future (the handle) if blocking is False.
+      Filename of saved checkpoint or concurrent.futures._base.Future (the handle) if `blocking=False`.
     """
     _kwargs = locals()
-    del _kwargs['blocking']
-    
     def _save_checkpoint(ckpt_dir: Union[str, os.PathLike],
                         target: PyTree,
                         step: int,
                         prefix: str = 'checkpoint_',
                         keep: int = 1,
-                        overwrite: bool = False) -> str:
+                        overwrite: bool = False,
+                        blocking=True) -> Union[str,Future]:
 
       ckpt_dir = os.fspath(ckpt_dir)  # Pathlib -> str
       # Write temporary checkpoint file.
       logging.info('Saving checkpoint at step: %s', step)
       # normalize path because gfile.glob() can modify path './', '//' ...
       ckpt_dir = safe_normpath(ckpt_dir)
-      ckpt_tmp_path = _checkpoint_path(ckpt_dir, 'tmp', prefix)
+      # Add the step number to the tmp checkpoint path to avoid overwritting in case of non-blocking.
+      ckpt_tmp_path = _checkpoint_path(ckpt_dir, 'tmp_'+str(step), prefix)
       ckpt_path = _checkpoint_path(ckpt_dir, step, prefix)
       gfile.makedirs(os.path.dirname(ckpt_path))
       base_path = os.path.join(ckpt_dir, prefix)
@@ -157,7 +166,7 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
       logging.info('Saved checkpoint at %s', ckpt_path)
 
       # Remove newer checkpoints
-      if overwrite:
+      if overwrite and blocking:
         ind = checkpoint_files.index(ckpt_path) + 1
         newer_ckpts = checkpoint_files[ind:]
         checkpoint_files = checkpoint_files[:ind]
@@ -166,13 +175,24 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
           try:
             gfile.remove(path)
           except errors.NotFoundError:
-            # Note this is likely to happen if this write was slower than the next one, but it doesn't matter if thread i or i+n deletes it.
-            logging.info('Removing checkpoint at %s failed, likely due to it being already deleted', path)
-
+            logging.info('Removing checkpoint at %s failed', path)
+      elif overwrite and not blocking:
+         logging.info('overwriting with non-blocking mode is not supported')
+         raise errors.OverwriteWithNonBlockingError(step)
       # Remove old checkpoint files.
+      last_kept = -float('inf')
       if len(checkpoint_files) > keep:
         old_ckpts = checkpoint_files[:-keep]
+        # Note: old_ckpts is sorted from oldest to newest.
         for path in old_ckpts:
+          if keep_every_n_steps:
+            step_number = _checkpoint_path_step(path)
+            if step_number and (step_number - last_kept) >= keep_every_n_steps:
+              logging.debug('Not deleting %s, because last_kept=%f and keeping '
+                            'every %d steps.',
+                            path, last_kept, keep_every_n_steps)
+              last_kept = step_number
+              continue
           logging.info('Removing checkpoint at %s', path)
           try:
             gfile.remove(path)
@@ -189,6 +209,9 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
     else:
       # if blocking is False, we will return the handle to the future.
       # NOTE: we set the max_workers to 1 because the work is serial but we want to have it done asynchronously.
+      if overwrite and not blocking:
+         logging.info('overwriting with non-blocking mode is not supported')
+         raise errors.OverwriteWithNonBlockingError(step)
       executor=thread.ThreadPoolExecutor(max_workers=1)
       future = executor.submit(_save_checkpoint, **_kwargs)
       return future
--- a/flax/training/checkpoints.py
+++ b/flax/training/checkpoints.py
@@ -129,6 +129,7 @@ def save_checkpoint(ckpt_dir: Union[str, os.PathLike],
                         prefix: str = 'checkpoint_',
                         keep: int = 1,
                         overwrite: bool = False,
+                        keep_every_n_steps: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


flax/training/checkpoints.py
```diff
@@ -89,92 +89,109 @@ def safe_normpath(path: str) -> str:
...
+      return ckpt_path
+
+    # Now we check if we are writing in blocking mode or not.
+    if blocking:
+      # if blocking is True, we will wait for the file write to be completed before returning.
+      return _save_checkpoint(**_kwargs)
+    else:
+      # if blocking is False, we will return the handle to the future.
+      # NOTE: we set the max_workers to 1 because the work is serial but we want to have it done asynchronously.
+      executor=thread.ThreadPoolExecutor(max_workers=1)
```
jheek, line 192: Because you create the executor inside save_checkpoint calling it while the previous checkpoint is still being stored results in a race. I'm not convinced we should make this a keyword argument. Why not add this snippet with the executor in the save_checkpoint docstring with a note explaining users about the performance implications of blocking vs non-blocking storage? This way users can also use the thread pool for other blocking stuff like serializing a tf.data iterator, storing metrics, etc.

jheek (changes requested):

flax/training/checkpoints.py
mjsML: @jheek 
On the doc string makes sense, will do.

On adding it as a keyword arg and the use of the executor, I'm following the design pattern of `restore_checkpoint` https://github.com/google/flax/blob/f458637584a460c139339e29fc0fe70c9ca15d49/flax/training/checkpoints.py#L253-L278 which adds the `parallel` keyword arg. I believe it needs to be a keyword argument because of the performance implications (mentioned in detail in #1543)

On the race, this is the result of the storage itself being slow to the point the compute of thread i+n might finish writing before thread i complete the write!... when the race happens the effect is thread i tries to delete an old checkpoint that doesn't exist anymore because thread i+n already did (which is not a bad thing , it doesn't matter who deletes checkpoint i-keep). Also this is handled in try-expect blocks around the gfile.remove https://github.com/google/flax/blob/1f7db1e19299655dd88c601c821c0bd129e62130/flax/training/checkpoints.py#L166-L181

But I see your point if this is used with the `overwrite=True` param.  I'll add more tests on my end and figure out how to support both `overwrite = True` and `overwrite=False`



thoughts?
Thanks again for the feedback!

jheek: The benefit of parallel is that the multithreading is contained within the restore (restore_checkpoint itself is always blocking). Also for save_checkpoint we need to be a bit more careful because it is usually in the train loop.

It might be worth adding a blocking=False option if we can avoid the race condition indeed. Although I'm still a bit unsure if the additional complexity is worth it. Keeping the thread pool in the user's train loop has the added benefit of letting users add their own blocking code to the thread pool. It is also more explicit so it should make it easier to reason about the code and understand its performance.

mjsML: I see your point about the complexity vs. the user's training loop ... I suggest we add a "block until" in the imagenet example (because IIUC that is the "example") ala https://github.com/google/flax/blob/f458637584a460c139339e29fc0fe70c9ca15d49/examples/imagenet/train.py#L380-L381
so something like
```python
for step in steps:
    future = save_checkpoint(...)
    futures.append(future)
...
# Wait until all file writes are done before exiting
[f.result() for f in futures]
# Wait until computations are done before exiting 
jax.random.normal(jax.random.PRNGKey(0), ()).block_until_ready() 
```
even 10% decrease of your wall time on a v3-128 for the imagenet example is worth it IMHO 😅 
And mind you, the default behavior is `blocking=True` to make sure it is indeed backward compatible without any surprises.

mjsML:
@jheek please review the latest commit 😄 
Thanks again in advance for the feedback.

mjsML:
Testing this on a v3-128, it seems that the IO for saving checkpoints is not the "biggest" factor in the ~10 secs between epochs / save:

```
I0917 00:47:08.033149 139687611345984 train.py:371] eval epoch: 9, loss: 4.1925, accuracy: 20.71
I0917 00:47:08.035786 139640577054464 logging_writer.py:35] [390] eval_accuracy=0.207061767578125, eval_loss=4.192461967468262
I0917 00:47:08.041632 139671520745216 logging_writer.py:35] [390] eval_accuracy=0.207061767578125, eval_loss=4.192461967468262
I0917 00:47:08.046746 140078102722304 logging_writer.py:35] [390] eval_accuracy=0.207061767578125, eval_loss=4.192461967468262
I0917 00:47:08.047906 140309568485120 logging_writer.py:35] [390] eval_accuracy=0.207061767578125, eval_loss=4.192461967468262
I0917 00:47:08.386637 140237985668160 checkpoints.py:213] writing in non-blocked mode
I0917 00:47:08.387206 140221541209856 checkpoints.py:137] Saving checkpoint at step: 390
I0917 00:47:08.387345 140237985668160 checkpoints.py:219] Dispatched the write operation, returning to compute thread now.
I0917 00:47:11.983901 139610051663616 logging_writer.py:35] [400] steps_per_second=2.806781, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.983829 139955167762176 logging_writer.py:35] [400] steps_per_second=2.806773, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.984067 140331489707776 logging_writer.py:35] [400] steps_per_second=2.806808, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.983868 140362236856064 logging_writer.py:35] [400] steps_per_second=2.806756, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.984073 139926848423680 logging_writer.py:35] [400] steps_per_second=2.806756, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.984109 139773167105792 logging_writer.py:35] [400] steps_per_second=2.806770, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.983339 140221557995264 logging_writer.py:35] [400] steps_per_second=2.806858, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.983726 139869607282432 logging_writer.py:35] [400] steps_per_second=2.806781, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.984069 140217105979136 logging_writer.py:35] [400] steps_per_second=2.806765, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.984224 140162594383616 logging_writer.py:35] [400] steps_per_second=2.806756, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.983890 139671520745216 logging_writer.py:35] [400] steps_per_second=2.807887, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.984282 139743318656768 logging_writer.py:35] [400] steps_per_second=2.808285, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.985023 140028843366144 logging_writer.py:35] [400] steps_per_second=2.806693, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.987699 139640577054464 logging_writer.py:35] [400] steps_per_second=2.807158, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.998755 140078102722304 logging_writer.py:35] [400] steps_per_second=2.807433, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:11.999929 140309568485120 logging_writer.py:35] [400] steps_per_second=2.807758, train_accuracy=0.27735474705696106, train_learning_rate=10.668883323669434, train_loss=3.6476545333862305
I0917 00:47:12.265046 140221541209856 checkpoints.py:167] Saved checkpoint at gs://BUCKET/non_blocking_imagenet_v3-128_8/checkpoint_390
I0917 00:47:12.265341 140221541209856 checkpoints.py:197] Removing checkpoint at gs://BUCKET/non_blocking_imagenet_v3-128_8/checkpoint_273
I0917 00:47:12.537031 140221541209856 checkpoints.py:220] Write operation completed.
I0917 00:47:21.934853 139942941269056 train.py:371] eval epoch: 10, loss: 4.2295, accuracy: 21.48
I0917 00:47:21.935297 139926848423680 logging_writer.py:35] [429] eval_accuracy=0.21478271484375, eval_loss=4.229455947875977
I0917 00:47:21.934717 139626147744832 train.py:371] eval epoch: 10, loss: 4.2295, accuracy: 21.48
```
but for most checkpoints it looks like:

```

```
Blocking 
END : 23:42:05.964823
START : 23:19:04.779000

~23:01

Non blocking 100 epochs / 10 checkpoints:
END : 01:53:58.942568
START :01:31:01.073212 

~22:57


Non blocking 100 epochs / 100 checkpoints:
END : 02:39:56.446720
START :02:17:00.443741

~22:56

I will probably have to profile the JAX code to see exactly where is it blocking.
I suspect there might be a bug with my code, so I'm marking this PR as draft for now.

marcvanzee:
@jheek have you had time to review this PR already>

marcvanzee:
@mjsML are you still planning on working on this? Otherwise @IvyZX is interested as well (she has a few improvements in mind based on your PR).

PR #1595, opened 2021-10-03
Fix typos.

dan-zheng:
- `form TensorFlow` → `from TensorFlow`
- `libraries than can be` → `libraries that can be`

---

- [x] This PR fixes a minor issue (e.g. typo or small bug) or improves the docs (you can dismiss the other checks if that's the case).


levskaya (approved):
Thank you!

PR #1621, opened 2021-10-16
Fixes RST link formatting.

andsteing:
Unfortunately, "source formatted" links like

see [`optax`](https://optax.readthedocs.io/)

are not supported in RST:

- https://github.com/spatialaudio/nbsphinx/issues/301
- https://docutils.sourceforge.io/FAQ.html#is-nested-inline-markup-possible

So removing the "source formatting" so these links render as normal text (but
with correct link formatting) in both Colab and RTD.

This will close overdue #659


docs/notebooks/annotated_mnist.ipynb
```diff
@@ -70,12 +70,14 @@
    "source": [
     "## 2. Define network\n",
     "\n",
+    "test:\n",
+    "\n",
```
marcvanzee, line 74:
```suggestion
```

marcvanzee (approved):

docs/notebooks/annotated_mnist.ipynb
andsteing: Done.

PR #1657, opened 2021-11-02
Fix GroupNorm type hint for param num_groups.

lkhphuc:
# What does this PR do?
In GroupNorm, `num_group` can be set to None if `group_size` is set. However, currently doing so the linter (LSP) will throw an error: "Argument of type None cannot be assigned to parameter "num_group" of type "int" ..."

This PR fix that.

## Checklist
- [x] This PR fixes a minor issue (e.g.: typo or small bug) or improves the docs

### DIFF
--- a/flax/linen/normalization.py
+++ b/flax/linen/normalization.py
@@ -293,7 +293,7 @@ class GroupNorm(Module):
       bias_init: Initializer for bias, by default, zero.
       scale_init: Initializer for scale, by default, one.
   """
-  num_groups: int = 32
+  num_groups: Optional[int] = 32
   group_size: Optional[int] = None
   epsilon: float = 1e-6
   dtype: Any = jnp.float32


flax/linen/normalization.py
```diff
@@ -293,7 +293,7 @@ class GroupNorm(Module):
       bias_init: Initializer for bias, by default, zero.
       scale_init: Initializer for scale, by default, one.
   """
-  num_groups: int = 32
+  num_groups: int | None = 32
```
jheek, line 296: nit: Optional[None] so it is in line with group_size type hint?

lkhphuc: That would make both num_groups and group_size optional, which isn't the case right? num_groups can only be None when group_size is not specified.

jheek: I don't follow. How does the current PR enforce that num_groups or group_size is specified?

lkhphuc: Yeah you're right. I was confused. Forced push with the Optional[int] style.

jheek (approved):

PR #1692, opened 2021-11-30
Fix child scope rng reuse.

jheek:
See test cases for bug repro.

Previously rng counters of child scopes would be forgotten when the parent was rewound or lifted.

This is fixed by tracking the rng counters recursively
from the root scope.


levskaya (approved):
Looks good!  Thanks for the fix!

PR #1696, opened 2021-12-02
Remove cyclic refs from scope

jheek:


flax/core/scope.py
```diff
@@ -466,9 +467,9 @@ def push(self,
                   name=name,
                   rngs=rngs,
                   parent=self,
+                  mutable=self.mutable,
                   path=self.path + (name,))
     scope.rng_counters = rng_counters
-    self._children[name] = scope
```
levskaya, line 471: Just double checking the reasoning here - this doesn't break scope aliasing in lift.py's `pack` because the child scopes still pull variables from their parents and thus the object identity of the child scopes is immaterial.

levskaya (approved):

flax/core/scope.py
jheek: Yes, actually this wasn't true before due the the Scope holding the rng counter state but that was fixed with the last PR about the duplicate RNGs. Relying on Scope identity was always a brittle design I think

PR #1698, opened 2021-12-06
Add error for empty scopes.

jheek:
Before, a ParamNotFoundError was raised.
This specializes the error to indicate the collection doesn't exist at all.
This should be easier to debug for users.

### DIFF
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -182,7 +182,7 @@ class ScopeCollectionNotFound(FlaxError):
     For example, you might have used ``module.apply(params, ...)`` instead
     of ``module.apply({'params': params}, ...)``.
   2. The collection is empty because the variables need to be initialized.
-    In this case, you have should made the collection mutable during
+    In this case, you should have made the collection mutable during
     apply (e.g.: ``module.apply(variables, ..., mutable=['state'])``.
   """
   def __init__(self, col_name, var_name, scope_path):
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -182,7 +182,7 @@ class ScopeCollectionNotFound(FlaxError):
     For example, you might have used ``module.apply(params, ...)`` instead
     of ``module.apply({'params': params}, ...)``.
   2. The collection is empty because the variables need to be initialized.
-    In this case, you have should make the collection mutable during
+    In this case, you have should made the collection mutable during
     apply (e.g.: ``module.apply(variables, ..., mutable=['state'])``.
   """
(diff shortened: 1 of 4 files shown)


flax/errors.py
```diff
@@ -173,6 +173,23 @@ def __init__(self, param_name, scope_path):
...
+class ScopeCollectionNotFound(FlaxError):
+  """
+  This error is thrown when trying to access a variable from an empty collection.
+
+  There are two common causes:
+  1. The collection was not passed to ``apply`` correctly.
+    For example, you might have used ``module.apply(params, ...)`` instead
+    of ``module.apply({'params': params}, ...)``.
+  2. The collection is empty because the variables need to be initialized.
+    In this case, you have should make the collection mutable during
```
marcvanzee, line 185:
```suggestion
    In this case, you have should made the collection mutable during
```

tests/core/core_scope_test.py
```diff
@@ -160,5 +160,14 @@ def f(scope):
     b = root.child(f)()
     self.assertFalse(jnp.allclose(a, b))
 
+  def test_empty_col_error(self):
+    root = Scope({})
+    with self.assertRaises(errors.ScopeCollectionNotFound):
+      root.param('test', nn.initializers.zeros)
+    root = Scope({'params': {}})
+    with self.assertRaises(errors.ScopeCollectionNotFound):
+      root.param('test', nn.initializers.zeros)
```
marcvanzee, line 169: In theory this error should be raised for any collection, but you are only testing it for the collection `params`. Maybe you could add a test doing something like `root.variable('col', 'test', nn.initializers.zeros)`?

marcvanzee (changes requested):

flax/errors.py
```diff
@@ -173,6 +173,23 @@ def __init__(self, param_name, scope_path):
...
+class ScopeCollectionNotFound(FlaxError):
+  """
+  This error is thrown when trying to access a variable from an empty collection.
+
+  There are two common causes:
+  1. The collection was not passed to ``apply`` correctly.
+    For example, you might have used ``module.apply(params, ...)`` instead
+    of ``module.apply({'params': params}, ...)``.
+  2. The collection is empty because the variables need to be initialized.
+    In this case, you have should made the collection mutable during
```
marcvanzee, line 185:
```suggestion
    In this case, you should have made the collection mutable during
```

marcvanzee (approved):
Thanks for adding this error message!

PR #1711, opened 2021-12-11
Fix failing lm1b tests.

levskaya:
Fix lm1b tests failing on a non-important JAX warning (from XLA: non usable donated buffers).


marcvanzee:
Sorry, I was out when this was sent for review, but thanks for the quick fix @levskaya!

PR #1720, opened 2021-12-13
add __eq__ and __hash__ for scopes

jheek:
This fixes a bug in dedup_scopes/dup_scopes which caused duplicate scopes
to be maintained because they are different
scope instances pointing to  the same data.


levskaya (dismissed):

levskaya (commented):
This _looks_ correct, but I see one failing test having to do w. the weird outer module "adoption" test cases.

levskaya (approved):

PR #1723, opened 2021-12-13
Implement LazyRNG

jheek:
No longer fold in static components of rngs (module names and rng counters) one-by-one
instead maintain a lazy rng which has a
jax prng and a suffix of static data that is
folded in only when make_rng is called

this should lead to better performance and much less cluter in jaxpr's and HLO.

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -23,6 +23,7 @@ from typing import Any, Callable, Container, Dict, Generic, Iterable, Mapping, O
 from . import tracers
 from flax import errors
 from flax import traceback_util
+from flax import struct
 from .frozen_dict import freeze
 from .frozen_dict import FrozenDict
 from .frozen_dict import unfreeze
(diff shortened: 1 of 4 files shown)


flax/core/scope.py
```diff
@@ -86,7 +104,13 @@ def _fold_in_str(rng: PRNGKey, data: str) -> PRNGKey:
    The newly generated PRNG key.
   """
   m = hashlib.sha1()
-  m.update(data.encode('utf-8'))
+  for x in data:
+    if isinstance(x, str):
+      m.update(x.encode('utf-8'))
+    elif isinstance(x, int):
+      m.update(x.to_bytes((x.bit_length() + 7) // 8, byteorder='big'))
```
levskaya, line 111: But will this match previous PRNG results?  This doesn't look the same as a jax threefry `fold_in`?  If we're going to suddenly change everyone's PRNG tree we may want to at least advertise this fact if we can't avoid it.  But given the actual use is always (str0, str1, ...) + (int,) we might preserve old PRNG by directly calling fold_in directly on ints instead of folding everything into the accumulated sha1 hash?

jheek: Yeah I thought that at first as well but then I realized that we are changing the rng anyway because we now take the sha1 hash of (str0, str1, ...) instead of folding in the sha1 hash of each string which already gives us a different RNG to beging with (before folding in the rng counter). In lifting there is also a subtle change because rngs will be split (vmap, scan) *before* the static part is folded in.

So lazy rng is going to affect change RNGs regardless because you cannot statically threefry the base rng with a string and threefry hash is not a associative operation (we cannot statically threefry the strings together first)

levskaya: Aha, yeah, I see.  Let's keep this simple new version then.  I guess we should just announce this change in case anyone optimized their seed as a hyperparameter, lol. :p

levskaya (approved):

flax/core/scope.py
jheek: It's really hard to guarantee determinism in a forward compatible way. XLA in particular can easily cause different round of errors due to changes in optimization. So updating JAX/flax dependencies should often change training results slightly.

flax/core/scope.py
```diff
@@ -71,9 +72,26 @@ class DenyList:
...
 MutableVariableDict = Dict[str, MutableCollection]
 
+PRNGFoldable = Union[int, str]
 
-def _fold_in_str(rng: PRNGKey, data: str) -> PRNGKey:
-  """Folds a string into a jax.random.PRNGKey using its SHA-1 hash.
+
+class LazyRng(struct.PyTreeNode):
+  rng: PRNGKey
+  suffix: Tuple[Any, ...] = struct.field(pytree_node=False)
```
marcvanzee, line 80: Nit: Shouldn't this have type `Iterable[PRNGFoldable]`, matching the type of `_fold_in_static`'s second argument?

marcvanzee (approved):
Nice!

flax/core/scope.py
jheek: It needs to be a tuple to make concat easy but Any should be PRNGFoldable indeed

levskaya:
@jheek - curious - are there any outstanding reasons we can't get this in?

PR #1745, opened 2021-12-30
Handle setup with transformed methods taking submodules of self.

levskaya:
For a setup-style module with helper methods, under named_call or other
transforms we previously would raise a name-reuse error when a
helper-method took a self.some_submodule as a method argument.  This
fixes scope-reuse behavior of setup under transforms to avoid this bug.

For instance the simplest breaking example, reduced from a user bug, is:
```
    class Foo(nn.Module):
      def setup(self):
        self.inner = nn.Dense(2)
      @nn.jit
      def helper(self, x, m):
        return m(x)   # we don't even need to call 'm' to trigger the issue!
      @nn.jit
      def __call__(self, x):
        return self.helper(x, self.inner)
    vs_bar = Foo().init(k, x)
```
Here, we pass local submodule `self.inner` as a helper-method argument.  This means that inside a transform `dedup_scopes` will already create the child scope for `self.inner` from the scope of `self`.  However, when `setup` is called again inside the transform this causes a name-reuse error.  This only occurs with setup-style modules, as `setup` is called anew at each transform boundary.

To solve this we generalize the `setup_called` state to be ternary, including an `INHERIT` state that allows scope reuse inside of `setup` during transforms if `setup` has already been called outside the transform boundary.  (name-reuse checks are still performed on the code inside `setup` the first time the module is run before the transform by abstract evaluation of the `setup` method.)

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -15,6 +15,7 @@
 """Flax Modules."""
 from contextlib import contextmanager
 import dataclasses
+import enum
 import functools
 import inspect
 import os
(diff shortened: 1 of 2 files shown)


levskaya:
Addresses #1746

jheek (approved):
Thanks for figuring this out!

PR #1753, opened 2022-01-04
Makes `FLAX_PROFILE=true` gotcha glaringly obvious.

andsteing:
Improves `TransformedMethodReturnValueError` description by explicitly mentioning `FLAX_PROFILE=true` environment variable.

This error can be quite confusing for example when it only happens when deploying the model into an environment with different environment variables so this should be mentioned explicitly.


jheek (approved):
thanks!

levskaya (approved):

PR #1774, opened 2022-01-11
Adds optax update guide.

andsteing:
This PR adds a detailed guide how `flax.optim` can be replaced with `optax`, including:

- simple case: switching `flax.optim` to `optax` + updating the training loop
- specifying the `optax` optimizer in terms of composable gradient updates
- weight decay
- gradient clipping
- learning rate schedules
- multiple optimizers

Additional to the RST documentation, the PR also contains a Colab that allows for execution, modification, and inspection of the code snippets.

### DIFF
--- a/docs/howtos/optax_update_guide.rst
+++ b/docs/howtos/optax_update_guide.rst
@@ -1,5 +1,11 @@
-Optax Update Guide
-==================
+Upgrading my Codebase to Optax
+==============================
+
+We have proposed to replace :py:mod:`flax.optim` with `Optax
+<https://optax.readthedocs.io>`_ in 2021 with `FLIP #1009
+<https://github.com/google/flax/blob/main/docs/flip/1009-optimizer-api.md>`_ and
+the Flax optimizers are now *effectively deprecated*. This guide is targeted
+towards :py:mod:`flax.optim` users to help them update their code to Optax.
 
 Code samples below are executable in
 `Colab <https://colab.research.google.com/github/google/flax/blob/main/docs/notebooks/optax_update_guide.ipynb>`_.
--- a/docs/howtos/optax_update_guide.rst
+++ b/docs/howtos/optax_update_guide.rst
@@ -4,6 +4,23 @@ Optax Update Guide
 Code samples below are executable in
 `Colab <https://colab.research.google.com/github/google/flax/blob/main/docs/notebooks/optax_update_guide.ipynb>`_.
 
+
+.. testsetup::
+
+  import flax
+  import jax
+  import jax.numpy as jnp
+  import flax.linen as nn
+  import optax
+
+  batch = {'image': jnp.ones([1, 28, 28, 1]), 'label': jnp.array([0])}
+  ds_train = [batch]
+  get_ds_train = lambda: [batch]
+  model = nn.Dense(1)
+  variables = model.init(jax.random.PRNGKey(0), batch['image'])
+  learning_rate, momentum, weight_decay, grad_clip_norm = .1, .9, 1e-3, 1.
+  loss = lambda params, batch: jnp.array(0.)
+
 Replacing ``flax.optim`` with ``optax``
 ---------------------------------------
 
@@ -151,6 +168,8 @@ argument to a function for scaling the updates added with
 |optax.scale_by_schedule()|_. Optax also allows to specify functions to
 inject arbitrary scalar values for other gradient updates via
 |optax.inject_hyperparams()|_.
+
+Read more about learning rate schedules in the :doc:`lr_schedule` guide.
 
 .. |optax.scale_by_schedule()| replace:: ``optax.scale_by_schedule()``
 .. _optax.scale_by_schedule(): https://optax.readthedocs.io/en/latest/api.html#optax.scale_by_schedule
--- a/docs/howtos/optax_update_guide.rst
+++ b/docs/howtos/optax_update_guide.rst
@@ -203,7 +203,7 @@ learning rate schedule as a parameter for `learning_rate`.
       optax.scale_by_schedule(lambda step: -schedule(step)),
   )
 
-Multiple Optimizers
+Multiple Optimizers or only updating a subset of parameters
 -------------------
 
 In Flax, Traversals are used to specify which parameters should be updated by an optimizer. And you can combine Traversals using :py:class:`flax.optim.MultiOptimizer` to apply different optimizers on different parameters. The equivalent in Optax is `optax.masked()` and `optax.chain()`.
--- a/docs/howtos/optax_update_guide.rst
+++ b/docs/howtos/optax_update_guide.rst
@@ -1,0 +1,230 @@
+Optax Update Guide
+==================
+
+Code samples below are executable in
+`Colab <https://colab.research.google.com/github/google/flax/blob/main/docs/notebooks/optax_update_guide.ipynb>`_.
+
+Replacing ``flax.optim`` with ``optax``
+---------------------------------------
+
+Optax has drop-in replacements for all of Flax's optimizers. Refer to Optax's
+documentation `Common Optimizers <https://optax.readthedocs.io/en/latest/api.html>`_
+for API details.
+
+The usage is very similar, with the difference that ``optax`` does not keep a
+copy of the ``params``, so they need to be passed around separately. Flax
+provides the utility :py:class:`~flax.training.train_state.TrainState` to store
+optimizer state, parameters, and other associated data in a single dataclass.
+
+.. codediff::
+  :title_left: ``flax.optim``
+  :title_right: ``optax``
+
+  @jax.jit
+  def train_step(optimizer, batch):
+    grads = jax.grad(loss)(optimizer.target, batch)
+    return optimizer.apply_gradient(grads)
+
+  optimizer_def = flax.optim.Momentum(
+      learning_rate, momentum)
+  optimizer = optimizer_def.create(variables['params'])
+
+  for batch in get_ds_train():
+    optimizer = train_step(optimizer, batch)
+
+  ---
+
+  @jax.jit
+  def train_step(params, opt_state, batch):
+    grads = jax.grad(loss)(params, batch)
+    updates, opt_state = tx.update(grads, opt_state)
+    params = optax.apply_updates(params, updates)
+    return params, opt_state
+
+  tx = optax.sgd(learning_rate, momentum)
+  params = variables['params']
+  opt_state = tx.init(params)
+
+  for batch in ds_train:
+    params, opt_state = train_step(params, opt_state, batch)
+
+
+Composable Gradient Transformations
+-----------------------------------
+
+Note though that |optax.sgd()|_ is simply a wrapper for the sequential
+application of two gradient transformations. Instead of using this alias, it is
+common to use |optax.chain()|_ to combine multiple of these generic building
+blocks.
+
+.. |optax.sgd()| replace:: ``optax.sgd()``
+.. _optax.sgd(): https://optax.readthedocs.io/en/latest/api.html#optax.sgd
+.. |optax.chain()| replace:: ``optax.chain()``
+.. _optax.chain(): https://optax.readthedocs.io/en/latest/api.html#chain
+
+.. codediff::
+  :title_left: Pre-defined alias
+  :title_right: Combining transformations
+
+  # Note that the aliases follow the convention to use positive
+  # values for the learning rate by default.
+  tx = optax.sgd(learning_rate, momentum)
+
+  ---
+
+  tx = optax.chain(
+      # 1. Step: keep a trace of past updates and add to gradients. 
+      optax.trace(decay=momentum),
+      # 2. Step: multiply result from step 1 with negative learning rate.
+      # Note that `optax.apply_updates()` simply adds the final updates to the
+      # parameters, so we must make sure to flip the sign here for gradient
+      # descent.
+      optax.scale(-learning_rate),
+  )
+
+Weight Decay
+------------
+
+Some of Flax's optimizers also include a weight decay. In Optax, the weight
+decay can be added as another "gradient transformation"
+|optax.add_decayed_weights()|_ that adds an update derived from the parameters.
+
+.. |optax.add_decayed_weights()| replace:: ``optax.add_decayed_weights()``
+.. _optax.add_decayed_weights(): https://optax.readthedocs.io/en/latest/api.html#optax.add_decayed_weights
+
+.. codediff::
+  :title_left: ``flax.optim``
+  :title_right: ``optax``
+
+  optimizer_def = flax.optim.Adam(
+      learning_rate, weight_decay=weight_decay)
+  optimizer = optimizer_def.create(variables['params'])
+
+  ---
+
+  tx = optax.chain(
+      optax.scale_by_adam(),
+      optax.add_decayed_weights(weight_decay),
+      # params -= learning_rate * (adam(grads) + params * weight_decay)
+      optax.scale(-learning_rate),
+  )
+  # Note that you'll need to specify `params` when computing the udpates:
+  # tx.update(grads, opt_state, params)
+
+Gradient Clipping
+-----------------
+
+Training can be stabilized by clipping gradients to a global norm (`Pascanu et
+al, 2012 <https://arxiv.org/abs/1211.5063>`_). In Flax this is often done by
+processing the gradients before passing them to the optimizer. With Optax this
+becomes just another gradient transformation |optax.clip_by_global_norm()|_.
+
+.. |optax.clip_by_global_norm()| replace:: ``optax.clip_by_global_norm()``
+.. _optax.clip_by_global_norm(): https://optax.readthedocs.io/en/latest/api.html#optax.clip_by_global_norm
+
+.. codediff::
+  :title_left: ``flax.optim``
+  :title_right: ``optax``
+
+  def train_step(optimizer, batch):
+    grads = jax.grad(loss)(optimizer.target, batch)
+    grads_flat, _ = jax.tree_flatten(grads)
+    global_l2 = jnp.sqrt(sum([jnp.vdot(p, p) for p in grads_flat]))
+    g_factor = jnp.minimum(1.0, grad_clip_norm / global_l2)
+    grads = jax.tree_map(lambda g: g * g_factor, grads)
+    return optimizer.apply_gradient(grads)
+
+  ---
+
+  tx = optax.chain(
+      optax.clip_by_global_norm(grad_clip_norm),
+      optax.trace(decay=momentum),
+      optax.scale(-learning_rate),
+  )
+
+Learning Rate Schedules
+-----------------------
+
+For learning rate schedules, Flax allows to overwrite hyper parameters when
+applying the gradients. Optax keeps a step counter and provides this as an
+argument to a function for scaling the updates added with
+|optax.scale_by_schedule()|_. Optax also allows to specify functions to
+inject arbitrary scalar values for other gradient updates via
+|optax.inject_hyperparams()|_.
+
+.. |optax.scale_by_schedule()| replace:: ``optax.scale_by_schedule()``
+.. _optax.scale_by_schedule(): https://optax.readthedocs.io/en/latest/api.html#optax.scale_by_schedule
+.. |optax.inject_hyperparams()| replace:: ``optax.inject_hyperparams()``
+.. _optax.inject_hyperparams(): https://optax.readthedocs.io/en/latest/api.html#optax.inject_hyperparams
+
+.. codediff::
+  :title_left: ``flax.optim``
+  :title_right: ``optax``
+
+  def train_step(step, optimizer, batch):
+    grads = jax.grad(loss)(optimizer.target, batch)
+    return step + 1, optimizer.apply_gradient(grads, learning_rate=schedule(step))
+
+  ---
+
+  tx = optax.chain(
+      optax.trace(decay=momentum),
+      # Note that we still want a negative value for scaling the updates!
+      optax.scale_by_schedule(lambda step: -schedule(step)),
+  )
+
+Multiple Optimizers
+-------------------
+
+Flax's :py:class:`flax.optim.MultiOptimizer` can be used to compose multiple
+optimizers and only apply them to a part of the params pytree. Optax provides
+an efficient masking mechanism to achieve the same goal.
+
+Note that below example is using :py:mod:`flax.traverse_util` to create the
+boolean masks required by |optax.masked()|_ - alternatively you could also
+create them manually, or use |optax.multi_transform()|_ that takes a
+multivalent pytree to specify gradient transformations.
+
+Beware that |optax.masked()|_ flattens the pytree internally and the inner
+gradient transformations will only be called with that partial flattened view of
+the params/gradients. This is not a problem usually, but it makes it hard to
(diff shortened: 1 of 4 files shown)


andsteing:
Review here:
https://flax--1774.org.readthedocs.build/en/1774/howtos/optax_update_guide.html

andsteing:
Oh, I just noticed that `codediff` does more than pretty formatting ... Fixing the PR to make it build again ...

andsteing:
Colab viewable here:
https://colab.research.google.com/github/google/flax/blob/1451cb8b11d438bfc65cfbfbe5fe4c3a1745cfa0/docs/notebooks/optax_update_guide.ipynb

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,249 @@
...
+
+  import flax
+  import jax
+  import jax.numpy as jnp
+  import flax.linen as nn
+  import optax
+
+  batch = {'image': jnp.ones([1, 28, 28, 1]), 'label': jnp.array([0])}
+  ds_train = [batch]
+  get_ds_train = lambda: [batch]
```
rosshemsley, line 18: nit: do we need both lines? (they seem to be the same thing)

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,249 @@
...
+  optimizer = optimizer_def.create(variables['params'])
+
+  ---
+
+  tx = optax.chain(
+      optax.scale_by_adam(),
+      optax.add_decayed_weights(weight_decay),
+      # params -= learning_rate * (adam(grads) + params * weight_decay)
+      optax.scale(-learning_rate),
+  )
```
rosshemsley, line 127: nit: optax defines `optax.adamw()` for this common case (https://github.com/deepmind/optax/blob/master/optax/_src/alias.py#L273)

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,249 @@
...
+  def train_step(step, optimizer, batch):
+    grads = jax.grad(loss)(optimizer.target, batch)
+    return step + 1, optimizer.apply_gradient(grads, learning_rate=schedule(step))
+
+  ---
+
+  tx = optax.chain(
+      optax.trace(decay=momentum),
+      # Note that we still want a negative value for scaling the updates!
+      optax.scale_by_schedule(lambda step: -schedule(step)),
```
rosshemsley, line 192: it could be worth mentioning that optax has a number of standard schedules predefined: https://optax.readthedocs.io/en/latest/optax-101.html#weight-decay-schedules-and-clipping

rosshemsley (approved):
Thanks! This looks great.

A couple comments to think about, but mainly nits.

It might also be worth linking to the optax quick start here: https://optax.readthedocs.io/en/latest/optax-101.html

docs/howtos/optax_update_guide.rst
rosshemsley: i.e. you can do:

```python
schedule = optax.warmup_cosine_decay_schedule(
  init_value=0.0,
  peak_value=1.0,
  warmup_steps=50,
  decay_steps=1_000,
  end_value=0.0,
)

optimizer = optax.adamw(learning_rate=schedule)
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,249 @@
...
+  ---
+
+  @jax.jit
+  def train_step(params, opt_state, batch):
+    grads = jax.grad(loss)(params, batch)
+    updates, opt_state = tx.update(grads, opt_state)
+    params = optax.apply_updates(params, updates)
+    return params, opt_state
+
+  tx = optax.sgd(learning_rate, momentum)
```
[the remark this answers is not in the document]
andsteing: Ah, interesting. I originally took the `tx` from the Optax README (under "Gradient Transformations"), and didn't realize that `optimizer` would be the more common name for the chain of transformations..

But then I started using that in our FLIP and all our examples. So I think for consistency I'll rather keep `tx`.

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,249 @@
+Optax Update Guide
+==================
+
+Code samples below are executable in
```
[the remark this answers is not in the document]
andsteing: Yeah, we have a couple of these Colabs.

But in this case I care more about the side-by-side of the `flax.optim` and `optax` code to make users' lives easier.

But I thought it's still very useful to have exactly the same code available in a Colab that runs end to end.

docs/howtos/optax_update_guide.rst
andsteing: Added a comment to clarify.

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,249 @@
...
+
+  for batch in get_ds_train():
+    optimizer = train_step(optimizer, batch)
+
+  ---
+
+  @jax.jit
+  def train_step(params, opt_state, batch):
+    grads = jax.grad(loss)(params, batch)
+    updates, opt_state = tx.update(grads, opt_state)
```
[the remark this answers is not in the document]
andsteing: Moved above.

docs/howtos/optax_update_guide.rst
andsteing: Added a comment.

docs/howtos/optax_update_guide.rst
andsteing: Good point, added a reference to the lr schedules, and mentioned that `learning_rate` of standard optimizers can be scalar or schedule.

andsteing:
Thanks for the comments, Ross!

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,263 @@
+Optax Update Guide
```
avital, line 1: As it reads now it sounds like it's about updating Optax.

I propose renaming to "Upgrading my codebase to Optax"

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,263 @@
...
+==================
+
```
avital, line 3: I think a short intro is in order here, e.g. "The built-in `flax.optim` package is being deprecated, and we instead recommend all users use the [Optax](link to github)] library. Follow this guide to upgrade your codebase."

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,263 @@
...
+  ---
+
+  tx = optax.chain(
+      optax.trace(decay=momentum),
+      # Note that we still want a negative value for scaling the updates!
+      optax.scale_by_schedule(lambda step: -schedule(step)),
+  )
+
+Multiple Optimizers
```
avital, line 206:
```suggestion
Multiple Optimizers or only updating a subset of parameters
```

(Since `optax.masked` can be used for that as well)

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,263 @@
...
+-------------------
+
+Flax's :py:class:`flax.optim.MultiOptimizer` can be used to compose multiple
```
avital, line 209:
```suggestion
In Flax, Traversals are used to specify which parameters should be updated by an optimizer. And you can combine Traversals using :py:class:`flax.optim.MultiOptimizer` to apply different optimizers on different parameters. The equivalent in Optax is `optax.masked()` and `optax.chain()`.
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,263 @@
+Optax Update Guide
+==================
+
+Code samples below are executable in
+`Colab <https://colab.research.google.com/github/google/flax/blob/main/docs/notebooks/optax_update_guide.ipynb>`_.
```
avital, line 5: What's the user journey where someone wants to run this Colab? (as opposed to copying code from the RTD page). I ask because I'd rather avoid having to remember to update two places if we change a variable name in this doc, etc.

avital: Or, another version of the question: Should we have colabs for *all* of our HOWTOs? If so, maybe we should find a way to automatically convert them. Or is this one special? I'm hesitant to introduce this new expectation by adding just one guide with a Colab, unless we want to go in that direction more broadly.

avital: (But no need to block merging this PR on this discussion, just wanted to bring it up. We can resolve separately)

avital (approved):
Excellent! Added a few small remarks.

docs/howtos/optax_update_guide.rst
andsteing: I think the user journey is to provide users with a working example that they can use to experiment with different gradient transformations to replicate their original setup. Of course they could do this on their own starting with the code blocks from the transition guide, but there's still some more code involved and it's annoying having to copy this together and a nicer experience to start with a working Colab.

I first tried to have a Colab that is automatically synced, but then realized two problems:

1. I think it's very useful to have the side-by-side in the update guide. This would require us to create a new Sphinx plugin that knows how to take two cells in the Colab, do some pre-processing, and then put them into a diff html table.
2. The Colab will necessarily contain some different code. For example, I want to provide a more realistic example that can be used int he Colab to play around with optimizer settings in a meaningful way. In RST on the other hand we want to have a cheap setup that does not load anything from `tdfs`. This would probably be a simpler Sphinx plugin replacing some code block with another code block when transcribing the Colab into RST.

If we have a Sphinx plugin implementing the 2 features mentioned above then I'll gladly use it. But it's too much work to implement, and I'd rather have to sync the two documents for the added functionality. I don't expect us to update this code frequently, so it's more about remembering (and even if we forgot to update a small thing, it wouldn't be too bad).

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+Optax has drop-in replacements for all of Flax's optimizers. Refer to Optax's
+documentation `Common Optimizers <https://optax.readthedocs.io/en/latest/api.html>`_
+for API details.
+
+The usage is very similar, with the difference that ``optax`` does not keep a
+copy of the ``params``, so they need to be passed around separately. Flax
+provides the utility :py:class:`~flax.training.train_state.TrainState` to store
+optimizer state, parameters, and other associated data in a single dataclass.
+
+.. codediff::
```
marcvanzee, line 46: Very cool you used the side-by-side view, great to see that it is actually useful 😄 . Not sure if you are aware of this, but if you add a "`#!`" to the end of a line, that line will be highlighted. (I can also see that it may not be necessary in this HOWTO, so feel free to ignore.)

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+  :title_left: ``flax.optim``
+  :title_right: ``optax``
+
+  @jax.jit
+  def train_step(optimizer, batch):
```
marcvanzee, line 51: Have you considered aligning this side-by-side view nicer? RIght now the two `train_step` functions are at different places, which makes it a bit less easy to compare.

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+    return params, opt_state
+
+  for batch in ds_train:
+    params, opt_state = train_step(params, opt_state, batch)
+
+
+Composable Gradient Transformations
+-----------------------------------
+
+Note though that |optax.sgd()|_ is simply a wrapper for the sequential
```
marcvanzee, line 82: "Note though" seems a bit of a strange way to start a new section. Maybe just say "`|optax.sgd()|_` is simply a wrapper for the..."?

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+      # Note that `optax.apply_updates()` simply adds the final updates to the
+      # parameters, so we must make sure to flip the sign here for gradient
+      # descent.
+      optax.scale(-learning_rate),
+  )
+
+Weight Decay
+------------
+
+Some of Flax's optimizers also include a weight decay. In Optax, the weight
```
marcvanzee, line 115: weight decay can even be added using a "decay mask", you could mention that since the Flax BERT finetuning example in HuggingFace uses it: https://github.com/huggingface/transformers/blob/master/examples/flax/text-classification/run_flax_glue.py#L189 (maybe just point to it as a reference, so users can look there if they want to see an example?)

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+  tx = optax.chain(
+      optax.scale_by_adam(),
+      optax.add_decayed_weights(weight_decay),
+      # params -= learning_rate * (adam(grads) + params * weight_decay)
+      optax.scale(-learning_rate),
+  )
+  # Note that you'll need to specify `params` when computing the udpates:
+  # tx.update(grads, opt_state, params)
+
+Gradient Clipping
```
marcvanzee, line 142: Great example!

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+  tx = optax.chain(
+      optax.clip_by_global_norm(grad_clip_norm),
+      optax.trace(decay=momentum),
+      optax.scale(-learning_rate),
+  )
+
+Learning Rate Schedules
+-----------------------
+
+For learning rate schedules, Flax allows to overwrite hyper parameters when
```
marcvanzee, line 176:
```suggestion
For learning rate schedules, Flax allows overwriting hyperparameters when
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+applying the gradients. Optax keeps a step counter and provides this as an
```
marcvanzee, line 177:
```suggestion
applying the gradients. Optax maintains a step counter and provides this as an
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+argument to a function for scaling the updates added with
+|optax.scale_by_schedule()|_. Optax also allows to specify functions to
```
marcvanzee, line 179:
```suggestion
|optax.scale_by_schedule()|_. Optax also allows specifying functions to
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+.. |optax.inject_hyperparams()| replace:: ``optax.inject_hyperparams()``
+.. _optax.inject_hyperparams(): https://optax.readthedocs.io/en/latest/api.html#optax.inject_hyperparams
+
+.. codediff::
+  :title_left: ``flax.optim``
+  :title_right: ``optax``
+
+  def train_step(step, optimizer, batch):
+    grads = jax.grad(loss)(optimizer.target, batch)
+    return step + 1, optimizer.apply_gradient(grads, learning_rate=schedule(step))
```
marcvanzee, line 202: You could consider making this line less wide since right now the side by side has a scrollbar on laptops, like this?

```suggestion
    grad = optimizer.apply_gradient(grads, learning_rate=schedule(step))
    return step+1, grad
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+|optax.scale_by_schedule()|_. Optax also allows to specify functions to
+inject arbitrary scalar values for other gradient updates via
+|optax.inject_hyperparams()|_.
+
+Read more about learning rate schedules in the :doc:`lr_schedule` guide.
+
+Read more about schedules defined in Optax under `Optimizer Schedules
+<https://optax.readthedocs.io/en/latest/api.html#optimizer-schedules>`_. the
+standard optimizers (like ``optax.adam()``, ``optax.sgd()`` etc.) also accept a
+learning rate schedule as a parameter for `learning_rate`.
```
marcvanzee, line 188:
```suggestion
learning rate schedule as a parameter for ``learning_rate``.
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+  tx = optax.chain(
+      optax.trace(decay=momentum),
+      # Note that we still want a negative value for scaling the updates!
+      optax.scale_by_schedule(lambda step: -schedule(step)),
+  )
+
+Multiple Optimizers or only updating a subset of parameters
+-------------------
+
+In Flax, Traversals are used to specify which parameters should be updated by an optimizer. And you can combine Traversals using :py:class:`flax.optim.MultiOptimizer` to apply different optimizers on different parameters. The equivalent in Optax is `optax.masked()` and `optax.chain()`.
```
marcvanzee, line 215:
```suggestion
In Flax, Traversals are used to specify which parameters should be updated by an optimizer. And you can combine Traversals using :py:class:`flax.optim.MultiOptimizer` to apply different optimizers on different parameters. The equivalent in Optax is ``optax.masked()`` and ``optax.chain()``.
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+optimizers and only apply them to a part of the params pytree. Optax provides
```
marcvanzee, line 216: This sentence doesn't seem to logically follow the previous one, maybe something went wrong?

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+an efficient masking mechanism to achieve the same goal.
+
+Note that below example is using :py:mod:`flax.traverse_util` to create the
```
marcvanzee, line 219:
```suggestion
Note that example below is using :py:mod:`flax.traverse_util` to create the
```

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,269 @@
...
+  :title_right: ``optax``
+
+  kernels = flax.traverse_util.ModelParamTraversal(lambda p, _: 'kernel' in p)
+  biases = flax.traverse_util.ModelParamTraversal(lambda p, _: 'bias' in p)
+  kernel_opt = flax.optim.Momentum(learning_rate, momentum)
+  bias_opt = flax.optim.Momentum(learning_rate * 0.1, momentum)
+  optimizer = flax.optim.MultiOptimizer(
+      (kernels, kernel_opt),
+      (biases, bias_opt)
+  ).create(variables['params'])
```
marcvanzee, line 247: Will this align nicer with the RHS?

```suggestion
  kernels = flax.traverse_util.ModelParamTraversal(lambda p, _: 'kernel' in p)
  biases = flax.traverse_util.ModelParamTraversal(lambda p, _: 'bias' in p)
  
  kernel_opt = flax.optim.Momentum(learning_rate, momentum)
  bias_opt = flax.optim.Momentum(learning_rate * 0.1, momentum)
  
  
  optimizer = flax.optim.MultiOptimizer(
      (kernels, kernel_opt),
      (biases, bias_opt)
  ).create(variables['params'])
```

marcvanzee (changes requested):

docs/howtos/optax_update_guide.rst
andsteing: Ah, I was wondering why no lines were highlighted (I somehow assumed it would be automatic).

Good to know, but I went through the diffs (or rather "side by sides") and didn't really find value in aligning the code snippets and highlighting the changes (there are too few common lines). But good to know for my next HOTWO.

docs/howtos/optax_update_guide.rst
andsteing: Thinking again about this, I'll undo an earlier change proposed by Ross to align things even more nicely. I think in this case it's more useful to have code aligned than define variables in global scope before defining the function uses them (see Ross's comment above for the motivation to rearrange in the first place)

docs/howtos/optax_update_guide.rst
andsteing: Done.

docs/howtos/optax_update_guide.rst
andsteing: Added reference to `optax.adamw()`. Also created https://github.com/deepmind/optax/pull/284 to clarify the usage.

docs/howtos/optax_update_guide.rst
andsteing: Ack. Leaving comment open for better visiblity.

docs/howtos/optax_update_guide.rst
andsteing: Linkified.

docs/howtos/optax_update_guide.rst
andsteing: Updated + linkified.

docs/howtos/optax_update_guide.rst
andsteing: Done.

docs/howtos/optax_update_guide.rst
andsteing: Done.

docs/howtos/optax_update_guide.rst
andsteing: I accepted a change from Avital and didn't proof read. Cleaned up.

docs/howtos/optax_update_guide.rst
marcvanzee: Ah sorry about that, I didn't realize Ross already commented on this and I should have checked. Ok I'm fine either way!

marcvanzee (approved):

docs/howtos/optax_update_guide.rst
andsteing: Done.

docs/howtos/optax_update_guide.rst
andsteing: Done.

docs/howtos/optax_update_guide.rst
andsteing: Done.

docs/howtos/optax_update_guide.rst
andsteing: Note that it's the `optimizer`.

Also, we already use the same pattern in the other snippets, so I'd leave it as is for consistency.

docs/howtos/optax_update_guide.rst
```diff
@@ -0,0 +1,281 @@
...
+
+
+  optimizer = flax.optim.MultiOptimizer(
+      (kernels, kernel_opt),
+      (biases, bias_opt)
+  ).create(variables['params'])
+
+  ---
+
+  kernels = flax.traverse_util.ModelParamTraversal(lambda p, _: 'kernel' in p)
```
avital, line 263: Oh, I'm sorry to have missed this in the first round of review but I'm curious -- why do we need to be using `ModelParamTraversal` at all? Presumably other users of Optax, e.g. those using Haiku, have some other pattern? I'd love to eventually deprecate `ModelParamTraversal` as I don't think it's needed anymore, and if needed replace it with something pretty much like a tree_map that gives you the full path of subkeys rather than just the values.

andsteing: There's plenty of different ways how you can create those masks.

I took `ModelParamTraversal` because it's the most straightforward transition from existing code, but feel free to update the example to use `traverse_utils.flatten_dict()` instead.

avital: OK I'll give it a shot.

PR #1798, opened 2022-01-20
Add design note about setup vs nn.compact

avital:
See https://flax--1798.org.readthedocs.build/en/1798/ for the rendered docs

### DIFF
--- a/docs/design_notes/setup_or_nncompact.rst
+++ b/docs/design_notes/setup_or_nncompact.rst
@@ -1,0 +1,78 @@
+Should I use ``setup`` or ``nn.compact``?
+=========================================
+
+In Flax's module system (named `Linen`_), submodules, parameters and/or other variables
+can be defined in two ways:
+
+1. **Explicitly** (using ``setup``):
+
+   Assign to ``self.<attr>`` inside a :meth:`setup <flax.linen.Module.setup>`
+   method. Then read ``self.<attr>`` from any "forward pass" method defined on the class.
+   This resembles how most objects work in Python, and how modules are defined in PyTorch.
+2. **In-line** (using ``nn.compact``):
+  
+   Declare directly within a single "forward pass" method annotated
+   with :meth:`nn.compact <flax.linen.compact>`. This allows you to define your whole module
+   in a single method, and "co-locate" submodules, parameters and variables next to 
+   where they are used.
+
+**Both of these approaches are perfectly valid, behave the same way, and interoperate with all of Flax**.
+
+Here is a short example of a module defined in both ways. 
+
+.. testsetup::
+
+  import flax.linen as nn
+
+.. codediff:: 
+  :title_left: Using ``setup``
+  :title_right: Using ``nn.compact``
+  
+  class MLP(nn.Module):
+    def setup(self):
+      self.dense1 = nn.Dense(32)
+      self.dense2 = nn.Dense(32)
+
+    def __call__(self, x):
+      x = self.dense1(x)
+      x = nn.relu(x)
+      x = self.dense2(x)
+      return x
+  ---
+  class MLP(nn.Module):
+
+
+
+    @nn.compact
+    def __call__(self, x):
+      x = nn.Dense(32, name="dense1")(x)
+      x = nn.relu(x)
+      x = nn.Dense(32, name="dense2")(x)
+      return x
+
+So, how would you decide which style to use? It can be a matter of taste, but here are some pros and cons:
+
+Reasons to prefer using ``nn.compact``:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Allows defining submodules, parameters and other variables next to where they are used: less    
+   scrolling up/down to see how everything is defined.
+2. Reduces code duplication when there are conditionals or for loops that conditionally define
+   submodules, parameters or variables.
+3. Code typically looks more like mathematical notation.
+4. Generally, shorter code.
+
+Reasons to prefer using ``setup``:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Closer to normal Python object semantics
+2. Easier when porting models from PyTorch
+3. Allows defining more than one "forward pass" method
+   (see :class:`MultipleMethodsCompactError <flax.errors.MultipleMethodsCompactError>`)
+
+
+
+
+
+
+.. _`Linen`: https://jax.readthedocs.io/en/latest/notebooks/thinking_in_jax.html#JIT-mechanics:-tracing-and-static-variables
--- a/docs/design_notes/setup_or_nncompact.rst
+++ b/docs/design_notes/setup_or_nncompact.rst
@@ -1,24 +1,28 @@
 Should I use ``setup`` or ``nn.compact``?
 =========================================
 
-In Flax's module system (named `Linen`_), submodules, parameters and/or other variables
+In Flax's module system (named `Linen`_), submodules and variables (parameters or others)
 can be defined in two ways:
 
 1. **Explicitly** (using ``setup``):
 
-   Assign to ``self.<attr>`` inside a :meth:`setup <flax.linen.Module.setup>`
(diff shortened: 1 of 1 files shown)


docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+can be defined in two ways:
+
+1. **Explicitly** (using ``setup``):
+
+   Assign to ``self.<attr>`` inside a :meth:`setup <flax.linen.Module.setup>`
+   method. Then read ``self.<attr>`` from any "forward pass" method defined on the class.
+   This resembles how most objects work in Python, and how modules are defined in PyTorch.
+2. **In-line** (using ``nn.compact``):
+  
+   Declare directly within a single "forward pass" method annotated
```
marcvanzee, line 14: "Declare directly" sounds a bit odd, maybe "Write your network's logic directly"?

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
+Should I use ``setup`` or ``nn.compact``?
+=========================================
+
+In Flax's module system (named `Linen`_), submodules, parameters and/or other variables
+can be defined in two ways:
+
+1. **Explicitly** (using ``setup``):
+
+   Assign to ``self.<attr>`` inside a :meth:`setup <flax.linen.Module.setup>`
```
marcvanzee, line 9: Assign variables (e.g., submodules) to `self`?

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+   method. Then read ``self.<attr>`` from any "forward pass" method defined on the class.
```
marcvanzee, line 10: Then use the variables assigned to `self` in `setup` from any ....?

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+2. **In-line** (using ``nn.compact``):
+  
+   Declare directly within a single "forward pass" method annotated
+   with :meth:`nn.compact <flax.linen.compact>`. This allows you to define your whole module
+   in a single method, and "co-locate" submodules, parameters and variables next to 
+   where they are used.
+
+**Both of these approaches are perfectly valid, behave the same way, and interoperate with all of Flax**.
+
+Here is a short example of a module defined in both ways.
```
marcvanzee, line 21: I would also explain how naming of submodules works for `setup`

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+So, how would you decide which style to use? It can be a matter of taste, but here are some pros and cons:
+
+Reasons to prefer using ``nn.compact``:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Allows defining submodules, parameters and other variables next to where they are used: less    
+   scrolling up/down to see how everything is defined.
+2. Reduces code duplication when there are conditionals or for loops that conditionally define
+   submodules, parameters or variables.
+3. Code typically looks more like mathematical notation.
```
marcvanzee, line 62: Why is this the case? I don't really get this one.

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+4. Generally, shorter code.
```
marcvanzee, line 63: Given 1 and 2 this one seems a bit redundant.

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+
+Reasons to prefer using ``setup``:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Closer to normal Python object semantics
+2. Easier when porting models from PyTorch
+3. Allows defining more than one "forward pass" method
+   (see :class:`MultipleMethodsCompactError <flax.errors.MultipleMethodsCompactError>`)
```
marcvanzee, line 71:
```suggestion
1. Closer to normal Python object semantics.
2. Easier when porting models from PyTorch.
3. Allows defining more than one "forward pass" method
   (see :class:`MultipleMethodsCompactError <flax.errors.MultipleMethodsCompactError>`).
```

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+
+
+    @nn.compact
+    def __call__(self, x):
+      x = nn.Dense(32, name="dense1")(x)
+      x = nn.relu(x)
+      x = nn.Dense(32, name="dense2")(x)
+      return x
+
+So, how would you decide which style to use? It can be a matter of taste, but here are some pros and cons:
```
marcvanzee, line 53: Why I see *extremely* often is that users use both `@nn.compact` and `setup()`. It would be useful to write down what the consequence is of that. (To be honest I'd like to know as well!)

marcvanzee (changes requested):

docs/design_notes/setup_or_nncompact.rst
```diff
@@ -0,0 +1,78 @@
...
+   scrolling up/down to see how everything is defined.
+2. Reduces code duplication when there are conditionals or for loops that conditionally define
+   submodules, parameters or variables.
+3. Code typically looks more like mathematical notation.
+4. Generally, shorter code.
+
+Reasons to prefer using ``setup``:
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+1. Closer to normal Python object semantics
```
andsteing, line 68: This is a tricky one. I think a lot of confusion stems from users thinking about Flax models too much like Python objects. As opposed to PyTorch, Flax modules behave quite differently though. For example, they exist in a bound and unbound state, and all methods (including `setup()`) are called in the bound state and cannot simply be called from the outside "unbound world". Also, the attribute assignment does not what it seems to do (since we overload `setattr()` and do some magic state handling).

Once Jonathan has published his "Linen lifecycle doc" then that will be a good place to link it. But in the meantime I'm wondering if a statement like this does not sow more confusion than clear things up.

docs/design_notes/setup_or_nncompact.rst
avital: well, submodules aren't variables...

docs/design_notes/setup_or_nncompact.rst
avital: ditto

docs/design_notes/setup_or_nncompact.rst
avital: yes, good point. I'll add something about that.

docs/design_notes/setup_or_nncompact.rst
avital: Well, in math you'd write something like $y=W*x+b$ rather than: $params=(W,b); y=params.W*x+params.b$

docs/design_notes/setup_or_nncompact.rst
avital: OK I'll remove it

docs/design_notes/setup_or_nncompact.rst
avital: Agreed.

docs/design_notes/setup_or_nncompact.rst
avital: Yeah good point. It actually works nicely. But maybe that should be in an footnote to this doc rather than in the main text?

docs/design_notes/setup_or_nncompact.rst
avital: Good point. I think I'll just downplay this as to not lead people to poor mismatch with their expectations.

avital:
@marcvanzee @andsteing I responded to the code review comments. PTAL.

marcvanzee (approved):

PR #1799, opened 2022-01-21
Fix scope rewind-aliasing issue under transforms.

levskaya:
For a Module under transformation, the same outside submodule can be
passed in multiple times.  Inside the transform boundary this means that
multiple copies of the same submodule can point to the same duplicated
Scope object.  For compact submodules, this scope is rewound by
returning a new rewound scope, but this doesn't correctly rewind the
second reference of the original un-rewound scope.  We solve this by
respecting aliasing relationship via memoization during our traversal
through all scopes on the module.

The traversal logic for get_module_scopes and set_module_scopes was
also harmonized to make the code clearer.

Several typing fixes were also added.


levskaya:
This is one way of addressing #1796 .  (There might be some other possible solution by forcing _dup_scopes to return unique clones of the roots?  But I'm not sure that'd be any easier or clearer.)

levskaya:
I also added a few fixes to pre-existing python type-checking issues that were breaking our builds.

jheek:
Are you sure we want a 1->N mapping from scopes to linen Modules? 
I'm worried this will break invariants of the Module internal state.
Another way to avoid this (I think) is to create always just one scope per Module so I think we need a referentially non-transparent tree_map to be used inside transforms.py:
```
def tree_map_ref(fn, tree):
  refs = {}
  def wrapper(x):
    if x not in refs:
      refs[x] = fn(x)
    return refs[x]
  return jax.tree_map(wrapper, tree)
```

Or am I missing something?

levskaya:
Oh, I'm not wed to any particular solution.  The issue is that we already have a 1->N scope-to-Module situation in the above edge-case, but I can look at your suggestion tomorrow - you just mean to suggest that during traversal over the Modules, never re-traversing through identical Module instances?  That would probably work.

marcvanzee:
@levskaya @jheek Can we submit the types fixes separately? Our builds are currently still failing and we shouldn't wait on this PR to merge for that.

jheek:
> @levskaya @jheek Can we submit the types fixes separately? Our builds are currently still failing and we shouldn't wait on this PR to merge for that.

the type fixes were already fixed in a seperate PR that's why this PR has merge conflicts

levskaya:
@jheek  - sorry to have been delayed on this.  I refined the `get_module_scopes`, `set_module_scopes` traversals to handle aliasing among the Modules.  This should preserve the 1-1 module-scope relationship.

@marcvanzee - sorry for the delay, hopefully this fixes the original issue!

In case this doesn't go in before I leave tomorrow and it looks good - please feel free to push through the merge.

PR #1803, opened 2022-01-24
Remove float32 dtype assumption

### LINKED ISSUE #1777: FLIP: default dtype
FLIP discussing a default dtype mechanism for linen layers

PR: #1776

NeilGirdhar:
Fixes #1777

## Checklist
- [x] This change includes necessary high-coverage tests.  (No quality testing = no merge!)

### DIFF
--- a/flax/linen/activation.py
+++ b/flax/linen/activation.py
@@ -67,8 +67,11 @@ class PReLU(Module):
     Returns:
       The transformed input.
     """
+    dtype = inputs.dtype
+    assert jnp.issubdtype(dtype, jnp.floating)
     negative_slope = self.param(
       'negative_slope',
-      lambda k: jnp.asarray(self.negative_slope_init, jnp.float32)
+      lambda k: jnp.asarray(self.negative_slope_init, dtype)
     )
-    return jnp.where(inputs >= 0, inputs, jnp.asarray(negative_slope, inputs.dtype) * inputs)
+    assert negative_slope.shape == ()
+    return jnp.where(inputs >= 0, inputs, negative_slope * inputs)
--- a/flax/linen/activation.py
+++ b/flax/linen/activation.py
@@ -15,41 +15,18 @@
 """Activation functions.
 """
 
+import jax.numpy as jnp
 # pylint: disable=unused-import
-# re-export activation functions from jax.nn
-from jax.nn import celu
-from jax.nn import elu
-from jax.nn import gelu
-from jax.nn import glu
-from jax.nn import leaky_relu
-from jax.nn import log_sigmoid
-from jax.nn import log_softmax
-from jax.nn import normalize
-from jax.nn import relu
-from jax.nn import sigmoid
-from jax.nn import soft_sign
-from jax.nn import softmax
-from jax.nn import softplus
-from jax.nn import swish
-from jax.nn import silu
-from jax.nn import selu
-from jax.nn import hard_tanh
-from jax.nn import relu6
-from jax.nn import hard_sigmoid
-from jax.nn import hard_swish
-
+# re-export activation functions from jax.nn and jax.numpy
+from jax.nn import (celu, elu, gelu, glu, hard_sigmoid, hard_swish, hard_tanh,
+                    leaky_relu, log_sigmoid, log_softmax, normalize, relu,
+                    relu6, selu, sigmoid, silu, soft_sign, softmax, softplus,
+                    swish)
 from jax.numpy import tanh
 # pylint: enable=unused-import
 
-from typing import Any
-
-from flax.linen.linear import canonicalize_inexact_dtypes
-from flax.linen.module import Module, compact
-import jax.numpy as jnp
-
-
-FloatingDType = Type[jnp.floating]
-Array = Any
+from .dtypes import Array, FloatingDType, canonicalize_inexact_dtypes
+from .module import Module, compact
 
 
 class PReLU(Module):
--- a/flax/linen/activation.py
+++ b/flax/linen/activation.py
@@ -43,7 +43,7 @@ from jax.numpy import tanh
 
 from typing import Any
 
-from flax.linen.linear import _canonicalize_dtypes
+from flax.linen.linear import canonicalize_inexact_dtypes
 from flax.linen.module import Module, compact
 import jax.numpy as jnp
 
@@ -77,8 +77,8 @@ class PReLU(Module):
     """
     assert jnp.issubdtype(inputs.dtype, jnp.floating)
     inputs = jnp.asarray(inputs, dtype)
-    param_dtype, dtype = _canonicalize_dtypes(inputs.dtype, param_dtype,
-                                              self.dtype)
+    param_dtype, dtype = canonicalize_inexact_dtypes(inputs.dtype, param_dtype,
+                                                     self.dtype)
     negative_slope = self.param(
       'negative_slope',
       lambda k: jnp.asarray(self.negative_slope_init, param_dtype)
--- a/flax/linen/activation.py
+++ b/flax/linen/activation.py
@@ -43,10 +43,12 @@ from jax.numpy import tanh
 
 from typing import Any
 
+from flax.linen.linear import _canonicalize_dtypes
 from flax.linen.module import Module, compact
 import jax.numpy as jnp
 
 
+FloatingDType = Type[jnp.floating]
 Array = Any
 
 
@@ -54,9 +56,15 @@ class PReLU(Module):
   """Parametric Rectified Linear Unit (PReLU) activation function.
 
   Attributes:
-    negative_slope_init: the value to initialize the negative slope.
+    dtype: the dtype of the computation (default: float32).
+    param_dtype: the dtype passed to parameter initializers (default: float32).
+    negative_slope_init: the value to initialize the negative slope
+      (default 0.01).
   """
+  dtype: Optional[FloatingDType] = jnp.float32
+  param_dtype: Optional[FloatingDType] = jnp.float32
   negative_slope_init: float = 0.01
+
   @compact
   def __call__(self, inputs: Array) -> Array:
     """Applies an activation to the inputs.
@@ -67,11 +75,13 @@ class PReLU(Module):
     Returns:
       The transformed input.
     """
-    dtype = inputs.dtype
-    assert jnp.issubdtype(dtype, jnp.floating)
+    assert jnp.issubdtype(inputs.dtype, jnp.floating)
+    inputs = jnp.asarray(inputs, dtype)
+    param_dtype, dtype = _canonicalize_dtypes(inputs.dtype, param_dtype,
+                                              self.dtype)
     negative_slope = self.param(
       'negative_slope',
-      lambda k: jnp.asarray(self.negative_slope_init, dtype)
+      lambda k: jnp.asarray(self.negative_slope_init, param_dtype)
     )
-    assert negative_slope.shape == ()
+    negative_slope = jnp.asarray(negative_slope, dtype)
     return jnp.where(inputs >= 0, inputs, negative_slope * inputs)
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -15,26 +15,18 @@
 """Attention core modules for Flax."""
 
 from functools import partial
-from typing import Any, Callable, Tuple, Type, Optional
+from typing import Callable, Optional
 
 import jax
-from jax import lax
-from jax import random
 import jax.numpy as jnp
-import numpy as np
+from jax import lax, random
 from typing_extensions import Protocol
 
-from flax.linen.initializers import zeros
-from flax.linen.linear import DenseGeneral
-from flax.linen.linear import canonicalize_inexact_dtypes
-from flax.linen.linear import default_kernel_init
-from flax.linen.module import Module, compact, merge_param
-
-PRNGKey = Any
-Shape = Tuple[int, ...]
-InexactDType = Type[jnp.inexact]
-Array = Any
-Initializer = Callable[[PRNGKey, Shape, InexactDType], Array]
+from .dtypes import (Array, InexactDType, Initializer, PRNGKey,
+                     canonicalize_inexact_dtypes)
+from .initializers import zeros
+from .linear import DenseGeneral, default_kernel_init
+from .module import Module, compact, merge_param
 
 
 class AttentionFunction(Protocol):
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -26,7 +26,7 @@ from typing_extensions import Protocol
 
 from flax.linen.initializers import zeros
 from flax.linen.linear import DenseGeneral
-from flax.linen.linear import _canonicalize_dtypes
+from flax.linen.linear import canonicalize_inexact_dtypes
 from flax.linen.linear import default_kernel_init
 from flax.linen.module import Module, compact, merge_param
 
@@ -265,10 +265,10 @@ class MultiHeadDotProductAttention(Module):
     Returns:
       output of shape `[batch_sizes..., length, features]`.
     """
-    param_dtype, dtype = _canonicalize_dtypes(jnp.result_type(inputs_q,
-                                                              inputs_kv),
-                                              self.param_dtype,
-                                              self.dtype)
+    param_dtype, dtype = canonicalize_inexact_dtypes(jnp.result_type(inputs_q,
+                                                                     inputs_kv),
+                                                     self.param_dtype,
+                                                     self.dtype)
     features = self.out_features or inputs_q.shape[-1]
     qkv_features = self.qkv_features or inputs_q.shape[-1]
     assert qkv_features % self.num_heads == 0, (
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -15,23 +15,42 @@
 """Attention core modules for Flax."""
 
 from functools import partial
-from typing import (Any, Callable, Tuple, Optional)
+from typing import Any, Callable, Tuple, Type, Optional
 
 import jax
 from jax import lax
 from jax import random
 import jax.numpy as jnp
 import numpy as np
+from typing_extensions import Protocol
 
-from flax.linen.linear import default_kernel_init
-from flax.linen.linear import DenseGeneral
-from flax.linen.module import Module, compact, merge_param
 from flax.linen.initializers import zeros
+from flax.linen.linear import DenseGeneral
+from flax.linen.linear import _canonicalize_dtypes
+from flax.linen.linear import default_kernel_init
+from flax.linen.module import Module, compact, merge_param
 
 PRNGKey = Any
-Shape = Tuple[int]
-Dtype = Any
+Shape = Tuple[int, ...]
+InexactDType = Type[jnp.inexact]
 Array = Any
+Initializer = Callable[[PRNGKey, Shape, InexactDType], Array]
+
+
+class AttentionFunction(Protocol):
+  @staticmethod
+  def __call__(query: Array,
+               key: Array,
+               value: Array,
+               bias: Optional[Array] = None,
+               mask: Optional[Array] = None,
+               broadcast_dropout: bool = True,
+               dropout_rng: Optional[PRNGKey] = None,
+               dropout_rate: float = 0.,
+               deterministic: bool = False,
+               dtype: InexactDType = jnp.float32,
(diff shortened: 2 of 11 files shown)


NeilGirdhar:
This is a first draft.

flax/linen/module.py
```diff
@@ -498,14 +502,14 @@ def __call__(self, x):
...
-      # this stub allows pytype to accept Modules as Callables.
-      pass
+  # if typing.TYPE_CHECKING:
+  #   def __init__(*args, **kwargs):
+  #     # this stub makes sure pytype accepts constructor arguments.
+  #     pass
+  #
+  #   def __call__(self, *args, **kwargs) -> Any:
+  #     # this stub allows pytype to accept Modules as Callables.
+  #     pass
```
NeilGirdhar, line 512: I can reintroduce this or remove it, however this is a Liskov violation.  Why do you need Modules to be arbitrary callables?

jheek: It's a bit lazy agreed, but common. In Pytorch Module you have:
```
forward: Callable[..., Any] = _forward_unimplemented
```
which amounts to the same.

NeilGirdhar:
@jheek 
I worked on this pull request as you requested, but I need some advice on how to complete it.

In particular, I'm having a problem sorting out how `RNNCellBase.initialize_carry` should be defined and used.

First, there's a Liskov violation.  The base class method is a promise that all child classes can be called using the base class interface, but `ConvLSTM` appears to change the size parameter from an integer to a tuple of integers.  There are a few options:
* Make all of the size parameters accept tuples,
* Add an _optional_ `input_shape` tuple to `ConvLSTM`, or
* Don't make `ConvLSTM` inherit from `RNNCellBase`.

I think removing the inheritance might make the most sense since there's no benefit to the polymorphism here.

The other problem has to do with passing the dtype to the initialization function.  Either:
* the user creates both the carry and the module, and must pass the exact same dtype to both,
* the user creates the module, and the module creates the carry, or
* the user creates an ordinary non-module dataclass, which creates both the module and the carry.

The second pattern seems more normal, but I can't figure out how to use it with Flax modules.  If you wouldn't mind taking a look at examples/sst2/models.py in this pull request—even if you don't prefer this solution—I'd like to learn how this can be accomplished.

Please let me know what you think!

NeilGirdhar:
Also, if possible, it would be nice to get https://github.com/google/flax/pull/1812 in before this pull request since being able to use a type checker without a all the spurious errors makes coding a lot faster.

NeilGirdhar:
I re-engineered the RNN classes so that initialization happens in one place rather than two.

NeilGirdhar:
@jheek When you have time, this pull request is ready for your review :smile:

PhilipVinc:
Bumpety 
Would love to see this merged!

jheek:
> Also, re-engineer RNN modules to simplify initialization.

Can you factor out re-engineering the carry initialization? This is a breaking change and will require some additional work

NeilGirdhar:
@jheek Sure!

NeilGirdhar:
@jheek Should be good now!  I didn't touch the recurrent modules since I don't believe there's a nicer solution than redesigning them.

flax/linen/attention.py
```diff
@@ -15,23 +15,42 @@
...
 from flax.linen.module import Module, compact, merge_param
-from flax.linen.initializers import zeros
 
 PRNGKey = Any
-Shape = Tuple[int]
-Dtype = Any
+Shape = Tuple[int, ...]
+InexactDType = Type[jnp.inexact]
 Array = Any
+Initializer = Callable[[PRNGKey, Shape, InexactDType], Array]
```
PhilipVinc, line 37: Not for this PR, but it might be useful also for downstream packages to put the definitions of those types in a `flax.typing` or `flax.types` submodule?

flax/linen/activation.py
```diff
@@ -67,8 +67,11 @@ def __call__(self, inputs: Array) -> Array:
     Returns:
       The transformed input.
     """
+    dtype = inputs.dtype
```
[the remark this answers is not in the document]
NeilGirdhar: What `param_dtype`?  There's no parameter.

flax/linen/attention.py
NeilGirdhar: Yeah, `Initializer` definitely belongs somewhere better.

FYI, I suggested that `Shape` be added to numpy ages ago, and they rightly turned down [the suggestion](https://github.com/numpy/numpy/issues/16657).  I think Stephen's reasoning applies here that the others are probably too short and simple?

NeilGirdhar: I added https://github.com/google/jax/pull/9596, which if accepted, would put `Initializer` in `jax.nn` where I think it belongs.

NeilGirdhar: By the way, I just found out that MyPy has a crazy bug if you name a file `typing` or `types`: https://github.com/python/mypy/issues/1876

flax/linen/activation.py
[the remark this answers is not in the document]
PhilipVinc: sorry, i mixed up the lines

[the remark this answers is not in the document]
NeilGirdhar: no worries :smile: happens to me all the time

PhilipVinc:
Some internal issues are holding this back?
I'm waiting on this to release a new version of netKet... do you have a rough timeline for when this will be merged?

flax/linen/linear.py
```diff
@@ -51,6 +54,38 @@ def _canonicalize_tuple(x: Union[Sequence[int], int]) -> Tuple[int, ...]:
     return (x,)
 
 
+def _canonicalize_dtypes(
+    input_dtype: InexactDType,
+    param_dtype: Optional[InexactDType],
+    computation_dtype: Optional[InexactDType]) -> Tuple[InexactDType,
+                                                        InexactDType]:
+  returned_param_dtype = input_dtype if param_dtype is None else param_dtype
```
jheek, line 62: param_dtype should still be float32 by default (see FLIP)

flax/linen/linear.py
```diff
@@ -51,6 +54,38 @@ def _canonicalize_tuple(x: Union[Sequence[int], int]) -> Tuple[int, ...]:
...
+    param_dtype: Optional[NumericDType],
+    computation_dtype: Optional[NumericDType]) -> Tuple[NumericDType,
+                                                        NumericDType]:
+  returned_param_dtype = input_dtype if param_dtype is None else param_dtype
+  dtype = (jnp.result_type(input_dtype, returned_param_dtype)
+           if computation_dtype is None else computation_dtype)
+
+  assert jnp.issubdtype(input_dtype, jnp.number)
+  if jnp.issubdtype(input_dtype, jnp.complexfloating):
+    assert jnp.issubdtype(returned_param_dtype, jnp.complexfloating)
```
jheek, line 84: complex numbers can still be projected by a real transformation

flax/linen/linear.py
```diff
@@ -51,6 +54,38 @@ def _canonicalize_tuple(x: Union[Sequence[int], int]) -> Tuple[int, ...]:
...
+    assert jnp.issubdtype(dtype, jnp.complexfloating)
```
jheek, line 85: casting back to real is currently allowed. We should look into disabling this in JAX instead of doing it inconsistently in this PR I think

flax/linen/linear.py
```diff
@@ -60,8 +95,8 @@ class DenseGeneral(Module):
       (-2, -1) will apply the transformation to the last two axes.
     batch_dims: tuple with batch axes.
     use_bias: whether to add a bias to the output (default: True).
-    dtype: the dtype of the computation (default: float32).
-    param_dtype: the dtype passed to parameter initializers (default: float32).
+    dtype: the dtype of the computation (default: None).
+    param_dtype: the dtype passed to parameter initializers (default: None).
```
jheek, line 99: default: float32

flax/linen/linear.py
```diff
@@ -51,6 +54,38 @@ def _canonicalize_tuple(x: Union[Sequence[int], int]) -> Tuple[int, ...]:
...
+           if computation_dtype is None else computation_dtype)
+
+  assert jnp.issubdtype(input_dtype, jnp.inexact)
+  if jnp.issubdtype(input_dtype, jnp.complexfloating):
+    assert jnp.issubdtype(returned_param_dtype, jnp.complexfloating)
+    assert jnp.issubdtype(dtype, jnp.complexfloating)
+  return returned_param_dtype, dtype
+
+
+def _canonicalize_numeric_dtypes(
```
jheek, line 73: I think it would be nice to expose and document these APIs publicly so people can implement their own layers according to the spec. 

I think you can have a canonicalize and canonicalize_dtype_inexact.

Also I think you should use the true dtypes of the params and not the param_dtype because users can cast params after construction

jheek (changes requested):

jheek:
@NeilGirdhar I finally found the time to look at the changes. Please have a look.

flax/linen/activation.py
```diff
@@ -67,8 +67,11 @@ def __call__(self, inputs: Array) -> Array:
...
+    assert jnp.issubdtype(dtype, jnp.floating)
     negative_slope = self.param(
       'negative_slope',
-      lambda k: jnp.asarray(self.negative_slope_init, jnp.float32)
+      lambda k: jnp.asarray(self.negative_slope_init, dtype)
```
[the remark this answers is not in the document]
NeilGirdhar: Sure, I can do that.  Here, I was assuming the parameter always has type `float` since that's how it's annotated.  And then I'm casting it to the input dtype.  But now that I think about it again, I guess it does make sense to store the parameter in a custom dtype.

flax/linen/activation.py
```diff
@@ -67,8 +67,11 @@ def __call__(self, inputs: Array) -> Array:
...
     )
-    return jnp.where(inputs >= 0, inputs, jnp.asarray(negative_slope, inputs.dtype) * inputs)
+    assert negative_slope.shape == ()
```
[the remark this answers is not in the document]
NeilGirdhar: Thanks for letting me know!

flax/linen/attention.py
```diff
@@ -323,38 +321,72 @@ def __call__(self,
...
+  @compact
+  def __call__(self,
+               inputs_q: Array,
+               inputs_kv: Array,
+               mask: Optional[Array] = None,
+               deterministic: Optional[bool] = None):
+    return self._apply(inputs_q, inputs_kv, mask, deterministic=deterministic)
+
+
+class SelfAttention(_BaseMultiHeadDotProductAttention):
```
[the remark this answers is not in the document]
NeilGirdhar: Fair point.  The issue is that `SelfAttention` cannot be a subclass of `MultiHeadDotProductAttention` because of a Liskov substitution error, so I just factored out the common behavior into a different class.  Would you like me to factor this change out into a different pull request?

flax/linen/linear.py
NeilGirdhar: Okay, sorry for missing this!

flax/linen/attention.py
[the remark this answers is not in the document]
NeilGirdhar: https://github.com/google/flax/pull/1962/files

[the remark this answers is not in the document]
NeilGirdhar: I added https://github.com/google/flax/pull/1962

NeilGirdhar:
@jheek Thanks for the thorough review.  It shouldn't take long to make these changes, but I'd like to wait on @avital to get back to me before I do that.  I emailed him a couple weeks ago, and haven't heard anything.  Just keeping you in the loop!

marcvanzee:
> @jheek Thanks for the thorough review. It shouldn't take long to make these changes, but I'd like to wait on @avital to get back to me before I do that. I emailed him a couple weeks ago, and haven't heard anything. Just keeping you in the loop!

Just a FYI that Avital is OOO for a few weeks, so I suppose he hasn't been checking his emails.

NeilGirdhar:
> Just a FYI that Avital is OOO for a few weeks, so I suppose he hasn't been checking his emails.

Ah okay!  No problem, I wasn't sure.  I'll try to get this done today then :smile:

flax/linen/linear.py
NeilGirdhar: Do you think it makes more sense to put the simply make the `param_dtype` default to `float32` on the modules?  That way these helper functions can remain completely type agnostic, and just defer to the input dtype if `None` is passed for the other types.

flax/linen/activation.py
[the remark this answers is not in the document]
NeilGirdhar: Done.

flax/linen/activation.py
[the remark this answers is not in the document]
NeilGirdhar: Removed.

flax/linen/linear.py
NeilGirdhar: Are you sure that differentiation with respect to such parameters will work in that case?  Consider:
```
from jax import grad
from functools import partial
import jax.numpy as jnp

x = jnp.zeros(3, dtype=jnp.complex64)

def f(x, w):
    return jnp.sum(w @ x)

grad(partial(f, x), holomorphic=True)(jnp.eye(3, dtype=jnp.complex64))  # Okay.
grad(partial(f, x), holomorphic=True)(jnp.eye(3))  # Not okay.
```

flax/linen/linear.py
NeilGirdhar: Oh, good point!  For some reason I thought it was a narrowing error.

I agree that we shouldn't try to fix this here.  Is there a Jax or numpy issue somewhere that you know of?  Maybe we should at least create one.

flax/linen/linear.py
NeilGirdhar: Done.

flax/linen/linear.py
NeilGirdhar: > I think it would be nice to expose and document these APIs publicly so people can implement their own layers according to the spec.

Yes, great point.

Did you want them in a new file?  I'll expose them here for now, but let me know if you want them somewhere else.

> Also I think you should use the true dtypes of the params 

Sorry, I'm not sure what you want me to do here.

NeilGirdhar:
@jheek First pass of the review is done.  Please let me know about the above questions when you find more time :smile:

flax/linen/linear.py
jheek: I would either make a dtypes.py or put it in module.py

> Sorry, I'm not sure what you want me to do here? 

You can pass the parameter values into jnp.result_type(*inputs, *params)

jheek: the canonicalize_dtype should only infer the dtype if it is None. param_dtype defaults to float32 and cannot be inferred (so it's not Optional[Dtype] but Dtype). This way you can init the params and then use the params + inputs to infer the dtype

NeilGirdhar: > the canonicalize_dtype should only infer the dtype if it is None

Right, that's what it does.

> This way you can init the params and then use the params + inputs to infer the dtype

What is the difference between
- passing the parameter d-type to the canonicalize function, versus
- initializing the parameters using the parameter dtype, and passing the parameters to the canonicalize function?

The exact same type inference will happen either way.

jheek: The param_dtype argument is not the same as the parameter dtype. In 99% of the cases users will stick to the default float32. For example during eval the two can diverge because a user does something like:
```
eval_params = jax.tree_map(lambda x: x.asarray(jnp.bfloat16), params)
```

The reasoning behind this common trick is that small updates during training don't accumulate well in half precision but during eval the weights are static and using half precision is a free gain in most cases.


The core idea behind canonicalize dtype is to reproduce what you normally expect from an equivalent numpy function. So imagine that we wrote nn.Dense as a pure numpy function:

```
def dense(input, kernel, bias, dtype=None):
  dtype = dtype or jnp.result_type(input, kernel, bias)
  return input @ kernel + bias
```

Here the user has to provide kernel and bias so the dtypes of the params are determined by the user and don't depend on input.dtype. The default dtype is just np.result_type(input, kernel, bias). 

The reason why we allow this dtype to be inferred is because it preserves the precision of the computation.
When inferring params however you are inferring behavior. e.g.: A complex linear mapping is something different than a linear real mapping even if you have a complex input. Similarly, a learned half precision dense does something very different in practice than a f32 dense layer even if the inputs are in half precision.

A second argument against a None default for param_dtype is that it would cause a big backwards incompatible change because many users really on f32 defaults for half precision inputs.

NeilGirdhar: > The reasoning behind this common trick is that small updates during training don't accumulate well in half precision but during eval the weights are static and using half precision is a free gain in most cases.

Okay!  I see what you're saying now.  I've started this work [here](https://github.com/NeilGirdhar/flax/blob/modified_dd/flax/linen/linear.py#L110), but it gets tricky [here](https://github.com/NeilGirdhar/flax/blob/modified_dd/flax/linen/attention.py#L281).  Hopefully, I'm on the right track now.

> A complex linear mapping is something different than a linear real mapping even if you have a complex input.

Yes that's true, but the parameter cotangents will still be complex, so I imagine that in most cases, complex inputs imply complex parameters.

> A second argument against a None default for param_dtype is that it would cause a big backwards incompatible change because many users really on f32 defaults for half precision inputs.

Right, that's why I switched all the defaults to float32.  This way, this PR just provides the ability to specify computation and parameter dtype.  It doesn't change behavior.

However, I still think you should deprecate this default in another PR.  As you say: "The core idea behind canonicalize dtype is to reproduce what you normally expect from an equivalent numpy function."  And the equivalent numpy function always produces outputs based on its input types.  It doesn't silently widen everything to float32.

Also, if someone does want half-precision (or complex, or double precision) computation throughtout their network, they will need to select it for every module.  I read the FLIP, but I think things can change in the future.  Dobule precision is already starting to be [as performant](https://blogs.nvidia.com/blog/2020/05/14/double-precision-tensor-cores/) as single precision on some GPUs.  It would be better not to bake in defaults that you can't easily remove.

jheek:
> . And as such the user should not be counting on any types or shapes of such parameters.

The variable collections are far from opaque. You make assumptions about its structure when doing transformations, taking gradients, when optimizing, etc. At the minimum they must be PyTree's of JAX arrays.

This approach is much more flexible than you would think at first glance though. For example, you can "box" a param in an arbitrary dataclass and add as much metadata as you want to parameters. Still once you do a jax.tree_map another part of the code can transform it and skip over the internal metadata without a problem.

NeilGirdhar:
@jheek 

Sorry, I've been meaning to get back to you on this, but I've had a lot of work to do.

> The variable collections are far from opaque. You make assumptions about its structure when doing transformations, taking gradients, when optimizing, etc. At the minimum they must be PyTree's of JAX arrays.

Yes, you're right that you can treat the variables as a PyTree.  What I was trying to get at by making it as opaque as possible is to make it so that users of a module X should not make assumptions about the objects that are placed into the variables tree by X.  Otherwise, the user code will break when X changes.  This ensures the _separation of concerns_.

Userfs of a module X can configure its functionality by adjusting its dataclass fields.  What you're proposing is a second way of changing its functionality by changing the data types of the contents of the variables.

I understand that people are already doing this, but it is absolutely horrible.  It's only one extra line of code to rebuild an appropriate module with the configuration you want, and then use that.  Trying to make all of the modules robust to changes in dataclass types is going to be way too much work for your team.  Besides actually making this work everywhere, you will need to test this (and there are no tests).

It's also not easy to make canonicalize do this.  You suggested that you can just put the array objects from the variables into the type inference.  That does work for simple modules like `Dense`.  It does not work for complex modules like the recursive network modules.  These modules have submodules that depend on the results of earlier computations (that need to know the dtype), and also other variables.  You'd have to collect all of the variables of all of the submodules, and there's no easy way to do that since there's no interface do that.  So, even if you wanted to do it this way, you can't do it for complex modules.

Therefore, I think this is bad design.  You should infer the computation dtype based on the module fields alone.  If people want to cast the parameters, then they'll also have to reconstruct the module.  Anyone who was relying on this undocumented behaviour will have to modify their code, unfortunately.

Best,

Neil

PhilipVinc:
I hate to bump this up again but... what's the status?

NeilGirdhar:
@PhilipVinc Unfortunately, I've decided not to spend any more time working on this.  You're welcome to lift my code into a pull request of your own if you like.

jheek:
@PhilipVinc @NeilGirdhar I'm taking over the implementation of the default dtype FLIP

PR #1812, opened 2022-01-25
Repair type errors

NeilGirdhar:
This is a minor change to fix various type annotation errors

## Checklist
      checks if that's the case).

### DIFF
--- a/flax/optim/dynamic_scale.py
+++ b/flax/optim/dynamic_scale.py
@@ -16,7 +16,7 @@
 """
 
 import functools
-from typing import NamedTuple, Any, Optional, Callable, Union, Sequence
+from typing import Any, Callable, NamedTuple, Optional, Sequence, Union
 
 from .. import struct
 
--- a/examples/linen_design_test/attention_simple.py
+++ b/examples/linen_design_test/attention_simple.py
@@ -31,7 +31,7 @@ class Dense(Module):
   kernel_init: Callable = initializers.lecun_normal()
   bias_init: Callable = initializers.zeros
   dtype: Any = jnp.float32
-  precision: Any = None
+  precision: Optional[lax.Precision] = None
 
   @compact
   def __call__(self, inputs):
--- a/examples/seq2seq/train.py
+++ b/examples/seq2seq/train.py
@@ -62,8 +62,6 @@ flags.DEFINE_integer(
     help=('Maximum length of a single input digit.'))
 
 PRNGKey = Any
-Shape = Tuple[int]
-Dtype = Any
 Array = Any
 
 
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -24,7 +24,8 @@ from flax import traceback_util
 import jax
 from jax import random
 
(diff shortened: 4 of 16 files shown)


flax/optim/dynamic_scale.py
```diff
@@ -86,7 +86,7 @@ def loss_fn(p):
   def value_and_grad(self, fun: Callable[..., Any],
                      argnums: Union[int, Sequence[int]] = 0,
                      has_aux: bool = False,
-                     axis_name: str = None,
+                     axis_name: Optional[str] = None,
```
NeilGirdhar, line 89: This was changed in the typing PEP, and type-checkers now ask that these be explicitly marked optional

NeilGirdhar:
It would also be really nice to get rid of:
```python
    def __init__(*args, **kwargs):
      # this stub makes sure pytype accepts constructor arguments.
      pass
    def __call__(self, *args, **kwargs) -> Any:
      # this stub allows pytype to accept Modules as Callables.
      pass
```
The `__call__` defintion causes MyPy to rightly complain for _every_ user-defined module:
```
cmm/haiku_module/linear.py:41: error: Signature of "__call__" incompatible with supertype "Module"  [override]
cmm/haiku_module/linear.py:41: note:      Superclass:
cmm/haiku_module/linear.py:41: note:          def __call__(self, *args: Any, **kwargs: Any) -> Any
cmm/haiku_module/linear.py:41: note:      Subclass:
cmm/haiku_module/linear.py:41: note:          def __call__(self, inputs: ndarray[Any, dtype[floating[Any]]]) -> ndarray[Any, dtype[floating[Any]]]
```
There must be an easier way to accomplish whatever it is that's trying to be accomplished with pytype?

NeilGirdhar:
(Not sure I can't reply to the review in line.)

For the comment about `Set[str]`, the cleanest would be in Python 3.9 to do `self.reservations = set[str]()`.  I'll move the annotation to a class level annotation if that seems cleaner to you.

As for the `get_variable`, I guess you could assert that the return value is a subclass of the default, but I don't think that's what you want?  The default could be None, and then I assume you don't expect `None` as the only possible return value?

NeilGirdhar:
@jheek Hey, I'm just about to rebase this again.   It's been a few weeks, any chance we could get this in soon so I don't have to keep rebasing it?

flax/optim/dynamic_scale.py
jheek: Makes sense

jheek (approved):
Thanks!

NeilGirdhar:
@marcvanzee Any idea how I fix this?  Since I'm not a Googler, I can't see the internal failure.

marcvanzee:
@jheek there seem to be internal tests failing related to this PR. Could you take a look?

NeilGirdhar:
I've rebased this off the latest pushes.

jheek:
@NeilGirdhar we had some internal issues with this PR that I fixed internally together with the merge conflicts.

NeilGirdhar:
@jheek Great, thank you!

PR #1825, opened 2022-01-27
Bumps Flax to v0.4.0

marcvanzee:
Moving from 0.3.6 --> 0.4.0 since we are removing flax.deprecated, which I think deserves a new number! 😄


jheek (approved):

PhilipVinc:
Is it too much to ask that https://github.com/google/flax/pull/1803 , which addresses most inconsitencies with complex numbers and dtypes in general, be merged before you release 0.4?

PhilipVinc:
#1803 is also technically breaking (albeit not in a huge fashion), in that as of master right now, results will always be downcasted to float32 while #1803 will switch to numpy promotion rules for output dtype of layers. It might be good for downstream users to minimise breaking changes in a single release.

jheek:
> Is it too much to ask that #1803 , which addresses most inconsitencies with complex numbers and dtypes in general, be merged before you release 0.4?

The PR is still in draft stage so I think we shouldn't block the PR. The release process is pretty lightweight so we can make a new release soon after it lands.

marcvanzee:
> #1803 is also technically breaking (albeit not in a huge fashion), in that as of master right now, results will always be downcasted to float32 while #1803 will switch to numpy promotion rules for output dtype of layers. It might be good for downstream users to minimise breaking changes in a single release.

Thanks for the feedback! I took a look at #1803, but that is a really big change and I currently can't quite determine when it will be done. Given that we want to remove our old API as soon as possible, like @jheek it seems better to go ahead and release new version now. 

Once the dtype issues have been resolved we can release another one!

avital:
And if it's a breaking change, that's a good reason to finish *a* release
before merging it, and then doing another release with the breaking
change...

PR #1842, opened 2022-02-01
Adds deprecation warnings to `flax.optim`.

andsteing:
This deprecation warning is quite noisy as it will be displayed every time a `flax.optim` instance is created.

Hopefully that will move users away from `flax.optim` in the coming months so we can remove it from Flax later this year without too much pain for external users.


avital (approved):
Thanks for seeing this refactor through!

PR #1844, opened 2022-02-01
Module tabulation

cgarciae:
# What does this PR do?
Adds mechanism to tabulate a Module, similar to Haiku's `tabulate` or Kera's `summary`.  This builds on the ideas/comments from #288 and aims to use `rich` to ultimately generate a nice table representation of the module. As previously discussed, the idea is to implement 2 steps:

1. Generate an intermediate representation
2. Render such representation

The PR has the following goals:

- [x] Create function to extract an intermediate representation for the Module (e.g. `module_info`).
- [x] Create a function that takes the intermediate representation and render is as a string (e.g. `render_table`).
- [x] Add a method to Module that directly render the module to a string (e.g. `Module.tabulate`).
- [x] Add docs
- [x] Add tests

**NOTE**: current names used are open for debate.

### DIFF
--- a/flax/linen/__init__.py
+++ b/flax/linen/__init__.py
@@ -116,4 +116,5 @@ from .transforms import (
   while_loop as while_loop,
   cond as cond
 )
+from .summary import tabulate
 # pylint: enable=g-multiple-import
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -20,23 +20,24 @@ import functools
 import inspect
 import threading
 import typing
-from typing import (Any, Callable, Dict, Iterable, List, Optional, Sequence,
-                    Set, Tuple, Type, TypeVar, Union, overload)
-from typing_extensions import dataclass_transform  # pytype: disable=not-supported-yet
 import weakref
+from typing import (Any, Callable, Dict, Iterable, List, Optional,
+                    Sequence, Set, Tuple, Type, TypeVar, Union, overload)
 
-from flax import config
-from flax import core
-from flax import errors
-from flax import serialization
-from flax import traceback_util
-from flax import traverse_util
+import jax
+import numpy as np
+from typing_extensions import \
+    dataclass_transform  # pytype: disable=not-supported-yet
+
+from flax import (config, core, errors, serialization, traceback_util,
+                  traverse_util)
 from flax.core import Scope
 from flax.core.frozen_dict import FrozenDict
-from flax.core.scope import (CollectionFilter, DenyList, FrozenVariableDict,  # pylint: disable=g-multiple-import
-                             Variable, VariableDict, union_filters)
+from flax.core.scope import (  # pylint: disable=g-multiple-import
+    CollectionFilter, DenyList, FrozenVariableDict, Variable, VariableDict,
+    union_filters)
 from flax.linen import summary 
-import jax
+
 
 
 traceback_util.register_exclusion(__file__)
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -20,7 +20,7 @@ import functools
 import inspect
 import threading
 import typing
-from typing import (Any, Callable, Dict, Iterable, List, Optional,
+from typing import (Any, Callable, Dict, Iterable, List, Optional, Sequence,
                     Set, Tuple, Type, TypeVar, Union, overload)
 from typing_extensions import dataclass_transform  # pytype: disable=not-supported-yet
 import weakref
@@ -35,6 +35,7 @@ from flax.core import Scope
 from flax.core.frozen_dict import FrozenDict
 from flax.core.scope import (CollectionFilter, DenyList, FrozenVariableDict,  # pylint: disable=g-multiple-import
                              Variable, VariableDict, union_filters)
+from flax.linen import summary 
 import jax
 
 
@@ -1353,6 +1354,95 @@ class Module:
     self.scope.put_variable(col, name, xs)
     return True
 
+  def tabulate(
+    self, 
+    rngs: Union[PRNGKey, RNGSequences],
+    *args,
+    method: Optional[Callable[..., Any]] = None,
+    mutable: CollectionFilter = True,
+    depth: Optional[int] = None,
+    output_methods: Optional[Union[Sequence[str], Set[str]]] = None,
+    **kwargs) -> str:
+    """Creates a summary of the Module represented as a table.
+
+    This method has the same signature as `init`, but instead of returning
+    the variables, it returns the string summarizing the Module in a table.
+    `tabulate` uses `jax.eval_shape` to run the forward computation without
+    consuming any FLOPs or allocating memory. 
+    
+    Example::
+
+      import jax
+      import jax.numpy as jnp
+      import flax.linen as nn
+
+      class Foo(nn.Module):
+          @nn.compact
+          def __call__(self, x):
+              h = nn.Dense(4)(x)
+              return nn.Dense(2)(h)
+
+      x = jnp.ones((16, 9))
+
+      print(Foo().tabulate(jax.random.PRNGKey(0), x))
+    
+
+    This gives the following output::
+      
+                         Foo Summary                    
+      ┏━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓
+      ┃ path    ┃ outputs       ┃ params               ┃
+      ┡━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
+      │ Inputs  │ float32[16,9] │                      │
+      ├─────────┼───────────────┼──────────────────────┤
+      │ Dense_0 │ float32[16,4] │ bias: float32[4]     │
+      │         │               │ kernel: float32[9,4] │
+      │         │               │                      │
+      │         │               │ 40 (160 B)           │
+      ├─────────┼───────────────┼──────────────────────┤
+      │ Dense_1 │ float32[16,2] │ bias: float32[2]     │
+      │         │               │ kernel: float32[4,2] │
+      │         │               │                      │
+      │         │               │ 10 (40 B)            │
+      ├─────────┼───────────────┼──────────────────────┤
+      │ Foo     │ float32[16,2] │                      │
+      ├─────────┼───────────────┼──────────────────────┤
+      │         │         Total │ 50 (200 B)           │
+      └─────────┴───────────────┴──────────────────────┘
+                                                        
+                Total Parameters: 50 (200 B)
+
+    **Note**: rows order in the table does not represent execution order, 
+    instead it aligns with the order of keys in `variables` which are sorted 
+    alphabetically.
+
+    Args:
+      rngs: The rngs for the variable collections.
+      *args: The arguments to the forward computation.
+      method: An optional method. If provided, applies this method. If not
+        provided, applies the ``__call__`` method.
+      mutable: Can be bool, str, or list. Specifies which collections should be
+        treated as mutable: ``bool``: all/no collections are mutable.
+        ``str``: The name of a single mutable collection. ``list``: A
+        list of names of mutable collections. By default all collections
+        except 'intermediates' are mutable.
+      depth: controls how many submodule deep the summary can go. By default its 
+        `None` which means no limit. If a submodule is not shown because of the 
+        depth limit, its parameter count and bytes will be added to the row of 
+        its first shown ancestor such that the sum of all rows always adds up to 
+        the total number of parameters of the Module.
+      output_methods: Method names in the `intermediates` collection that should be
+        included as outputs in the summary. The `'__call__'` method is always included.
+      **kwargs: keyword arguments to pass to the forward computation.
+
+    Returns:
+      A string summarizing the Module.
+    """
+
+    tabulate_fn = summary.tabulate(self, rngs, method=method, mutable=mutable, 
+                           depth=depth, output_methods=output_methods)
+    return tabulate_fn(*args, **kwargs)
+    
 
 def merge_param(name: str, a: Optional[T], b: Optional[T]) -> T:
   """Merges construction and call time argument.
--- a/flax/linen/summary.py
+++ b/flax/linen/summary.py
@@ -1,15 +1,28 @@
+# Copyright 2022 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
+"""Flax Module summary library."""
 import dataclasses
 import io
 from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Sequence, Set, Tuple, Union
 
-import flax
+from flax.core.scope import CollectionFilter, DenyList
 import jax
+import rich.console
 import rich.table
 import rich.text
-import rich.console
 import yaml
-from flax.core.scope import CollectionFilter, DenyList
 
 PRNGKey = Any  # pylint: disable=invalid-name
 RNGSequences = Dict[str, PRNGKey]
@@ -23,57 +36,63 @@ class Row:
   Attributes:
     path: A tuple of strings that represents the path to the module.
     outputs: Output of the Module as reported by `capture_intermediates`.
-    module_variables: Dictionary of variables in the module (no submodules included).
-    counted_variables: Dictionary of variables that should be counted for this row,
-      if no summarization is done (e.g. `depth=None` in `module_summary`) then this field
-      is the same as `module_variables`, however if a summarization is done then this
-      dictionary potentially contains parameters from submodules depending on the depth
-      of the Module in question.
+    module_variables: Dictionary of variables in the module (no submodules
+      included).
+    counted_variables: Dictionary of variables that should be counted for this
+      row, if no summarization is done (e.g. `depth=None` in `module_summary`)
+      then this field is the same as `module_variables`, however if a
+      summarization is done then this dictionary potentially contains parameters
+      from submodules depending on the depth of the Module in question.
   """
   path: Tuple[str, ...]
   outputs: Any
-  module_variables: Dict[str, Dict[str, Array]] 
+  module_variables: Dict[str, Dict[str, Array]]
   counted_variables: Dict[str, Dict[str, Array]]
 
-  def size_and_bytes(self, collections: Iterable[str]) -> Dict[str, Tuple[int, int]]:
+  def size_and_bytes(self,
+                     collections: Iterable[str]) -> Dict[str, Tuple[int, int]]:
     return {
-      col: _size_and_bytes(self.counted_variables[col])
-      if col in self.counted_variables else (0, 0)
-      for col in collections
+        col: _size_and_bytes(self.counted_variables[col])
+        if col in self.counted_variables else (0, 0) for col in collections
     }
+
 
 class Table(List[Row]):
   """A list of Row objects.
 
   Table inherits from `List[Row]` so it has all the methods of a list, however
   it also contains some additional fields:
-  
+
   * `module`: the module that this table is summarizing
   * `collections`: a list containing the parameter collections (e.g. 'params', 'batch_stats', etc)
   """
-  def __init__(self, module: 'flax.linen.Module', collections: List[str], rows: Iterable[Row]):
+
+  def __init__(self, module: 'flax.linen.Module', collections: List[str],
+               rows: Iterable[Row]):
     super().__init__(rows)
     self.module = module
     self.collections = collections
 
 
 def tabulate(
-  module: 'flax.linen.Module', 
-  rngs: Union[PRNGKey, RNGSequences],
-  method: Optional[Callable[..., Any]] = None,
-  mutable: CollectionFilter = DenyList('intermediates'),
-  depth: Optional[int] = None,
-  output_methods: Optional[Union[Sequence[str], Set[str]]] = None,
-  ) -> Callable[..., str]:
+    module: 'flax.linen.Module',
+    rngs: Union[PRNGKey, RNGSequences],
+    method: Optional[Callable[..., Any]] = None,
+    mutable: CollectionFilter = DenyList('intermediates'),
+    depth: Optional[int] = None,
+    output_methods: Optional[Union[Sequence[str], Set[str]]] = None,
+) -> Callable[..., str]:
   """Returns a function that creates a summary of the Module represented as a table.
 
   This function accepts most of the same arguments as `Module.init`, except that
-  it returns a function of the form `(*args, **kwargs) -> str` where `*args` and `**kwargs`
+  it returns a function of the form `(*args, **kwargs) -> str` where `*args` and
+  `**kwargs`
   are passed to `method` (e.g. `__call__`) during the forward pass.
-  
-  `tabulate` uses `jax.eval_shape` under the hood to run the forward computation without
-  consuming any FLOPs or allocating memory. 
-  
+
+  `tabulate` uses `jax.eval_shape` under the hood to run the forward computation
+  without
+  consuming any FLOPs or allocating memory.
+
   Example::
 
     import jax
@@ -90,11 +109,11 @@ def tabulate(
     tabulate_fn = nn.tabulate(Foo(), jax.random.PRNGKey(0))
 
     print(tabulate_fn(x))
-  
+
 
   This gives the following output::
-    
-                          Foo Summary                       
+
+                          Foo Summary
       ┏━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓
       ┃ path    ┃ outputs       ┃ params               ┃
       ┡━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩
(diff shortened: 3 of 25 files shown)


cgarciae:
@avital @marcvanzee 

The proposed `module_info` function takes in a Module and sample inputs and returns a `Table` which is just a list of `Row`s, where Row is:

```python
@dataclass
class Row:
  path: Tuple[str, ...]
  outputs: Any # outputs of the module, shapes and dtypes only
  module_variables: Dict[str, Dict[str, Array]] # module's immediate variables only 
  counted_variables: Dict[str, Dict[str, Array]] # all variables that should be counted for this row
```
The idea is that with this information you can later create a meaningful representation. Under the hood its implemented using `jax.eval_shape` over `Module.init` with `capture_intermediate=True`. It has a `depth` argument which can be used to abbreviate the table e.g. `depth=1` will stop the recursion after 1 level so it will only show the top Module and its immediate submodules. `counted_variables` specifies which variables should be counted for that row (e.g. total number or total bytes), in most cases it will be the same as `module_variables` except for when max depth is reached in which case `counted_variables` contains the variables for that module and all of its sub-modules.

cgarciae:
Next I'll implement a `render_table` function that uses rich to create a nice table representation. Design-wise it will look somewhat like this but without the `module` and `inputs` columns:
150032598-3e39f1a4-1a88-4303-9a95-6a8784a8b894

The last columns are going to be variable depending on the collections present e.g `Rng` will be removed and `Parameter` and `BatchStat` will be replace with `params` and `batch_stats`.

cgarciae:
Gave it a go at an initial implementation of `render_table`, here is a sample of how it currently looks:

Screenshot from 2022-02-02 09-18-38

Some notes:
* `variables` and `outputs` are formatted with `yaml`.
* arrays are represented as `{dtype}[{shape}]`, similar how JAX represents tracers.
* need to add checks to avoid empty collections being shown

Feedback on design appreciated.

cgarciae:
@avital 

Even though `ConvBlock` is defined like this:

```python
@dataclasses.dataclass
class ConvBlock(nn.Module):
  features: int
  kernel_size: List[int]
  training: bool

  @nn.compact
  def __call__(self, x: Array) -> Array:
      x = nn.Conv(self.features, self.kernel_size)(x)
      x = nn.BatchNorm(use_running_average=not self.training)(x)
      x = nn.Dropout(0.5, deterministic=not self.training)(x)
      x = nn.relu(x)
      return x
```
Since the `variables` keys are alphabetically ordered e.g:

```
    intermediates: {
        ConvBlock_0: {
            BatchNorm_0: {
                __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 32), dtype=float32),),
            },
            Conv_0: {
                __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 32), dtype=float32),),
            },
            Dropout_0: {
                __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 32), dtype=float32),),
            },
            __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 32), dtype=float32),),
        },
        ConvBlock_1: {
            BatchNorm_0: {
                __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 64), dtype=float32),),
            },
            Conv_0: {
                __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 64), dtype=float32),),
            },
            Dropout_0: {
                __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 64), dtype=float32),),
            },
            __call__: (ShapeDtypeStruct(shape=(32, 28, 28, 64), dtype=float32),),
        },
        Dense_0: {
            __call__: (ShapeDtypeStruct(shape=(32, 10), dtype=float32),),
        },
        __call__: ((ShapeDtypeStruct(shape=(32, 10), dtype=float32), {'a': ShapeDtypeStruct(shape=(32, 10), dtype=float32), 'b': ShapeDtypeStruct(shape=(32, 10), dtype=float32)}),),
```
It would appear as if `BatchNorm` happened before `Conv` which can be confusing if you don't know about this.

cgarciae:
I thought that it would be nice if we showed the shapes of the inputs to the main module so I added a row with that. Here is how the latest version looks:

Screenshot from 2022-02-03 08-31-48
 
 Here is the same Module but simplified with `depth=1`:
 
 
Screenshot from 2022-02-03 08-32-18

avital:
This looks really cool! One question about the design -- why make separate columns for each variable collection? Seems potentially wasteful as these columns are always only one line, whereas the others may be multiple lines. Instead, maybe you could fit the parameter count and memory footprint directly within the column for variables, something like this maybe?

[...oh! you did the layout this way so that you could summarize the total at the bottom...]

So, maybe another proposal is to drop the `variables` column entirely and embed all of the information that's there into the respective column for each of the variable collections, e.g.:

```
batch_stats
---

64 (256 B)
  mean: float32[32]
  var: float32[32]
```

cgarciae:
@avital we could strive for removing the `variables` column, however, it might look strange when you use the `depth` parameter to get less rows. Currently the `variables` column shows the variables created by that module only (not the variables by its submodules) and columns like `params` or `batch_stats` shows the counts for that module and all its submodules when its submodules are filtered out by `depth`.

This is under the assumption we want to have the `depth` filter / summarization mechanism (do we?). I kinda like `depth` because when for big models it can give you a better overview if you set depth to low values like `1` or `2`. That said, this is very subjective an I am open to ideas.

cgarciae:
@avital I implemented your suggestion of removing the `variables` column and I quick like it:

Screenshot from 2022-02-04 10-06-47

Here is with `depth=1`:

Screenshot from 2022-02-04 10-07-44

avital:
Hi @cgarciae and sorry for the late reply. I agree, I think this looks very nice!

cgarciae:
@avital I proceeded to create the `tabulate` method. Will be adding tests next.

Also, this PR introduced a lot of helper function and some classes, organization-wise are they ok there or should we move them?

avital:
@cgarciae I propose moving all of the new code to flax/linen/summary.py rather than module.py. I guess we should keep the `tabulate` method on Module for sake of discoverability but all of its implementation is probably best in a separate file. And I propose treating `module_info` (maybe rename `module_summary` in line with my proposed file name? Or we could unify on some different name) and `render_table` as public APIs as people may want to, e.g. run `module_info` on one machine and `render_table` on another, e.g. when serializing...

cgarciae:
@avital thanks for the feedback! 

* Moved helper functions/classes to `summary.py` as suggested
* Rename `module_info` to `module_summary`. 
* Exported `module_summary` and `render_table` as public attributes in the `linen` module.
* Added 3 tests.
* Added docs for `module_summary` and `render_table`.

cgarciae:
Hey @jheek! I had a talk with @marcvanzee, he suggested that I asked you if you could review this PR since both him and @avital are unavailable right now 🙏‍

marcvanzee:
@cgarciae apologies for the superlate reply. As you know we have been very busy and understaffed lately, but hopefully that will become better now! I will review your PR now. Can you please rebase and resolve the conflicts in the meantime?

cgarciae:
Hey @marcvanzee! Fixed merge issues.

Check the `test_tabulate` test if you want to play with it. One last pending issue to discuss here is that rows are not ordered by call order but alphabetically instead, in the future it would nice to modify the `capture_intermediates` feature to somehow indicate call order as well.

cc: @jheek

marcvanzee (changes requested):
First of all, this is going to be a really great addition! Users often struggle with printing their params and they always use their own little util functions.

As a high-level comment: I am having some problems rendering this table in the Python interpreter:

```
>>> nn.Dense(4).tabulate(random.PRNGKey(0), jnp.ones((3, 4)))
'\n\x1b[3m                 Dense Summary                  \x1b[0m\n┏━━━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━┓\n┃\x1b[1m \x1b[0m\x1b[1mpath  \x1b[0m\x1b[1m \x1b[0m┃\x1b[1m \x1b[0m\x1b[1moutputs     \x1b[0m\x1b[1m \x1b[0m┃\x1b[1m \x1b[0m\x1b[1mparams              \x1b[0m\x1b[1m \x1b[0m┃\n┡━━━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━┩\n│ Inputs │ \x1b[2mfloat32\x1b[0m[3,4] │                      │\n├────────┼──────────────┼──────────────────────┤\n│ *      │ \x1b[2mfloat32\x1b[0m[3,4] │ bias: \x1b[2mfloat32\x1b[0m[4]     │\n│        │              │ kernel: \x1b[2mfloat32\x1b[0m[4,4] │\n│        │              │                      │\n│        │              │ \x1b[1m20 \x1b[0m\x1b[1;2m(80 B)\x1b[0m            │\n├────────┼──────────────┼──────────────────────┤\n│\x1b[1m \x1b[0m\x1b[1m      \x1b[0m\x1b[1m \x1b[0m│\x1b[1m \x1b[0m\x1b[1m       Total\x1b[0m\x1b[1m \x1b[0m│\x1b[1m \x1b[0m\x1b[1m20 \x1b[0m\x1b[1;2m(80 B)\x1b[0m\x1b[1m           \x1b[0m\x1b[1m \x1b[0m│\n└────────┴──────────────┴──────────────────────┘\n\x1b[1m                                                \x1b[0m\n\x1b[1m          Total Parameters: 20 \x1b[0m\x1b[1;2m(80 B)\x1b[0m\x1b[1m           \x1b[0m\n\n'
```

This happens both in my ITerm terminal and in my VSCode terminal.

I am using UTF-8 encoding and I have coloring enabled in Iterm. Any idea why this happens?

I suppose something is wrong with my setup, but given that I run into this now, I fear that it may happen to other users as well...

flax/linen/summary.py
```diff
@@ -0,0 +1,435 @@
...
+  * `module`: the module that this table is summarizing
+  * `collections`: a list containing the parameter collections (e.g. "params", "batch_stats", etc)
+  """
+  def __init__(self, module: "flax.linen.Module", collections: List[str], rows: Iterable[Row]):
+    super().__init__(rows)
+    self.module = module
+    self.collections = collections
+
+
+def module_summary(
```
marcvanzee, line 59: The name seems to suggests this will return a summary, but it actually returns a Table object, which isn't very useful, so I wonder if it is useful to expose this function at all. 

Alternatively you could add a function `tabulate` to this file that calls `module_summary` (I would rename it to `_get_module_table`) and `render_table` (I would rename it to `_render_table`).

Otherwise, users may try to call `summary.module_summary` expecting a summary of the model.

marcvanzee (changes requested):

flax/linen/summary.py
```diff
@@ -0,0 +1,435 @@
...
+    class Foo(nn.Module):
+        @nn.compact
+        def __call__(self, x):
+            h = nn.Dense(4)(x)
+            return nn.Dense(2)(h)
+
+    x = jnp.ones((16, 9))
+    model = Foo()
+
+    table = nn.module_summary(model, jax.random.PRNGKey(0), x))
```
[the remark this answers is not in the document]
cgarciae: This is shorter but I am not sure its clearer 🤔

flax/linen/module.py
```diff
@@ -1359,6 +1362,93 @@ def __call__(self, x):
     self.scope.put_variable(col, name, xs)
     return True
 
+  def tabulate(
+    self, 
+    rngs: Union[PRNGKey, RNGSequences],
+    *args,
+    method: Optional[Callable[..., Any]] = None,
+    mutable: CollectionFilter = DenyList("intermediates"),
```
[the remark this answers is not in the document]
cgarciae: Sounds good, I was aiming for `tabulate` to have (almost) the same signature as `init` so this was copied over, but it wouldn't hurt to have `tabulate` be as flexible as possible.

flax/linen/summary.py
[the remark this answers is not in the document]
cgarciae: I like this! Will do.

BTW: I wish `Module.init/apply` did this as well.

flax/linen/__init__.py
```diff
@@ -35,6 +35,8 @@
 from .pooling import avg_pool, max_pool, pool
 from .recurrent import ConvLSTM, GRUCell, LSTMCell, OptimizedLSTMCell
 from .stochastic import Dropout
+from .initializers import ones, zeros
```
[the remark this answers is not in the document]
cgarciae: Doesn't register as a change from this PR 🤔

flax/linen/summary.py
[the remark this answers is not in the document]
cgarciae: Ok, will use the short version.

flax/linen/summary.py
cgarciae: Plan:
- rename `module_summary` => `_get_module_table`
- rename `render_table` => `_render_table`
- create `summary.tabulate` that follows @jheek transformation style suggestion.

flax/linen/summary.py
```diff
@@ -0,0 +1,435 @@
...
+  module_outputs = {
+    key: value
+    for key, value in variables["intermediates"].items()
+    if key in output_methods
+  }
+
+  if len(module_outputs) == 0:
+    output = None
+  elif len(module_outputs) > 1:
+    moduel_path = "/".join(path)
```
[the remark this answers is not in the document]
cgarciae: I'll rename it to `path_repr` which makes more sense.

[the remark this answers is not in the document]
cgarciae: On second thoughts I just inlined it.

tests/linen/linen_module_test.py
```diff
@@ -17,26 +17,25 @@
 import dataclasses
```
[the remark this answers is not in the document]
cgarciae: sounds good!

tests/linen/linen_module_test.py
```diff
@@ -1382,6 +1381,170 @@ class MyComponent2(Generic[T], nn.Module):
...
+
+    table = module_summary(
+      module, 
+      {"dropout":random.PRNGKey(0), "params": random.PRNGKey(1)}, 
+      x, 
+      training=True,
+    )
+
+    # 11 rows = 1 Inputs + 4 ConvBlock_0 + 4 ConvBlock_1 + 1 Dense_0 + 1 Module output
+    assert len(table) == 11
```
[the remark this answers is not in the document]
cgarciae: Pytest somehow overrides `assert` such that if you change this `11` to a `12` the error message when you run this test like this:

```bash
pytest tests/linen/summary_test.py
```
is presented in a rich form as:
```
>     assert len(table) == 12
E     AssertionError: assert 11 == 12
E      +  where 11 = len([Row(path=('Inputs',), outputs=(DeviceArray([[[[1.],\n               [1.],\n               [1.],\n               ...,\n   ...2), {'a': ShapeDtypeStruct(shape=(32, 10), dtype=float32), 'b': ShapeDtypeStruct(shape=(32, 10), dtype=float32)}),)}})])
```
Using `assertEquals` looks like this:
```
>     self.assertEqual(len(table), 12)
E     AssertionError: 11 != 12
```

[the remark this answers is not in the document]
cgarciae: Ok, I see `assertEqual` is used throughout the code base so I will change it to match the style.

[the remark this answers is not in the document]
cgarciae: I do prefer the `assert` style as it looks more natural.

flax/linen/module.py
```diff
@@ -1353,6 +1355,93 @@ def __call__(self, x):
...
+        treated as mutable: ``bool``: all/no collections are mutable.
+        ``str``: The name of a single mutable collection. ``list``: A
+        list of names of mutable collections. By default all collections
+        except "intermediates" are mutable.
+      depth: controls how many submodule deep the summary can go. By default its 
+        `None` which means no limit. If a submodule is not shown because of the 
+        depth limit, its parameter count and bytes will be added to the row of 
+        its first shown ancestor such that the sum of all rows always adds up to 
+        the total number of parameters of the Module.
+      output_methods: Method names in the `intermediates` collection that should be
```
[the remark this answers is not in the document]
cgarciae: Added brief description for `*args` and `**kwargs`.

cgarciae:
Hey @marcvanzee and @jheek!
Fixed a couple issues based on your feedback. Its ready for another round 😀

jheek:
@cgarciae you can have at most 5 commits in a PR. Please squash / rebase to fix the build

jheek (changes requested):

flax/linen/module.py
```diff
@@ -1353,6 +1355,95 @@ def __call__(self, x):
     self.scope.put_variable(col, name, xs)
     return True
 
+  def tabulate(
```
jheek, line 1358: tabulate does not yet show up in the docs. You still need to add it to flax.linen.rst Note you can click "details" on the RTD action to see a preview of how the docs will look

jheek: I think inspection/summarization makes sense

marcvanzee:
> I think inspection/summarization makes sense

+1 to inspection!

cgarciae:
Created the `Inspection` section.

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+    (array.size * array.dtype.itemsize
+    for array in jax.tree_leaves(pytree)),
+    0
+  )
+  return size, bytes
+
+def _get_rich_repr(obj):
+  import io
+
+  import rich.console
```
marcvanzee, line 389: Imports should be at the top.

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+
+  return module_variables
+
+def _render_table(table: Table) -> str:
+  rich_table = rich.table.Table(
+      show_header=True,
+      show_lines=True,
+      show_footer=True,
+      title=f"{table.module.__class__.__name__} Summary",
+      # box=rich.box.HORIZONTALS,
```
marcvanzee, line 300: Nit: Remove?

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+  )
+
+  rich_table.add_column("path")
+  # rich_table.add_column("variables")
```
marcvanzee, line 304: Nit: Remove?

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+      if collection in row.module_variables:
+        col_repr += _as_yaml_str(jax.tree_map(_format_value, row.module_variables[collection]))
+        col_repr += "\n\n"
+      
+      col_repr += f"[bold]{_size_and_bytes_repr(size_bytes)}[/bold]"
+      collections_size_repr.append(col_repr)
+
+    rich_table.add_row(
+      "/".join(row.path) if row.path else table.module.__class__.__name__,
+      # _as_yaml_str(jax.tree_map(_format_value, row.module_variables)),
```
marcvanzee, line 325: Nit: Remove?

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+      _as_yaml_str(jax.tree_map(_format_value, row.outputs)),
+      *collections_size_repr
+    )
+
+
+  # add footer with totals
+  rich_table.columns[1].footer = rich.text.Text.from_markup("Total", justify="right")
+
+  # get collection totals
+  colletion_total = {col: (0, 0) for col in table.collections}
```
marcvanzee, line 335: typo: `colletion_total` --> `collection_total`

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+    caption_totals = (
+      caption_totals[0] + size,
+      caption_totals[1] + bytes,
+    )
+
+  rich_table.caption_style = "bold"
+  rich_table.caption = f"\nTotal Parameters: {_size_and_bytes_repr(caption_totals)}"
+
+  return "\n" + _get_rich_repr(rich_table) + "\n"
+
```
marcvanzee, line 359:
```suggestion
```

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+    for array in jax.tree_leaves(pytree)),
+    0
+  )
+  # number of bytes
+  bytes = sum(
+    (array.size * array.dtype.itemsize
+    for array in jax.tree_leaves(pytree)),
+    0
+  )
+  return size, bytes
```
marcvanzee, line 384: Note: `bytes` is a reserved keyword in Python.

```suggestion
  leaves = jax.tree_leaves(pytree)
  size = sum(x.size for x in leaves)
  num_bytes = sum(x.size * x.dtype.itemsize for x in leaves)
```

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+
+  f = io.StringIO()
+  console = rich.console.Console(file=f, force_terminal=True)
+  console.print(obj)
+
+  return f.getvalue()
+
+def _as_yaml_str(value) -> str:
+  
+  if hasattr(value, "__len__") and len(value) == 0:
```
marcvanzee, line 399: Isn't this more general?

```suggestion
  if not value:
```

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+      explicit_end=False,
+  )
+  return file.getvalue().replace("\n...", "").replace("'", "").strip()
+
+def _format_value(value):
+  if hasattr(value, "shape") and hasattr(value, "dtype"):
+    shape_str = ",".join(map(str, value.shape))
+    return f"[dim]{value.dtype}[/dim][{shape_str}]"
+  else:
+    return str(value)
```
marcvanzee, line 418: How about this? It will output shapes as `(3, 4)` instead of `[3, 4]`, but that is actually quite standard. Or did you have a reason why you want to use square brackets?

```suggestion
def _format_value(value):
  if isinstance(value, (np.ndarray, jnp.ndarray)):
    return f"[dim]{value.dtype}[/dim][{value.shape}]"
  else:
    return str(value)
```

marcvanzee (changes requested):
Some overall comments:

* The build is failing in this PR, but the error looks unrelated. Maybe just sync to main will resolve it?
* We use single quotes in the Flax codebase, so please change that everywhere.

flax/linen/module.py
```diff
@@ -1353,6 +1355,95 @@ def __call__(self, x):
...
+        the total number of parameters of the Module.
+      output_methods: Method names in the `intermediates` collection that should be
+        included as outputs in the summary. The `"__call__"` method is always included.
+      **kwargs: keyword arguments to pass to the forward computation.
+
+    Returns:
+      A string summarizing the Module.
+    """
+
+    tabulate_fn = tabulate(self, rngs, method=method, mutable=mutable,
```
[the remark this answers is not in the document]
cgarciae: Sound good.

flax/linen/summary.py
```diff
@@ -0,0 +1,431 @@
...
+    counted_variables: Dictionary of variables that should be counted for this row,
+      if no summarization is done (e.g. `depth=None` in `module_summary`) then this field
+      is the same as `module_variables`, however if a summarization is done then this
+      dictionary potentially contains parameters from submodules depending on the depth
+      of the Module in question.
+  """
+  path: Tuple[str, ...]
+  outputs: Any
+  module_variables: Dict[str, Dict[str, Array]] # variables of the row's corresponding module
+  counted_variables: Dict[str, Dict[str, Array]] # all variables that should be counted for this row
```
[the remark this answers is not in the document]
cgarciae: Makes sense! I think they where created before the docs strings just so I knew what they did 😅

flax/linen/summary.py
cgarciae: good catch!

flax/linen/summary.py
cgarciae: Changed to `_bytes`

flax/linen/summary.py
cgarciae: I'll be keeping
```python
if hasattr(value, "shape") and hasattr(value, "dtype"):
```
as `ShapeDtypeStruct` created by `eval_shape` is not detected as an instance of `jnp.ndarray`.

flax/linen/summary.py
cgarciae: Done!

cgarciae:
Hey @marcvanzee! Thanks for detailed review.

I think the only thing missing is how do we want to render arrays:

With this change values will render like this:

**original**
```
bias: float32[2]
kernel: float32[4,2]
```
**new proposal**
```
bias: float32[(2,)]
kernel: float32[(4, 2)]
```
**alternative proposal**
```
bias: float32(2)
kernel: float32(4, 2)
```

flax/linen/summary.py
jheek: A square bracket shape is quite standard in rendering I think. See for example how jaxpr's are rendered when you print the output of jax.make_jaxpr. I think the intuition borrows from the fact that you always index a multi-dimensional array with square brackets but instead of the index you now put the length.

marcvanzee:
> Hey @marcvanzee! Thanks for detailed review.
> 
> I think the only thing missing is how do we want to render arrays:
> 
> **original**
> 
> ```
> bias: float32[2]
> kernel: float32[4,2]
> ```
> 
> **new proposal**
> 
> ```
> bias: float32[(2,)]
> kernel: float32[(4, 2)]
> ```
> 
> **alternative proposal**
> 
> ```
> bias: float32(2)
> kernel: float32(4, 2)
> ```

Having read @jheek , I think the original was actually fine, so feel free to keep using that!

flax/linen/summary.py
cgarciae: If `value` is something like `0` or `None` it would give the wrong output.

cgarciae:
@marcvanzee reverted back to square brackets `[...]`. Only one comment pending for resolution.

flax/linen/summary.py
marcvanzee: Make sense! Let's keep it as is then.

tests/linen/linen_module_test.py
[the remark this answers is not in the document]
marcvanzee: `self.assertEqual` provides a bit more functionality than just `assert`. It provides better error messages in the case of failure, and it will automatically call `assertListEqual`, `assertDictEqual`, ... if the two inputs have the same type.

marcvanzee (approved):
Looks good to me!

Minor comment for the next time: It wasn't entirely clear to me whether I should review the PR again or whether you were still addressing my comments. You can explicitly re-request a review in the main tab of the PR by clicking the cycled arrows next to a reviewer's name. This is quite helpful since it can happen that you are already addressing some of my comments, but you do not yet want me to do the review again. So then it helps if we both know that I won't do another review until you've explicitly requested it.

jheek (approved):
Looks great!

marcvanzee:
@cgarciae this is currently blocked on internal tests failing. We should just make sure we link in `rich` and get access to it, I'll look into it.

marcvanzee:
@cgarciae this change seems to add back all deprecated HOWTOs. perhaps something went wrong when syncing to main, could you please revert those changes?

marcvanzee:
@cgarciae Update: I will update it in the internal version so no action needed.

cgarciae:
Ok, or should I just sync again with main?

cgarciae:
@marcvanzee I cleaned the branch, don't know why it caught these other changes.

marcvanzee:
This change is committed but the PR is still open: https://github.com/google/flax/commit/ef6bf4054c30271a58bfabb58f3d0049ef5d851a

I am closing this PR, thank you @cgarciae!!! Awesome change!

trvs2cool:
This works really nicely for Modules using the call method.  

However, for modules using other method, it does not work.  The issue is that module.init with capture_intermediates=True only captures [__call__](https://github.com/google/flax/blob/8a148dbd5de4c327dd6b7c931a8b649431e8274e/flax/linen/module.py#L1145) intermediates (See [summary.py:202](https://github.com/google/flax/blob/8a148dbd5de4c327dd6b7c931a8b649431e8274e/flax/linen/summary.py#L202)).

I think the fix is fairly simple.  In [summary.py:184](https://github.com/google/flax/blob/8a148dbd5de4c327dd6b7c931a8b649431e8274e/flax/linen/summary.py#L184)) just pass the output_methods into the capture_intermediates by a lambda:

```  
    capture_intermediates_fn = lambda x,y: y in output_methods or y == '__call__'

    shape_variables = jax.eval_shape(lambda: module.init(
        rngs,
        *args,
        method=method,
        mutable=mutable,
        capture_intermediates=capture_intermediates_fn,
        **kwargs,
    ))```

marcvanzee:
@trvs2cool thanks for spotting this! Do you mind creating a separate issue for this problem?

PR #1848, opened 2022-02-02
Adds HOWTO "Convert PyTorch Models to Flax" to docs

matthias-wright:
This HOWTO shows how to convert PyTorch models to Flax.
The idea for this HOWTO originated in #1770, specifically [this](https://github.com/google/flax/discussions/1770#discussioncomment-1942875) comment by @marcvanzee.

### DIFF
--- a/docs/howtos/convert_pytorch_to_flax.rst
+++ b/docs/howtos/convert_pytorch_to_flax.rst
@@ -1,0 +1,276 @@
+Convert PyTorch Models to Flax
+==============================
+
+.. testsetup::
+
+  import numpy as np
+  import jax
+  from jax import random, numpy as jnp
+  import flax
+
+  from flax import linen as nn
+  from flax.core import freeze
+
+  import torch
+
+We will show how to convert PyTorch models to Flax. We will cover convolutions, fc layers, batch norm, and average pooling.
+
+
+FC Layers
+--------------------------------
+
+Let's start with fc layers. The only thing to be aware of here is that the PyTorch kernel has shape [outC, inC]
+and the Flax kernel has shape [inC, outC]. Transposing the kernel will do the trick.
+
+.. testcode::
+
+  t_fc = torch.nn.Linear(in_features=3, out_features=4)
+  
+  kernel = t_fc.weight.detach().cpu().numpy()
+  bias = t_fc.bias.detach().cpu().numpy()
+  
+  # [outC, inC] -> [inC, outC]
+  kernel = jnp.transpose(kernel, (1, 0))
+  
+  key = random.PRNGKey(0)
+  x = random.normal(key, (1, 3))
+  
+  params = freeze({'params': {'kernel': kernel, 'bias': bias}})
+  j_fc = nn.Dense(features=4)
+  j_out = j_fc.apply(params, x)
+  
+  t_x = torch.from_numpy(np.array(x))
+  t_out = t_fc(t_x)
+  t_out = t_out.detach().cpu().numpy()
+  
+  assert np.all(np.abs(j_out - t_out) < 1e-06)
+
+
+
+Convolutions
+--------------------------------
+
+Let's now look at 2D convolutions. PyTorch uses the NCHW format and Flax uses NHWC.
+Consequently, the kernels will have different shapes. The kernel in PyTorch has shape [outC, inC, kH, kW]
+and the Flax kernel has shape [kH, kW, inC, outC]. Transposing the kernel will do the trick.
+
+.. testcode::
+
+  t_conv = torch.nn.Conv2d(in_channels=3, out_channels=4, kernel_size=2, padding='valid')
+  
+  kernel = t_conv.weight.detach().cpu().numpy()
+  bias = t_conv.bias.detach().cpu().numpy()
+  
+  # [outC, inC, kH, kW] -> [kH, kW, inC, outC]
+  kernel = jnp.transpose(kernel, (2, 3, 1, 0))
+  
+  key = random.PRNGKey(0)
+  x = random.normal(key, (1, 6, 6, 3))
+  
+  params = freeze({'params': {'kernel': kernel, 'bias': bias}})
+  j_conv = nn.Conv(features=4, kernel_size=(2, 2), padding='valid')
+  j_out = j_conv.apply(params, x)
+  
+  # [N, H, W, C] -> [N, C, H, W]
+  t_x = torch.from_numpy(np.transpose(np.array(x), (0, 3, 1, 2)))
+  t_out = t_conv(t_x)
+  # [N, C, H, W] -> [N, H, W, C]
+  t_out = np.transpose(t_out.detach().cpu().numpy(), (0, 2, 3, 1))
+  
+  assert np.all(np.abs(j_out - t_out) < 1e-06)
+
+
+   
+Convolutions and FC Layers
+--------------------------------
+
+We have to be careful, when we have a model that uses convolutions followed by fc layers (ResNet, VGG, etc).
+In PyTorch, the activations will have shape [N, C, H, W] after the convolutions and are then
+reshaped to [N, C * H * W] before being fed to the fc layers.
+When we port our weights from PyToch to Flax, the activations after the convolutions will be of shape [N, H, W, C] in Flax.
+Before we reshape the activations for the fc layers, we have to transpose them to [N, C, H, W].
+
+Consider this PyTorch model:
+
+.. testcode::
+
+  class TModel(torch.nn.Module):
+
+    def __init__(self):
+      super(TModel, self).__init__()
+      self.conv = torch.nn.Conv2d(in_channels=3, out_channels=4, kernel_size=2, padding='valid')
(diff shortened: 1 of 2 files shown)


docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+the zero-padding will not be considered for the average calculation. There does not exist a similar
+parameter for |flax.linen.avg_pool()|_. However, we can easily implement a wrapper around the pooling
+operation.
+
+.. |flax.linen.avg_pool()| replace:: ``flax.linen.avg_pool()``
+.. _flax.linen.avg_pool(): https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.avg_pool.html
+
+.. testcode::
+
+   def avg_pool(inputs, window_shape, strides=None, padding='VALID'):
```
andsteing, line 229: should we also test this with an assert?

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+
+
+
+Batch Norm
+--------------------------------
+
+``torch.nn.BatchNorm2d`` uses ``0.1`` as the default value for the momentum parameter while
+|flax.linen.BatchNorm|_ uses ``0.9``. However, this corresponds to the same computation, because PyTorch multiplies
+the estimated statistic with ``(1 − momentum)`` and the new observed value with ``momentum``,
+while Flax multiplies the estimated statistic with momentum and the new observed value with ``(1 − momentum)``.
```
andsteing, line 180: nit: ``momentum``

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+
+  import numpy as np
+  import jax
+  from jax import random, numpy as jnp
+  import flax
+
+  from flax import linen as nn
+  from flax.core import freeze
+
+  import torch
```
andsteing, line 14: this is currently failing because `torch` is not listed as a `[testing]` dependency.

could you add it to to `setup.py` in the same PR?
(we'd have to look at overall check setup and run time to see if this causes a significant change before deciding if we want to add it for real)

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+  kernel = t_fc.weight.detach().cpu().numpy()
+  bias = t_fc.bias.detach().cpu().numpy()
+  
+  # [outC, inC] -> [inC, outC]
+  kernel = jnp.transpose(kernel, (1, 0))
+  
+  key = random.PRNGKey(0)
+  x = random.normal(key, (1, 3))
+  
+  params = freeze({'params': {'kernel': kernel, 'bias': bias}})
```
andsteing, line 38: you can omit freeze here

either you should say
```
variables = {'params': {'kernel': kernel, 'bias': bias}}
```

or
```
params = {'kernel': kernel, 'bias': bias}
j_fc.apply({'params': params}, x)
```

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+  j_fc = nn.Dense(features=4)
+  j_out = j_fc.apply(params, x)
+  
+  t_x = torch.from_numpy(np.array(x))
+  t_out = t_fc(t_x)
+  t_out = t_out.detach().cpu().numpy()
+  
+  assert np.all(np.abs(j_out - t_out) < 1e-06)
```
andsteing, line 46: nit:

```
np.testing.assert_almost_equal(j_out, t_out) 
```

docs/howtos/convert_pytorch_to_flax.rst
andsteing: (same comment applies below as well)

docs/howtos/convert_pytorch_to_flax.rst
andsteing: (same comment applies below as well)

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+
+
+
+
+Batch Norm
+--------------------------------
+
+``torch.nn.BatchNorm2d`` uses ``0.1`` as the default value for the momentum parameter while
+|flax.linen.BatchNorm|_ uses ``0.9``. However, this corresponds to the same computation, because PyTorch multiplies
```
andsteing, line 178: nit: in the text you use `flax.linen` but in the code you use `nn` - consider using the same name in both text and code to avoid any confusion in that matter

andsteing, line 178: nit: in the text you use `flax.linen` but in the code you use `nn` - consider using the same name in both text and code to avoid any confusion in that matter

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,276 @@
...
+     y = y / counts 
+     return y
+
+
+
+
+Transposed Convolutions
+--------------------------------
+
+``torch.nn.ConvTranspose2d`` and |flax.linen.ConvTranspose|_ are not compatible.
```
andsteing, line 267: This paragraph will make the reader wonder how a `torch.nn.ConvTranspose2d` should be converted. Do you know how?

andsteing:
This is super useful! Thanks so much for writing this up...

andsteing (changes requested):
Great HOWTO, added some comments...

matthias-wright:
Thanks for the quick review!

Regarding the paragraph on ``torch.nn.ConvTranspose2d``: there is an implementation of a gradient based transposed convolution but the [PR](https://github.com/google/jax/pull/5772) is currently pending. For now, I just linked to the PR in the HOWTO. In order to create a code example, I would have to copy the implementation from the PR into the HOWTO. But the implementation consists of 207 lines of code. I think that might make the HOWTO a bit messy. What do you think?

docs/howtos/convert_pytorch_to_flax.rst
andsteing: Note: adding `torch` adds about 30 seconds to the `Install Dependencies` step in the build. The build time variance is rather large (12 min ... 20 min)

docs/howtos/convert_pytorch_to_flax.rst
andsteing: Thanks for adding the reference. I created #1872 to track this.

andsteing (changes requested):
Thanks for the updates! One pending comment about the PyTorch compatible `avg_pool()` implementation, otherwise PR looks good to me.

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,351 @@
...
+``torch.nn.AvgPool2d`` and |nn.avg_pool()|_ are compatible when using default parameters.
+However, ``torch.nn.AvgPool2d`` has a parameter ``count_include_pad``. When ``count_include_pad=False``,
+the zero-padding will not be considered for the average calculation. There does not exist a similar
+parameter for |nn.avg_pool()|_. However, we can easily implement a wrapper around the pooling
+operation.
+
+
+.. testcode::
+
+  def pool(inputs, init, reduce_fn, window_shape, strides, padding):
```
andsteing, line 224: In the interest of keeping this short and easy to read, let's instead export the function as `nn.pool()`

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,351 @@
...
+.. _nn.pooling: https://github.com/google/flax/blob/main/flax/linen/pooling.py
+
+.. testcode::
+  
+  def avg_pool(inputs, window_shape, strides=None, padding='VALID'):
+    """
+    Pools the input by taking the average over a window.
+    In comparison to nn.avg_pool(), this pooling operation does not
+    consider the padded zero's for the average computation.
+    Args:
```
andsteing, line 293: nit: still in the interest of keeping this short and easy to read, let's just remove the `Args: ...` pydoc

docs/howtos/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,351 @@
...
+      window_shape: a shape tuple defining the window to reduce over.
+      strides: a sequence of `n` integers, representing the inter-window
+          strides (default: `(1, ..., 1)`).
+      padding: either the string `'SAME'`, the string `'VALID'`, or a sequence
+        of `n` `(low, high)` integer pairs that give the padding to apply before
+        and after each spatial dimension (default: `'VALID'`).
+    Returns:
+      The average for each window slice.
+    """
+    assert inputs.ndim == 4
```
andsteing, line 304: nit: I think the updated implementation should work with arbitrary `ndim` like the `nn.avg_pool()`

matthias-wright:
Thanks for the feedback!

andsteing (approved):

Issue #1205, opened 2021-04-06
Expanding flax.linen with common ML primitives.


phoenix-meadowlark:
I had to translate some TensorFlow models into JAX and found the following layers didn't exist:

- `nn.flatten`: In TensorFlow as [`tf.keras.layers.Flatten`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Flatten) and Torch as [`nn.Flatten`](https://pytorch.org/docs/stable/generated/torch.nn.Flatten.html#torch.nn.Flatten). Could also be implemented with `start_dim` and `end_dim` as in Torch.

```python
def flatten(inputs: Array) -> Array:
  """Flattens 'inputs', preserving the batch dim."""
  return inputs.reshape((inputs.shape[0], -1))
```

- `nn.global_avg_pool`: In TensorFlow as [`tf.keras.layers.GlobalAveragePoolingND`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/GlobalAveragePooling2D) (and not in Torch).

```python
def global_avg_pool(inputs: Array) -> Array:
  """Averages over the spacial dims of 'inputs'."""
  return jnp.mean(inputs, axis=list(range(1, len(inputs.shape) - 1)))
```

- `nn.ZeroPad`: In TensorFlow as [`tf.keras.layers.ZeroPaddingND`](https://www.tensorflow.org/api_docs/python/tf/keras/layers/ZeroPadding2D) and Torch as [`nn.ZeroPad2d`](https://pytorch.org/docs/stable/generated/torch.nn.ZeroPad2d.html#torch.nn.ZeroPad2d)

```python
class ZeroPad(nn.Module):
  padding: Sequence[Union[int, Sequence[int]]]

  @nn.compact
  def __call__(self, inputs: Array) -> Array:
    # If an int is given for a dimension's padding, apply it symmetrically.
    padding = [[p, p] if isinstance(p, int) else p for p in self.padding]
    # Don't pad the batch dim or features dim.
    padding = [[0, 0]] + padding + [[0, 0]]
    return jnp.pad(inputs, padding)
```

[...]

PR #1856, opened 2022-02-05
Add `flax.linen.ConvLocal`, a.k.a. unshared convolution

romanngg:
# What does this PR do?

Add local/unshared convolution, a.k.a. locally-connected layer - a convolution without weight sharing, i.e. having an individual kernel applied at each output spatial location (as opposed to a single kernel shared between all locations, which is used in standard convolutions).

Related to #1205 - this is equivalent to [`tf.keras.layers.LocallyConnected2D`](https://keras.io/api/layers/locally_connected_layers/), or [`neural_tangents.stax.ConvLocal`](https://neural-tangents.readthedocs.io/en/latest/neural_tangents.stax.html#neural_tangents.stax.ConvLocal).

## Checklist
      checks if that's the case).
      link).


romanngg:
Is there a way to run tests at JAX head? This relies on a recent merge https://github.com/google/jax/pull/4671

jheek:
We do need this to be in a JAX release before we can expose it in Flax

flax/linen/linear.py
```diff
@@ -199,56 +201,22 @@ def _conv_dimension_numbers(input_shape):
   return lax.ConvDimensionNumbers(lhs_spec, rhs_spec, out_spec)
 
 
-class Conv(Module):
-  """Convolution Module wrapping lax.conv_general_dilated.
+def _get_conv_call(shared_weights: bool) -> Callable:
```
jheek, line 204: why not have a shared_weights arg on Conv and perhaps a `LocalConv = partial(Conv, shared_weights=False)`?

romanngg: I agree it would be cleaner, but when I try
```python
class _Conv(Module):
  """...
  """
  shared_weights: bool
  features: int
  kernel_size: Iterable[int]
  strides: Union[None, int, Iterable[int]] = 1
  padding: Union[str, Iterable[Tuple[int, int]]] = 'SAME'
  input_dilation: Union[None, int, Iterable[int]] = 1
  kernel_dilation: Union[None, int, Iterable[int]] = 1
  feature_group_count: int = 1
  use_bias: bool = True
  dtype: Dtype = jnp.float32
  param_dtype: Dtype = jnp.float32
  precision: Any = None
  kernel_init: Callable[[PRNGKey, Shape, Dtype], Array] = default_kernel_init
  bias_init: Callable[[PRNGKey, Shape, Dtype], Array] = zeros

  ...


Conv = functools.partial(_Conv, True)


ConvLocal = functools.partial(_Conv, False)
```
I get `flax:checkpoints_test` failure as
```python
  File "/build/work/710b6a4d0078c1beccde29236a65e8895303/google3/runfiles/google3/third_party/py/flax/tests/checkpoints_test.py", [line 312](https://cs.corp.google.com/piper///depot/google3/third_party/py/flax/tests/checkpoints_test.py?l=312&ws=romann/2404&snapshot=239), in test_convert_checkpoint
    checkpoints.convert_pre_linen(state_dict['optimizer'])))
  File "/build/work/710b6a4d0078c1beccde29236a65e8895303/google3/runfiles/google3/third_party/py/flax/optim/base.py", [line 350](https://cs.corp.google.com/piper///depot/google3/third_party/py/flax/optim/base.py?l=350&ws=romann/2404&snapshot=239), in restore_state
    self.target, self.state, state)
  File "/build/work/710b6a4d0078c1beccde29236a65e8895303/google3/runfiles/google3/third_party/py/flax/optim/base.py", [line 155](https://cs.corp.google.com/piper///depot/google3/third_party/py/flax/optim/base.py?l=155&ws=romann/2404&snapshot=239), in restore_state
    opt_target = serialization.from_state_dict(opt_target, state_dict['target'])
  File "/build/work/710b6a4d0078c1beccde29236a65e8895303/google3/runfiles/google3/third_party/py/flax/serialization.py", [line 51](https://cs.corp.google.com/piper///depot/google3/third_party/py/flax/serialization.py?l=51&ws=romann/2404&snapshot=239), in from_state_dict
    return ty_from_state_dict(target, state)
  File "/build/work/710b6a4d0078c1beccde29236a65e8895303/google3/runfiles/google3/third_party/py/flax/core/frozen_dict.py", [line 199](https://cs.corp.google.com/piper///depot/google3/third_party/py/flax/core/frozen_dict.py?l=199&ws=romann/2404&snapshot=239), in _restore_frozen_dict
    for key, value in xs.items()})
  File "/build/work/710b6a4d0078c1beccde29236a65e8895303/google3/runfiles/google3/third_party/py/flax/core/frozen_dict.py", [line 199](https://cs.corp.google.com/piper///depot/google3/third_party/py/flax/core/frozen_dict.py?l=199&ws=romann/2404&snapshot=239), in <dictcomp>
    for key, value in xs.items()})
KeyError: '_Conv_0'
```
So it looks like there's some naming issue where the "children" names `Conv` and `ConvLocal` aren't used, and `_Conv` is used, do you have ideas how to fix it?

romanngg:
Updated following JAX 0.3 release - please take a look!

flax/linen/linear.py
jheek: It might be easier to use subclassing here. You could make shared_weights an abstract property of the base class. `partial` indeed gives issues with class names (and documentation rendering)

romanngg: Thanks for the suggestion, done--ptal!

examples/imagenet/models.py
```diff
@@ -131,5 +132,21 @@ def __call__(self, x, train: bool = True):
                     block_cls=BottleneckResNetBlock)
 
 
+ResNet18Local = partial(ResNet, stage_sizes=[2, 2, 2, 2],
+                        block_cls=ResNetBlock, conv=nn.ConvLocal)
+ResNet34Local = partial(ResNet, stage_sizes=[3, 4, 6, 3],
+                        block_cls=ResNetBlock, conv=nn.ConvLocal)
+ResNet50Local = partial(ResNet, stage_sizes=[3, 4, 6, 3],
```
jheek, line 139: Is there a paper link for these networks? Intuitively I would guess these networks have a huge number of params? I think even the first conv is like 100M+ params?

flax/linen/linear.py
```diff
@@ -302,26 +311,101 @@ def maybe_broadcast(x):
...
 
+class Conv(_Conv):
+  """Convolution Module wrapping `lax.conv_general_dilated`."""
+
+  @property
+  def shared_weights(self) -> bool:
+    return True
+
+
+class ConvLocal(_Conv):
```
jheek, line 401: This one still needs to be exposed in docs/flax.linen.rst

jheek (changes requested):
Overall looks good. I'm just wondering whether the ResnetLocal variants are reasonable default models to present to users.

Also the docs and CHANGELOG need an entry for the new ConvLocal

examples/imagenet/models.py
romanngg: Good point - removed ResNets with depth > 18 that aren't tested (_ResNet1 and ResNet18 are tested).

Re references, AFAIK first mentions are the old results in
https://link.springer.com/article/10.1007/BF00342633
http://yann.lecun.com/exdb/publis/pdf/lecun-89.pdf

But I am not aware of Locally-connected resnets specifically, and added these only for the purpose of simple examples / tests.

flax/linen/linear.py
romanngg: Thanks, done!

romanngg:
Thanks @jheek, addressed your suggestions, ptal!

PR #1878, opened 2022-02-09
Make it harder to accidentally pass badly structured pytree to `variables`.

### LINKED ISSUE #1768: flax.errors.ScopeParamNotFoundError: No parameter named "kernel" exists in "/MLP_0/Dense_0" when attempting to use Jax2TF with a pre-trained JAX NeRF Model
**Redirected from the JAX repo  (https://github.com/google/jax/issues/9139#issue-1096888310)**

Tensorflow vers: 2.7; JAX vers: 0.2.24; jaxlib vers: 0.1.72+cuda111; FLAX vers: 0.3.6

The following code is based on the MNIST FLAX jax2tf example, which I adapted for JAX NeRF:

```python
import collections
from os import path
from absl import app
from absl import flags
from flax.training import checkpoints
from jax import random
from jax.experimental.jax2tf.examples import saved_model_lib
from nerf import models
from nerf import utils
import tensorflow as tf

FLAGS = flags.FLAGS

utils.define_flags()

def main(unused_argv):
    rng = random.PRNGKey(20200823)
    rng, key = random.split(rng)
    utils.update_flags(FLAGS)
    utils.check_flags(FLAGS)    
    model, state = models.get_model_state(key, FLAGS, restore=False)
    print('Loading model')
    state = checkpoints.restore_checkpoint(FLAGS.train_dir, state)
    params = state.optimizer.target
    predict_fn = lambda params, input: model.apply({"params": params}, input)
    Rays = collections.namedtuple("Rays", ("origins", "directions", "viewdirs"))
    input_signatures = [Rays(origins=tf.TensorSpec((3,),tf.float32),directions=tf.TensorSpec((3,),tf.float32),viewdirs=tf.TensorSpec((3,),tf.float32))]
    saved_model_lib.convert_and_save_model(
        predict_fn,
        params,
        '/any/path/',
        input_signatures=input_signatures) 

if __name__ == "__main__":
    app.run(main)
```
In order to simplify the inputs to the network, and since I am only interested in running inference in TF, I initialize the RNG keys and `randomized` NeRF model inputs to `None` and `False` respectively, so that only the `rays` are inputted. This is the only change over the original JAX NeRF code:

```python
def __call__(self, rays, rng_0 = None, rng_1=None, randomized=False, depth_gt = None, rgb_only = False,depth_sampling = False):
        """Nerf Model.

        Args:
          rng_0: jnp.ndarray, random number generator for coarse model sampling.
          rng_1: jnp.ndarray, random number generator for fine model sampling.
          rays: util.Rays, a namedtuple of ray origins, directions, and viewdirs.
          randomized: bool, use randomized stratified sampling.
          rgb_only: bool, return only rgb

        Returns:
          ret: list, [(rgb_coarse, disp_coarse, acc_coarse), (rgb, disp, acc)]
        """
        # Stratified sampling along rays
        if (randomized):  
            key, rng_0 = random.split(rng_0)
        else:
            key = None
```
(also, every call to `model.apply()`  has its args order inverted to match this)

The error is prompted when attempting to compute the TF graph in this line of 'saved_model_lib.py':

```python
tf_graph = tf.function(lambda inputs: tf_fn(param_vars, inputs),
                         autograph=False,
                         experimental_compile=compile_model)
```

Full error stack: 

```
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/jorge/jaxnerf/nerf/save_jax_as_tf.py", line 45, in <module>
    app.run(main)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "/home/jorge/jaxnerf/nerf/save_jax_as_tf.py", line 38, in main
    saved_model_lib.convert_and_save_model(
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/jax/experimental/jax2tf/examples/saved_model_lib.py", line 114, in convert_and_save_model
    tf_graph.get_concrete_function(input_signatures[0])
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 1259, in get_concrete_function
    concrete = self._get_concrete_function_garbage_collected(*args, **kwargs)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 1239, in _get_concrete_function_garbage_collected
    self._initialize(args, kwargs, add_initializers_to=initializers)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 780, in _initialize
    self._stateful_fn._get_concrete_function_internal_garbage_collected(  # pylint: disable=protected-access
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 3157, in _get_concrete_function_internal_garbage_collected
    graph_function, _ = self._maybe_define_function(args, kwargs)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 3557, in _maybe_define_function
    graph_function = self._create_graph_function(args, kwargs)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 3392, in _create_graph_function
    func_graph_module.func_graph_from_py_func(
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/framework/func_graph.py", line 1143, in func_graph_from_py_func
    func_outputs = python_func(*func_args, **func_kwargs)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 672, in wrapped_fn
    out = weak_wrapped_fn().__wrapped__(*args, **kwds)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/jax/experimental/jax2tf/examples/saved_model_lib.py", line 107, in <lambda>
    tf_graph = tf.function(lambda inputs: tf_fn(param_vars, inputs),
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/jax/experimental/jax2tf/jax2tf.py", line 418, in converted_fun
    out_with_avals = _interpret_fun(flat_fun, args_flat, args_avals_flat,
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/jax/experimental/jax2tf/jax2tf.py", line 486, in _interpret_fun
    fun.call_wrapped(*in_vals)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/jax/linear_util.py", line 166, in call_wrapped
    ans = self.f(*args, **dict(self.params, **kwargs))
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/jax/experimental/jax2tf/jax2tf.py", line 272, in fun_no_kwargs
    return fun(*args, **kwargs)
  File "/home/jorge/jaxnerf/nerf/save_jax_as_tf.py", line 35, in <lambda>
    predict_fn = lambda params, input: model.apply({"params": params}, input)
  File "/home/jorge/jaxnerf/nerf/nerf/models.py", line 268, in __call__
    raw_rgb, raw_sigma = self.MLP_0(samples_enc)
  File "/home/jorge/jaxnerf/nerf_sh/nerf/model_utils.py", line 70, in __call__
    x = dense_layer(self.net_width)(x)
  File "/home/jorge/anaconda3/envs/jaxnerf/lib/python3.8/site-packages/flax/linen/linear.py", line 171, in __call__
    kernel = self.param('kernel',
flax.errors.ScopeParamNotFoundError: No parameter named "kernel" exists in "/MLP_0/Dense_0". (https://flax.readthedocs.io/en/latest/flax.errors.html#flax.errors.ScopeParamNotFoundError)
```

Has anyone else attempted to save a JAX NeRF model using jax2tf and encountered any such issue?

melissatan:
Try to detect if user is passing a incorrectly structured pytree to nn.Module apply() `variables` arg.
If so, print hopefully helpful error message. 

Fixes #1768

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -769,7 +769,7 @@ def bind(variables: VariableDict,
   across the JAX software ecosystem.
   """
   if not _is_valid_variables(variables):
-    raise errors.ApplyScopeInvalidVariablesError()
+    raise errors.ApplyScopeInvalidVariablesTypeError()
   if rngs is not None and not _is_valid_rngs(rngs):
     raise errors.InvalidRngError(
       'rngs should be a dictionary mapping strings to `jax.PRNGKey`.')
@@ -794,6 +794,12 @@ def apply(fn: Callable[..., Any],
               *args,
               rngs: Optional[RNGSequences] = None,
               **kwargs) -> Union[Any, Tuple[Any, VariableDict]]:
+    # Try to detect if user accidentally passed {'params': {'params': ...}.
+    if 'params' in variables and isinstance(
+        variables['params'], 
+        (dict, FrozenDict)) and 'params' in variables['params']:
+      raise errors.ApplyScopeInvalidVariablesStructureError(variables)
+
     with bind(variables, rngs=rngs, mutable=mutable).temporary() as root:
       y = fn(root, *args, **kwargs)
     if mutable is not False:
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -122,7 +122,7 @@ class InvalidRngError(FlaxError):
     super().__init__(msg)
 
 
-class ApplyScopeInvalidVariablesError(FlaxError):
+class ApplyScopeInvalidVariablesTypeError(FlaxError):
   """
   When calling :meth:`Module.apply() <flax.linen.Module.apply>`, the first
   argument should be a variable dict. For more explanation on variable dicts,
@@ -132,6 +132,18 @@ class ApplyScopeInvalidVariablesError(FlaxError):
     super().__init__('The first argument passed to an apply function should be '
                      'a dictionary of collections. Each collection should be a '
                      'dictionary with string keys.')
+
+
+class ApplyScopeInvalidVariablesStructureError(FlaxError):
+  """
+  This error is thrown when the dict passed as `variables` to apply() has an
+  extra 'params' layer, i.e. {'params': {'params': ...}}.
+  For more explanation on variable dicts, please see :mod:`flax.core.variables`.
+  """
+  def __init__(self, variables):
+    super().__init__(f'Expected the first argument passed to an apply function '
+                     'to be a dictionary containing a \'params\' key at the '
+                     'root level, but got "{variables}".')
 
 
 class ScopeParamNotFoundError(FlaxError):
@@ -176,7 +188,7 @@ class ScopeCollectionNotFound(FlaxError):
   def __init__(self, col_name, var_name, scope_path):
     super().__init__(
       f'Tried to access "{var_name}" from collection "{col_name}"" in '
-      f'"{scope_path}" but the collection is emtpy.')
+      f'"{scope_path}" but the collection is empty.')
 
 
 class ScopeParamShapeError(FlaxError):
--- a/tests/core/core_scope_test.py
+++ b/tests/core/core_scope_test.py
@@ -111,6 +111,21 @@ class ScopeTest(absltest.TestCase):
     with self.assertRaisesRegex(errors.ScopeParamShapeError, msg):
       apply(f)(freeze({'params': {'test': np.ones((2,))}}))
 
+  def test_apply_variables_bad_pytree(self):
+    def f(scope):
+      scope.param('kernel', nn.initializers.ones, (4,))
+
+    params = freeze({
+        'params': {
+            'kernel': np.ones((4,)),
+        },
+    })
+    apply(f)(params)  # Valid.
+    msg = 'dictionary containing a \'params\' key at the root level'
+    with self.assertRaisesRegex(errors.ApplyScopeInvalidVariablesStructureError,
+                                msg):
+      apply(f)({'params': params})
+      
   def test_mutate_undefined_collection(self):
     def f(scope):
       scope.put_variable('state', 'test', 123)


flax/linen/module.py
```diff
@@ -1104,6 +1104,17 @@ def other_fn(instance, ...):
     if method is None:
       method = self.__call__
     method = _get_unbound_fn(method)
+    # Try to detect if user accidentally passed {'params': {'params': ...} or
+    # {'kernel': ...} as `variables`.
+    if 'params' in variables and isinstance(
+        variables['params'], FrozenDict) and 'params' in variables['params']:
+      raise ValueError(
```
marcvanzee, line 1111: I'm thinking it may be nice to add a new error class to errors.py: https://github.com/google/flax/blob/main/flax/errors.py

Then we can add a new class covering both cases, and we explain how a variable dict should look like and what mistakes users typically make (i.e., the two mistakes we find here).

We could also add a separate error class for each case, that may be even more clear.

WDYT?

flax/linen/module.py
```diff
@@ -1104,6 +1104,17 @@ def other_fn(instance, ...):
     if method is None:
       method = self.__call__
     method = _get_unbound_fn(method)
+    # Try to detect if user accidentally passed {'params': {'params': ...} or
+    # {'kernel': ...} as `variables`.
+    if 'params' in variables and isinstance(
```
marcvanzee, line 1109: I think you can remove the check for `FrozenDict`. `variables` is not required to be a FrozenDict either, so I think this error should also be raised if it is a regular dict.

flax/linen/module.py
```diff
@@ -1104,6 +1104,17 @@ def other_fn(instance, ...):
...
+        variables['params'], FrozenDict) and 'params' in variables['params']:
+      raise ValueError(
+          'Found nested params in variables dict. '
+          'Pass variables=params instead of `{\'params\':params}`.'
+      )
+    if 'kernel' in variables and 'params' not in variables:
```
marcvanzee, line 1115: Why did you add a specific check for `kernel`? Don't we want this error to be raised always if `params` is not in variables?

Your test now passes because `kernel` is in the variable dict, but there are many modules that don't have a `kernel` param, and then we still want to throw an error.

marcvanzee (changes requested):
Thanks for fixing this lightning fast!

flax/linen/module.py
melissatan: I see, had mistakenly thought there was something special about the {"kernel": ...} case mentioned in the task description. 
Do I now understand correctly that {"kernel":...} was mentioned only as an example of the general case where the pytree is missing 'params'?

flax/linen/module.py
melissatan: if i check only `isinstance(variables['params'], dict), it won't catch flax.core.frozen_dict.FrozenDict since FrozenDict is not a dict, so the test would fail. If a real user would ever pass a FrozenDict, how about `isinstance(variables['params'], (dict, FrozenDict))`?

flax/linen/module.py
melissatan: Happy to add a new error class to cover badly structured variable dicts, including explanation as you mentioned.
I think it might be overkill to make a separate error class per case, given that these two cases are really instances of the same kind of user mistake, which means we would have to (more or less) duplicate the explanation text for those two and thus increase the maintenance burden.

flax/linen/module.py
```diff
@@ -1104,6 +1104,17 @@ def other_fn(instance, ...):
     if method is None:
       method = self.__call__
     method = _get_unbound_fn(method)
+    # Try to detect if user accidentally passed {'params': {'params': ...} or
```
jheek, line 1107: This test could go in core/scope.py (apply method) to be more general.

flax/linen/module.py
melissatan: @marcvanzee while updating this conditional I realised that the `{"kernel": ...}` case, where I assume the user had a correctly structured params dict but accidentally passed `variables=variables['params']`, could be tricky to cover properly. That's because I can't simply make apply() enforce that `variables` is a dict that contains a 'params' key, since in our tests we have something like this being used as a legitimate value for `variables`:
```
(Pdb) variables
FrozenDict({
    test: {
        inner: {
            baz: DeviceArray([1.], dtype=float32),
        },
    },
})
``` 
I'm not so familiar with real usecases yet.  Could someone advise on whether it's better to:
A) Enforce the 'params' key constraint, and update all the tests, or
B) Don't go out of our way to check for the {"kernel": ...} case, assuming it's much rarer than `{'params': {'params': ...}}`, or
C) Come up with a better heuristic? If so, any thoughts? 
Thanks!

melissatan: Taking option B for now. Filed #1909 to document findings.

flax/linen/module.py
melissatan: Done.

marcvanzee (approved):
Awesome, looks great!

jheek (approved):

Issue #1577, opened 2021-09-27
Need to update the LR schedule HOWTO to use Optax and TrainState


avital:
Currently it uses the deprecated `flax.optim` Optimizer pattern.

matthias-wright:
Is anyone working on this?

andsteing:
No, it's up for grabs!

Happy to review it if you decide to update the guide.

andsteing:
Part of #1357

PR #1903, opened 2022-02-10
Updates LR schedule HOWTO for optax

matthias-wright:
This PR addresses #1577.

Note that I replaced the triangular scheduler with the cosine scheduler from the [imagenet example](https://github.com/google/flax/blob/cb5ebcd3b9ab8db10c79e88fcdb2dc67ce2668e2/examples/imagenet/train.py#L88) because the triangular scheduler did not play well with ``jit``. I think ``jit`` cannot trace the ``create_triangular_schedule`` function because of the if else that depends on the input.

### DIFF
--- a/docs/howtos/lr_schedule.rst
+++ b/docs/howtos/lr_schedule.rst
@@ -19,7 +19,10 @@ We will show you how to...
 .. testsetup::
 
   import jax
+  import jax.numpy as jnp
+  from flax.training import train_state
   import optax
+  import functools
 
 .. testcode::
   
--- a/docs/howtos/lr_schedule.rst
+++ b/docs/howtos/lr_schedule.rst
@@ -1,138 +1,155 @@
 Learning Rate Scheduling
 =============================
 
-Note: See "Learning Rate Schedules" in :doc:`optax_update_guide` for
-implementing learning rate schedules using ``optax``.
-
-
 The learning rate is considered one of the most important hyperparameters for
 training deep neural networks, but choosing it can be quite hard.
-To simplify this, one can use a so-called *cyclic learning rate*, which
-virtually eliminates the need for experimentally finding the best value and
-schedule for the global learning rate. Instead of monotonically decreasing the
-learning rate, this method lets the learning rate cyclically vary between
-reasonable boundary values.
-Here we will show you how to implement a triangular learning rate scheduler,
-as described in the paper  `"Cyclical Learning Rates for Training Neural Networks" <https://arxiv.org/abs/1506.01186>`_.
+Rather than simply using a fixed learning rate, it is common to use a learning rate scheduler.
+In this example, we will use the *cosine scheduler*.
+Before the cosine scheduler comes into play, we start with a so-called *warmup* period in which the
+learning rate increases linearly for ``warmup_epochs`` epochs.
+For more information about the cosine scheduler, check out the paper
+`"SGDR: Stochastic Gradient Descent with Warm Restarts" <https://arxiv.org/abs/1608.03983>`_.
 
 We will show you how to...
 
 * define a learning rate schedule
 * train a simple model using that schedule
 
-The triangular schedule makes your learning rate vary as a triangle wave during training, so over the course of a period (``steps_per_cycle``
-training steps) the value will start at ``lr_min``, increase linearly to ``lr_max``, and then decrease again to ``lr_min``.
 
 .. testsetup::
 
   import jax
+  import optax
 
 .. testcode::
   
-  def create_triangular_schedule(lr_min, lr_max, steps_per_cycle):
-    top = (steps_per_cycle + 1) // 2
-    def learning_rate_fn(step):
-      cycle_step = step % steps_per_cycle
-      if cycle_step < top:
-        lr = lr_min + cycle_step/top * (lr_max - lr_min)
-      else:
-        lr = lr_max - ((cycle_step - top)/top) * (lr_max - lr_min)
-      return lr
-    return learning_rate_fn
+  def create_learning_rate_fn(config, base_learning_rate, steps_per_epoch):
+    """Create learning rate schedule."""
+    warmup_fn = optax.linear_schedule(
+        init_value=0., end_value=base_learning_rate,
+        transition_steps=config.warmup_epochs * steps_per_epoch)
+    cosine_epochs = max(config.num_epochs - config.warmup_epochs, 1)
+    cosine_fn = optax.cosine_decay_schedule(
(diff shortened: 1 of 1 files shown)


docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,158 @@
...
-        lr = lr_max - ((cycle_step - top)/top) * (lr_max - lr_min)
-      return lr
-    return learning_rate_fn
-
-
-To use the schedule, one must create a learning rate function by passing the hyperparameters to the
-create_triangular_schedule function and then use that function to compute the learning rate for your updates.
-For example using this schedule on MNIST would require changing the train_step function
+  def create_learning_rate_fn(config, base_learning_rate, steps_per_epoch):
+    """Create learning rate schedule."""
```
andsteing, line 30: nit: use descriptive-style in function pydoc:
https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,158 @@
...
+        init_value=base_learning_rate,
+        decay_steps=cosine_epochs * steps_per_epoch)
+    schedule_fn = optax.join_schedules(
+        schedules=[warmup_fn, cosine_fn],
+        boundaries=[config.warmup_epochs * steps_per_epoch])
+    return schedule_fn
+
+To use the schedule, we must create a learning rate function by passing the hyperparameters to the
+create_learning_rate_fn function and then pass the function to your ``optax`` optimizer.
+For example using this schedule on MNIST would require changing the train_step function:
```
andsteing, line 45: nit: ``train_step``

docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,158 @@
...
+    cosine_fn = optax.cosine_decay_schedule(
+        init_value=base_learning_rate,
+        decay_steps=cosine_epochs * steps_per_epoch)
+    schedule_fn = optax.join_schedules(
+        schedules=[warmup_fn, cosine_fn],
+        boundaries=[config.warmup_epochs * steps_per_epoch])
+    return schedule_fn
+
+To use the schedule, we must create a learning rate function by passing the hyperparameters to the
+create_learning_rate_fn function and then pass the function to your ``optax`` optimizer.
```
andsteing, line 44: nit: ``create_learning_rate_fn``

docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,158 @@
...
-    optimizer = optimizer.apply_gradient(grad, {"learning_rate": lr}) #!
+    (_, logits), grads = grad_fn(state.params)
+    new_state = state.apply_gradients(grads=grads)
     metrics = compute_metrics(logits, batch['label'])
-    return optimizer, metrics
+    lr = learning_rate_fn(state.step) #!
+    metrics['learning_rate'] = lr #!
+    return new_state, metrics
 
 And the train_epoch function:
```
andsteing, line 81: nit: ``train_epoch``

docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,158 @@
...
-
+  
     logging.info('train epoch: %d, loss: %.4f, accuracy: %.2f', epoch,
-                epoch_metrics['loss'], epoch_metrics['accuracy'] * 100)
+                 epoch_metrics['loss'], epoch_metrics['accuracy'] * 100)
+  
+    return state, epoch_metrics
+
+
+And the create_train_state function:
```
andsteing, line 136: nit: ``create_train_state``

docs/howtos/lr_schedule.rst
andsteing, line 44: consider adding a link to the Optax docs somewhere:
https://optax.readthedocs.io/en/latest/api.html#optimizer-schedules

andsteing (changes requested):
So it seems the former HOWTO didn't actually work?
(Since the if..else problem is not something that has to do with Optax but rather a limitation of JIT)

This kind of proves the point that we should test these functions via doctest. Could you add some test code that covers the functions?

(sometimes we also link Colabs as starting points like in the `howtos/optax_update_guide.rst` but that's probably not needed here since this is rather short)

matthias-wright:
Thanks for the feedback! No, it didn't work, at least not in my environment. With doctests, do you mean blocks of ``.. testcode::`` and ``.. testoutput::`` that are visible in the HOWTO?

matthias-wright:
I only tested the ``train_step`` function for now because testing the ``train_epoch`` function takes around 2 minutes (on my machine).

docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,230 @@
...
+  train_ds_size = config.train_ds_size
+  steps_per_epoch = train_ds_size // config.batch_size
+  learning_rate_fn = create_learning_rate_fn(config, config.learning_rate, steps_per_epoch)
+  
+  rng = jax.random.PRNGKey(0)
+  state = create_train_state(rng, config, learning_rate_fn)
+  rng, _ = jax.random.split(rng)
+
+  batch = get_dummy_batch(config.batch_size)
+  state, metrics = train_step(state, batch, learning_rate_fn)
```
andsteing, line 225: Why not set `config.train_ds_size=128` and then call `train_epoch()` instead? That shouldn't take significantly longer and would test all your code.

andsteing (changes requested):
Looks good, thanks for the updates!

I think you could indeed test `train_epoch()` (see file comment) and then the PR would be good to go.

docs/howtos/lr_schedule.rst
```diff
@@ -1,138 +1,230 @@
...
+    """Creates initial `TrainState`."""
+    cnn = CNN()
+    params = cnn.init(rng, jnp.ones([1, 28, 28, 1]))['params']
+    tx = optax.sgd(learning_rate_fn, config.momentum) #!
+    return train_state.TrainState.create(
+        apply_fn=cnn.apply, params=params, tx=tx)
+
+
+
+.. testcode::
```
andsteing, line 212: I think if you used `testsetup` here then it would still test it but not show it in the HOWTO, which would be preferable.

matthias-wright:
Thanks for the feedback! I had to use ``testcleanup`` instead of ``testsetup`` but the result is the same. The tests are executed during build but aren't visible in the HOWTO.

andsteing (approved):
Thanks for updating the HOWTO!
It's great to have it properly tested now...

PR #1920, opened 2022-02-17
Improves seq2seq example

marcvanzee:
Makes the following improvements:

- Factors out model and input pipeline code into separate files. Before it was a single file, but it was actually too much code so it made it hard to navigate around
- Splits rngs inside jitted function
- No longer use a global CharacterTable, but pass around a local one instead.
- Fixes tests with new setup

Further follow-up improvements:

- Update Colab
- Put relevant stuff of CharacterTable inside a train_state so we only have to pass around that (and it can also be passed to jitted functions)
- Add type annotations everywhere

### DIFF
--- a/examples/seq2seq/input_pipeline.py
+++ b/examples/seq2seq/input_pipeline.py
@@ -14,17 +14,17 @@
 
 """Input pipeline for seq2seq addition example."""
 
-import numpy as np
 import random
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 
 
-class CharacterTable(object):
-  """Encode/decodes between strings and integer representations."""
+class CharacterTable:
+  """Encodes/decodes between strings and integer representations."""
 
-  def __init__(self, chars, max_len_query_digit = 3):
+  def __init__(self, chars, max_len_query_digit=3):
     self._chars = sorted(set(chars))
     self._char_indices = dict(
         (ch, idx + 2) for idx, ch in enumerate(self._chars))
@@ -66,12 +66,12 @@ class CharacterTable(object):
     return (1, self.max_output_len, self.vocab_size)
 
   def encode(self, inputs):
-    """Encode from string to list of integers."""
+    """Encodes from string to list of integers."""
     return np.array(
         [self._char_indices[char] for char in inputs] + [self.eos_id])
 
   def decode(self, inputs):
-    """Decode from list of integers to string."""
+    """Decodes from list of integers to string."""
     chars = []
     for elem in inputs.tolist():
       if elem == self.eos_id:
@@ -80,7 +80,7 @@ class CharacterTable(object):
     return ''.join(chars)
 
   def encode_onehot(self, batch_inputs, max_len=None):
-    """One-hot encode a string input."""
+    """One-hot encodes a string input."""
 
     if max_len is None:
       max_len = self.max_input_len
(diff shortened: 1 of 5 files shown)


examples/seq2seq/input_pipeline.py
```diff
@@ -0,0 +1,139 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Input pipeline for seq2seq addition example."""
+
+import numpy as np
```
andsteing, line 17: nit: first standard library, then other packages, i.e.

```
import random


import jax
import jax.numpy as jnp
import numpy as np
```

examples/seq2seq/input_pipeline.py
```diff
@@ -0,0 +1,139 @@
...
+import random
+
+import jax
+import jax.numpy as jnp
+
+
+class CharacterTable(object):
```
andsteing, line 24: "new style classes" is a Python 2.x thing ...

examples/seq2seq/input_pipeline.py
```diff
@@ -0,0 +1,139 @@
...
+  def encoder_input_shape(self):
+    max_input_len = self._max_len_query_digit * 2 + 2  # includes EOS
+    return (1, max_input_len, self.vocab_size)
+
+  @property
+  def decoder_input_shape(self):
+    return (1, self.max_output_len, self.vocab_size)
+
+  def encode(self, inputs):
+    """Encode from string to list of integers."""
```
andsteing, line 69: nit: function pydoc is "descriptive style" i.e. `Encodes` and not `Encode` (here and elsewhere)

examples/seq2seq/input_pipeline.py
```diff
@@ -0,0 +1,139 @@
...
+
+    return np.array([encode_str(inp) for inp in batch_inputs])
+
+  def decode_onehot(self, batch_inputs):
+    """Decode a batch of one-hot encoding to strings."""
+    decode_inputs = lambda inputs: self.decode(inputs.argmax(axis=-1))
+    return np.array(list(map(decode_inputs, batch_inputs)))
+
+  def generate_examples(self, num_examples):
+    """Returns @num_examples examples."""
```
andsteing, line 104: nit: `Yields`

examples/seq2seq/models.py
```diff
@@ -0,0 +1,163 @@
...
+    carried_lstm_state = tuple(
+        select_carried_state(*s) for s in zip(new_lstm_state, lstm_state))
+    # Update `is_eos`.
+    is_eos = jnp.logical_or(is_eos, x[:, self.eos_id])
+    return (carried_lstm_state, is_eos), y
+
+
+  @staticmethod
+  def initialize_carry(batch_size, hidden_size):
+    # use dummy key since default state init fn is just zeros.
```
andsteing, line 59: Would `None` work instead?

examples/seq2seq/models.py
```diff
@@ -0,0 +1,163 @@
...
+    return nn.LSTMCell.initialize_carry(
+        jax.random.PRNGKey(0), (batch_size,), hidden_size)
+
+
+class Encoder(nn.Module):
+  """LSTM encoder, returning state after EOS is input."""
```
andsteing, line 65: Can't parse this sentence; should `is` be `in` ?

andsteing (approved):
Thanks for updating this example!

Added mostly small nits/suggestions.

examples/seq2seq/input_pipeline.py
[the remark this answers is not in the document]
marcvanzee: Hmm no me neither, removed it.

examples/seq2seq/models.py
marcvanzee: No, because LSTMCell.initialize_carry splits the rng.

examples/seq2seq/models.py
marcvanzee: input is used as a verb, but I'll rephrase it to make it more clear.

examples/seq2seq/models.py
```diff
@@ -0,0 +1,163 @@
...
+import functools
+from typing import Any, Tuple
+
+from flax import linen as nn
+import jax
+import jax.numpy as jnp
+import numpy as np
+
+PRNGKey = Any
+Shape = Tuple[int]
```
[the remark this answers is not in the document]
marcvanzee: Hmm no actually none of these seem to be used. Removed them.

PR #1944, opened 2022-02-25
Simplify and fix setattr logic.

levskaya:
Simplifies setattr logic and allows users to set attributes inside `__post_init__` without spurious errors.

There was too much confusing special-casing going on inside setattr, this simplifies that logic and removes some needlessly irritating constraints on our users.  Added reference to `__post_init__` usage in our error docs.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -665,27 +665,27 @@ class Module(metaclass=ModuleMeta):
     fields = self.__dataclass_fields__  # pytype: disable=attribute-error
     is_dataclass_attr = name in fields and fields[name].init
 
-    if not self._state.in_setup and self._state.is_initialized:
-      # Raises a TypeError just like frozen python dataclasses.
-      raise errors.SetAttributeFrozenModuleError(self.__class__.__name__, name,
-                                                 val)
+    if not self._state.in_setup:
+      if not self._state.is_initialized:
+        # Setting attributes before end of Module.__post_init__()
+        object.__setattr__(self, name, val)
+        return
+      else:
+        # We're past all initialization and setup logic:
+        # Raises a TypeError just like frozen python dataclasses.
+        raise errors.SetAttributeFrozenModuleError(
+            self.__class__.__name__, name, val)
+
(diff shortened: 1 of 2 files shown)


marcvanzee (approved):
Very nice and useful change! One thing that would be useful to clarify somehow is in what order `setup()` and `__post_init__()` are called. I suppose `setup()` runs before `__post_init__()`, but it is not directly clear if you just look at the names of the methods.

I can imagine it is useful to add a test that sets an attribute on `self` in `__post_init__()`, which uses some information defined in `setup()`.

flax/errors.py
```diff
@@ -423,6 +423,26 @@ def __call__(self, x):
 
     variables = SomeModule(features=3).init(random.PRNGKey(0), jnp.ones((1, )))
 
+  Alternatively, if you need to introduce canonicalization or automatically derive
+  other attributes, you can add logic to `__post_init__`:
+
+    class Foo(nn.Module):
```
jheek, line 429: This only works if `canonicalize(canonicalize(orig_attributes)) == canonicalize(orig_attributes)` otherwise cloning can cause side-effects. @avital has a PR that stores the initial attr values to solve this case (and enables atttribute updates in setup)

levskaya:
actually `__post_init__` is always called -before- `setup` - it is called automatically by the `init` method generated by the dataclass decorator.  Recall that `setup` is called -lazily- only before any user-defined method or `getattr`.

flax/errors.py
levskaya: ah right of course,  I'll axe the text in the docs for this PR then - I still think the logic of this PR is an improvement on the setattr jumble that we had before and should be compatible with any such further fixes.

levskaya: note that the motivating case was one of deriving other attributes from dataclass attribs rather than normalizing dataclass attribs directly.

levskaya: @avital - you have a link to that PR?

lucasb-eyer:
Thanks for this quick change Anselm!

PR #1948, opened 2022-02-27
Fix `get_type_hints()` for linen modules

### LINKED ISSUE #1947: `typing.get_type_hints()` is broken for linen modules
I have some serialization code that involves a recursive call to `get_type_hints()`, which breaks for flax modules:
```python
from typing import get_type_hints

from flax import linen as nn


class Network(nn.Module):
    layers: int

# Fails!
# NameError: name 'Module' is not defined
print(get_type_hints(Network))
```

The reason for this seems to be that forward references are (seemingly unnecessarily) used when fields are being dynamically added to the module dataclass, but the typing module tries to resolve these names in the wrong local namespace:
https://github.com/google/flax/blob/96c78cd1bb43dfacfb8a999f3155facec00ecb3b/flax/linen/module.py#L533-L534

This can be confirmed because adding one extra line fixes the error:
```python
from typing import get_type_hints

from flax import linen as nn
from flax.linen.module import Module, Scope, _Sentinel  # New


class Network(nn.Module):
    layers: int

# Works!
# {'layers': <class 'int'>, 'parent': typing.Union[typing.Type[flax.linen.module.Module], typing.Type[flax.core.scope.Scope], typing.Type[flax.linen.module._Sentinel], NoneType], 'name': <class 'str'>}
print(get_type_hints(Network))
```

brentyi:
# What does this PR do?

Fixes #1947 

I removed some unnecessary forward references in `nn.Module`, which were preventing the types of the module's `parent` field from being resolved correctly. Happy to add tests if it's desired/appropriate.

## Checklist
      checks if that's the case).
      [discussion](https://github.com/google/flax/discussions) #1947 

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -530,8 +530,8 @@ class Module(metaclass=ModuleMeta):
     """Handles final optional dataclass attributes: `parent` and `name`."""
     # Use cls.__dict__ to get annotations of cls itself (no parent class).
     annotations = dict(cls.__dict__.get('__annotations__', {}))
-    parent_annotation = Union[Type["Module"], Type["Scope"],
-                              Type["_Sentinel"], None]
+    parent_annotation = Union[Type[Module], Type[Scope],
+                              Type[_Sentinel], None]
     if ('parent' in annotations
         and annotations['parent'] != parent_annotation):
       raise errors.ReservedModuleAttributeError(annotations)


marcvanzee (approved):
Looks good, but I will have to run internal tests on this to verify it doesn't break anything.

PR #1949, opened 2022-02-28
Lifted while loop

jheek:
- adds while_loop to lift.py
- Adds mutable_filter to scope_fn.
  This way collections can be made
  immutable in part of the lifted transform.
  (Like the carry variables in the cond_fn of a while_loop)
- Adds while_loop to transforms.py
- Adds multi fns lifted transform support

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -770,7 +770,7 @@ def while_loop(cond_fn: Callable[[Scope, C], bool],
       local_rng_groups = []
       for rng_group, rng_split in zip(rng_groups, rng_splits):
         if rng_split:
-          rng_group = tree_map_rngs(lambda rng: random.fold_in(rng, -i), rng_group)
+          rng_group = tree_map_rngs(lambda rng: random.fold_in(rng, i), rng_group)
         local_rng_groups.append(rng_group)
       return local_rng_groups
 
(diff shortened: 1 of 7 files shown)


flax/core/lift.py
```diff
@@ -720,7 +721,90 @@ def scanned(broadcast_vars, carry, scan_variable_groups, rng_groups, args):
...
+
+  def inner(scope_fn, repack_fn,
+            variable_groups, rng_groups, carry_init):
+    carry_variables, broadcast_variables = variable_groups
+
+    def make_loop_rngs(i):
+      local_rng_groups = []
+      for rng_group, rng_split in zip(rng_groups, rng_splits):
+        if rng_split:
+          rng_group = tree_map_rngs(lambda rng: random.fold_in(rng, -i), rng_group)
```
levskaya, line 779: why flip the sign?

levskaya (approved):
Looks great, just a few things that might need deleted?

flax/core/lift.py
jheek: Otherwise cond_fn and body_fn share rngs.

levskaya: sure, I guess I was wondering why have the negative here rather than just below.  it's not important :)

jheek: Oh I see you're right. This one is redundant

PR #1951, opened 2022-02-28
Update doc tests requirements.txt

levskaya:
We were pinning to a super old JAX version and this was beginning to cause spurious test failures when docs referenced newer symbols, etc.


levskaya:
Trying to fix failure seen in #1950

marcvanzee (approved):

PR #1955, opened 2022-03-02
Define Sequential module.

### LINKED ISSUE #1155: Implement a Sequential Module

melissatan:
# Adds a Sequential module and some unit tests.

Fixes #1155

## Checklist
      checks if that's the case).
      [discussion on an earlier PR](https://github.com/google/flax/pull/1156) 

### DIFF
--- a/flax/linen/__init__.py
+++ b/flax/linen/__init__.py
@@ -17,22 +17,26 @@
 
 # pylint: disable=g-multiple-import
 # re-export commonly used modules and functions
-from .activation import (celu, elu, gelu, glu, leaky_relu, log_sigmoid,
-                         log_softmax, relu, sigmoid, soft_sign, softmax,
-                         softplus, swish, silu, tanh, PReLU)
+from .activation import (PReLU, celu, elu, gelu, glu, leaky_relu, log_sigmoid,
+                         log_softmax, relu, sigmoid, silu, soft_sign, softmax,
+                         softplus, swish, tanh)
 from .attention import (MultiHeadDotProductAttention, SelfAttention,
-                        dot_product_attention, dot_product_attention_weights,
-                        make_attention_mask, make_causal_mask, combine_masks)
-from ..core import broadcast, DenyList, FrozenDict
-from .linear import Conv, ConvTranspose, Dense, DenseGeneral, Embed
-from .module import (Module, compact, nowrap, enable_named_call,
-                     disable_named_call, override_named_call, Variable, init,
-                     init_with_output, apply, merge_param)
+                        combine_masks, dot_product_attention,
+                        dot_product_attention_weights, make_attention_mask,
+                        make_causal_mask)
+from .combinators import Sequential
+from ..core import DenyList, FrozenDict, broadcast
+from .initializers import ones, zeros
+from .linear import Conv, ConvLocal, ConvTranspose, Dense, DenseGeneral, Embed
+from .module import (Module, Variable, apply, compact,
+                     disable_named_call, enable_named_call, init,
+                     init_with_output, merge_param, nowrap, override_named_call)
 from .normalization import BatchNorm, GroupNorm, LayerNorm
-from .pooling import avg_pool, max_pool
-from .recurrent import GRUCell, LSTMCell, ConvLSTM, OptimizedLSTMCell
+from .pooling import avg_pool, max_pool, pool
+from .recurrent import ConvLSTM, GRUCell, LSTMCell, OptimizedLSTMCell
 from .stochastic import Dropout
-from .transforms import jit, named_call, checkpoint, remat, remat_scan, scan, vmap, map_variables, vjp, jvp, custom_vjp
-from .initializers import zeros, ones
+from .transforms import (checkpoint, custom_vjp, jit, jvp, map_variables,
+                         named_call, remat, remat_scan, scan, vjp, vmap,
+                         while_loop)
 
 # pylint: enable=g-multiple-import
--- a/flax/linen/combinators.py
+++ b/flax/linen/combinators.py
@@ -1,0 +1,39 @@
+"""Combinators of modules, such as a Sequential."""
+
+from typing import Callable, Sequence
+
+from flax.linen.module import Module
+
+class Sequential(Module):
+  """Applies a linear chain of Modules.
+
+  Meant to be used only for the simple case of fusing together callables where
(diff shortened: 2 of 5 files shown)


flax/linen/__init__.py
```diff
@@ -17,22 +17,26 @@
 
 # pylint: disable=g-multiple-import
 # re-export commonly used modules and functions
-from .activation import (celu, elu, gelu, glu, leaky_relu, log_sigmoid,
-                         log_softmax, relu, sigmoid, soft_sign, softmax,
-                         softplus, swish, silu, tanh, PReLU)
+from .activation import (PReLU, celu, elu, gelu, glu, leaky_relu, log_sigmoid,
```
melissatan, line 20: these changes were the result of running a formatter. i figured it's probably easier to send through this change, rather than the linter complaining.

melissatan (commented):
This "Merge branch 'main'..." commit got created by Github when I resolved a merge conflict on the web UI, I tried to squash it on the command line in IDE but this commit didn't show up when I did `git rebase -i HEAD~2`. Would appreciate a hint for how to squash it. Thanks!

marcvanzee (changes requested):
Wow, you're lightning fast, great stuff!

flax/linen/combinators.py
```diff
@@ -0,0 +1,33 @@
+"""Combinators of modules, such as a Sequential."""
+
+from typing import Callable, Sequence
+
+from flax.linen import module
+
+
+class Sequential(module.Module):
+  """Applies a linear chain of callables.
```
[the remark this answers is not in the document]
melissatan: Done.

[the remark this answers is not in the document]
melissatan: Done.

flax/linen/combinators.py
```diff
@@ -0,0 +1,33 @@
+"""Combinators of modules, such as a Sequential."""
+
+from typing import Callable, Sequence
+
+from flax.linen import module
```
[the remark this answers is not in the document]
melissatan: Done.

flax/linen/combinators.py
```diff
@@ -0,0 +1,33 @@
...
+    @nn.compact
+    def __call__(self, x):
+      return Sequential([nn.Dense(layer_size, name=f'layers_{idx}')
+                         for idx, layer_size
+                         in enumerate(self.feature_sizes)])(x)
+  ```
+  """
+  layers: Sequence[Callable]
+
+  def __call__(self, x):
```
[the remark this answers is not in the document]
melissatan: Done.

flax/linen/combinators.py
```diff
@@ -0,0 +1,44 @@
...
+  Modules will be applied in the order that they are passed in the constructor.
+
+  The apply() method of Sequential accepts any input and forwards it to the
+  first module it contains. It chains the output sequentially to the input of
+  the next module and returns the output of the final module.
+
+  Example usage:
+
+    class Foo(nn.Module):
+        feature_sizes: List[int]
```
marcvanzee, line 22: Nit: for annotating arguments abstract collection types are preferred (see [here](https://docs.python.org/3/library/typing.html#typing.List)]

```suggestion
        feature_sizes: Sequence[int]
```

flax/linen/combinators.py
```diff
@@ -0,0 +1,44 @@
...
+      if not callable(layer):
+        raise ValueError(
+            f"Each layer should implement `__call__`, but layer `{layer}` is "
+            "not callable."
+        )
+      if idx == 0:
+        outputs = layer(outputs, *args, **kwargs)
+      else:
+        outputs = layer(outputs)
+    return outputs
```
marcvanzee, line 44: Maybe we can get rid of `inputs` all together, since we can assume it will be contained in `*args`? Also, I wonder whether it is correct to return the input if the module is empty... maybe we should throw an error in that case? Otherwise it could potentially introduce bugs for the users (I don't think anyone would like to use an empty Sequential module).

How about something like this? We then don't explicitely check for `callable`, but that should be covered by the type checker. I believe the original PR had that check in as an attempt to stop the type checker to complain when using `Sequence[Module]`, but we are not using that anymore here.

```suggestion
  def __call__(self, *args, **kwargs):
    if not self.layers:
      raise ValueError(f'Empty Sequential module {self.name}.')

    y = self.layers[0](*args, **kwargs)
    for layer in self.layers[1:]:
      y = layer(y)
    return y
```

tests/linen/linen_combinators_test.py
```diff
@@ -0,0 +1,93 @@
+"""Tests for flax.linen.combinators."""
```
marcvanzee, line 1: These tests look really great!

marcvanzee (changes requested):

flax/linen/combinators.py
melissatan: Done.

tests/linen/linen_combinators_test.py
melissatan: Thanks for taking a look! Updated to add test for the empty module.

marcvanzee (approved):
Awesome, looks really good! One last question: could you add an entry to `CHANGELOG.md` for this change?

flax/linen/combinators.py
```diff
@@ -0,0 +1,39 @@
...
+  the next module and returns the output of the final module.
+
+  Example usage::
+
+    class Foo(nn.Module):
+        feature_sizes: Sequence[int]
+
+        @nn.compact
+        def __call__(self, x):
+        return nn.Sequential([nn.Dense(layer_size, name=f'layers_{idx}')
```
avital, line 26: Can we make the example include a non-linearity like `relu`? That's going to be necessary for pretty much all use-cases of `nn.Sequential` and people will want to see how it's done.

melissatan: Ok, doing this in #1978 .

PR #1960, opened 2022-03-03
Fix type for user-facing precision arguments.

levskaya:
The type for lax precision arguments was too narrow, this
adjusts the type to cover all possible cases.

### DIFF
--- a/examples/linen_design_test/attention_simple.py
+++ b/examples/linen_design_test/attention_simple.py
@@ -19,6 +19,7 @@ from flax.core import Scope
 from flax.core.frozen_dict import freeze, unfreeze
 from flax.linen import initializers
 from flax.linen import Module, compact, vmap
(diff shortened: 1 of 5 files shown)


marcvanzee (approved):
LGTM. Out of curiosity, how did you determine the values for `PrecisionLike`?

levskaya:
It's a copy of the private type definition used internally by JAX (and the set of arguments that `jax._src.lax.lax.canonicalize_precision` handles)

marcvanzee:
> It's a copy of the private type definition used internally by JAX

Ah yes I see it [here](https://github.com/google/jax/blob/main/jax/_src/lax/lax.py#L610). Isn't it less error prone to just import their definition?

levskaya:
It's not a public definition, so I don't want to import it unless they're down to export it.

PR #1961, opened 2022-03-04
Adds Linen Upgrade guide as a full HOWTO

marcvanzee:
I recently upgraded a rather large codebase to Linen and found the current upgrade guide to be quite outdated (e.g., it still used the old optimizers, it did not use TrainState, ...). I've updated it, and immediately moved it to our native HOWTO format, so it will be easier for us to maintain.


marcvanzee:
Changing reviewer to @avital since he's back and wrote most of the original upgrade guide.

avital (approved):
Thanks for doing this! It's nice to also keep a history of our migrations docs, as a habit.

PR #203, opened 2020-04-13
Update VAE input pipeline.

us:
### Old : 
**reconstruction.png:**
reconstruction_29
**sample.png:**
sample_29
**last two epoch metrics:**
```
eval epoch: 29, loss: 101.0106, BCE: 73.4271, KLD: 27.5835
eval epoch: 30, loss: 101.1614, BCE: 73.3810, KLD: 27.7803
```
### New:
**reconstruction.png:**
reconstruction_29
**sample.png:**
sample_29

**last two epoch metrics:**
```
eval epoch: 29, loss: 100.8061, BCE: 73.3755, KLD: 27.4306
eval epoch: 30, loss: 100.8589, BCE: 73.1393, KLD: 27.7196
```
**Related issues:** #138
**Old pr:** #185

PR #1962, opened 2022-03-04
Repair Liskov error with attention modules

NeilGirdhar:
Fixes # (issue)

## Checklist
      checks if that's the case).

### DIFF
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -179,31 +179,7 @@ def dot_product_attention(query: Array,
                     precision=precision)
 
 
-class MultiHeadDotProductAttention(Module):
-  """Multi-head dot-product attention.
-
-    Attributes:
-      num_heads: number of attention heads. Features (i.e. inputs_q.shape[-1])
-        should be divisible by the number of heads.
-      dtype: the dtype of the computation (default: float32)
-      param_dtype: the dtype passed to parameter initializers (default: float32).
-      qkv_features: dimension of the key, query, and value.
-      out_features: dimension of the last projection
-      broadcast_dropout: bool: use a broadcasted dropout along batch dims.
-      dropout_rate: dropout rate
-      deterministic: if false, the attention weight is masked randomly
-        using dropout, whereas if true, the attention weights
-        are deterministic.
-      precision: numerical precision of the computation see `jax.lax.Precision`
-        for details.
-      kernel_init: initializer for the kernel of the Dense layers.
-      bias_init: initializer for the bias of the Dense layers.
-      use_bias: bool: whether pointwise QKVO dense transforms use bias.
-      attention_fn: dot_product_attention or compatible function. Accepts
-        query, key, value, and returns output of shape
-        `[bs, dim1, dim2, ..., dimN,, num_heads, value_channels]``
-      decode: whether to prepare and use an autoregressive cache.
-  """
+class _BaseMultiHeadDotProductAttention(Module):
   num_heads: int
   dtype: Dtype = jnp.float32
   param_dtype: Dtype = jnp.float32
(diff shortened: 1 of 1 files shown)


NeilGirdhar:
This is a Liskov error because method overrides should accept a superset of whatever their overridden methods accept.  Otherwise, the child class cannot be used polymorphically.

flax/linen/attention.py
```diff
@@ -338,13 +313,46 @@ def __call__(self,
...
+      kernel_init: initializer for the kernel of the Dense layers.
+      bias_init: initializer for the bias of the Dense layers.
+      use_bias: bool: whether pointwise QKVO dense transforms use bias.
+      attention_fn: dot_product_attention or compatible function. Accepts
+        query, key, value, and returns output of shape
+        `[bs, dim1, dim2, ..., dimN,, num_heads, value_channels]``
+      decode: whether to prepare and use an autoregressive cache.
+  """
+  @compact
+  def __call__(self,
```
marcvanzee, line 342: Can you please format the arguments similar to `SelfAttention`? I am fine with both notations, just would be nice to be consistent.

flax/linen/attention.py
```diff
@@ -338,13 +313,46 @@ def __call__(self,
     return out
 
 
-class SelfAttention(MultiHeadDotProductAttention):
-  """Self-attention special case of multi-head dot-product attention."""
+class MultiHeadDotProductAttention(_BaseMultiHeadDotProductAttention):
+  """Multi-head dot-product attention.
 
+    Attributes:
+      num_heads: number of attention heads. Features (i.e. inputs_q.shape[-1])
```
marcvanzee, line 320: Why don't we put the attribute descriptions on the base class? I think that these should appear correctly then if you are using an auto-suggestion tool. Seems to be a bit odd to put them here, or did you have a reason for that?

Looking at RTD for `SelfAttention` now doesn't have the descriptions of the attributes: https://flax--1962.org.readthedocs.build/en/1962/_autosummary/flax.linen.SelfAttention.html#flax.linen.SelfAttention

marcvanzee (changes requested):
Nice! Just some small comments on the docstrings / layout.

flax/linen/attention.py
NeilGirdhar: Done as requested.  Just wondering though: will these show up in the documentation?  If not, we could make two copies of the arguments.

flax/linen/attention.py
NeilGirdhar: Done.

flax/linen/attention.py
marcvanzee: Did you know that you can view the documentation in your change by going to the bottom of the conversations tab and clicking on more details in the  "docs/readthedocs.org:flax" check?

This is `SelfAttention` in your branch: https://flax--1962.org.readthedocs.build/en/1962/_autosummary/flax.linen.SelfAttention.html#flax.linen.SelfAttention

marcvanzee (changes requested):
I looked at this some more, but I'm not entirely sure if this change will improve things. If we want to proceed with this change then: 

* For things to appear correctly on RTD, we will have to copy the Attributes doc strings on `BaseMultiHeadDotProductAttention`) to both `SelfAttention` and `MultiHeadDotProductAttention` (24 lines), because inheritance of parent docstrings doesn't work in RTD. This seems like quite a lot of duplication, and feels a bit odd since the *actual* attributes aren't defined in any of those classes, but in the base class. There is issue [#203](https://github.com/readthedocs/sphinx-autoapi/issues/203) in the RTD repo, which claims this should work but I tried some simple examples and it does not seem to work, and I don't see any easy way to fix this. Inheritance of docstrings from parent classes seems to be a bit tricky anyway (see e.g. [this SO thread](https://stackoverflow.com/questions/2025562/inherit-docstrings-in-python-class-inheritance)). 

* We will also have to add docstring to both `__call__` functions since these are no longer annotated either -- this seemed this does work now for `SelfAttention`, probably because it only calls `super.__call__` right now, so RTD just copied the docstring.

So if fixing the Liskov violation comes at the cost of adding all these docstrings, I am not sure whether this will make things better or worse...

@levskaya @jheek any thoughts?

PR #1964, opened 2022-03-04
Module lifecycle note

jheek:
RTD link: https://flax--1964.org.readthedocs.build/en/1964/design_notes/module_lifecycle.html

### DIFF
--- a/docs/design_notes/module_lifecycle.rst
+++ b/docs/design_notes/module_lifecycle.rst
@@ -12,7 +12,7 @@ The Module lifecycle
 
 This design note is intended for users who are already familiar with linen Modules but want to understand more about the design principles behind the abstraction. This note should give you a good understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, check out the `MNIST Tutorial <https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html>`_.
 
-linen Modules offer a pythonic abstracton on top of Flax core. The ``Module`` abstraction allows you to create classes that have state, parameters and randomness on top of JAX. This is a practical guide to the design and behavior of the ``Module`` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
+linen Modules offer a Pythonic abstracton on top of Flax core. The `Module <https://flax.readthedocs.io/en/latest/flax.linen.html#module>`_ abstraction allows you to create classes that have state, parameters and randomness on top of JAX. This is a practical guide to the design and behavior of the ``Module`` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
 
 
 Overview
@@ -21,7 +21,7 @@ Overview
 Definition
 =============
 
-Let's start with a high-level overview of the Module lifecycle. We start by defining a simple Module:
+Let's start with a high-level overview of the Module lifecycle. First, define a simple Module:
 
 
 .. testcode::
@@ -65,47 +65,46 @@ Now we want to construct and use the ``MLP`` Module:
 
 First, we construct an instance of ``MLP`` and pass the construction attributes. Note that construction here is different from what you might expect if you are not used to Functional Programming patterns. The ``MLP`` constructor does not actually create variables or any internal state whatsoever. It's best to think of it as a specification or template of the Module that contains functionality but no data.
 
-Let's first take a closer look at initialization. Surprisingly, there actually is no special initialization path in Flax. Calling ``init`` is just a special case of ``apply`` and we could also code it as:
+Let's take a closer look at initialization. Surprisingly, there actually is no seperate initialization path in Flax. Calling ``init`` is just a special case of ``apply``, which you can also write as:
 
 
-.. testcode:
+.. testcode::
 
   # equivalent to: variables = mlp.init(random.PRNGKey(0), x)
   _, variables = mlp.apply({}, x, rngs={"params": random.PRNGKey(0)}, mutable=True)
 
 
-Thus, ``init`` is nothing more than a special case of ``apply`` where:
+Thus, ``init`` is nothing more than a wrapper around ``apply`` where:
 
 #. We call a Module without any initial variables (an empty dict).
-#. A PRNG generator named “params” is always passed for randomly initializing parameters (using the parameter initialization function).
+#. A PRNG generator named ``"params"`` is always passed for randomly initializing parameters (using the parameter initialization function).
 #. All variable collections are set to mutable (``mutable=True``). When a collection is mutable, existing variables can be updated and new variables can be created. Thus, inside ``init`` variables can be initialized in any variable collection and they are all added to the returned variable dictionary.
 
 Lifecycle
 =============
 
 
-Now that we understand ``init`` is just a special case of ``apply``, let's analyze what ``.apply(...)`` does in more detail. In fact, most of the complexity of Modules resides in the ``apply`` method. The "Module lifecycle" consists of constructing and ``apply``-ing a Module. We can summarize the Module lifecycle as follows:
+Now that you have learned about ``init`` being a special case of ``apply``, let's look at ``.apply(...)`` in more detail. In fact, most of the complexity of Modules resides in the ``apply`` method. The "Module lifecycle" consists of constructing and ``apply``-ing a Module. We can summarize the Module lifecycle as follows:
 
 
-#. We construct `mlp = MLP(hidden_size=5, out_size=3)`, such that `mlp.hidden_size=5` and `mlp_out_size=3`.
+#. We construct ``mlp = MLP(hidden_size=5, out_size=3)``, such that ``mlp.hidden_size=5`` and ``mlp.out_size=3``.
 
-#. Then, call ``mlp.apply`` which performs the following:
+#. Then, call ``mlp.apply``, whichg:
 
-   #. Makes a clone of ``mlp``, let's call it `mlp_copy`.
+   #. Makes a clone of ``mlp``, let's call it ``mlp_copy``.
 
-   #. Calls `mlp_copy.setup()`.
+   #. Calls ``mlp_copy.setup()``.
 
    #. Returns the output of ``mlp_copy.__call__()`` and optionally the variable collections that were specified as mutable using the keyword argument ``mutable=``.
 
-
-You might be surprised that the lifecycle includes cloning the Module instance. You might be surprised that the lifecycle includes cloning the Module instance. In short, this is done to ensure that `apply` can be treated as a pure function (i.e., if you pass the same arguments in, it will return the same outputs). This will be explained in more detail later in the  :ref:`Top-level Modules` section.
+Notice that the lifecycle includes cloning the Module instance. This is done to ensure that ``apply`` can be treated as a pure function (i.e., if you pass the same arguments in, it will return the same outputs). You will learn about this in more detail later in the  :ref:`Top-level Modules` section.
 
 Variables
 ==========
 
-The word “variable” is ubiquitous in programming and math. However, it's important to have a good understanding of what variables are in the context of JAX and Flax. Inside Flax Modules, `variables <https://flax.readthedocs.io/en/latest/flax.linen.html#module-flax.core.variables>`_ act like you expect from Python. They are initialized once, read, and perhaps even updated every so often. However, JAX has no concept of variables. Instead, values are stored in arrays similar to NumPy arrays  -- with one important difference: they are immutable.
+The word “variable” is ubiquitous in programming and math. However, it's important to have a good understanding of what variables are in the context of JAX and Flax. Inside Flax Modules, `variables <https://flax.readthedocs.io/en/latest/flax.linen.html#module-flax.core.variables>`_ act like you expect from Python. They are initialized once, read, and perhaps even updated every so often. However, JAX has no concept of variables. Instead, values are stored in arrays similar to NumPy arrays - with one important difference: they are immutable.
 
-The ``init`` and ``apply`` methods return the variables as a nested dictionary with string keys and JAX arrays at the leaves. They are immutable and therefore really just a snapshot of state the variables are in. When ``apply`` is called again, the variable dict is passed as an argument. Such that the variables are in the same state as when the previous ``init`` / ``apply`` call finished.
+The ``init`` and ``apply`` methods return the variables as a nested dictionary with string keys and JAX arrays at the leaves. At the top level each key corresponds to a variable collection. Inside each collection the nested dict structure corresponds with the ``Module`` hierarchy. The variable dict is immutable and therefore really just a snapshot of state the variables are in. When ``apply`` is called again, the variable dict is passed as an argument. Such that the variables are in the same state as when the previous ``init`` / ``apply`` call finished.
 
 
 .. note::
@@ -131,7 +130,7 @@ Linen provides an alternative API for defining modules more compactly. This is e
       return nn.Dense(self.out_size)(h)
 
 
-A compact ``Module`` is similar in spirit to a function. It offers a concise notation and restricts external interaction to the inputs and return values of the function. In this case the concise notation might make it easier for others to understand what the Module is doing. There is no need to jump back and forth between the ``setup`` and ``__call__`` method to understand what the submodules are doing. Instead, simply reading the ``__call__`` method from top to bottom once should provide a clear picture of what the Module is doing. This can make a significant difference if you are implementing complex Modules with many hyperparameters. See `setup or compact <https://flax.readthedocs.io/en/latest/design_notes/setup_or_nncompact.html>`_ for a practical guide on decding between setup and compact.
+A compact ``Module`` is similar in spirit to a function. It offers a concise notation and restricts external interaction to the inputs and return values of the function. In this case the concise notation might make it easier for others to understand what the Module does. There is no need to jump back and forth between the ``setup`` and ``__call__`` method to understand what the submodules are doing. Instead, simply reading the ``__call__`` method from top to bottom once should provide a concise overview. This can make a significant difference if you are implementing complex Modules with many hyperparameters. See `setup or compact <https://flax.readthedocs.io/en/latest/design_notes/setup_or_nncompact.html>`_ for a practical guide on decding between setup and compact.
 
 Another benefit of defining submodules and/or variables inline is that you can add arguments to your method when constructing variables. The most common example of this is using shape information to determine the shape of a parameter like this:
 
@@ -222,24 +221,28 @@ In the above example the construction order is fixed. After construction the sub
 Top-level Modules
 *****************
 
-When a Module instance is created at the "top-level", it will be in an "unbound" state - that is, it has no variables attached. "Top-level" means it is not constructed as a sub-Module inside another Module class. Apart from calling ``init`` and ``apply``, there is not much you can do with an unbound Module. Note also that ``setup`` is not called on unbound Modules, so you can really only access the construction arguments. Refer to the :ref:`Future work` section to learn how this might change in the future.
+When a Module instance is created at the "top-level", it will be in an "unbound" state - that is, it has no variables attached. "Top-level" means it is not constructed as a sub-Module inside another Module class. Apart from calling ``init`` and ``apply``, there is not much you can do with an unbound Module. Note also that ``setup`` is not called on unbound Modules, so you can only access the construction arguments. Refer to the :ref:`Future work` section to learn how this might change in the future.
 
-Why are top-level modules always unbound?
+Why are top-level Modules always unbound?
 ===============================================
 
-When we call ``apply``, a copy of the top-level Module is created which will actually hold the variables and PRNG sequences. This stateful, "bound", clone only exists while we are executing the apply method. The reason for this is that if you create a stateful object and destroy it before the apply function returns, the ``apply`` function itself behaves like a pure function (pure means if you put the same arguments in, it will return the same outputs).
+When we call ``apply``, a copy of the top-level Module is created which will actually hold the variables and PRNG sequences. This stateful, "bound", clone only exists while we are executing the apply method. The reason for this is that if you create a stateful object and destroy it before the apply function returns, the ``apply`` function itself behaves like a pure function. A pure function has two constraints:
+
+#. If you put the same arguments in, it will return the same outputs
+#. It does not change anything outside the function. This means you cannot manipulate stateful objects that are accessible outside the pure function.
+
 
 Pure functions have many advantages but when using JAX they are often essential. For example, most code requires compilation using ``jax.jit`` to be fast and once you created a Module you probably want to optimize its parameters using ``jax.grad``. However, these APIs expect a pure function and don't work on stateful bound ``Module`` instances directly. Moreover, pure functions allow for flexible interoperability with other libraries. For example, We recommend `Optax <https://github.com/deepmind/optax>`_ for optimizing parameters. The optimizers in Optax expect and return a PyTree of JAX arrays to optimize, just like the ``apply`` function of a Linen Module.
 
 Cloning
 ===============================================
 
-To make this approach work reliably we need well-defined cloning behavior. Rather than relying on a complex nested cloning procedure like Python's ``deepcopy``, Flax enforces that a ``Module`` is exactly defined by its construction arguments. Therefore cloning a Module reduces to calling the constructor with its original construction arguments. Because ``Module`` acts as an immutable dataclass, the construction arguments are mapped 1:1 to instance attributes.
+To make this approach work reliably we need well-defined cloning behavior. Rather than relying on a complex nested cloning procedure like Python's ``deepcopy``, Flax enforces that a ``Module`` is exactly defined by its construction arguments. Therefore cloning a Module reduces to calling the constructor with its original construction arguments. Because ``Module`` acts as an immutable dataclass, the construction arguments are mapped directly to instance attributes. Non-construction attributes that are computed in ``setup`` or ``__post_init__`` should also depend only on the construciton arguments to ensure a well-defined clone.
 
 Bind
 ===============================================
 
-Sometimes it's useful to have a bound, top-level Module without having to wrap the code in a function. For example: to interact with a Module inside a Jupyter notebook. The `bind <https://flax.readthedocs.io/en/latest/flax.linen.html?highlight=bind#flax.linen.Module.bind>`_ method returns a bound clone with an unlimited lifetime. The downside of this is that you cannot combine it with JAX transformations or integrate it into a vanilla JAX codebase that expects stateless code.
+Sometimes it's useful to have a bound, top-level Module without having to wrap the code in a function. For example: to interact with a Module inside a Jupyter notebook. The `bind <https://flax.readthedocs.io/en/latest/flax.linen.html?highlight=bind#flax.linen.Module.bind>`_ method returns a bound clone with an unlimited lifetime. The downside of this is that you cannot combine it with JAX transformations or integrate it into a vanilla JAX codebase that expects stateless code. For example, `Optax <https://github.com/deepmind/optax>`_ can optimze a Pytree of parameters but it cannot directly optimize a bound ``Module`` instance created with ``.bind`` (because that's not a Pytree). Thus, you cannot combine the ``bind`` API with a functional optimizer API like Optax.
 
 
 Setup
@@ -311,7 +314,7 @@ For the most part functionalization is something that is handled automatically f
 
 Here ``inner`` takes a function that closes over a Module instance. In this example, that works fine because we are not transforming the inner method with a lifted transformation. Most methods are not transformed but it is good to know how to make Module methods transformable.
 
-The main obstacle for transformability are types that JAX does not recognize. By default JAX will recognize containers (lists, dictionaries, and tuples) of JAX/NumPy arrays and python scalars (bool, int, float, etc.). The containers are called `pytrees <https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html>`_ and the pytree API is extensible with custom types. With the flax.struct package you can also define pytree compatible dataclasses.
+The main obstacle for transformability are types that JAX does not recognize. JAX only understands `Pytree <https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html>`_ arguments. That's arbitrarily nested Python containers (dict, list, tuple) of (Jax) numpy ndarrays and Python numbers/bools. Flax allows to define dataclasses which are Pytree compatible using the `flax.struct <https://flax.readthedocs.io/en/latest/flax.struct.html>`_ API.
 
 Function closure is the most common way to accidentally hide a JAX array or Linen Module from a transformation. There is however an easy workaround if you want to pass closures that are also compatible with JAX and Linen transformations:
 
--- a/docs/design_notes/module_lifecycle.rst
+++ b/docs/design_notes/module_lifecycle.rst
@@ -1,0 +1,378 @@
+The Module lifecycle
+######################
+
+.. testsetup::
+
+  from typing import Any, Callable, Iterable
+  import flax
+  from flax import linen as nn
+  from jax import random
+  import jax
+
+
+This design note is intended for users who are already familiar with linen Modules but want to understand more about the design principles behind the abstraction. This note should give you a good understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, check out the `MNIST Tutorial <https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html>`_.
+
+linen Modules offer a pythonic abstracton on top of Flax core. The ``Module`` abstraction allows you to create classes that have state, parameters and randomness on top of JAX. This is a practical guide to the design and behavior of the ``Module`` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
+
+
+Overview
+***********
+
+Definition
+=============
+
+Let's start with a high-level overview of the Module lifecycle. We start by defining a simple Module:
+
+
+.. testcode::
+
+  class MLP(nn.Module):
+    # 1. Attribute annotations
+    hidden_size: int
+    out_size: int
+
+    # 2. The ``setup`` method
+    def setup(self):
+      self.hidden = nn.Dense(self.hidden_size)
+      self.out = nn.Dense(self.out_size)
+
+    # 3. User methods  
+    def __call__(self, x):
+      a = self.hidden(x)
+      h = nn.relu(a)
+      return self.out(h)
+
+
+This Module consists of:
+
+#. **Attribute annotations**, defined as `dataclass <https://docs.python.org/3/library/dataclasses.html>`_ fields. These annotations automatically define a constructor.
+#. **The ``setup`` method**, which creates submodules and assigns them to attributes.
+#. **User methods**. By convention, most Modules have just one  ``__call__`` method, but you can define multiple methods or use different method names.
+
+Construction/initialization
+=============================
+
+Now we want to construct and use the ``MLP`` Module:
+
+
+.. testcode::
+
+  mlp = MLP(hidden_size=5, out_size=3)
+  x = jax.numpy.ones((1, 2))
+  variables = mlp.init(random.PRNGKey(0), x)
+  y = mlp.apply(variables, x)
+
+
+First, we construct an instance of ``MLP`` and pass the construction attributes. Note that construction here is different from what you might expect if you are not used to Functional Programming patterns. The ``MLP`` constructor does not actually create variables or any internal state whatsoever. It's best to think of it as a specification or template of the Module that contains functionality but no data.
+
+Let's first take a closer look at initialization. Surprisingly, there actually is no special initialization path in Flax. Calling ``init`` is just a special case of ``apply`` and we could also code it as:
+
+
+.. testcode:
+
+  # equivalent to: variables = mlp.init(random.PRNGKey(0), x)
+  _, variables = mlp.apply({}, x, rngs={"params": random.PRNGKey(0)}, mutable=True)
+
+
+Thus, ``init`` is nothing more than a special case of ``apply`` where:
+
+#. We call a Module without any initial variables (an empty dict).
+#. A PRNG generator named “params” is always passed for randomly initializing parameters (using the parameter initialization function).
+#. All variable collections are set to mutable (``mutable=True``). When a collection is mutable, existing variables can be updated and new variables can be created. Thus, inside ``init`` variables can be initialized in any variable collection and they are all added to the returned variable dictionary.
+
+Lifecycle
+=============
+
+
+Now that we understand ``init`` is just a special case of ``apply``, let's analyze what ``.apply(...)`` does in more detail. In fact, most of the complexity of Modules resides in the ``apply`` method. The "Module lifecycle" consists of constructing and ``apply``-ing a Module. We can summarize the Module lifecycle as follows:
+
+
+#. We construct `mlp = MLP(hidden_size=5, out_size=3)`, such that `mlp.hidden_size=5` and `mlp_out_size=3`.
+
+#. Then, call ``mlp.apply`` which performs the following:
+
+   #. Makes a clone of ``mlp``, let's call it `mlp_copy`.
+
+   #. Calls `mlp_copy.setup()`.
+
+   #. Returns the output of ``mlp_copy.__call__()`` and optionally the variable collections that were specified as mutable using the keyword argument ``mutable=``.
+
+
+You might be surprised that the lifecycle includes cloning the Module instance. You might be surprised that the lifecycle includes cloning the Module instance. In short, this is done to ensure that `apply` can be treated as a pure function (i.e., if you pass the same arguments in, it will return the same outputs). This will be explained in more detail later in the  :ref:`Top-level Modules` section.
+
+Variables
+==========
+
+The word “variable” is ubiquitous in programming and math. However, it's important to have a good understanding of what variables are in the context of JAX and Flax. Inside Flax Modules, `variables <https://flax.readthedocs.io/en/latest/flax.linen.html#module-flax.core.variables>`_ act like you expect from Python. They are initialized once, read, and perhaps even updated every so often. However, JAX has no concept of variables. Instead, values are stored in arrays similar to NumPy arrays  -- with one important difference: they are immutable.
+
+The ``init`` and ``apply`` methods return the variables as a nested dictionary with string keys and JAX arrays at the leaves. They are immutable and therefore really just a snapshot of state the variables are in. When ``apply`` is called again, the variable dict is passed as an argument. Such that the variables are in the same state as when the previous ``init`` / ``apply`` call finished.
+
+
+.. note::
+   Module fields are declared using the `field_name: TypeHint` syntax (same as dataclasses). Without a type hint, an attribute is considered a static property of the class. In case you cannot specify the type you can use ``typing.Any`` as a wildcard type.
+
+
+Compact Modules
+******************
+
+Linen provides an alternative API for defining modules more compactly. This is especially useful for the common case where the Module consists of only one method that uses parameters and/or sub-modules. Using the compact API the MLP can be rewritten as follows:
+
+
+.. testcode::
+
+  class CompactMLP(nn.Module):
+    hidden_size: int
+    out_size: int
+
+    @nn.compact
+    def __call__(self, x):
+      a = nn.Dense(self.hidden_size)(x)
+      h = nn.relu(a)
+      return nn.Dense(self.out_size)(h)
+
+
+A compact ``Module`` is similar in spirit to a function. It offers a concise notation and restricts external interaction to the inputs and return values of the function. In this case the concise notation might make it easier for others to understand what the Module is doing. There is no need to jump back and forth between the ``setup`` and ``__call__`` method to understand what the submodules are doing. Instead, simply reading the ``__call__`` method from top to bottom once should provide a clear picture of what the Module is doing. This can make a significant difference if you are implementing complex Modules with many hyperparameters. See `setup or compact <https://flax.readthedocs.io/en/latest/design_notes/setup_or_nncompact.html>`_ for a practical guide on decding between setup and compact.
+
+Another benefit of defining submodules and/or variables inline is that you can add arguments to your method when constructing variables. The most common example of this is using shape information to determine the shape of a parameter like this:
+
+
+.. testcode::
+
+  class CompactScaledMLP(nn.Module):
+    hidden_size: int
+    out_size: int
+
+    @nn.compact
+    def __call__(self, x):
+      scale = self.param("scale", nn.initializers.ones, x.shape[-1:])
+      x *= scale[None]
+      a = nn.Dense(self.hidden_size)(x)
+      h = nn.relu(a)
+      return nn.Dense(self.out_size)(h)
+
+
+.. testcode::
+  :hide:
+
+  mdl = CompactScaledMLP(hidden_size=4, out_size=5)
+  x = jax.numpy.ones((3, 2))
+  vars = mdl.init(random.PRNGKey(0), x)
+  assert vars["params"]["scale"].shape == (2,)
+
+Many of the standard Linen Modules like ``nn.Dense`` use shape inference already to avoid the need to specify input shapes (like the number of input features to a Dense layer).
+
+Compact control flow
+=====================
+
+The order in which you define submodules determines the name of a submodule if none is provided explicitly (using the ``name=`` keyword argument passed to the Module's constructor). Because the ``name`` determines how parameters are mapped to submodules, you must be careful about mixing control flow with auto-generated names. Using control flow can change the order or remove certain submodules altogether. This is useful in case a submodule should only exist depending on some construction argument. However, when control flow depends on the input arguments to the Module, you should be careful. For example, the following Module will break:
+
+
+.. testcode::
+
+  class WrongModule(nn.Module):
+    @nn.compact
+    def __call__(self, x, mode):
+      if mode == "encode":
+        return nn.Dense(features=8)(x)
+      elif mode == "decode":
+        return nn.Dense(features=4)(x)
+
+
+The above Module will break because either the encoder or decoder path will construct a Module named "Dense_0". This means the two Modules will share parameters which is not intented here. Actually, the two Modules cannot share parameters because they each have a different number of features.
+
+This problem can be solved in various ways:
+ - Provide explicit names
+ - create the modules in ``setup``
+ - or move the constructor out of the control flow.
+
+The latter is done as follows:
+
+.. testcode::
+
+  class CorrectModule(nn.Module):
+    @nn.compact
+    def __call__(self, x, mode):
+      encoder = nn.Dense(8)
+      decoder = nn.Dense(4)
+      if mode == "encode":
+        return encoder(x)
+      elif mode == "decode":
+        return decoder(x)
+
+.. testcode::
+  :hide:
+
+  def init_fn(mdl):
+    x = jax.numpy.ones((3, 2))
+    z = mdl(x, "encode")
+    return mdl(z, "decode")
+
+  mdl = CorrectModule()
+  vars = nn.init(init_fn, mdl)(random.PRNGKey(0))
+  assert vars["params"]["Dense_0"]["kernel"].shape == (2, 8)
+  assert vars["params"]["Dense_1"]["kernel"].shape == (8, 4)
+
+
+In the above example the construction order is fixed. After construction the submodules can be used in an arbitrary order.
+
+.. note::
+   compact modules show a strong resemblance to `React hooks <https://reactjs.org/docs/hooks-custom.html>`_.
+
+
+Top-level Modules
+*****************
+
+When a Module instance is created at the "top-level", it will be in an "unbound" state - that is, it has no variables attached. "Top-level" means it is not constructed as a sub-Module inside another Module class. Apart from calling ``init`` and ``apply``, there is not much you can do with an unbound Module. Note also that ``setup`` is not called on unbound Modules, so you can really only access the construction arguments. Refer to the :ref:`Future work` section to learn how this might change in the future.
+
+Why are top-level modules always unbound?
+===============================================
+
+When we call ``apply``, a copy of the top-level Module is created which will actually hold the variables and PRNG sequences. This stateful, "bound", clone only exists while we are executing the apply method. The reason for this is that if you create a stateful object and destroy it before the apply function returns, the ``apply`` function itself behaves like a pure function (pure means if you put the same arguments in, it will return the same outputs).
+
+Pure functions have many advantages but when using JAX they are often essential. For example, most code requires compilation using ``jax.jit`` to be fast and once you created a Module you probably want to optimize its parameters using ``jax.grad``. However, these APIs expect a pure function and don't work on stateful bound ``Module`` instances directly. Moreover, pure functions allow for flexible interoperability with other libraries. For example, We recommend `Optax <https://github.com/deepmind/optax>`_ for optimizing parameters. The optimizers in Optax expect and return a PyTree of JAX arrays to optimize, just like the ``apply`` function of a Linen Module.
+
+Cloning
+===============================================
+
+To make this approach work reliably we need well-defined cloning behavior. Rather than relying on a complex nested cloning procedure like Python's ``deepcopy``, Flax enforces that a ``Module`` is exactly defined by its construction arguments. Therefore cloning a Module reduces to calling the constructor with its original construction arguments. Because ``Module`` acts as an immutable dataclass, the construction arguments are mapped 1:1 to instance attributes.
+
+Bind
+===============================================
+
+Sometimes it's useful to have a bound, top-level Module without having to wrap the code in a function. For example: to interact with a Module inside a Jupyter notebook. The `bind <https://flax.readthedocs.io/en/latest/flax.linen.html?highlight=bind#flax.linen.Module.bind>`_ method returns a bound clone with an unlimited lifetime. The downside of this is that you cannot combine it with JAX transformations or integrate it into a vanilla JAX codebase that expects stateless code.
+
+
+Setup
+**********
+
+The ``setup`` method is often used like the constructor hook (`__init__`) in normal Python classes. However, for more advanced use cases it's good to realize that it is not quite the same as a constructor.
+
+``setup`` is only called after a Module becomes bound. Normally, this is not an issue because most Modules are bound (almost) immediately (as part of ``init`` and ``apply``). Inside ``setup``, sub-modules become bound when they are assigned to an attribute. Inside an ``nn.compact`` decorated method, sub-modules are bound immediately when constructed. As explained in the previous section, top-level Modules are never bound and thus setup is not called when they are constructed. This means you cannot access attributes assigned in setup from an unbound, top-level module.
+
+.. testcode::
+
+  class TopLevelAccess(nn.Module):
+
+    def setup(self):
+      self.foo = nn.Dense(2)
+  
+  mdl = TopLevelAccess()
+  assert not hasattr(mdl, "foo")  # foo is not defined because setup is not called
+
+The ``setup`` method is not called immediately after the ``Module`` becomes bound but only when you interact with the ``Module`` instance (e.g.: call a method or access an attribute). This should not impact the behavior of a ``Module`` but the lazy execution does sometimes affect log statements and stack traces during debugging. The section on functionalization will explain why we need ``setup`` to be lazy in the first place.
+
+
+Functionalization
+******************
+
+So far we had a pure ``apply`` function that is typically transformed with some JAX transformations and inside ``apply`` we have a stateful Module instance to work with. In other words: Outside of a Module we are in a functional world where we have the power of JAX's functional transformations and inside the Module we get the power of Flax's stateful variables and PRNG sequence, and the ``apply`` method is our bridge between these two worlds.
+
+But what if we want to use JAX transformations **inside** Modules? The answer to this is functionalization.
+
+This procedure itself is tedious and error-prone but handled internally by Flax. At a high-level we can summarize it as follows. For a method ``fn`` defined within a Module:
+
+#. Collect the state (variables & PRNG sequences) of the Module(s) that should be available inside the JAX transformation and take a snapshot of it.
+
+#. Call the JAX transformation with the original arguments and the collected state. Then inside the transformation:
+
+   #. Unpack the state and recreate the Modules
+
+   #. Call the user code ``fn``
+
+   #. Collect the updated variables and rng and return it together with the original return values from ``fn``
+
+#. Update the original state with the updated state returned from the transformation.
+
+A more in depth explanation of functionalization and lifting can be found in the `Lifted Transformation <https://flax.readthedocs.io/en/latest/design_notes/lift.html>`_ design note.
+
+Practical consequences
+==========================
+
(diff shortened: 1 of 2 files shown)


docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+
+## Setup
+
+The `setup` method is often used like the constructor hook (`__init__`) in normal Python classes. However, for more advanced use cases it’s good to realize that it is not quite the same as a constructor.
+
+`setup` is only called after a Module becomes bound. Normally, this is not an issue because most Modules are bound (almost) immediately (as part of `init` and `apply`). Inside `setup`, sub-modules become bound when they are assigned to an attribute. nside an `nn.compact` decorated method, sub-modules are bound immediately when constructed. As explained in the previous section, top-level Modules are never bound and thus setup is not called when they are constructed. This means that you cannot access attributes assigned in setup from an unbound, top-level module.
```
8bitmp3, line 197: Nit: "nside" -> "Inside"

```suggestion
`setup` is only called after a Module becomes bound. Normally, this is not an issue because most Modules are bound (almost) immediately (as part of `init` and `apply`). Inside `setup`, sub-modules become bound when they are assigned to an attribute. Inside an `nn.compact` decorated method, sub-modules are bound immediately when constructed. As explained in the previous section, top-level Modules are never bound and thus setup is not called when they are constructed. This means that you cannot access attributes assigned in setup from an unbound, top-level module.
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+Here the closure is implemented using a Flax dataclass. The function itself is annotated with `flax.struct.field(pytree_node=False)` to indicate that it does not contain Jax Arrays or linen Modules. The partially applied `args` on the other hand is treated as a pytree container. We rewrite the closure to use Partial. Now the inner method can be transformed using lifted transformations.
+
+
+## Future work
+
+
+### Setup for unbound Modules
+
+The current Module abstraction is particularly restrictive when it comes to initializing fields after construction. In the current Module API, the setup method is the place to initialize the fields of  the Module instance. Because `setup` is only called on a bound Module, the full Module API is available inside setup, including variable declaration. However, oftentimes we don’t actually require any stateful API’s to initialize a field. In fact, most commonly we simply want to declare a submodule. More importantly, it’s often useful to inspect submodules for debugging or to partially run the model. Consider for example:
```
8bitmp3, line 274: Nit: "inside `setup`" <- "inside setup"

```suggestion
The current Module abstraction is particularly restrictive when it comes to initializing fields after construction. In the current Module API, the setup method is the place to initialize the fields of  the Module instance. Because `setup` is only called on a bound Module, the full Module API is available inside `setup`, including variable declaration. However, oftentimes we don’t actually require any stateful API’s to initialize a field. In fact, most commonly we simply want to declare a submodule. More importantly, it’s often useful to inspect submodules for debugging or to partially run the model. Consider for example:
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
+# The Module Lifecycle
+
+This design note is intended for people who are somewhat familiar with Modules but want to understand more about the design principles behind the abstraction. After reading this, you should have a solid understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, Please take a look at the [MNIST tutorial](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html).
```
8bitmp3, line 3: Nit: suggestion (more accessibility-friendly language) + capitalization

```
If you have no practical experience with Modules yet, check out the [MNIST tutorial](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html).
```

```suggestion
This design note is intended for people who are somewhat familiar with Modules but want to understand more about the design principles behind the abstraction. After reading this, you should have a solid understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, check out the [MNIST tutorial](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html).
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
+# The Module Lifecycle
```
8bitmp3, line 1: Suggestion: Use sentence case

```suggestion
# The Module lifecycle
```

docs/design_notes/module_lifecycle.md
8bitmp3, line 3: More suggestions:

```suggestion
This design note is intended for users who are already familiar with Flax Modules and want to understand the design principles behind the abstraction. This note should give you a strong understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules, check out the [MNIST tutorial](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html) to get started.
```

Note: https://developers.google.com/style/translation?hl=en#be-inclusive (writing for a global audience)

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+This design note is intended for people who are somewhat familiar with Modules but want to understand more about the design principles behind the abstraction. After reading this, you should have a solid understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, Please take a look at the [MNIST tutorial](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html).
+
+Modules are the core of Flax. This is a practical guide to the design and behavior of the `Module` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
```
8bitmp3, line 5: Suggestion:


```suggestion
Modules are at the core of Flax. This note is a practical guide about the design and behavior of the `Module` class. After studying this note, you should feel more comfortable with Modules and use them in custom ways suited for your tasks.
```

Note: https://developers.google.com/style/translation?hl=en#be-inclusive (writing for a global audience)

8bitmp3:
The review is still work in progress @jheek. I've left a few suggestions. LMKWYT!

@marcvanzee @avital @saeta @andsteing @gnecula @levskaya

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+```python
+# equivalent to: variables = mlp.init(random.PRNGKey(0), x)
+_, variables = mlp.apply({}, x, rngs={"params": random.PRNGKey(0)}, mutable=True)
+```
+
+
+Thus, `init` is nothing more than a special case of `apply` where:
+
+1. We call a module without any initial variables (an empty dict)
```
marcvanzee, line 66:
```suggestion
1. We call a module without any initial variables (an empty dict).
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+2. A PRNG generator named “params” is always passed for randomly initializing parameters.
```
marcvanzee, line 67:
```suggestion
2. A PRNG generator named `params` is always passed for randomly initializing parameters (using the parameter initialization functions).
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+3. All variable collections are set to mutable (`mutable=True`). When a collection is mutable, existing variables can be updated and new variables can be created.
```
marcvanzee, line 68:
```suggestion
3. All variable collections are set to mutable (`mutable=True`). When a collection is mutable, existing variables can be updated and new variables can be created. In this case, this is necessary because `init` will initialize all parameters which are added to the returned variable dictionary.
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Lifecycle
+
+Thus, what remains is to actually understand what `.apply(...)` does. All the interesting stuff in a Module happens inside the `apply` method. We will refer to the various stages before and during `apply` as the “Module lifecycle”. We can summarize this lifecycle in a number of steps:
```
marcvanzee, line 72: It is a bit odd to start a new section with "Thus". Maybe like this?

```suggestion
Now that we understand `init` is just a special case of `apply`, let's analyze what `.apply(...)` does in more detail. In fact, most of the complexity of Modules resides inside the `apply` method. We will refer to the various stages before and during `apply` as the "Module lifecycle". We can summarize this lifecycle in a number of steps:
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+
+1. We construct `mlp = MLP(5, 3)`, such that `mlp.hidden_size=5` and `mlp_out_size=3`
```
marcvanzee, line 75: This is not really part of the `apply` function right? It seels like 2.1, 2.2, 2.3 below are actually the `apply` lifecycle.

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+2. Then, `mlp.apply`:
+    1. Makes a clone of `mlp`, let’s call it `mlp_copy`
```
marcvanzee, line 77:
```suggestion
    1. Makes a clone of `mlp`, let’s call it `mlp_copy`.
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+    2. Calls `mlp_copy.setup()`.
+    3. Returns the output of `mlp_copy.__call__()` and optionally the variable collections that were specified as mutable using `mutable=` kwarg.
```
marcvanzee, line 79:
```suggestion
    3. Returns the output of `mlp_copy.__call__()` and optionally the variable collections that were specified as mutable using the keyword argument `mutable=`.
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+
+You might be surprised that the lifecycle includes cloning the Module instance. This will be explained later in the **Top-level Modules** section.
```
marcvanzee, line 82: I think you should provide at least a hint here why we do the clone to help the reader somewhat.

```suggestion
You might be surprised that the lifecycle includes cloning the Module instance. In short, this is done to ensure that `apply` can be treated as a pure function (i.e., if you pass the same arguments in, it will return the same outputs). This will be explained in more detail later in the **Top-level Modules** section.
```

marcvanzee, line 82: I think you can add links to subsections as well, so maybe do that for "Top-level Modules"

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Variables
```
marcvanzee, line 84: Consider linking to the documentation of Linen's Variables as well: https://flax.readthedocs.io/en/latest/flax.linen.html#module-flax.core.variables

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+The word “variable” is ubiquitous in programming and math. However, it’s important to have a good understanding of what that means in the context of JAX and Flax. Inside Flax modules, variables act like you expect from Python. They are initialized once, read, and perhaps even updated every so often. However, JAX has no concept of variables. Instead, values are stored in arrays similar to NumPy arrays  -- with one important difference: they are immutable.
+
+`init` and `apply` return the variables as a dictionary of ordinary JAX arrays. They are immutable and therefore really just a snapshot of the variables. Next time we call apply we can feed back the snapshot so the variables are again in the same state as where we finished on the last call to apply.
```
marcvanzee, line 88: > `init` and `apply` return the variables as a dictionary of ordinary JAX arrays.

Dictionaries with string keys and JAX arrays in the leafs, right?

marcvanzee, line 88: I don't really get the "Next time" sentence. what does feeding back the snapshot mean?

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+
+> **NOTE** \
+Module fields are declared using the `field_name: TypeHint` syntax (same as dataclasses). Without a type hint, an attribute is considered a static property of the class. In case you cannot specify the type you can use `typing.Any` as a wildcard type.
+
+
+## Compact Modules
```
marcvanzee, line 95: Consider linking to this design note somewhere: https://flax.readthedocs.io/en/latest/design_notes/setup_or_nncompact.html

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+  def __call__(self, x):
+    scale = self.param("scale", x.shape[-1:], nn.initializers.ones)
+    x *= scale
+    a = nn.Dense(self.hidden_size)(x)
+    h = nn.relu(a)
+    return nn.Dense(self.out_size)(h)
+```
+
+
+Many of the standard linen layers like `nn.Dense` use this type of shape inference already to avoid the need to specify input shapes (like the number of input features to a Dense layer).
```
marcvanzee, line 134: Maybe also explain this isn't always possible and why (otherwise: why would we have a `setup` method at all?) I think our `Embed` layer uses setup so you can consider given that as an example.

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Compact control flow
+
+The order in which you define submodules determines the name of a submodule if none is provided explicitly (using the name= construction argument). Because the name determines how parameters are mapped to submodules, you must be careful about mixing control flow with auto-generated names. Using control flow can change the order or remove certain submodules altogether. This is useful in the case where certain submodules should only exist depending on some construction argument. However, when control flow depends on the input arguments to the Module you should be careful. For example the following Module will break:
```
marcvanzee, line 138:
```suggestion
The order in which you define submodules determines the name of a submodule if none is provided explicitly (using the `name=` keyword argument pass to the module's constructor). Because the name determines how parameters are mapped to submodules, you must be careful about mixing control flow with auto-generated names. Using control flow can change the order or remove certain submodules altogether. This is useful in the case where certain submodules should only exist depending on some construction argument. However, when control flow depends on the input arguments to the Module you should be careful. For example the following Module will break:
```

marcvanzee, line 138: > For example the following Module will break

Why does it break? What happens when you use it?

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+    if mode == "encode"
+      return encoder(x)
+    elif mode == "decode":
+      return decoder(x)
+```
+
+
+In the above example the construction order is fixed. After construction the submodules can be used in an arbitrary order.
+
+> Note \
```
marcvanzee, line 170:
```suggestion
> **NOTE** \
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+compact modules show a strong resemblance to [React hooks](https://reactjs.org/docs/hooks-custom.html)
```
marcvanzee, line 171:
```suggestion
compact modules show a strong resemblance to [React hooks](https://reactjs.org/docs/hooks-custom.html).
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+
+## Top-level Modules
+
+When a Module instance is created at the top-level it will be in an "unbound" state, meaning it has no variables attached. By top-level we mean it is not constructed as a sub-Module inside another Module class. Apart from calling `init` and `apply` there is not much you can do with an unbound Module. Note also that `setup` is not called on unbound Modules so you can really only access the construction arguments. See the Future work section for how this might change in the future.
```
marcvanzee, line 176:
```suggestion
When a Module instance is created at the top level it will be in an "unbound" state, meaning it has no variables attached. By top-level we mean it is not constructed as a sub-Module inside another Module class. Apart from calling `init` and `apply` there is not much you can do with an unbound Module. Note also that `setup` is not called on unbound Modules so you can really only access the construction arguments. See the Future work section for how this might change in the future.
```

marcvanzee, line 176: > Note also that `setup` is not called on unbound Modules

I don't get this, when it is not called? Do you mean it is not called when you construct it?

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Why are top-level modules always unbound?
+
+When we call `apply`, a copy of the top-level Module is created which will actually hold the variables and PRNG sequences. This stateful, "bound", clone only exists while we are executing the apply method. The reason for this is that if you create a stateful object and destroy it before the apply function returns, the `apply` function itself behaves like a pure function (pure means if you put the same arguments in, it will return the same outputs).
+
+Pure functions have many advantages but when using JAX they are often essential. For example, most code requires compilation using `jax.jit` to be fast and once you created a Module you probably want to optimize its parameters using `jax.grad`. However, these APIs expect a pure function and don't work on stateful bound `Module` instances directly. Moreover, pure functions allow for flexible interoperability with other libraries.
```
marcvanzee, line 182: > Moreover, pure functions allow for flexible interoperability with other libraries.

Can you give an example? Maybe mention Optax?

docs/design_notes/module_lifecycle.md
marcvanzee, line 197:
```suggestion
`setup` is only called after a Module becomes bound. Normally, this is not an issue because most Modules are bound (almost) immediately (as part of `init` and `apply`). Inside `setup`, sub-modules become bound when they are assigned to an attribute. Inside an `nn.compact` decorated method, sub-modules are bound immediately when constructed. As explained in the previous section, top-level Modules are never bound and thus setup is not called when they are constructed. This means that you cannot access attributes assigned in setup from an unbound, top-level module.
```

marcvanzee, line 197: I feel this is really hard to grasp without an example, would it be possible to add a very simple one?

marcvanzee (changes requested):
Finished for ~80%, still need to review the last two sections.

docs/design_notes/module_lifecycle.md
jheek: I think it applies for example to the line below: "off the beaten track". The link is interesting but also makes things more complicated :) One thing I observed is that many communities have moved to a much more engaged/personal/funny way of writing docs rather than the dry reference material that was common before. Yet what is funny for one person is confusing to another. And what is boring for some people is clear and to the point for others.

8bitmp3: @marcvanzee I think it's things like "colloquialisms, idioms, or slang" or things that may come across as such to users who aren't as proficient in English/"computer science" English (https://developers.google.com/style/translation)

@jheek Yup, it's not easy to balance

docs/design_notes/module_lifecycle.md
8bitmp3: The only reason IMO we should use sentence case, is because when you have Flax names that use common English names—such as `Module`—titles may not make it obvious to users.

So, here:

```suggestion
# The Module lifecycle
```

it's more apparent that `Module` is a name, and not just a word, such as "lifecycle". If @jheek introduces a class called `Lifecycle`, this title may get confusing (e.g. when users start searching for `Lifecycle` on Flax docs/search engines)

Reference: https://developers.google.com/style/capitalization?hl=en#capitalization-in-titles-and-headings

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+This design note is intended for people who are somewhat familiar with Modules but want to understand more about the design principles behind the abstraction. After reading this, you should have a solid understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, Please take a look at the [MNIST tutorial](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html).
+
+Modules are the core of Flax. This is a practical guide to the design and behavior of the `Module` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
+
+
+## Overview
+
+### Definition
+
+Let’s start with a high-level overview of the Module lifecycle. We start by defining a simple Module:
```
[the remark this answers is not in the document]
8bitmp3: +1 @marcvanzee

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+    h = nn.relu(a)
+    return self.out(h)
+```
+
+
+This Module consists of:
+
+
+1. **_Attribute annotations_**, defined as [dataclass](https://docs.python.org/3/library/dataclasses.html) fields. These annotations automatically define a constructor.
+2. **The `setup` method,** that creates submodules and assigns them to attributes.
```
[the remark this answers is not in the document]
8bitmp3: +1 @marcvanzee

docs/design_notes/module_lifecycle.md
8bitmp3: Would this be even more applicable here @marcvanzee @jheek 

```suggestion
1. Call a Module without any initial variables (an empty dict).
```

(capitalization)

docs/design_notes/module_lifecycle.md
8bitmp3: +1

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+2. A PRNG generator named “params” is always passed for randomly initializing parameters.
+3. All variable collections are set to mutable (`mutable=True`). When a collection is mutable, existing variables can be updated and new variables can be created.
+
+### Lifecycle
+
+Thus, what remains is to actually understand what `.apply(...)` does. All the interesting stuff in a Module happens inside the `apply` method. We will refer to the various stages before and during `apply` as the “Module lifecycle”. We can summarize this lifecycle in a number of steps:
+
+
+1. We construct `mlp = MLP(5, 3)`, such that `mlp.hidden_size=5` and `mlp_out_size=3`
+2. Then, `mlp.apply`:
```
8bitmp3, line 76: WDYT @jheek @marcvanzee @saeta @andsteing @gnecula 

```suggestion
2. Then, call `mlp.apply` which performs the following:
```

docs/design_notes/module_lifecycle.md
8bitmp3, line 88: Nit:

```suggestion
Methods `init` and `apply` return the variables as a dictionary of ordinary JAX arrays. They are immutable and, therefore, really are just a snapshot of the variables. Next time we call `apply`, we can feed the snapshot back, so the variables are again in the same state as where we finished on the last `apply` call.
```

or something along these lines. LMKWYT @jheek @avital @marcvanzee

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+  @nn.compact
+  def __call__(self, x):
+    a = nn.Dense(self.hidden_size)(x)
+    h = nn.relu(a)
+    return nn.Dense(self.out_size)(h)
+
+```
+
+
+A compact `Module` is similar in spirit to a function. It offers a concise notation and restricts external interaction to the inputs and return values of the function. In this case the concise notation might make it easier for others to understand what the Module is doing. There is no need to jump back and forth between the setup and call function to understand what the submodules are doing. Instead, simply reading the call method from top to bottom once should give a clear picture of what the Module is doing. This can make a significant difference if you are implementing complex Modules with many hyper parameters.
```
8bitmp3, line 114: Nit: LMKWYT

```suggestion
A compact `Module` is similar in spirit to a function. It offers a concise notation and restricts external interaction to the inputs and return values of the function. In this case the concise notation might make it easier for others to understand what the Module is doing. There is no need to jump back and forth between the `setup` and `call` methods to understand what the submodules are doing. Instead, simply reading the `call` method from top to bottom once should provide a clear picture of what the Module is doing. This can make a significant difference if you are implementing complex Modules with many hyperparameters.
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+Another benefit of defining submodules and/or variables inline is that you can use arguments to your method when constructing variables. The most common example of this is using shape information to determine the shape of a parameter like this:
```
8bitmp3, line 116: Suggestion: do you mean "use arguments **in** your method" or "**add** arguments to your method..."?

e.g.

```suggestion
Another benefit of defining submodules and/or variables inline is that you can add arguments to your method when constructing variables. The most common example of this is using shape information to determine the shape of a parameter like this:
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+      x = fn(x)
+    return x
+```
+
+
+Here `inner` takes a function that closes over a Module instance. In this example, that works fine because we are not transforming the inner method with a lifted transformation. Most methods are not transformed but it is good to know how to make Module methods transformable.
+
+The main obstacle for transformability are types that JAX does not recognize. By default JAX will recognize containers (lists, dictionaries, and tuples) of JAX/NumPy arrays and python scalars (bool, int, float, etc.). The containers are called [pytrees](https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html) and the pytree API is extensible with custom types. With the flax.struct package you can also define pytree compatible dataclasses.
+
+Function closure is the most common way to accidentally hide a JAX array or linen Module from a transformation. There is however an easy workaround if you want to pass closures that are also compatible with JAX and Linen transformations:
```
8bitmp3, line 243:
```suggestion
Function closure is the most common way to accidentally hide a JAX array or Flax Linen Module from a transformation. There is however an easy workaround if you want to pass closures that are also compatible with JAX and Linen transformations:
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+  def __call__(self, x):
+    dense = nn.Dense(x.shape[-1])
+    return self.inner(x, Partial(lambda mdl, x: mdl(x) + 1, dense)
+  
+  def inner(self, x, fn):
+    ...
+```
+
+
+Here the closure is implemented using a Flax dataclass. The function itself is annotated with `flax.struct.field(pytree_node=False)` to indicate that it does not contain Jax Arrays or linen Modules. The partially applied `args` on the other hand is treated as a pytree container. We rewrite the closure to use Partial. Now the inner method can be transformed using lifted transformations.
```
8bitmp3, line 266:
```suggestion
Here the closure is implemented using a Flax dataclass. The function itself is annotated with `flax.struct.field(pytree_node=False)` to indicate that it does not contain JAX Arrays or Flax Linen Modules. The partially applied `args` on the other hand is treated as a pytree container. We rewrite the closure to use Partial. Now the inner method can be transformed using lifted transformations.
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+`init` and `apply` return the variables as a dictionary of ordinary JAX arrays. They are immutable and therefore really just a snapshot of the variables. Next time we call apply we can feed back the snapshot so the variables are again in the same state as where we finished on the last call to apply.
+
+
+> **NOTE** \
+Module fields are declared using the `field_name: TypeHint` syntax (same as dataclasses). Without a type hint, an attribute is considered a static property of the class. In case you cannot specify the type you can use `typing.Any` as a wildcard type.
+
+
+## Compact Modules
+
+Linen provides an alternative API for defining modules more compactly. This is especially useful for the common case where the Module consists of only one method that uses parameters and/or sub-modules. Using the compact API the MLP can be rewritten as follows:
```
8bitmp3, line 97:
```suggestion
Flax Linen provides an alternative API for defining modules more compactly. This is especially useful for the common case where the Module consists of only one method that uses parameters and/or sub-modules. Using the `compact` API the MLP can be rewritten as follows:
```

(for more clarity e.g. for new users)

docs/design_notes/module_lifecycle.md
8bitmp3, line 114:
```suggestion
A `compact` `Module` is similar in spirit to a function. It offers a concise notation and restricts external interaction to the inputs and return values of the function. In this case the concise notation might make it easier for others to understand what the Module is doing. There is no need to jump back and forth between the setup and call function to understand what the submodules are doing. Instead, simply reading the call method from top to bottom once should give a clear picture of what the Module is doing. This can make a significant difference if you are implementing complex Modules with many hyper parameters.
```

Note that "compact", like "Module", are common English words, so using backticks here and there may help users understand the context better.

docs/design_notes/module_lifecycle.md
8bitmp3: @marcvanzee your version + some formatting

```suggestion
The order in which you define submodules determines the name of a submodule if `none` is provided explicitly (using the `name=` keyword argument pass to the module's constructor). Because the `name` determines how parameters are mapped to submodules, you must be careful about mixing control flow with auto-generated names. Using control flow can change the order or remove certain submodules altogether. This is useful in cases where certain submodules should only exist depending on some construction argument. However, when control flow depends on the input arguments to the Module, you should be careful. For example, the following Module will break:
```

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+  @nn.compact
+  def __call__(self, x, mode):
+    if mode == "encode"
+      return nn.Dense(8)(x)
+    elif mode == "decode":
+      return nn.Dense(4)(x)
+```
+
+
+This problem can be solved in various ways: Provide explicit names, create the modules in setup, or move the constructor out of the control flow. The latter is done as follows:
```
8bitmp3, line 152: Suggestion: use a list + formatting

```suggestion
This problem can be solved in various ways:
- Provide explicit names
- Ceate the modules in `setup`, or
- Move the constructor out of the control flow.

The latter is done as follows:
```

docs/design_notes/module_lifecycle.md
8bitmp3: A modification of @marcvanzee 's suggestion with some formatting:

```suggestion
When a Module instance is created at the "top-level", it will be in an "unbound" state—that is, it has no variables attached. "Top-level" means it is not constructed as a sub-Module inside another Module class. Apart from calling `init` and `apply`, there is not much you can do with an unbound Module. Note also that `setup` is not called on unbound Modules, so you can really only access the construction arguments. Refer to the **Future work** section to learn how this might change in the future.
```

NOTE: "sub-Module" or "submodule" - should pick one for consistency across all docs IMO @jheek @marcvanzee @avital

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+This procedure itself is tedious and error-prone but handled internally by Flax. At a high-level we can summarize it as follows. For a method `fn` defined within a Module:
+
+1. Collect the state (variables & PRNG sequences) of the Module(s) that should be available inside the JAX transformation and take a snapshot of it.
+2. Call the JAX transformation with the original arguments and the collected state. Then inside the transformation:
+    1. Unpack the state and recreate the Modules
+    2. Call the user code `fn`
+    3. Collect the updated variables and rng and return it together with the original return values from `fn`
+3. Update the original state with the updated state returned from the transformation.
+
+A more in depth explanation of functionalization and lifting can be found [here](https://flax.readthedocs.io/en/latest/design_notes/lift.html).
```
8bitmp3, line 217:
```suggestion
A more in depth explanation of functionalization and lifting can be found in the [Lifted Transformations](https://flax.readthedocs.io/en/latest/design_notes/lift.html) design note.
```

(This may be "better" than "here")

docs/design_notes/module_lifecycle.md
jheek: I agree it creates some ambiguity between class names and normal "words"

docs/design_notes/module_lifecycle.md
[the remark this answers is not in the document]
jheek: Why is that relevant to the lifecycle? Note that the doc is expliclty a follow up after learning the APIs so users should already now compact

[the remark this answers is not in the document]
marcvanzee: I was just thinking to say something like: "(note we will discuss [compact modules](link-to-compact-modules) below)", just to guide the reader, but if was just a minor thing so if you think it isn't necessary feel free to leave it out.

docs/design_notes/module_lifecycle.md
jheek: Note that users should already now the linen API (including compact) when reading this doc

docs/design_notes/module_lifecycle.md
jheek: agreed

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+
+## Overview
+
+### Definition
+
+Let’s start with a high-level overview of the Module lifecycle. We start by defining a simple Module:
+
+
+```python
```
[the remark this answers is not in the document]
jheek: rewrote to rst syntax

docs/design_notes/module_lifecycle.md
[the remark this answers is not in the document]
jheek: The compact chapter is a bit of a de-tour already. I'm trying to minimize it's impact on the overall narrative. It seems to cause plenty of distraction already :)

docs/design_notes/module_lifecycle.md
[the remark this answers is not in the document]
jheek: the clause isn't non-essential because here you are giving a definition.

docs/design_notes/module_lifecycle.md
jheek: looks better

docs/design_notes/module_lifecycle.md
jheek: fair point

docs/design_notes/module_lifecycle.md
jheek: clarified

docs/design_notes/module_lifecycle.md
jheek: I think this is best covered in the setup vs compact note?

docs/design_notes/module_lifecycle.md
jheek: added clarification

docs/design_notes/module_lifecycle.md
jheek: done

docs/design_notes/module_lifecycle.md
jheek: added

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+mlp = MLP(5, 3)
+variables = mlp.init(random.PRNGKey(0), x)
+y = mlp.apply(variables, x)
+```
+
+
+First, we construct an instance of `MLP` and pass the construction attributes. Note that construction here is different from what you might expect if you are not used to Functional Programming patterns. The `MLP` constructor does not actually create variables or any internal state whatsoever. It’s best to think of it as a specification or template of the Module that contains functionality but no data.
```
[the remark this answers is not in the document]
jheek: > In most imperative language is it expected that the constructor does not "do" much

I'm not saying that imperative languages do a lot of work in the constructor but that the constructor creates internal state. I think how much work is done is not what really matters for example pytorch inits the params in the constructor but they could also create an "empty" param and have a seperate init routine so the constructor does less work. But at the end of the day you are creating a stateful object by calling the Module constructor whether it is empty or not.

docs/design_notes/module_lifecycle.md
andsteing: What about linking `Module` to `linen.Module` ?
(First, it's nice to have a code link ready when reading the tutorial, second I like having pages that anybody can land on and understand what is meant)

docs/design_notes/module_lifecycle.md
andsteing: Don't we have `flax.core` that has precisely nothing to do with Modules? Or is this an intended pun?

More generally, I think might be interesting to say in the beginning of this doc that this is about `Linen Modules` which are one (opinionated) way of how the underlying functional core can be used in Flax to write compact modules with some nice features.

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+    a = self.hidden(x)
+    h = nn.relu(a)
+    return self.out(h)
+```
+
+
+This Module consists of:
+
+
+1. **_Attribute annotations_**, defined as [dataclass](https://docs.python.org/3/library/dataclasses.html) fields. These annotations automatically define a constructor.
```
andsteing: It's mysterious to me why this one is in italics but not the others...

It looks a bit uneven in the numbered list (but maybe there's a good reason making up for that)

Similarly, consider using the same pattern (i.e. with comma, or with period) for all three points.

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+2. **The `setup` method,** that creates submodules and assigns them to attributes.
+3. A **user-defined method**. By convention, most Modules have just one  `__call__` method, but you can define multiple methods or use different method names.
```
andsteing: Still nit-ting here, but why don't we use the same `3. User methods` as in the code above to keep things in parallel?

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Construction/initialization
+
+Now we want to construct and use the `MLP` Module:
+
+
+```python
+mlp = MLP(5, 3)
```
andsteing: In terms of style, should we encourage passing the arguments as keywords here?

docs/design_notes/module_lifecycle.md
andsteing: `mlp.out_size=3`

docs/design_notes/module_lifecycle.md
andsteing: Make it a link instead of bold face?
(here and also below where you use document-internal links)

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+    1. Makes a clone of `mlp`, let’s call it `mlp_copy`
+    2. Calls `mlp_copy.setup()`.
+    3. Returns the output of `mlp_copy.__call__()` and optionally the variable collections that were specified as mutable using `mutable=` kwarg.
+
+
+You might be surprised that the lifecycle includes cloning the Module instance. This will be explained later in the **Top-level Modules** section.
+
+### Variables
+
+The word “variable” is ubiquitous in programming and math. However, it’s important to have a good understanding of what that means in the context of JAX and Flax. Inside Flax modules, variables act like you expect from Python. They are initialized once, read, and perhaps even updated every so often. However, JAX has no concept of variables. Instead, values are stored in arrays similar to NumPy arrays  -- with one important difference: they are immutable.
```
andsteing: nit: maybe use "–" instead of "--"

docs/design_notes/module_lifecycle.md
andsteing: "nested dictionaries" might also clarify this; maybe even mention what the structure corresponds to?

docs/design_notes/module_lifecycle.md
andsteing: (and linking from there to this new document)

docs/design_notes/module_lifecycle.md
andsteing: nit: this paragraph uses "doing" three times, which feels like a lot

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+
+In the above example the construction order is fixed. After construction the submodules can be used in an arbitrary order.
+
+> Note \
+compact modules show a strong resemblance to [React hooks](https://reactjs.org/docs/hooks-custom.html)
+
+
+## Top-level Modules
```
andsteing: Should the main title here be "top-level modules"? Or rather something along the lines "Bound/unbound modules" ?

I think about "Flax code" (i.e. code inside modules) to always run in a "bound" state. It cannot be called "from normal Python code" because the top-level modules are in an "unbound" state. So one needs to call the bridging init/apply functions that will call the code inside the module in a "bound" state.

It's an interesting fact that "top-level modules" can never be in a bound state, but rather are cloned on-the-fly and the original copy (instance) remains unbound all the time ... But it feels not like the most important part of this story

docs/design_notes/module_lifecycle.md
andsteing: "is not called" -> "has not been called" ?

(because when one calls `top_level_module.init()` then setup will be called)

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+When a Module instance is created at the top-level it will be in an "unbound" state, meaning it has no variables attached. By top-level we mean it is not constructed as a sub-Module inside another Module class. Apart from calling `init` and `apply` there is not much you can do with an unbound Module. Note also that `setup` is not called on unbound Modules so you can really only access the construction arguments. See the Future work section for how this might change in the future.
+
+### Why are top-level modules always unbound?
+
+When we call `apply`, a copy of the top-level Module is created which will actually hold the variables and PRNG sequences. This stateful, "bound", clone only exists while we are executing the apply method. The reason for this is that if you create a stateful object and destroy it before the apply function returns, the `apply` function itself behaves like a pure function (pure means if you put the same arguments in, it will return the same outputs).
```
andsteing: This paragraph is not really clear to me. It seems to me that a function is pure when its outputs only depend on its inputs, regardless of whether you create stateful objects (and - or not - destroy them). So what's the motivation for the cloning exactly?
(An example might help)

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+Pure functions have many advantages but when using JAX they are often essential. For example, most code requires compilation using `jax.jit` to be fast and once you created a Module you probably want to optimize its parameters using `jax.grad`. However, these APIs expect a pure function and don't work on stateful bound `Module` instances directly. Moreover, pure functions allow for flexible interoperability with other libraries.
+
+### Cloning
+
+To make this approach work reliably we need well-defined cloning behavior. Rather than relying on a complex nested cloning procedure like Python's `deepcopy`, Flax enforces that a `Module` is exactly defined by its construction arguments. Therefore cloning a Module reduces to calling the constructor with its original construction arguments. Because `Module` acts as an immutable dataclass, the construction arguments are mapped 1:1 to instance attributes.
```
andsteing: what about post init & computed fields ?

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+### Bind
+
+Sometimes it's useful to have a bound, top-level Module without having to wrap the code in a function. For example: to interact with a Module inside a Jupyter notebook. The [bind](https://flax.readthedocs.io/en/latest/flax.linen.html?highlight=bind#flax.linen.Module.bind) method returns a bound clone with an unlimited lifetime. The downside of this is that you cannot combine it with JAX transformations or integrate it into a vanilla JAX codebase that expects stateless code.
```
andsteing: An example highlighting what can go wrong would help here.

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+`setup` is only called after a Module becomes bound. Normally, this is not an issue because most Modules are bound (almost) immediately (as part of `init` and `apply`). Inside `setup`, sub-modules become bound when they are assigned to an attribute. nside an `nn.compact` decorated method, sub-modules are bound immediately when constructed. As explained in the previous section, top-level Modules are never bound and thus setup is not called when they are constructed. This means that you cannot access attributes assigned in setup from an unbound, top-level module.
+
+The `setup` method is not called immediately after the `Module` becomes bound but only when you interact with the `Module` instance (e.g.: call a method or access an attribute). This should not impact the behavior of a `Module` but the lazy execution does sometimes affect log statements and stack traces during debugging. The section on functionalization will explain why we need `setup` to be lazy in the first place.
+
+
+# Functionalization
```
andsteing: This whole section: should it really be part of this document, or is it maybe a better fit for the lifted transformations design note?

It seems not so much related to "Linen module lifecycle" to me, but rather about "pure functions that can directly be transformed vs. functions inside bound modules that must take extra care to transform state that is implicitly handled by Flax"

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+```python
+class Foo(nn.Module):
+
+  @nn.compact
+  def __call__(self, x):
+    dense = nn.Dense(x.shape[-1])
+    return self.inner(x, lambda x: dense(x) + 1)
+  
+  def inner(self, x, fn):
```
andsteing: This code snippet (and the one below) are actually perfectly fine, as long as the method is not transformed.

Consider making this point a bit clearer by adding the transformation (maybe commented out), and making that point more explicitly in a comment or the main text.

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+  def __call__(self, *args, **kwargs):
+    return self.fn(*(tuple(self.args) + args), **kwargs)
+
+class Foo(nn.Module):
+
+  @nn.compact
+  def __call__(self, x):
+    dense = nn.Dense(x.shape[-1])
+    return self.inner(x, Partial(lambda mdl, x: mdl(x) + 1, dense)
```
andsteing: should it be `[dense]` ?

also, it's missing a closing parens `)`

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+  def inner(self, x, fn):
+    for i in range(3):
+      x = fn(x)
+    return x
+```
+
+
+Here `inner` takes a function that closes over a Module instance. In this example, that works fine because we are not transforming the inner method with a lifted transformation. Most methods are not transformed but it is good to know how to make Module methods transformable.
+
+The main obstacle for transformability are types that JAX does not recognize. By default JAX will recognize containers (lists, dictionaries, and tuples) of JAX/NumPy arrays and python scalars (bool, int, float, etc.). The containers are called [pytrees](https://jax.readthedocs.io/en/latest/jax-101/05.1-pytrees.html) and the pytree API is extensible with custom types. With the flax.struct package you can also define pytree compatible dataclasses.
```
andsteing: simpler might be something like "JAX only understands [Pytree](...) arguments. That's arbitrarily nested Python containers (dict, list, tuple) of (Jax) numpy ndarrays and Python numbers/bools. Flax allows to define dataclasses that ..."

nit: maybe link and format `flax.struct`

docs/design_notes/module_lifecycle.md
andsteing: +1

also consider adding a link back to this document from that other guide

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+
+Function closure is the most common way to accidentally hide a JAX array or linen Module from a transformation. There is however an easy workaround if you want to pass closures that are also compatible with JAX and Linen transformations:
+
+
+```python
+class Partial(flax.struct.PytreeNode):
```
andsteing: `PyTreeNode`

docs/design_notes/module_lifecycle.md
andsteing: This code snippet (and the one below) are actually perfectly fine, as long as the method is not transformed.

Consider making this point a bit clearer by adding the transformation (maybe commented out), and making that point more explicitly in a comment or the main text.

docs/design_notes/module_lifecycle.md
andsteing: +1

also consider adding a link back to this document from that other guide

docs/design_notes/module_lifecycle.md
andsteing: This whole section: should it really be part of this document, or is it maybe a better fit for the lifted transformations design note?

It seems not so much related to "Linen module lifecycle" to me, but rather about "pure functions that can directly be transformed vs. functions inside bound modules that must take extra care to transform state that is implicitly handled by Flax"

docs/design_notes/module_lifecycle.md
```diff
@@ -0,0 +1,289 @@
...
+```
+
+
+Here the closure is implemented using a Flax dataclass. The function itself is annotated with `flax.struct.field(pytree_node=False)` to indicate that it does not contain Jax Arrays or linen Modules. The partially applied `args` on the other hand is treated as a pytree container. We rewrite the closure to use Partial. Now the inner method can be transformed using lifted transformations.
+
+
+## Future work
+
+
+### Setup for unbound Modules
```
andsteing: This section feels like it would be better a separate FLIP.

Not sure why this would be relevant to users reading about Linen module lifecycle. And publishing it as part of a design note will just archive the discussion, instead of discussing the pros and cons of its implementation.

andsteing (changes requested):

docs/design_notes/module_lifecycle.md
jheek: I see your point... My intention here is to give an idea about what is still being actively worked on / up for discussion. Of course, this shouldn't go in depth about the actual implementation, backward comp, etc. And if we eventually do make this change it should be part of a FLIP.

docs/design_notes/module_lifecycle.md
jheek: Done

docs/design_notes/module_lifecycle.md
jheek: I do think this is part of the lifecycle because functionalization affects multiple design decisions (lazy setup, cloning, ...) and effectively inside a lifted transformation there is a mini/recurisve lifecycle.

docs/design_notes/module_lifecycle.md
jheek: Yeah this is strange, It is fixed now

docs/design_notes/module_lifecycle.md
jheek: done

docs/design_notes/module_lifecycle.md
jheek: done

docs/design_notes/module_lifecycle.md
jheek: rephrased

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+  from typing import Any, Callable, Iterable
+  import flax
+  from flax import linen as nn
+  from jax import random
+  import jax
+
+
+This design note is intended for users who are already familiar with linen Modules but want to understand more about the design principles behind the abstraction. This note should give you a good understanding of the assumptions and guarantees the Module API is built upon. If you have no practical experience with Modules yet, check out the `MNIST Tutorial <https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html>`_.
+
+linen Modules offer a pythonic abstracton on top of Flax core. The ``Module`` abstraction allows you to create classes that have state, parameters and randomness on top of JAX. This is a practical guide to the design and behavior of the ``Module`` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
```
8bitmp3, line 15: Nit: "Pythonic" (capitalized)

```suggestion
linen Modules offer a Pythonic abstracton on top of Flax core. The ``Module`` abstraction allows you to create classes that have state, parameters and randomness on top of JAX. This is a practical guide to the design and behavior of the ``Module`` class. By the end, you should feel comfortable to go off the beaten track and use Modules in new ways.
```

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+
+
+Overview
+***********
+
+Definition
+=============
+
+Let's start with a high-level overview of the Module lifecycle. We start by defining a simple Module:
```
8bitmp3, line 24: Nit: to avoid repeating "start", maybe try

```suggestion
Let's start with a high-level overview of the Module lifecycle. First, define a simple Module:
```

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+
+  mlp = MLP(hidden_size=5, out_size=3)
+  x = jax.numpy.ones((1, 2))
+  variables = mlp.init(random.PRNGKey(0), x)
+  y = mlp.apply(variables, x)
+
+
+First, we construct an instance of ``MLP`` and pass the construction attributes. Note that construction here is different from what you might expect if you are not used to Functional Programming patterns. The ``MLP`` constructor does not actually create variables or any internal state whatsoever. It's best to think of it as a specification or template of the Module that contains functionality but no data.
+
+Let's first take a closer look at initialization. Surprisingly, there actually is no special initialization path in Flax. Calling ``init`` is just a special case of ``apply`` and we could also code it as:
```
8bitmp3, line 68: Nit: to avoid repeating "first" in 2 consecutive paragraphs, maybe try this 

```suggestion
Let's take a closer look at initialization. Surprisingly, there actually is no special initialization path in Flax. Calling ``init`` is just a special case of ``apply``, which you can also write as:
```

And maybe try rephrasing the first or second "special" to avoid repeating "special" twice.

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+.. testcode:
+
+  # equivalent to: variables = mlp.init(random.PRNGKey(0), x)
+  _, variables = mlp.apply({}, x, rngs={"params": random.PRNGKey(0)}, mutable=True)
+
+
+Thus, ``init`` is nothing more than a special case of ``apply`` where:
+
+#. We call a Module without any initial variables (an empty dict).
+#. A PRNG generator named “params” is always passed for randomly initializing parameters (using the parameter initialization function).
```
8bitmp3, line 80: Nit: maybe use backticks for code - e.g. ``params`` - instead of "params"

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+#. All variable collections are set to mutable (``mutable=True``). When a collection is mutable, existing variables can be updated and new variables can be created. Thus, inside ``init`` variables can be initialized in any variable collection and they are all added to the returned variable dictionary.
+
+Lifecycle
+=============
+
+
+Now that we understand ``init`` is just a special case of ``apply``, let's analyze what ``.apply(...)`` does in more detail. In fact, most of the complexity of Modules resides in the ``apply`` method. The "Module lifecycle" consists of constructing and ``apply``-ing a Module. We can summarize the Module lifecycle as follows:
```
8bitmp3, line 87: Nit: 

- "Special" [``init``] is repeated 3 or 4 times here (in the `apply` context), so maybe rephrase it a little.
- "We understand" is a presumption - maybe use "you learned about"

e.g.

```suggestion
Now that you have learned about ``init`` (a special case of ``apply``), let's analyze what ``.apply(...)`` does in more detail. In fact, most of the complexity of Modules resides in the ``apply`` method. The "Module lifecycle" consists of constructing and ``apply``ing a Module. We can summarize the Module lifecycle as follows:
```

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+
+
+#. We construct `mlp = MLP(hidden_size=5, out_size=3)`, such that `mlp.hidden_size=5` and `mlp_out_size=3`.
+
+#. Then, call ``mlp.apply`` which performs the following:
```
8bitmp3, line 92: Nit: this can be shortened a little 

e.g.

```suggestion
#. Then, call ``mlp.apply``, which:
```

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+
+   #. Makes a clone of ``mlp``, let's call it `mlp_copy`.
+
+   #. Calls `mlp_copy.setup()`.
+
+   #. Returns the output of ``mlp_copy.__call__()`` and optionally the variable collections that were specified as mutable using the keyword argument ``mutable=``.
+
+
+You might be surprised that the lifecycle includes cloning the Module instance. You might be surprised that the lifecycle includes cloning the Module instance. In short, this is done to ensure that `apply` can be treated as a pure function (i.e., if you pass the same arguments in, it will return the same outputs). This will be explained in more detail later in the  :ref:`Top-level Modules` section.
```
8bitmp3, line 101: Nit: repeating sentences.

```suggestion
Notice that the lifecycle includes cloning the Module instance. This is done to ensure that `apply` can be treated as a pure function (i.e., if you pass the same arguments in, it will return the same outputs). You will learn about this in more detail later in the  :ref:`Top-level Modules` section.
```

(+ some minor refactoring)

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,378 @@
...
+.. note::
+   compact modules show a strong resemblance to `React hooks <https://reactjs.org/docs/hooks-custom.html>`_.
+
+
+Top-level Modules
+*****************
+
+When a Module instance is created at the "top-level", it will be in an "unbound" state - that is, it has no variables attached. "Top-level" means it is not constructed as a sub-Module inside another Module class. Apart from calling ``init`` and ``apply``, there is not much you can do with an unbound Module. Note also that ``setup`` is not called on unbound Modules, so you can really only access the construction arguments. Refer to the :ref:`Future work` section to learn how this might change in the future.
+
+Why are top-level modules always unbound?
```
8bitmp3, line 227: Nit: maybe

```suggestion
Why are top-level Modules always unbound?
```

docs/design_notes/module_lifecycle.md
jheek: done

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,377 @@
...
+  variables = mlp.init(random.PRNGKey(0), x)
+  y = mlp.apply(variables, x)
+
+
+First, we construct an instance of ``MLP`` and pass the construction attributes. Note that construction here is different from what you might expect if you are not used to Functional Programming patterns. The ``MLP`` constructor does not actually create variables or any internal state whatsoever. It's best to think of it as a specification or template of the Module that contains functionality but no data.
+
+Let's take a closer look at initialization. Surprisingly, there actually is no seperate initialization path in Flax. Calling ``init`` is just a special case of ``apply``, which you can also write as:
+
+
+.. testcode:
```
levskaya, line 71: I'm not seeing this rendered at:
 https://flax--1964.org.readthedocs.build/en/1964/design_notes/module_lifecycle.html
is there a subtle formatting error here that's accidentally hiding it?

levskaya (approved):
It reads great, thanks so much for all the work on this doc, and sorry that I've been away while it's been written.

docs/design_notes/module_lifecycle.rst
jheek: yup missing double colon. Ugh I'm not a fan of RestructuredText

docs/design_notes/module_lifecycle.md
jheek: done

docs/design_notes/module_lifecycle.md
jheek: but it will be called on the clone...

docs/design_notes/module_lifecycle.md
jheek: > I think about "Flax code" (i.e. code inside modules) to always run in a "bound" state.

This inside / outside Module code is exactly the intuition I'm trying to cultivate here. bound/unbound is the jargon top-level vs inside Module is the practical distinction. I would prefer to keep it intuitive.

docs/design_notes/module_lifecycle.md
jheek: adding some clarification

docs/design_notes/module_lifecycle.md
jheek: added

docs/design_notes/module_lifecycle.md
jheek: I refer to the lifted note for details. The main point here is that lifting repeats the lifecycle and justifies some constraints further (like cloning)

docs/design_notes/module_lifecycle.md
jheek: fixed

docs/design_notes/module_lifecycle.md
jheek: fixed

docs/design_notes/module_lifecycle.md
andsteing: ok then I think I need to update my thinking here :)

docs/design_notes/module_lifecycle.md
andsteing: ... but the function would still be pure if it returned the stateful object, no? if true, then the requirement of wanting to have a pure function would not necessitate you to destroy the bound module inside the function ...

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,381 @@
...
+
+Bind
+===============================================
+
+Sometimes it's useful to have a bound, top-level Module without having to wrap the code in a function. For example: to interact with a Module inside a Jupyter notebook. The `bind <https://flax.readthedocs.io/en/latest/flax.linen.html?highlight=bind#flax.linen.Module.bind>`_ method returns a bound clone with an unlimited lifetime. The downside of this is that you cannot combine it with JAX transformations or integrate it into a vanilla JAX codebase that expects stateless code. For example, `Optax <https://github.com/deepmind/optax>`_ can optimze a Pytree of parameters but it cannot direclty optimize a bound ``Module`` instance created with ``.bind`` (because that's not a Pytree). Thus, you cannot combine the ``bind`` API with a functional optimizer API like Optax.
```
andsteing, line 245: direclty -> directly

docs/design_notes/module_lifecycle.md
jheek: Here you get towards the problem of referential transparency. JAX expects a PyTree (freely cloneable because it doesn't have identity) object as output to a function. Cloning a stateful object is itself a side effect because the clone has it's own state. This is also why using dicts and lists with as Pytrees can lead to weird behavior (because they are actually stateful). For example if you have a function like this:

```python
def add_metrics(metrics, loss):
  metrics["loss"] = loss
  return metrics
```
The behavior will change if you add @jit to this function because the metrics dict will no longer be shared.
I try to avoid using referential transparency though as people find it very confusing. So I think it's a good to encourage pure function with immutable inputs/outputs. And anything else is "outside warranty" :)

marcvanzee (approved):

docs/design_notes/module_lifecycle.rst
```diff
@@ -0,0 +1,381 @@
...
+
+
+Setup
+**********
+
+The ``setup`` method is often used like the constructor hook (`__init__`) in normal Python classes. However, for more advanced use cases it's good to realize that it is not quite the same as a constructor.
+
+``setup`` is only called after a Module becomes bound. Normally, this is not an issue because most Modules are bound (almost) immediately (as part of ``init`` and ``apply``). Inside ``setup``, sub-modules become bound when they are assigned to an attribute. Inside an ``nn.compact`` decorated method, sub-modules are bound immediately when constructed. As explained in the previous section, top-level Modules are never bound and thus setup is not called when they are constructed. This means you cannot access attributes assigned in setup from an unbound, top-level module.
```
zhangqiaorjc, line 253: When does the Module become bound? I don't see it getting mentioned in LifeCycle section or anywhere before this Setup section.

jheek: In this paragraph it says modules become bound when the are constructed in compact or assigned to an attribute in self. In the section on top-level modules it's explained that the top-level Module itself is always unbound and init/apply operate on a clone that is bound

PR #1970, opened 2022-03-08
Fix example in nn.while_loop docstring

levskaya:
The old while loop docstring example was broken in several ways, added a tested one.


jheek (approved):
Thanks!

PR #1990, opened 2022-03-13
 PEP484-compatible export of names in flax & its public modules

### LINKED ISSUE #1542: PEP484-compatible export of names in Flax
Now that [Flax provides a marker file according PEP 561](https://github.com/google/flax/pull/1483), mypy is able to find type annotations. The issue is that nothing is exported at the moment – mypy requires public API of a package being exported using either `__all__` or `import ... as ...` syntax.

```bash
python3 -m venv env
source env/bin/activate
pip install -U pip mypy

git clone git@github.com:google/flax.git
pip install ./flax
python -m mypy --install-types --non-interactive --strict -c "from flax import linen as nn; nn.Dense(1)"

<string>:1: error: Module has no attribute "Dense"
Found 1 error in 1 file (checked 1 source file)
```

Currently, projects that are using Flax need to set`--no-implicit-reexport=False`.

There are some references on that matter.

- [PEP-0484](https://www.python.org/dev/peps/pep-0484/#stub-files)
  > - Modules and variables imported into the stub are not considered exported from the stub unless the import uses the import ... as ... form or the equivalent from ... import ... as ... form. 
  > - However, as an exception to the previous bullet, all objects imported into a stub using from ... import * are considered exported. 
- [MyPy docs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport)
  > `--no-implicit-reexport` always treated as enabled for stub files.

[JAX recently started to export type annotations](https://github.com/google/jax/pull/7606), it would be great to have similar changes to Flax if possible.

lkhphuc:
# What does this PR do?

Fixes #1542, explicit export of modules following this [PR in Jax](https://github.com/google/jax/pull/7606/files).
All __init__.py file are using the `from a import b as b`, with the exception of `optimizer/__init.py`, which is not touched in this PR, since it already exports through `__all__ = [...]` list. I can change that optimizer's export file to the same format if desired.

There might be some Google's internal tooling to handle such code style change so feel free to close if you can't accept this PR. Just too annoyed by the Language server complaining and have a bit of free time. 

## Checklist
      checks if that's the case).
      link).


levskaya (approved):
Sorry for the long delay, this all looks great - thanks so much for the help improving this usability issue!

PR #1994, opened 2022-03-15
Update GroupNorm error msg to account for default value of num_groups

sudhakarsingh27:
Since the default value of num_groups attribute is set to 32, passing
group_size results in ValueError. To be able to only pass the value of
group_size, additionally, num_groups=None should also be passed. The
ValueError message is now updated to reflect this fact.

Fixes: #1704


marcvanzee (approved):
Thanks, looks great!

PR #2007, opened 2022-03-22
Enable support for non-string keys in state dicts.

### LINKED ISSUE #1925: Can not assign dict whos key is not string as module attribute
Hi,
It seems current flax.linen not allow assign a dict with non-string keys to the module attribute.
See below simple example:
it will trigger error:
`AssertionError: A state dict must only have string keys.`

Questions:
1. Is it an intended behavior? Why?
2. If it is intended, is there any work around? As it is quite possible we need assign the information contained by a dict to the module, and the key of the dict may not string.

```python

import flax.linen as nn
import jax
import jax.numpy as jnp


class Foo(nn.Module):
    def setup(self):
        self.a = {(1, 2): 3} # here the dict using tuple as key

    @nn.compact
    def __call__(self, x):
        return x


foo = Foo()
rng = jax.random.PRNGKey(0)
x = jnp.ones(shape=(3, 3))
vars = foo.init({"params": rng}, x)
out = foo.apply(vars, x)
print(out)
```

jheek:
This implies Modules can have a dictionary assignment with non-stirng keys.
Fixes #1925

### DIFF
--- a/flax/serialization.py
+++ b/flax/serialization.py
@@ -117,11 +117,14 @@ def _restore_list(xs, state_dict: Dict[str, Any]) -> List[Any]:
 
 
 def _dict_state_dict(xs: Dict[str, Any]) -> Dict[str, Any]:
-  return {key: to_state_dict(value) for key, value in xs.items()}
+  str_keys = set(str(k) for k in xs.keys())
+  if len(str_keys) != len(xs):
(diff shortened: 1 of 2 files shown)


levskaya (approved):
Thanks!

PR #2009, opened 2022-03-22
Canonicalize conv padding

### LINKED ISSUE #2000: flax.linen.Conv needs better error checking of 'padding' argument.
Hi!

The following code leads to mysterious error message `RuntimeError: UNKNOWN: -:4:130: error: expected '['` :

```
x = np.random.normal(size=(7, 48, 48, 96)).astype(np.float32)
model_def = nn.Conv(
    features=96, kernel_size=(7, 7),
    strides=(4, 4),
    padding=(3, 3))
model_state, conv_params = model_def.init({'params': jax.random.PRNGKey(42)}, x).pop('params')

out = model_def.apply({"params": conv_params}, x)
```

The mistake here is that I was using `padding=(3, 3)` instead of `padding=((3, 3), (3, 3))`, but the error message is not informative. It would be great if that could be improved. Ideally, a simpler padding spec. like `padding=(3, 3)` or even `padding=3` could directly be supported.

jheek:
Fixes #2000

### DIFF
--- a/tests/linen/linen_linear_test.py
+++ b/tests/linen/linen_linear_test.py
@@ -15,6 +15,7 @@
 """Tests for flax.deprecated.nn.linear."""
 
 import functools
+from multiprocessing.sharedctypes import Value
 
 from absl.testing import absltest
 from absl.testing import parameterized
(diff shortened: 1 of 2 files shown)


tests/linen/linen_linear_test.py
```diff
@@ -842,6 +843,18 @@ def __call__(self, x):
...
+      if expected is None:
+        with self.assertRaises(ValueError):
+          nn.linear.canonicalize_padding(pad, rank)
+      else:
+        self.assertEqual(nn.linear.canonicalize_padding(pad, rank), expected)
+    test_pad("SAME", 2, "SAME")
+    test_pad(2, 3, [(2, 2), (2, 2), (2, 2)])
+    test_pad((2, 2), 3)
+    test_pad((2, 2), 1)
+    test_pad([1, (2, 3)], 2, [(1, 1), (2, 3)])
```
andsteing, line 857: maybe test this codepath as well:

```python
test_pad([None, (2, 3)], 2)
```

andsteing (approved):
SG, added single comment about additional test

PR #2012, opened 2022-03-23
Bumps flax/jax versions in examples/imagenet.

andsteing:
Tested that training starts with TPU pod commands from

https://github.com/google/flax/blob/main/examples/imagenet/README.md

### DIFF
--- a/examples/imagenet/requirements.txt
+++ b/examples/imagenet/requirements.txt
@@ -1,9 +1,9 @@
 absl-py==1.0.0
 clu==0.0.6
-flax==0.3.6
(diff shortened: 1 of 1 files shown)


andsteing:
This is a follow-up to #1856 that added a dependency on new Flax features in the imagenet example without updating the `requirements.txt`, which started our regression tests to fail.

Note that before this PR we had to create a new Flax release in #2010

FYI @jheek @romanngg

romanngg:
Thanks Andreas, missed that req file!

jheek (approved):

PR #2025, opened 2022-04-04
Fix bug where the linen Module state is reused.

jheek:
For example in scan the transform_fn is called twice.
This led to mis counting compact call default names (see test case)


levskaya (approved):
Ha - thanks so much for catching that.  Hopefully no one depends on wonky default names in checkpoints - I don't think too many people use scanned layers yet...  (I wonder if it's worth doing a global test internally to be safe?)

jheek:
> Hopefully no one depends on wonky default names in checkpoints

You can't depend on it because the behavior changes between init and apply (only init was wrong and apply would fail with param not found)

PR #2039, opened 2022-04-08
Removes deprecated filterwarning from pytest.ini

marcvanzee:
Since we removed the deprecated API this filter warning can be removed as well.


jheek (approved):
thanks!

PR #2055, opened 2022-04-18
Handle edge-case of rate==1.0 in Dropout layer.

levskaya:
A number of users would benefit from treating the edge-case of dropout rate==1.0 acting like complete zero-ing of the input array rather than returning NaNs from the rescaling operation.  We handle this edge-case statically here based on the rate parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ vNext
 -
 -
 - Added lifted conditional `nn.cond`
--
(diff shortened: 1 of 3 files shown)


levskaya:
build failing due to upstream chex bug, should already be fixed at chex HEAD and will be fixed by a new chex pypi release.

marcvanzee (approved):

Issue #2051, opened 2022-04-15
Annotated MNIST: The implementation of `cross_entropy_loss` is incorrect


ayaka14732:
Currently the `cross_entropy_loss` in the [Annotated MNIST](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html) example is implemented as follows:

```python
def cross_entropy_loss(*, logits, labels):
  one_hot_labels = jax.nn.one_hot(labels, num_classes=10)
  return -jnp.mean(jnp.sum(one_hot_labels * logits, axis=-1))
```

The correct implementation should be:

```python
from operator import getitem

@jax.jit
def cross_entropy_loss(logits, labels):
    logits = nn.log_softmax(logits)
    loss = jax.vmap(getitem)(logits, labels)
    loss = -loss.mean()
    return loss
```

Test:

```python
import jax
from jax import nn
import jax.numpy as np
jnp = np
import jax.random as rand
import numpy as onp
from operator import getitem
import torch
import torch.nn.functional as F

jax.config.update('jax_platform_name', 'cpu')

key = rand.PRNGKey(42)

x = rand.normal(key, (16, 10))
y = rand.randint(key, (16,), 0, 10)

x_ = torch.from_numpy(onp.asarray(x))
y_ = torch.from_numpy(onp.asarray(y)).long()

print(F.cross_entropy(x_, y_).numpy())  # PyTorch implementation: 2.888901

def cross_entropy_loss(*, logits, labels):
    one_hot_labels = jax.nn.one_hot(labels, num_classes=10)
    return -jnp.mean(jnp.sum(one_hot_labels * logits, axis=-1))

print(cross_entropy_loss(logits=x, labels=y))  # Current implementation: -0.0056607425

@jax.jit
def cross_entropy_loss(logits, labels):
    logits = nn.log_softmax(logits)
    loss = jax.vmap(getitem)(logits, labels)
    loss = -loss.mean()
    return loss

print(cross_entropy_loss(x, y))  # Correct implementation: 2.8889012
```

andsteing:
Thanks for opening this inquiry!

Note that the [Annotated MNIST](https://flax.readthedocs.io/en/latest/notebooks/annotated_mnist.html) example already returns `nn.log_softmax()` from the model:

```python
class CNN(nn.Module):
  """A simple CNN model."""

  @nn.compact
  def __call__(self, x):
    x = nn.Conv(features=32, kernel_size=(3, 3))(x)
    x = nn.relu(x)
[...]

PR #2071, opened 2022-04-26
Makes annotated_mnist use Optax's xent loss.

andsteing:
As mentioned in #2051, it is a bit confusing that we return `nn.log_softmax(x)` from the model and then define the loss as `-jnp.mean(jnp.sum(one_hot_labels * logits, axis=-1))`

This PR updates the [Annotated MNIST](https://colab.sandbox.google.com/github/google/flax/blob/main/docs/notebooks/annotated_mnist.ipynb) to use the same `optax.softmax_cross_entropy()` for loss computation that we already use in [`examples/minst/`](https://github.com/google/flax/tree/main/examples/mnist)

Note that #2051 also has some discussions about a `vmap()` based implementation and performance comparisons, but the current implementation seems to perform better on TPU and similarly on GPU.

### DIFF
--- a/docs/notebooks/annotated_mnist.ipynb
+++ b/docs/notebooks/annotated_mnist.ipynb
@@ -1,598 +1,655 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "id": "9q9JZjt16AMf"
-   },
-   "source": [
-    "# Annotated MNIST\n",
-    "\n",
-    "This tutorial demonstrates how to construct a simple convolutional neural\n",
-    "network (CNN) using the [Flax](https://flax.readthedocs.io) Linen API and train\n",
-    "the network for image classification on the MNIST dataset.\n",
-    "\n",
-    "Note: This notebook is based on Flax's official\n",
-    "[MNIST Example](https://github.com/google/flax/tree/main/examples/mnist).\n",
-    "If you see any changes between the two feel free to create a\n",
-    "[pull request](https://github.com/google/flax/compare)\n",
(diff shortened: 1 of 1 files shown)


docs/notebooks/annotated_mnist.ipynb
```diff
@@ -1,598 +1,655 @@
...
+        "[MNIST Example](https://github.com/google/flax/tree/main/examples/mnist).\n",
+        "If you see any changes between the two feel free to create a\n",
+        "[pull request](https://github.com/google/flax/compare)\n",
+        "to synchronize this Colab with the actual example."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "source": [
+        "# Check GPU\n",
```
marcvanzee, line 25: This cell seems to be new, why did you add it?

marcvanzee (approved):

docs/notebooks/annotated_mnist.ipynb
andsteing: To make clear that the example runs on a Colab with GPU.

But looking at the RTD this is actually quite visible now ...
https://flax--2071.org.readthedocs.build/en/2071/notebooks/annotated_mnist.html

What would you prefer? Removing it or changing to `!nvidia-smi -L` that will have a sinle line output?

marcvanzee: Makes sense, I actually think it is quite nice it shows the GPU info, so I'm fine keeping it as is.

marcvanzee (approved):

ayaka14732 (approved):

PR #2088, opened 2022-04-28
Linen activations import

onnoeberhard:
# What does this PR do?

Addresses #2087 

These changes make `flax.linen` more consistent with `jax.nn` in the sense that all functions one might import from `jax.nn` can now directly be imported from `flax.linen` as well. This PR includes 3 new imports from `jax.nn` into `flax.linen.activations` and additional imports from `flax.linen.activations` into `flax.linen.__init__`.

The only import left out is `jax.nn.standardize` which will replace `jax.nn.normalize` (deprecated) in the future, but had not been introduced in JAX version 0.3 (Flax's dependency).

Tagging @marcvanzee

## Checklist
      checks if that's the case).
- [x] This change is discussed in a Github discussion: #2087 


marcvanzee (approved):
Thanks a lot!

marcvanzee:
Note: I squashed your commits into 1 to keep our commit tree clean.

flax/linen/__init__.py
```diff
@@ -23,10 +23,19 @@
...
   glu as glu,
+  hard_sigmoid as hard_sigmoid,
+  hard_silu as hard_silu,
+  hard_swish as hard_swish,
+  hard_tanh as hard_tanh,
   leaky_relu as leaky_relu,
   log_sigmoid as log_sigmoid,
   log_softmax as log_softmax,
+  logsumexp as logsumexp,
+  normalize as normalize,
```
levskaya, line 34: could we add the `standardize` alias for `normalize` as well? (this name change follows stats literature naming)

onnoeberhard: I initially added it, but then removed it again when I saw that it had only been added to JAX very recently (not in version 0.3, which I understand to be Flax's dependency).

I guess we could do something like `try: import standardize; except: pass`, but this might be bad style. Or we could define the alias ourselves.

onnoeberhard: What do you think would be the best option here?

marcvanzee: It seems cleaner to create an issue to add this once JAX rolls out a new version.

marcvanzee:
Sorry, I forgot about this PR. I will get it in ASAP and create an issue for adding the additional activation functions once JAX rolls out a new version.

flax/linen/__init__.py
marcvanzee: Added #2119

PR #2090, opened 2022-04-28
Improves early stopping docstring

marcvanzee:
#1249 suggested to add an early stopping HOWTO, but the content seems small enough to put in the docstring, which is what I did in this PR.

@ayush-1506: FYI, thanks for adding the initial HOWTO!


marcvanzee:
Assigning to @andsteing since @jheek is out in vacation for a week.

flax/training/early_stopping.py
```diff
@@ -45,8 +59,9 @@ def update(self, metric):
     """Update the state based on metric.
     
     Returns:
-      Whether there was an improvement greater than min_delta from
-          the previous best_metric and the updated EarlyStop object.
+      A pair (has_improved, early_stop), where `has_improved` is True when there
```
andsteing, line 62: why don't we store the `has_improved` in the dataclass like the `should_stop` ?

would simplify the usage and make it harder to mix up the `has_improved` with something like `should_stop` or some other boolean

andsteing (approved):
change sgtm

maybe also add it to
https://github.com/google/flax/blob/main/docs/flax.training.rst

so it shows up in RST search?

flax/training/early_stopping.py
marcvanzee: TBH I was also confused about the API but didn't think about it so carefully. I propose we get this PR in and improve the API in a follow-up

marcvanzee:
@andsteing I added it to RTD and created a new issue for improving the API. Will merge the PR now!

PR #2106, opened 2022-05-05
Add typehint to Module.scope

cgarciae:
# What does this PR do?
Small but useful change that adds typehint information to `Module.scope`. Currently editors and static analyzers don't provide any useful information about `self.scope`. Users should be able to better autocompletion when referencing scope now.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -530,7 +530,7 @@ class Module:
     cls._wrap_module_methods()
     # Set empty class defaults.
     cls._state = _uninitialized_module_internal_state
-    cls.scope = None
+    cls.scope: Optional[Scope] = None
 
   @classmethod
   def _customized_dataclass_transform(cls):


flax/linen/module.py
```diff
@@ -530,7 +530,7 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
     cls._wrap_module_methods()
     # Set empty class defaults.
     cls._state = _uninitialized_module_internal_state
-    cls.scope = None
+    cls.scope: Optional[Scope] = None
```
jheek, line 533: This will not actually put a attribute type hint on the class right?

cgarciae: It will not, it is just a function-local typehint. Just to double check
```python
class Foo(nn.Module):
    pass
print(Foo.__annotations__)
```
yields
```
{'parent': typing.Union[typing.Type[flax.linen.module.Module], typing.Type[flax.core.scope.Scope], typing.Type[flax.linen.module._Sentinel], NoneType], 'name': <class 'str'>}
```
So I think we are good.

jheek: So Python supports a local annotation on something that isn't a local variable?
I guess I don't really understand what this annotation actually does

cgarciae: For Python itself it does nothings, but linters take the hint and infer that `Module.scope` has this type.
After this change VSCode (pylance/pyright) is able to give better auto-completion because it now understands the type of `scope` so it now suggests attributes and methods from `Scope` and its also aware its an `Optional` so warns you when you don't check if its `None` before using it.

jheek (approved):

PR #2111, opened 2022-05-10
Adds "full eval" HOWTO.

andsteing:
Adds a HOWTO explaining the problem and possible solutions for "full dataset processing", which is especially relevant for computing evaluation metrics.

The added function `flax.jax_utils.pad_shard_unpad()` is copied verbatim from [`big_vision`](https://github.com/google-research/big_vision) and was created by @lucasb-eyer  - thanks!

For reviewing this PR, see:

https://flax--2111.org.readthedocs.build/en/2111/howtos/full_eval.html

https://colab.research.google.com/github/andsteing/flax/blob/doc/docs/notebooks/full_eval.ipynb

Note that for running the Colab you'll need to replace the line

```
!pip install -q git+https://github.com/google/flax
```

with

```
!pip install -q git+https://github.com/andsteing/flax@doc
```


docs/howtos/full_eval.rst
```diff
@@ -0,0 +1,161 @@
...
+      x, '(d b) ... -> d b ...', d=jax.local_device_count())
+  unshard = lambda x: einops.rearrange(x, 'd b ... -> (d b) ...')
+
+  correct = total = 0
+  for batch in ds.as_numpy_iterator():
+    images = batch['image']
+    n = len(images)
+    padding = np.zeros([per_host_batch_size - n, *images.shape[1:]], images.dtype)
+    padded_images = np.concatenate([images, padding])
+    preds = unshard(get_preds(vs_p, shard(padded_images)))[:n]
```
cgarciae, line 102: Hey! What does `vs_p` mean / represent? Is it short for "variables_predict"? If it is, it took me a while to figure out, maybe a more descriptive name would help readers 😀

andsteing: I abbreviated `variables` with `vs`. Then often when I replicate things I use the `_p` suffix. But I'm not sure if that use is really widespread.

In either case, the `_p` is not needed here since it's an empty set, so I removed it, and replaced `vs` with the full `variables`.

docs/howtos/full_eval.rst
```diff
@@ -0,0 +1,161 @@
...
+Processing the entire Dataset
+=============================
+
+For efficiency reasons, we form batches that contain multiple examples and
+process them in parallel. Especially when evaluating a model, it is important
+that we process all examples and **avoid losing the remainder** of examples that
+does not form a complete batch at the end.
+
+
+The problem
```
marcvanzee, line 13: It seems the problem is two-fold: double compilation (both in training and eval), and incorrect metric results (in eval). Is this correct? Maybe it is worth emphasizing this. Currently you state that is especially important during eval but you don't explain why.

docs/howtos/full_eval.rst
```diff
@@ -0,0 +1,161 @@
...
+
+The solution: padding
+---------------------
+
+Even though it's possible to solve this problem by cleverly adjusting the number
+of batches executed by different devices on different hosts, such a solution
+quickly becomes complicated and makes the main eval loop hard to read with a lot
+of cumbersome logic.
+
+The more straight forward solution to this problem is to use padding at the end
```
marcvanzee, line 78:
```suggestion
The more straightforward solution to this problem is to use padding at the end
```

marcvanzee (approved):
Looks great!

docs/howtos/full_eval.rst
andsteing: Done.

docs/howtos/full_eval.rst
andsteing: I see it more like this
1. In eval, we care about processing the full dataset because otherwise the metrics are off. (in training we usually do multiple epochs and using some examples 1x less for training does not matter)
2. When we want to avoid loosing data at the end, we run into other problems (like e.g. multiple compliations)

I think 1. is mentioned in the first paragraph "Especially when evaluating a model, it is important that we process all examples", and 2. is mentioned further down as disadvantage of some solutions.

andsteing:
> One case that might be worth discussing is what to do if you update the metric state inside a compiled `train_step`/`test_step`. It would seem that you would have to use padding plus some form of masking?

@cgarciae, please check out the added section about using a `eval_step()` with the new argument `pad_shard_unpad(static_return)`

docs/howtos/full_eval.rst
```diff
@@ -0,0 +1,197 @@
...
+
+
+.. |tfds.split_for_jax_process()| replace:: ``tfds.split_for_jax_process()``
+.. _tfds.split_for_jax_process(): https://www.tensorflow.org/datasets/api_docs/python/tfds/split_for_jax_process
+
+
+The solution: padding
+---------------------
+
+Even though it's possible to solve this problem by cleverly adjusting the number
```
jheek, line 73: pmap is not guaranteed to work when hosts have different shapes potentially causing deadlocks so I don't think you can actually do this right now. Unless you remove the parallel primitives from the batch dependent program

docs/howtos/full_eval.rst
```diff
@@ -0,0 +1,197 @@
...
+
+
+.. |clu.metrics| replace:: ``clu.metrics``
+.. _clu.metrics: https://github.com/google/CommonLoopUtils/blob/main/clu/metrics.py
+
+
+Adding "infinite padding"
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Above solution works in most cases, but it has some limitations:
```
jheek, line 163: another solution is to let each host process indepedently and do the pmean(metrics) add the very end in a seperate pmapped program

docs/howtos/full_eval.rst
andsteing: yes, this would only work in some cases - e.g. when you evaluate a model that does not have any parallel primitives inside, and then have custom code that calls psum etc when inside pmap, but not when jit-ted on single host ... making it even messier

I don't think it adds anything here to go into too many details since it's not the proposed solution though

docs/howtos/full_eval.rst
andsteing: yes, this is actually very similar to what I do with `count_correct_p()` below

note that the above two sections `Adding "infinite padding"` and `Computing metrics in eval_step()` can easily be combined

so I think the different usecases are covered with the subsections, but if you feel there is a specific combination that should be added, feel free to add some more specific comments and I'll write it up.

docs/howtos/full_eval.rst
jheek: Yeah I guess the easiest thing to say is: Just keep your batch size static unless you know what you are doing. I can agree with that

docs/howtos/full_eval.rst
jheek: Ah indeed a parallel stoppinig criteria is not so different when you use padding. I was thinking more towards a note on what to do when you don't want to pad. You might not want to include that route for simplicity though

andsteing: but without padding different hosts would have different batch sizes? and this would also lead to re-compilations?

(in that case I think we should keep it simple with "solution=padding" since it seems superior, at the cost of a little more complicated code, but that added complication is quite small, especially in the case where one compute the metrics in the main eval loop)

docs/howtos/full_eval.rst
cgarciae: > In either case, the _p is not needed here since it's an empty set, so I removed it, and replaced `vs` with the full `variables`.

I still see `vs_p` here, maybe I am wrong but is the intention is to rename it to `variables`?

flax/jax_utils.py
```diff
@@ -243,3 +243,81 @@ def body_wrapper(c, xs):
...
+  """Wraps a function with code that pads, shards, then un-shards, un-pads.
+
+  Args:
+    wrapped: the function to be wrapped. Signature is `params, *args, *kwargs`.
+    static_argnums: indices of arguments to `wrapped` that should _not_ be
+      padded and sharded, but instead be forwarded as-is. The default is (0,)
+      because by far the most common use-case is to pass `params` first.
+    static_argnames: names of kwargs to `wrapped` that should _not_ be padded
+      and sharded, but instead be forwarded as-is.
+    static_return: whether not to un-shard, and un-pad the return value; static
```
cgarciae, line 260: Wondering if `static_return` should also accepted a tuple of ints so in case there are multiple outputs you can choose which ones to let pass untouched. Maybe this could be a future improvement.

cgarciae:
Left new comment about `static_return` which is not urgent.
A previous comment that was market as resolved says that maybe you were going to change the name `vs_p` to `variables` but I still see `vs_p`. Maybe I misunderstood, just wanted to clarify that in case its not intended.

flax/jax_utils.py
andsteing: I considered this, but couldn't come up with a usecase. With evaluation, usually we
- either want to get the values to do something with them (e.g. custom metric computation outside JAX)
- or return metrics from an `eval_step()`

Note that for training the padding is not needed.

So I thought YAGNI and went for the simpler solution.

cgarciae: Yeah, I also couldn't come up with a scenario where it was needed. Could be expanded in the future if someone finds the need.

cgarciae (commented):
LGTM @andsteing! Enjoyed the read, also made me realize all `clu` and `jax_metrics` metrics should support masking.

andsteing:
@cgarciae yes exactly, ideally all metrics support a single `mask` feature that is expected exactly for this reason; how the `mask` is incorporated for intermediate value updates then depends on the individual metrics ...

jheek (approved):

cgarciae (approved):

PR #2126, opened 2022-05-16
RNN Layer

### LINKED ISSUE #1677: RNN Layers
### Description of the model to be implemented
Is there a plan to add a simple RNN cell and a Layer wrapper that can take an RNN Cell, output an RNN Layer like in TensorFlow

### Reference implementations in other frameworks
https://www.tensorflow.org/api_docs/python/tf/keras/layers/RNN

I am interested in contributing this feature

cgarciae:
# What does this PR do?
Implements  a`RNNBase` module that serves as a base for common RNN layers. Also add a  `RNN` layer which can take in `RNNCellBase` instances and perform the full computation.

Closes #1677


jheek:
> (planned: no) unroll: "If True, the network will be unrolled, else a symbolic loop will be used. Unrolling can speed-up a RNN, although it tends to be more memory-intensive. Unrolling is only suitable for short sequences."

Could you just pass on the unroll= keyword arg to scan?

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+    cell: RNNCellBase,
+    x: jnp.ndarray,
+    *,
+    init_key: jnp.ndarray,
+    time_axis: int,
+    num_feature_dims: int,
+    stateful: bool,
+    return_sequences: bool,
+    return_state: bool,
+    go_backwards: bool,
```
jheek, line 519: I would just use the same term (reverse) here as lax.scan

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+      variable_axes=variable_axes,
+      variable_broadcast=variable_broadcast,
+      variable_carry=variable_carry,
+      split_rngs=split_rngs,
+    )
+
+    carry, x = scan(cell, initial_carry, x)
+
+    # if the module is initializing keep the initial carry
+    if is_initializing:
```
jheek, line 626: I wonder if a user should just pas a reset flag for the memory. Perhaps even one that can be a boolean arry with the same shape as the batch dims.

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+      raise ValueError(
+        f'\'num_feature_dims\' dimensions must be smaller than ' 
+        f'number of dimensions, got num_feature_dims \'{num_feature_dims}\' '
+        f'for input of shape \'{x.shape}\''
+      )
+
+    num_batch_dims = x.ndim - (num_feature_dims + 1) # features dims + time dim
+
+    # move time dimension to the first axis
+    x = x.swapaxes(0, time_axis)
```
jheek, line 587: nit: transforms.scan does already support passing time_axis as the in axis

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+    )
+
+    carry, x = scan(cell, initial_carry, x)
+
+    # if the module is initializing keep the initial carry
+    if is_initializing:
+      carry = initial_carry
+
+    if stateful:
+      self.put_variable('cache', 'carry', carry)
```
jheek, line 630: Is this really cache or should the name be something more discriptive like "memory"?

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+      slices[time_axis] = -1
+      # apply slices
+      x = x[tuple(slices)]
+
+    if return_state:
+      return x, carry
+    else:
+      return x
+
+class RNN(RNNBase):
```
jheek, line 648: What does the BaseRNN vs RNN class distinction do exactly?

flax/linen/recurrent.py
cgarciae: `memory` sounds good but I don't see a previous use of this collection, are we ok adding a new collection type?

flax/linen/recurrent.py
cgarciae: `RNNBase` is a base class exposes the `rnn_forward` method that contains the main logic that all rnn implementation can use to define their `__call__`. The `RNN` class is a very generic instance that accepts an arbitrary `RNNCellBase`.

flax/linen/recurrent.py
cgarciae: Agreed, will make the change.

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+        size = x.shape[-num_feature_dims:] # spatial + feature dim
+      
+      initial_carry = cell.initialize_carry(
+        init_key,
+        batch_dims=x.shape[1 : 1 + num_batch_dims], 
+        size=size,
+      )
+
+    scan = transforms.scan(
+      cell.__class__.__call__,
```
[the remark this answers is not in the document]
cgarciae: Done

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+    # select last time step
+    if not return_sequences:
+      # build slices
+      slices: List[Union[slice, int]] = [slice(None)] * x.ndim
+      slices[time_axis] = -1
+      # apply slices
+      x = x[tuple(slices)]
+
+    if return_state:
+      return x, carry
```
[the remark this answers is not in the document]
cgarciae: Done

cgarciae:
@jheek `unroll` is currently not supported by `nn.scan`:

https://github.com/google/flax/blob/18be4d4dbf8ad18fda099355f1a698dfe94c8989/flax/linen/transforms.py#L665-L675

Maybe we should open an issue so it does.

flax/linen/recurrent.py
cgarciae: Right now "for free" the user could globally reset the memory like this:

```python
variables = module.init(...)
initial_memory = variables['memory']
...
# reset
variables = variables.copy({'memory': initial_memory)
```
Your proposed mechanism sounds very flexible, we could indeed accept a boolean or a mask. However, would be nice to have a use case. Maybe this feature can be added later.

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+    if stateful:
+      self.put_variable('cache', 'carry', carry)
+
+    # revert time axis to its original position
+    x = x.swapaxes(0, time_axis)
+
+    # select last time step
+    if not return_sequences:
+      # build slices
+      slices: List[Union[slice, int]] = [slice(None)] * x.ndim
```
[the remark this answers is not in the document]
cgarciae: Implemented it using `index_in_dim`

flax/linen/recurrent.py
```diff
@@ -497,3 +500,189 @@ def initialize_carry(rng, batch_dims, size, init_fn=zeros):
...
+      carry = initial_carry
+
+    if stateful:
+      self.put_variable('cache', 'carry', carry)
+
+    # revert time axis to its original position
+    x = x.swapaxes(0, time_axis)
+
+    # select last time step
+    if not return_sequences:
```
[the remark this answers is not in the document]
cgarciae: @marcvanzee said something similar, I like the idea. I mainly implemented it just to keep up with keras but I totally agree that the user could do this. Will remove it.

flax/linen/recurrent.py
cgarciae: Done.

flax/linen/recurrent.py
jheek: I think the use case is pretty common. If you want to concatenate variable size sequences without padding all sequences to the maximum length you need a flexible reset mechanism that works per item and not for the whole batch. A similair trick is used in attention where the sequence id is used to mask out attention between input examples

cgarciae: Last change introduced `reset_state` as suggested to address this.

marcvanzee:
@cgarciae as discussed offline, I am sorry but I don't think we can get this PR merged as it is now. You are adding a lot of code at once to our core library (over 500 LOC added in `recurrent.py`, which currently is <500 LOC), and you are introducing a lot of complexity: the function `rnn_forward` you are introducing contains 18 arguments.

I think it helps to always keep the [Flax philosophy](https://flax.readthedocs.io/en/latest/philosophy.html) in mind, in particular: 
* _"Generally, prefer duplicating code over adding options to functions"_
* _"If we expose an abstraction to our developers, we own the mental overhead"_

Our abstractions are purposely minimal and simple. Your PR goes somewhat against this philosophy, so before reviewing it we should first discuss it in detail and reach consensus about what we want to add. Therefore, going forward I think a better strategy for this project is to first file an issue (maybe a FLIP, but could also just be an issue) where you explain:

1. the motivation for this change: why do we need it?
2. possible implementations: what are the alternatives? what are their tradeoffs? Which one is preferred?
3. an implementation plan: what are the steps? how long will they take?

We then can iterate on this issue until we reach consensus on what we want to implement.

Again, I am truly sorry you put so much time in writing this PR, and I wished we could have warned you earlier, but let's see this as a learning lesson for all of us!

cgarciae:
Hey @marcvanzee, thanks for the feedback! I will be opening an issue with proposal (FLIP-like) answering these questions and offering a list of features so the team can choose what is best initially. I think the time was well spent regardless since a lot was learned and lead to some communication with other teams, hopefully all this is transferred to a future implementation.

I'll be closing the PR as it will not be merged, but will keep the branch alive.

PR #2132, opened 2022-05-21
Upgrade to modern python idioms using pyupgrade.

levskaya:
I ran the command
```
pyupgrade --py37-plus --keep-runtime-typing **.py
```
across all of the flax subdirectories and inspected all changes by eye.

All the suggestions seemed good to me with the exception of the tool's
preference for python "set" literal notation (i.e. using `{...}` instead
of `set(...)`) which I think makes several bits of code less clear to readers
who aren't so familiar with python set vs dict notation.

Most of the changes are modern `super()` notation, string formatting
fixes, and modernization of tuple/iterator syntax.


cgarciae:
`pyupgrade` is available as a `pre-commit` hook if we want to continually apply these format conventions.

https://github.com/asottile/pyupgrade

PR #2159, opened 2022-05-27
Fix missing seperator in static fold in

### LINKED ISSUE #2157: LazyRNG can accidentally have key sharing across layers.
User reported repo:
```
import jax
from flax import linen as nn

class Leaf(nn.Module):
  def __call__(self, x):
    return x + jax.random.randint(self.make_rng("rng"), (), 0, 100)

class Node(nn.Module):
  leaf_name: str
  @nn.compact
  def __call__(self, x):
    return Leaf(name=self.leaf_name)(x)

class Model(nn.Module):
  @nn.compact
  def __call__(self, x):
    return (Node(name="ab", leaf_name="cdef")(x),
            Node(name="abc", leaf_name="def")(x),
    )

print(Model().apply({}, 0, rngs={"rng": jax.random.PRNGKey(33)})
# (DeviceArray(23, dtype=int32), DeviceArray(23, dtype=int32))
```

This occurs because we don't add a separator at:
https://github.com/google/flax/blob/main/flax/core/scope.py#L102

jheek:
fixes #2157

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -138,6 +138,8 @@ def _fold_in_static(rng: PRNGKey,
     return rng
   m = hashlib.sha1()
   for x in data:
+    # encode seperate to avoid collision slike for example: ("ab", "c") and ("a", "bc")
+    m.update(b'\00')
     if isinstance(x, str):
       m.update(x.encode('utf-8'))
     elif isinstance(x, int):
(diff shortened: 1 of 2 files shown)


levskaya (approved):
Looks good.  Seeing a number of internal tests fail due to RNG change... will have to think about how to deal with that.

levskaya (approved):
Loooks good.  I'm seeing this break a bunch of internal tests due to overtight tolerances, golden tests, etc... so we'll need to decide how to integrate it.

marcvanzee:
@levskaya any update on this?

PR #2162, opened 2022-05-29
Improve tabulate

cgarciae:
# What does this PR do?
Adds various small improvements to `tabulate` including:
* Removes `output_methods` argument, it is no longer necessary since now internally `capture_intermediates` is set to a function that collects them automatically (inspired by [this comment](https://github.com/google/flax/pull/1844#issuecomment-1140321111), thanks @trvs2cool!). Also solves the issue of intermediates for methods other than `__call__` not being captured.
* Adds the `exclude_methods` argument which lets you exclude certain methods, `setup` is excluded by default.
* Adds a test for modules with methods
* Sets `summary.tabulate`'s `mutable` argument to `True` by default to be consistent with `Module.tabulate`'s signature.
* Fixes `None`s being rendered as `null`.
* Removes argument default values from `_get_module_table` to avoid potential errors.

### DIFF
--- a/tests/linen/summary_test.py
+++ b/tests/linen/summary_test.py
@@ -29,44 +29,73 @@ from flax.linen.summary import _get_module_table
 # Parse absl flags test_srcdir and test_tmpdir.
 jax.config.parse_flags_with_absl()
 
-def _maybe_get_shape(x):
-  return x.shape if hasattr(x, "shape") else x
+def _get_shapes(pytree):
+  return jax.tree_map(lambda x: x.shape if hasattr(x, 'shape') else x, pytree)
 
+class ConvBlock(nn.Module):
+  features: int
+  kernel_size: List[int]
+  test_sow: bool
 
-def _get_tabulate_cnn(test_sow: bool = False) -> Type[nn.Module]:
-  @dataclasses.dataclass
-  class ConvBlock(nn.Module):
-    features: int
-    kernel_size: List[int]
-    training: bool
+  def setup(self) -> None:
+    self.conv = nn.Conv(self.features, self.kernel_size)
+    self.bn = nn.BatchNorm()
+    self.dropout = nn.Dropout(0.5)
 
-    @nn.compact
-    def __call__(self, x: Array) -> Array:
-      x = nn.Conv(self.features, self.kernel_size)(x)
+  def block_method(self, x: Array, training: bool) -> Array:
+    x = self.conv(x)
 
-      if test_sow:
-        self.sow('intermediates', 'INTERM', x)
+    if self.test_sow:
+      self.sow('intermediates', 'INTERM', x)
 
-      x = nn.BatchNorm(use_running_average=not self.training)(x)
-      x = nn.Dropout(0.5, deterministic=not self.training)(x)
-      x = nn.relu(x)
-      return x
+    x = self.bn(x, use_running_average=not training)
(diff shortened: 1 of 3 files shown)


tests/linen/summary_test.py
```diff
@@ -33,38 +33,65 @@ def _maybe_get_shape(x):
   return x.shape if hasattr(x, "shape") else x
 
 
-def _get_tabulate_cnn(test_sow: bool = False) -> Type[nn.Module]:
+def _get_tabulate_cnn(test_sow: bool = False, use_method: bool = False) -> Type[nn.Module]:
   @dataclasses.dataclass
   class ConvBlock(nn.Module):
```
marcvanzee, line 38: Can't you just have one class with both `__call__` and `block_method` implemented? (maybe with different behavior)

Then you can just make `test_sow` a class attribute as well and you don't need `_get_tabulate_cnn` anymore.

tests/linen/summary_test.py
```diff
@@ -179,8 +207,8 @@ def test_module_summary_with_depth(self):
 
     table = _get_module_table(
       module, 
-      {"dropout":random.PRNGKey(0), "params": random.PRNGKey(1)}, 
-      depth=1,
+      {"dropout":random.PRNGKey(0), "params": random.PRNGKey(1)},
+      method=CNN.__call__, mutable=True, depth=1,
```
marcvanzee, line 211: Why did you add the `method` kwarg here? It should be the default right?

marcvanzee (changes requested):
One thought about `_maybe_get_shape`: do you really need it? It seems the outputs of the `table[i].outputs` is always either an array or a pair `(array, dict)`. Can't you just compare them directly? E.g. instead of:

```py
self.assertEqual(
      jax.tree_map(_maybe_get_shape, table[0].outputs),
      (x.shape, dict(training=True)),
    )
```

You can do:


```py
self.assertEqual(table[0].outputs, 
      (x.shape, dict(training=True)),
```

Or am I missing something?

tests/linen/summary_test.py
cgarciae: Sounds reasonable. Will refactor.

tests/linen/summary_test.py
cgarciae: I removed all defaults from this private `_get_module_table` function, however, `method` here could be set to `None` so that `__call__` is internally used by default.0

cgarciae:
@marcvanzee you'd have to do it like this:

```python
self.assertEqual(
  (table[0].outputs[0].shape, table[0].outputs[1]),
  (x.shape, dict(training=True)
)
```
Doesn't look as bad as I thought. Will refactor.

marcvanzee:
> @marcvanzee you'd have to do it like this:
> 
> ```python
> self.assertEqual(
>   (table[0].outputs[0].shape, table[0].outputs[1]),
>   (x.shape, dict(training=True)
> )
> ```
> 
> Doesn't look as bad as I thought. Will refactor.

Or just as two separate asserts?

```py
self.assertEqual(table[0].outputs[0].shape, x.shape)
self.assertEqual(table[0].outputs[1], dict(training=True))
```

cgarciae:
@marcvanzee refactored `test_summary.py` based on your comments.
Also added the `excluded_methods` argument which lets you exclude certain method, `setup` is excluded by default.

marcvanzee (approved):
Looks good!

Issue #1676, opened 2021-11-17
Further improve Flax Modules on ReadTheDocs


marcvanzee:
Following #1416 we should further improve Flax Modules on ReadTheDocs.

(Content below from @andsteing)

We now have an `Parameters` section with only names/types:
![image](https://user-images.githubusercontent.com/16050001/142163481-2c386e17-e53b-49b0-a04c-053335bf9783.png)

And then the attributes after the return type:
![image](https://user-images.githubusercontent.com/16050001/142163590-d733ca8b-580e-4abe-8642-21fa1105e4e9.png)

And then only methods defined in the module (which in this case actually are "functional attributes"):
![image](https://user-images.githubusercontent.com/16050001/142163650-3e4879f9-9d97-4e98-863e-88e69797ff52.png)

In the future this **could be further improved** with

1. Removing the `Parameters` section
2. Showing all attributes concisely with their type + doc (i.e. making better use of horizontal screen estate)
3. Showing "functional attributes" as attributes and not as methods without docstring.

marcvanzee:
Closing this after @cgarciae 's awesome improvements to RTD!! 🎉

PR #2167, opened 2022-05-31
Improve module docs

cgarciae:
# What does this PR do?

Solves some of the "further improvements" points from #1676:
* Removes the `Parameters` section.
* Removes "functional attributes" from the methods table.

## Notes
To remove the "functional attributes" a patch is made to `sphinx.ext.autosummary.generate` inside `conf.py`. We can send a PR to sphinx with this patch so we don't have to maintain this code in the future. The patch consists on providing an `annotations` variable to `autoclass` with a list of the class `__annotations__` keys, we use this to exclude false methods by checking if they appear as annotations.

## Sample
Screenshot from 2022-06-02 12-56-40

After this change the `Methods` table in `Dense` is now empty.

Screenshot from 2022-06-02 12-58-21

We should consider moving `__call__` inside `Methods`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -35,6 +35,8 @@ sys.path.insert(0, os.path.abspath('..'))
 # Include local extension.
 sys.path.append(os.path.abspath('./_ext'))
 
+# patch sphinx
+import docs.conf_sphinx_patch
 # -- Project information -----------------------------------------------------
 
 project = 'Flax'
@@ -76,7 +78,7 @@ autosummary_generate = True
 
 master_doc = 'index'
 
-autodoc_typehints = 'description'
+autodoc_typehints = 'none'
 
 
 # -- Options for HTML output -------------------------------------------------
--- a/docs/_templates/autosummary/flax_module.rst
+++ b/docs/_templates/autosummary/flax_module.rst
@@ -15,7 +15,7 @@
    .. autosummary::
 
    {% for item in methods %}
-   {%- if item not in inherited_members and not item in ['__init__'] %}
+   {%- if item not in inherited_members and item not in annotations and not item in ['__init__'] %}
        ~{{ name }}.{{ item }}
    {%- endif %}
    {%- endfor %}
--- a/docs/conf_sphinx_patch.py
+++ b/docs/conf_sphinx_patch.py
@@ -1,0 +1,173 @@
+# Copyright 2022 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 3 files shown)


docs/conf.py
```diff
@@ -119,3 +119,162 @@
 # Tell sphinx-autodoc-typehints to generate stub parameter annotations including
 # types, even if the parameters aren't explicitly documented.
 always_document_param_types = True
+
+
+# -- Patch sphinx.ext.autosummary.generate ----------------------------------
```
marcvanzee, line 124: Did you write all this code yourself, or did you just modify a specific line? I think it is useful to add a link to the original code and mark which lines you changed in this patch.

marcvanzee (changes requested):

docs/conf.py
cgarciae: As stated a little bit below in that comment, this patch only adds a single line to `generate_autosummary_content`:

```python
ns['annotations'] = list(getattr(obj, '__annotations__', {}).keys())
```
I've added the link to the source to the comment.

marcvanzee: Ahh apologies! I wasn't really clear to me what that line meant, thanks for the clarification.

docs/conf.py
```diff
@@ -119,3 +119,163 @@
...
+
+# This patch is needed to make autosummary provide the "annotations"
+# variable so we can exclude function attributes from the methods list
+# in flax_module.rst. The patch as such only adds this single line:
+#
+#     ns['annotations'] = list(getattr(obj, '__annotations__', {}).keys())'
+#
+# We should consider sending a PR to sphinx so we can get rid of this.
```
andsteing, line 132: Could you add a `TODO` and reference to an open issue assigned to somebody for this?

docs/conf.py
```diff
@@ -119,3 +119,163 @@
 # Tell sphinx-autodoc-typehints to generate stub parameter annotations including
 # types, even if the parameters aren't explicitly documented.
 always_document_param_types = True
+
+
+# -- Patch sphinx.ext.autosummary.generate ----------------------------------
+
+# This patch is needed to make autosummary provide the "annotations"
```
andsteing, line 126: Should we split the patch into a separate module for better readability?
That way the `config.py` would remain short and be about config, and the (temporary) patch would live in a separate file.

andsteing (approved):
Changes look great, thanks!

Added two comments about how to best integrate the temporary patch. Otherwise SGTM.

marcvanzee (approved):
LGTM (pending Andreas' suggestions)

docs/conf.py
cgarciae: Thanks for bringing this up! That was my original idea but I was hesitant to introduce a new file here, since we agree on this I will be adding this code to a new `sphinx_patch.py` file.

cgarciae: Created `conf_sphinx_patch.py`.

marcvanzee:
@cgarciae the build of this PR is failing

marcvanzee (approved):

PR #1666, opened 2021-11-09
Draft: re-do docs landing page

avital:
Try a first attempt at a full re-haul of the RTD landing page with more explicit calls to action. This is a starting point for future exploration.

See a preview at https://flax--1666.org.readthedocs.build/en/1666/

PR #1626, opened 2021-10-18
Moves docs/examples.rst to examples/README.md

andsteing:

PR #2177, opened 2022-06-08
RDT Redesign

cgarciae:
# What does this PR do?

This PR continues #1666 by @avital and expands it in the following ways:

* Reorganizes navigation, tries to linearize user journey
  * Annotated MNIST was renamed to Getting Started, should serve as the starting point.
  * Documents in `notebooks` and `howtos` were reorganized into `Guides` and `Advanced Topics`. `Guides` is the next sections users view and is sorted as best as possible.
  * Examples section was added with links to all examples.
* Updates landing page content
  * Added the [sphinx-design](https://sphinx-design.readthedocs.io) extension to facilitate creating responsive bootstrap-like elements.
  * Added a `Features` sections.
  * Design and content of the whole page was improved.

Live demo: https://flax--2177.org.readthedocs.build/en/2177/


docs/examples.rst
```diff
@@ -0,0 +1,17 @@
+Examples
```
andsteing, line 1: We moved our example docs from RTD to Github's README in #1626 because we thought that "examples are all about the code", so navigating it in Github makes more sense, because individual examples have their docs as READMEs, and because when people start from Github they'll probably miss the RTD.

Having some docs in RTD and some other docs in Github can make it easy to miss one or the other part, so while I don't think there's a clearcut answer where to put docs, we should always think of this.

(Note that having the examples docs on Github is not set in stone, so we should definitely reconsider if there are good reasons for that)

docs/index.rst
```diff
@@ -2,78 +2,110 @@
...
-==================
+   .. grid-item::
+      :columns: 7 5 7 5
+      :margin: 0 2 auto auto 
 
-Flax is a neural network library and ecosystem for JAX that is
-designed for flexibility. Flax is in use by a growing community of
-researchers and engineers at Google who happily use Flax for their
-daily research.
+      .. image:: ./flax.png
```
andsteing, line 14: the logo is cool, but it doesn't add anything.
also, we now have it twice on this page (top left, center).

let's instead use the precious screen estate to show links / text / code ...

cgarciae: I knew this would be controversial 😅 I took inspiration from the [FastAPI](https://fastapi.tiangolo.com/) landing page, I like that yields a little bit of a dramatic entrance, it looks better in this other theme where the nav bar is at the top and the icon is small so it doesn't look repetitive:

Screenshot 2022-06-30 at 09-43-46 Flax

I wont remove it immediately but I am aware this won't go through.

andsteing: Not per se against a dramatic entrance ;)

Note also that FastAPI has a bit of a different setup on the page - their actual logo (circle with flash) is much smaller, and they have large text next to it. (not saying we should do that, but if they simply had the huge green circle with flash logo I wouldn't like that too much)

docs/examples.rst
andsteing: I think more generally the question is how to make best use of RTD vs. Github.

IMO we should stick to Github when possible (e.g. for viewing code / browsing files), and only use RTD for documentation.

andsteing: as discussed offline the other day, let's instead do the following:

- move `examples/README.md` to this `examples.rst`
- make a BOLD LARGE LINK in `examples/README.md` that points users to the `examples.rst`

this way we have our doc in RST and searchable and all that, but we also make sure that people browsing the repo won't miss the examples documentation

docs/api_reference/_autosummary/flax.linen.BatchNorm.rst
```diff
@@ -0,0 +1,20 @@
+﻿flax.linen.BatchNorm
```
andsteing, line 1: Regarding all the `/api_reference/*.rst` files: are these autogenerated? if yes, consider mentioning them in the PR (I don't immediately see why we have different paths, and why so many new files appeared)

docs/examples.rst
cgarciae: Great idea @andsteing! I've opened #2331 to do this after we merge this, I don't want to do it right now as the PR is already too large but we should do it soon after.

docs/api_reference/_autosummary/flax.optim.AdaBelief.rst
```diff
@@ -0,0 +1,30 @@
+﻿flax.optim.AdaBelief
```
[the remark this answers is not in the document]
cgarciae: Thanks for bringing this up, these files shouldn't be committed, I've updated the `.gitignore` and removed these files.

docs/index.rst
```diff
@@ -3,77 +3,265 @@
...
+
+
+----
+
+Installation
+^^^^^^^^^^^^
+
+.. code-block:: bash
+
+   pip install flax
```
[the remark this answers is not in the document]
cgarciae: Added!

docs/index.rst
```diff
@@ -3,77 +3,265 @@
...
+.. testsetup::
+
+   import jax
+   from jax.random import PRNGKey
+   import flax.linen as nn
+   import jax.numpy as jnp
+
+.. testcode::
+
+   class MLP(nn.Module):
```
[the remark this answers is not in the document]
cgarciae: I like the idea of using this section to mention more features. I've created this issue for now #2332.
We can work on this soon after the merge

docs/api_reference/_autosummary/flax.linen.BatchNorm.rst
cgarciae: Yes, thanks for the heads up, they have been removed!

docs/_static/css/flax_theme.css
```diff
@@ -11,7 +11,7 @@
 .rst-content table.docutils td {
   vertical-align: top;
   padding: 0;
-  width: 50%;
+/*  width: 50%; avitalo: why was this here??? */
```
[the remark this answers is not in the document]
cgarciae: removed!

docs/index.rst
```diff
@@ -3,77 +3,265 @@
...
+      :columns: 12 12 12 6
+
+      .. card:: Functional API
+         :class-card: sd-text-black sd-border-0
+         :shadow: none
+         :class-title: sd-fs-5
+
+         .. div:: sd-font-normal
+            
+            Flax's functional API radically redefines what Modules can do via lifted transformations like vmap, scan, etc, while also enabling seamless integration with other JAX libraries like Optax and Chex.
```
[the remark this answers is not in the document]
andsteing: resolved, but no changes?

andsteing (approved):

PR #2199, opened 2022-06-14
Exclude pseudo-fields of dataclass

YouJiacheng:
pseudo-fields inlclude ClassVar's and InitVar's:
ClassVar is shared by all instances of the class by design, i.e. it is a class attribute. 
InitVar is not regarded as an attribute of an instance, it will be passed to `__post_init__ `hook only.
They shouldn't be data or meta-data of pytree.
This change will also make flax don't depend on dataclasses' internal implementation "`__dataclass_fields__`".
possible problem: There is not a runtime promise that ClassVar/InitVar can not be used as a normal instance attribute, i.e. user can read and mutate them freely. This problem might be not important, since flax use `frozen=True` and mutation is not allowed.

## Checklist
      checks if that's the case).

### DIFF
--- a/flax/struct.py
+++ b/flax/struct.py
@@ -95,16 +95,15 @@ def dataclass(clz: _T) -> _T:
   Returns:
     The new class.
   """
-  # workaround for pytype not recognizing __dataclass_fields__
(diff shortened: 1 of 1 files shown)


jheek (approved):
This looks much better :)

PR #2204, opened 2022-06-17
Fix codediff overflow

cgarciae:
# What does this PR do?

TODO


docs/_ext/codediff.py
```diff
@@ -27,6 +27,7 @@
 In order to highlight a line of code, prepend it with "#!".
```
marcvanzee, line 27:
```suggestion
In order to highlight a line of code, append "#!" to it.
```

docs/conf.py
```diff
@@ -88,6 +89,7 @@
 #
 html_theme = 'sphinx_book_theme'
 # html_style = 'css/flax_theme.css'
+html_css_files = ["css/flax_theme.css"]
```
marcvanzee, line 92: So it was actually not being used?
Also, remove the line above?

docs/conf.py
```diff
@@ -61,7 +61,8 @@
     'nbsphinx',
     'recommonmark',
     'codediff',
-    'sphinx_markdown_tables'
+    'sphinx_markdown_tables',
```
marcvanzee, line 64: I think this is used to render markdown tables. Maybe try removing it and see if the diffs still render correctly. If so you can remove it.

marcvanzee (changes requested):
I really like this change, it looks even better than before! Thanks a lot.

docs/conf.py
cgarciae: Seems that it was not being used with the new theme.

docs/conf.py
cgarciae: Since we don't know what its being used for maybe lets leave it? Sphinx will not fail if a pluggin is not present.

marcvanzee: Let's just remove it then, we can always put it back if we find something has broken.

marcvanzee (approved):

Issue #2200, opened 2022-06-14
Support for lifted nn.switch module


marcociccone:
Hi,  I think it would be useful to add a lifted version of `jax.lax.switch` to flax linen as done for other libraries such as haiku.
As a use case consider having a list of functions (in our case `nn` modules) and you want to apply them over an axis of your input. https://github.com/google/jax/issues/673#issuecomment-894955037

See the [original](https://colab.research.google.com/gist/tomhennigan/86fe7b7d46a930b58a5d8d76a75d75ea/fix-for-300.ipynb) haiku example from https://github.com/deepmind/dm-haiku/issues/300#issuecomment-1028195090 that I readapted to how I expect it to work in flax.

````python
class MultiDense(nn.Module):
    features: nn.Module
    num_classes: int
    num_heads: int = 1
    dtype: Any = jnp.float32 

    @nn.compact
    def __call__(self,
                 x,
                 train: bool = True,
                 heads_idx: List = None):
        
        # first time calling for init shoudl pass all indexes 
        self.heads = [nn.Dense(num_classes, dtype=self.dtype) for range(num_heads)]
        vmap_fns = nn.vmap(lambda i, x: nn.switch(i, self.heads, x))
        x = vmap_fns(heads_idx, x)
        return x

cgarciae:
Implemented in #2205.

cgarciae:
See example how to implement multi-head architecture here: https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.switch.html?highlight=switch#flax.linen.switch

jheek:
Btw do keep in mind that if you vmap over the switch index than jax will rewrite the switch into a select meaning it will execute *all the branches for each input* so in this example the conditional doesn't actually avoid any computation

jheek:
You could use an nn.scan together with an nn.cond but that only works efficiently if the heads are large enough and make efficient use of flops if they aren't batched together

PR #2205, opened 2022-06-17
Add nn.switch

cgarciae:
# What does this PR do?

Add `nn.switch` as a lifted version of `jax.lax.switch`. 

Solves #2200

### DIFF
--- a/flax/linen/transforms.py
+++ b/flax/linen/transforms.py
@@ -1122,6 +1122,89 @@ def cond(
       pred, *operands,
       variables=variables, rngs=rngs)
 
+def _switch_wrapper(*args, variables, rngs, n_branches):
+  # first n_branches arguments are branches.
+  # then scope, index, and the rest are *operands
+  branches = args[:n_branches]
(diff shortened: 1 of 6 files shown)


flax/linen/transforms.py
```diff
@@ -1262,3 +1262,40 @@ def core_fn(scopes, *args, **kwargs):
       return trafo_fn(module_scopes, *args, **kwargs)
 
   return wrapped_fn
+
+def switch(index, branches: Sequence[Module], *args, **kwargs):
+  """Lifted version of `jax.lax.switch`."""
+
+  all_variables = [module.scope.variables() for module in branches]
```
jheek, line 1269: Switch is esentially a generalization of cond, right? Probably the code between nn.cond and nn.switch can be mostly shared. This implementation doesn't actually uses the lifting logic so it's missing all the weird edge cases it handles

jheek (changes requested):

jheek (commented):
Looks good!

jheek (approved):

Issue #900, opened 2021-01-19
Writing asynchronous checkpoints


marcvanzee:
Currently checkpoints are not written asynchronously in Flax, which means that by default, the program waits with execution until the checkpoint is saved. Writing a checkpoint asynchronously should be as simple as launching a new thread on a function that saves the checkpoint:

```
pool.apply_async(checkpoints.save_checkpoints, (ckpt_dir, optimizer, step))
```

Important: copy optimizer to CPU before calling `apply_async`, otherwise it may happen that you need 2x TPU memory to store parameters, if you are training the model while transferring data (asynchronously) from TPU->CPU.

Users have argued that it may be worthwhile to just add a utility function that does this to make sure it is done right.

ayush-1506:
@marcvanzee Hi! Can I work on this issue? If so, anything in particular I should be aware of? (Apart from the details mentioned in your comment above).

marcvanzee:
Hi @ayush-1506, definitely, please give it a try! Thanks for picking this up.

ayush-1506:
@marcvanzee Thanks! I'm wondering what the API should look like over here: We want to continue training while also saving the checkpoints, since these two are independent of each other. Ideally the user could spawn a new process that saves the checkpoint, but this is going to be manual. What kind of a utility function are we looking at over here? (What should this look like?)

marcvanzee:
Actually, thinking about this some more, I think writing a utility function that writes the checkpoint asynchronous safely would require a full copy of the parameters, which seems a like a bad idea. Otherwise, as you say, we should somehow disallow modifying the parameters, but I'm not sure if there is a clean solution for this.

So on second thought I'd actually be somewhat hesitant to implement this as a standard utility function into Flax. @jheek, @levskaya: any thoughts?

jheek:
Generally you want to copy the params from device to host in the main thread (to avoid OOM). Then use a async thread to do the serialisation & IO. It's mostly safe there is just the one edge case where you should block the save_checkpoint if the previous one is still busy.

PR #2233, opened 2022-06-23
Add single-host async save to save_checkpoint.

IvyZX:
Add single-host async save functionality to save_checkpoint.

User needs to create an AsyncManager instance outside the save_checkpoint() call and use it to make sure async saves still happen in order (aka previous saves will block subsequent saves).

Fixes https://github.com/google/flax/issues/900 https://github.com/google/flax/issues/1543

## Checklist
- [x] This change is discussed inGithub [issue](https://github.com/google/flax/issues/900) and [issue](https://github.com/google/flax/issues/1543)


levskaya (approved):
Looks great!

PR #2234, opened 2022-06-23
Add a method for detecting the use of "init" functions.

levskaya:
We add an optional frozen-dict-valued `flags` field to Scope objects that are copied onwards, and use
it solely to implement a flag to indicate the use of top-level "init" functions.

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -367,6 +367,11 @@ class _ChildRNGSentinel:
 # used to identify that an rng counter is meant for a child scope
 child_rng_token = _ChildRNGSentinel()
 
+class _DefaultSentinel:
+  pass
+# used to denote no default flag value on scope
+no_flag = _DefaultSentinel()
+
 
 class Scope:
   """A Scope allows easy access to variables and manages RNGS of a neural network layer.
@@ -387,7 +392,8 @@ class Scope:
                name: Optional[str] = None,
                mutable: CollectionFilter = False,
                parent: Optional['Scope'] = None,
-               path: Iterable[str] = ()):
+               path: Iterable[str] = (),
+               flags: Optional[Dict] = None):
     """Initializes a Scope.
 
     Args:
@@ -397,6 +403,7 @@ class Scope:
       mutable: A CollectionFilter determining which variables are mutable.
       parent: The parent scope.
       path: The path in the variable tree from the root scope to this scope.
+      flags: internal flags.
     """
     rngs = {k: LazyRng.create(v) for k, v in rngs.items()} if rngs else {}
     self._variables = variables
@@ -405,6 +412,7 @@ class Scope:
     self.path = tuple(path)
     self.rngs = rngs
     self.mutable = mutable
+    self.flags = freeze({} if flags is None else flags)
 
     self._root = parent.root if parent else None
     self.trace_level = tracers.trace_level(tracers.current_trace())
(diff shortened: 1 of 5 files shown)


flax/linen/module.py
```diff
@@ -1033,6 +1033,14 @@ def make_rng(self, name: str) -> PRNGKey:
       raise ValueError("Can't use RNGs on unbound modules")
     return self.scope.make_rng(name)
 
+  def is_under_init(self) -> bool:
+    if self.scope is None:
+      raise ValueError("Can't check if under init on unbound modules")
+    if self.scope.flags and 'initializing' in self.scope.flags:
```
andsteing, line 1039: nit:
```python
return self.scope.flags and 'initializing' in self.scope.flags
```

flax/core/scope.py
```diff
@@ -387,7 +387,8 @@ def __init__(self,
                name: Optional[str] = None,
                mutable: CollectionFilter = False,
                parent: Optional['Scope'] = None,
-               path: Iterable[str] = ()):
+               path: Iterable[str] = (),
+               flags: Optional[Set] = None):
```
cgarciae, line 391: Maybe the signature could be 

```python
flags: Sequence[str] = ()
```
and have
```
self.flags = set(flags)
```
This way `self.flags` is always a `set`.

flax/linen/module.py
jheek, line 1039: I think it would be less error-prone if flags have no defaults. Here you have default behavior and typos default to False. It would be more robust to define flags as a dict without the options to define defaults at the call site. So if a user requires self.is_initializing the flag **must** be set

flax/core/scope.py
levskaya: did something similar

flax/core/scope.py
```diff
@@ -761,6 +765,13 @@ def _populate_collections(self):
     for col in collections:
       self._collection(col)
 
+  def has_flag(self, key) -> bool:
+    return key in self.flags
+
+  def get_flag(self, key) -> Any:
```
cgarciae, line 771: Maybe `get_flag` could accept a default in case flag doesn't exist?

```python
  def get_flag(self, key, default=None) -> Any:
    if key not in self.flags:
      if default is None:
        return ValueError(f'Flag {key} not present in scope.')
      return default
    return self.flags[key]
```

flax/linen/module.py
```diff
@@ -1033,6 +1033,12 @@ def make_rng(self, name: str) -> PRNGKey:
       raise ValueError("Can't use RNGs on unbound modules")
     return self.scope.make_rng(name)
 
+  def running_init(self) -> bool:
+    """Returns True if running under self.init(...) or nn.init(...)()."""
+    if self.scope is None:
+      raise ValueError("Can't check if running under init() on unbound modules")
+    return self.scope.has_flag('initializing')
```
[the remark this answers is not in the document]
levskaya: updated this.

flax/core/scope.py
levskaya: I added an optional default argument - if none provided it throws an error on missing flag.

levskaya:
Note: there are 2 internal test failures which are pre-existing and unrelated.

jheek (commented):
Overall looks good. I do think we should also replace the cases where we use is_mutable_collection("params") as a substitute for is_initializing. Otherwise the new init check isn't actually used by the core Modules themselves

PR #2277, opened 2022-07-07
Split build into multiple jobs

cgarciae:
# What does this PR do?

Splits the `build` job into multiple jobs. This has two main benefits:
* Isolates the `pytest` tests from the `notebooks` which often try to install additional dependencies.
* Speed: some tasks are independent and can be ran in parallel.

### DIFF
--- a/tests/run_all_tests.sh
+++ b/tests/run_all_tests.sh
@@ -39,9 +39,6 @@ handle_errors () {
     fi
 }
 
-# Run embedded tests inside docs
-sphinx-build -M doctest docs docs/_build
-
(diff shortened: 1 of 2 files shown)


tests/run_all_tests.sh
```diff
@@ -39,9 +39,6 @@ handle_errors () {
     fi
 }
 
-# Run embedded tests inside docs
```
marcvanzee, line 42: Why did you remove this? Don't we want to be able to test this locally?

marcvanzee (changes requested):
Nice!

PR #2313, opened 2022-07-21
Replace use of id() with global counter-based id.

levskaya:
Historically we key'd on id() to record sharing relationships during
lifting and outer module adoption.  This was dumb, and after recently
fixing one bad bug arising from id reuse, we should use something
sound instead.

### DIFF
--- a/flax/ids.py
+++ b/flax/ids.py
@@ -1,0 +1,57 @@
+# Copyright 2022 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""UUIDs for Flax internals."""
+
+import threading
+
+
+class UUIDManager:
+  """Globally unique counter-based id manager.
+
+  We need globally unique key ids for Module and Variable object instances
+  to preserve and recreate sharing-by-reference relationship when lifting
+  transforms and adopting outside Modules.
+  - Use of id() is unacceptable because these identifiers are literally
+    pointers which can be recycled, so we rely on a globally unique counter id
+    instead.
+  - We need to handle copy/deepcopy uniqueness via a wrapped type.
+  """
+  def __init__(self):
+    self._lock = threading.Lock()
+    self._id = 0
+
+  def __call__(self):
+    with self._lock:
+      self._id += 1
+      return FlaxId(self._id)
+
+uuid = UUIDManager()
+
+
+class FlaxId:
+  """Hashable wrapper for ids that handles uniqueness of copies."""
+  def __init__(self, rawid):
+    self.id = rawid
+  def __eq__(self, other):
+    return isinstance(other, FlaxId) and other.id == self.id
+  def __hash__(self):
+    return hash(self.id)
+  def __repr__(self):
+    return f"FlaxId({self.id})"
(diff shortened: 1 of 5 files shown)


levskaya:
Seeing an odd, seemingly unrelated doctest failure in one case here that confuses me, I'll have to look into it later.

flax/core/lift.py
```diff
@@ -232,6 +232,7 @@ def repack(inner_scope_tree):
             # Some lifted transforms like scan return redundant variables.
             continue
           for var_name, value in collection.items():
+            print('pack', col_name, var_name, value)
```
marcvanzee, line 235: Remove?

flax/ids.py
```diff
@@ -0,0 +1,43 @@
...
+  """Globally unique counter-based id manager."""
+  def __init__(self):
+    self._lock = threading.Lock()
+    self._id = 0
+
+  def __call__(self):
+    with self._lock:
+      self._id += 1
+      cur_id = self._id
+    return cur_id
```
marcvanzee, line 40: Does this work as well?

```suggestion
    with self._lock:
      self._id += 1
    return self._id
```

flax/core/lift.py
levskaya: lol.  thanks for noticing this, I was going crazy wondering what that doctest failure was about - I think this is an old edit that vscode held onto and was accidentally saved into this PR.

flax/ids.py
levskaya: I'm probably being overly paranoid due to old C programming trauma and weird lock/mutex edge conditions around what is supposed to be in the critical section, but I just tried to break this in python and I couldn't, so I think the simpler form is ok - I replaced it.

levskaya: thinking about it again and looking at some python bytecode disassembly -- technically the simpler version could be wrong if a thread switch occurs between exiting the with-block and re-accessing the value of _id before returning.... so even though it's an edge cases and we don't even really want to support multithreaded module construction, I'd rather do the correct thing.

levskaya:
ha - I noticed from a failing user's unit test that this implementation of a unique id will break under `deepcopy`, so I need to add a custom deepcopy method to fix this.

flax/ids.py
marcvanzee: Makes sense, returning directly seems like the best option.

marcvanzee:
> ha - I noticed from a failing user's unit test that any construction-created/self-bound unique id scheme will break under default `deepcopy` (which will copy the supposedly unique id), so I need to add a custom `__deepcopy__` method to fix this.

Is this actually what we want? Supposing a deepcopy will return an identical copy of the object, so maybe the id should be the same as well?

levskaya:
> Is this actually what we want? Supposing a deepcopy will return an identical copy of the object, so maybe the id should be the same as well?

Probably not since it would break the previous behavior, we really want the semantics to be "share by reference" which meant "key on object instance identity" following what we intended w. the old id()-based sharing.  In any case there are internal code/tests that assume the previous behavior: deepcopy should return a _new_ non-shared instance. (some people seem to use it as a hack to clone layer definitions)

jheek:
> deepcopy should return a new non-shared instance.

So is id() the right approach after all? Or perhaps we should error out on deep copies?

levskaya:
It's pretty easy to maintain the per-instance behavior (non shared ids) by defining `__copy__`, `__deep_copy__` on an wrapped, hashable id type.  I'm hesitant to break user code immediately, though we could choose to have Module have error-raising copy, deepcopy implementations to prevent its use and encourage use of `clone` instead in the few existing use sites.

levskaya:
Oops, this was accidentally merged early because of a miscommunication, we can roll it back if this change is ultimately unwanted (it should have no user-visible effects either way).

marcvanzee:
I have no problem with this change, your explanation makes sense to me so I am fine with the change as it is.

For reference: this is a strict improvement over the id() approach which was flawed since it could result in multiple objects having the same id. This is because id() just returns the address of the C pointer to the object instance, which can be reused after garbage collection. This can lead to bugs.

PR #2317, opened 2022-07-21
Document repo analytics

cgarciae:
# What does this PR do?
Documents how to run the repo analytics and does some cleanup.

### DIFF
--- a/.github/analytics/get_repo_metrics.py
+++ b/.github/analytics/get_repo_metrics.py
@@ -17,7 +17,7 @@ import os
 from datetime import datetime
 from pathlib import Path
 from typing import Callable, List
-import typer
+from absl import app, flags
 
 import pandas as pd
 import requests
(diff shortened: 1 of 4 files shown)


.github/analytics/get_repo_metrics.py
```diff
@@ -407,4 +408,4 @@ def main(
   plt.show()
 
 if __name__ == '__main__':
-  main()
+  typer.run(main)
```
gnecula, line 411: I believe that we use "from absl import app" and "app.main" in other places in the project to run the main function. Perhaps then we do not need the `typer` dependency.

cgarciae: Let me try changing it to `absl` for consistency.

cgarciae: Ok, update to use `absl`.

cgarciae:
Thanks @gnecula for reviewing it. Updated the script to use `absl`.

gnecula (approved):

PR #2325, opened 2022-07-23
Replace all jax.tree_* calls with jax.tree_util.tree_*

### LINKED ISSUE #2324: Fix jax.tree_blahblah FutureWarning warnings due to jax symbol cleanup.
Newer versions of JAX moved all tree utility methods to `jax.tree_util` and emit warnings for using the old locations under `jax.`
We added a temporary warning exemption for these FutureWarnings, but we should fix our ~400 use sites of the old locations.

levskaya:
JAX has moved all tree utils methods under jax.tree_util and will require using these in the near future.

Fixes #2324

### DIFF
--- a/docs/advanced_topics/lift.md
+++ b/docs/advanced_topics/lift.md
@@ -86,7 +86,7 @@ class ManualVmapMLP(nn.Module):
 
 xs = jnp.ones((3, 4))
 variables = ManualVmapMLP().init(random.PRNGKey(0), xs)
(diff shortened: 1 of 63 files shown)


gnecula (approved):

gnecula:
There seem to be a few more places to update, see failed test.

levskaya:
The other failures are actually chex warnings that we should just exempt from strict warning check.

PR #2371, opened 2022-08-04
Add path_aware_map function

cgarciae:
# What does this PR do?

Adds a `path_aware_map` function to `traverse_util`, this function makes it easy to use create Flax with Optax's`multi_transform` and `masked` combinators.

### DIFF
--- a/flax/traverse_util.py
+++ b/flax/traverse_util.py
@@ -43,13 +43,16 @@ returns a copy of the data including the provided updates.
 import abc
 import copy
 import dataclasses
+from typing import Any, Callable, Dict, Tuple
 import warnings
 
 import jax
 import flax
+from flax.core.scope import VariableDict
 
 from . import struct
 
+Path = Tuple[str, ...]
 
 # the empty node is a struct.dataclass to be compatible with JAX.
 @struct.dataclass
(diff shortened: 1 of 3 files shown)


flax/traverse_util.py
```diff
@@ -159,6 +164,45 @@ def unflatten_dict(xs, sep=None):
     cursor[path[-1]] = value
   return result
 
+def variables_path_partitions(
```
jheek, line 167: This function doesn't do anything with variables or partitioning specifically?
This function maps over a flattened dict so I think it should have a name and docstring that explains that.

flax/traverse_util.py
```diff
@@ -159,6 +164,45 @@ def unflatten_dict(xs, sep=None):
...
+    partition_fn: Callable[[Path, Array], Partition]
+) -> Callable[[VariableDict], Dict[Path, Partition]]:
+  """Creates a function that maps a variables structure to a path partitions dictionary 
+  compatible with Optax ``multi_transform`` and ``masked``.
+
+  Here is an example usage with ``multitransform``::
```
jheek, line 173: This module doesn't really have a direct connection to Optax. I think specific examples like this where you combine multiple independent APIs belong in seperate documentation. For example a HOWTO use Optax with Flax.

cgarciae: Point taken. It will be featured in the transfer learning tutorial I am creating so maybe that is that. Will modify the docstrings.

flax/traverse_util.py
```diff
@@ -159,6 +162,31 @@ def unflatten_dict(xs, sep=None):
...
+    {'a': {'x': 15, 'y': -3}, 'b': {'x': 25}}
+
+  Args:
+    f: A callable that takes in ``(path, value)`` arguments and maps them to a new value.
+    nested_dict: A nested dictionary structure.
+  
+  Returns:
+    A new nested dictionary structure with the mapped values.
+  """
+  flat = flatten_dict(nested_dict)
```
jheek, line 188: You should pass keep_empty_nodes=True here to make sure the structure isn't changes when there's empty dicts in the input. And make sure f isn't passed `empty_node` inputs

flax/traverse_util.py
```diff
@@ -159,6 +162,31 @@ def unflatten_dict(xs, sep=None):
     cursor[path[-1]] = value
   return result
 
+def path_aware_map(
+    f: Callable[[Path, Any], Any], nested_dict: VariableDict) -> VariableDict:
+  """A path-aware map that operates over nested dictionary structures. Works somewhat
+  similar to ``tree_map`` but the function ``f`` is addtionally provided with the
```
cgarciae, line 168: Improve description

jheek (approved):
Looks great!

PR #2372, opened 2022-08-04
Test docstrings with autodoc on CI

cgarciae:
# What does this PR do?

* Runs `pytest` with the `--doctest-modules` flag on CI, this runs `doctest` over the docstrings (tests code examples using `>>>`).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,7 +88,6 @@ jobs:
         sphinx-build -M doctest docs docs/_build -T
         # test docstrings
         pytest -n auto flax --doctest-modules --suppress-no-test-exit-code
-          
     - name: Build documentation
       run: |
         sphinx-build -M html docs docs/_build -T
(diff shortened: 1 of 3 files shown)


levskaya:
Does the `--doctest-modules` actually do what you say it does above?  From the pytest docs: 

> By default, all files matching the test*.txt pattern will be run through the python standard [doctest](https://docs.python.org/3/library/doctest.html#module-doctest) module. You can change the pattern by issuing... `--doctest-glob="*.rst"`

It's not clear it actually runs over python files?

cgarciae:
Later down they line they state:

> In addition to text files, you can also execute doctests directly from docstrings of your classes and functions, including from test modules:

So in summary, the new command

```bash
pytest -n auto flax --doctest-modules --suppress-no-test-exit-code
```
doesn't run any `txt` files because we don't have any, and doesn't run any tests either because its pointed at the `flax` folder which doesn't have any, but runs the docstrings which is the only thing we want from that line.

levskaya:
Aha thanks for clarifying!  I missed that.

levskaya (approved):
Thank you!

PR #2425, opened 2022-08-30
Fix Conv docstrings

### LINKED ISSUE #2156: Conv docs page doesn't show attribute/argument's description
`Conv` and `ConvLocal` inherit from `_Conv` but their docstrings don't "re-expose" the `Attributes` section so Sphinx doesn't show the description of each attribute to the users. An easy solution would be to just duplicate these sections, else somehow dynamically modifying `{Conv, ConvLocal}.__docs__` to add the common attributes section *might* work. 

Screenshot 2022-05-26 at 13-59-08 flax linen Conv — Flax documentation

cgarciae:
# What does this PR do?

Fixes #2156. Patches `Conv` and `ConvLocal` docstrings with the `Attributes` information form `_Conv`'s docstrings, this makes sure attributes are properly shown in the documentation page. Current documentation for `Conv` and `ConvLocal` doesn't show the attribute information:
Screenshot from 2022-08-30 12-16-12

After the change it now render attributes correctly:
Screenshot from 2022-08-30 12-17-12

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -467,7 +467,42 @@ class _Conv(Module):
 
 
 class Conv(_Conv):
-  """Convolution Module wrapping `lax.conv_general_dilated`."""
+  """Convolution Module wrapping `lax.conv_general_dilated`.
+  
+  Attributes:
+    features: number of convolution filters.
+    kernel_size: shape of the convolutional kernel. For 1D convolution,
+      the kernel size can be passed as an integer. For all other cases, it must
+      be a sequence of integers.
+    strides: an integer or a sequence of `n` integers, representing the
+      inter-window strides (default: 1).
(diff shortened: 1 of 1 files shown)


flax/linen/linear.py
```diff
@@ -481,6 +481,17 @@ class ConvLocal(_Conv):
   def shared_weights(self) -> bool:
     return False
 
+# get docstring from _Conv, delete leading spaces
```
jheek, line 484: This works for RTD but I think static analysis tools won't understand this kind of docstring manipulation. I would prefer to keep things simple and just copy paste the attributes

cgarciae: Got it, will do!

jheek (approved):

PR #2440, opened 2022-09-06
Improve attribute error msg for unbounded modules

### LINKED ISSUE #1014: Raise clearer Exception when calling method of unbound module

cgarciae:
# What does this PR do?

Fixes #1014. `AttributeError`'s message raised by `__getattr__` for unknown fields now reminds the user it should call `apply` when looking for attributes that might be defined inside `.setup()`if  `self.scope` is None`.


flax/linen/module.py
```diff
@@ -775,8 +775,11 @@ def __getattr__(self, name: str) -> Any:
     if name in self.__dict__:
       return self.__dict__[name]
     else:
-      raise AttributeError(
-          f'"{self.__class__.__name__}" object has no attribute "{name}"')
+      msg = f'"{self.__class__.__name__}" object has no attribute "{name}".'
+      if self.scope is None:
+        msg += (f' If "{name}" is defined in \'.setup()\', remember to call '
+          'the current method inside \'apply\'.')
```
jheek, line 781: '..., it is only accessible from inside "init" or "apply"'

jheek (approved):

PR #2447, opened 2022-09-07
Annotate Module fields

### LINKED ISSUE #2416: `__getattr__` and pytype checking
Currently `nn.Module` implement `__getattr__`. This disable any type checking when making a typo in attribute:

```python
my_module = MyModule()

my_module.non_existing_attribute  # << pytype do not detect this
```

It would be nice to wrap `__getattr__` in a `if not typing.TYPE_CHECKING:` so that pytype actually check the attribute types.

Note that `class Module` would have to be updated so that the `Module` fields are defined (note doing this today would also allow type checking on `module.state`,...):

```python
class Module:
  state: State
```

cgarciae:
# What does this PR do?

Fixes #2416. Adds a base class `_ModuleBase` which contains the annotations for `Module`'s fields, this pattern avoids a couple of pitfalls with dataclasses and makes static analysis tools happy. Thanks @Conchylicultor for proposing the solution!

The approach works but adding the `TYPE_CHECKING` check over `__getattr__` opened the lid to a barricade of `pytype` errors, question is do we want to solve all typing errors right now?

### DIFF
--- a/flax/linen/experimental/layers_with_named_axes.py
+++ b/flax/linen/experimental/layers_with_named_axes.py
@@ -238,18 +238,17 @@ def _normalize(mdl: nn.Module, x: Array, mean: Array, var: Array,
   y = x - mean
   mul = lax.rsqrt(var + epsilon)
   if use_scale:
-    scale = mdl.param_with_axes(
-        'scale',
-        scale_init,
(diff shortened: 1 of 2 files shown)


Conchylicultor:
Why was it closed ?

PR #2465, opened 2022-09-14
Use proper scikit-learn dependency

cgarciae:
Depend on `scikit-learn` instead of `sklean` in `docs/requirements.txt`. `sklearn` seems to be some form of mirror for `scikit-learn` but its not the proper dependency.


jheek (approved):

cgarciae:
Merged with master to fix CI.

PR #2476, opened 2022-09-22
Add perturb() to allow capturing intermediate gradients

IvyZX:
Add a perturb() function in nn.Module, which could be used to capture the intermediate gradients inside a module's __call__.

See example in the docstring. There will be a future Colab/notebook with better examples.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1462,6 +1462,50 @@ class Module:
     self.scope.put_variable(col, name, xs)
     return True
 
+  def perturb(self, name: str, value: T, collection: str = 'perturbations') -> T:
+    """Add an zero-value variable ('perturbation') to the intermediate value.
+    
+    The gradient of `value` would be the same as the gradient of this
+    perturbation variable. Therefore, if you define your loss function with
+    both params and perturbations as standalone arguments, you can get the
+    intermediate gradients of `value` by running `jax.grad` on the perturbation
+    argument.
+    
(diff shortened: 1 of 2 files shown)


cgarciae:
Looks good! Should we add a simple test?

IvyZX:
> Looks good! Should we add a simple test?

Test added!

flax/linen/module.py
```diff
@@ -1462,6 +1462,48 @@ def __call__(self, x):
     self.scope.put_variable(col, name, xs)
     return True
 
+  def perturb(self, col: str, name: str, value: T) -> T:
```
levskaya, line 1465: I wonder if we should encourage a default (but still configurable) collection name "perturbations" or some such here...

levskaya (approved):
Looks good as a start.  I still think we should think about whether there are ways to simplify using "perturb" in practice w.o. incurring too much extra memory burden.  We may wish to reserve the right to modify this API as we understand how people will use it - perhaps it's worth marking as "experimental" in docstrings to be clear that we may need to tweak it in the future.

flax/linen/module.py
IvyZX: Fixed both! Made "collections" a keyword arg with default, and modified the docstring.

cgarciae:
@IvyZX @levskaya  we forgot to add this new method to the sphinx page in `docs/api_reference/flax.linen.rst`, its currently not being rendered.

PR #2477, opened 2022-09-22
Only accept None target for unknown types

### LINKED ISSUE #2474: from_state_dict is too lenient
I was checking #1261 and found this pattern:
```python
with open(model_file, "rb") as state_f:
  state = from_bytes(AwesomeModel, state_f.read())
```
However setting `target` to a regular `Module` subclass type in `from_bytes` just results in no checks being performed as state is returned as is for unregistered types:

https://github.com/google/flax/blob/6fd24460e88c833fa481228582937c692e9c182c/flax/serialization.py#L63-L64

Using an unregistered type just gives a false sense of security that validation is happening, it would be more correct to accept `target=None` as a way of opting-out of this check but anything else should require the type to be registered.

cgarciae:
# What does this PR do?

Fixes #2474.  Raises an error in `from_state_dict` for unregistered types.

### DIFF
--- a/flax/serialization.py
+++ b/flax/serialization.py
@@ -61,7 +61,10 @@ def from_state_dict(target, state: Dict[str, Any]):
   else:
     ty = type(target)
   if ty not in _STATE_DICT_REGISTRY:
-    return state
+    if target is None:
+      return state
+    else:
+      raise ValueError(f'No serialization handler for target of type "{ty.__name__}"')
   ty_from_state_dict = _STATE_DICT_REGISTRY[ty][1]
   return ty_from_state_dict(target, state)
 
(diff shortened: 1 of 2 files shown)


cgarciae:
@jheek the use of `from_state_dict` inside `_map_over_modules_in_tree` via `_try_setup` was a bit unexpected, lots of errors arising from this fact, not sure what type of validation we are going with it.

https://github.com/google/flax/blob/45c29559f9c105a6d7e97d218d8557f8c1600716/flax/linen/module.py#L259-L269

cgarciae:
Will try adding a `accept_unregistered_type` flag to fallback.

cgarciae:
Given that `from_state_dict`'s current behaviour is to check on a best-effort basis and its widely used inside the code base for various auxiliary tasks, efforts to fix have felt a bit brittle and since there are no strong issues I feel we should close this PR.

PR #2486, opened 2022-09-27
Fix to use `is_initializing` for init-detection.

yotarok:
# What does this PR do?

The logic used for detecting initialization-mode in `linen.BatchNorm` is currently based on mutability of `params`.
There's a clean approach to do it by `Module.is_initializing` and the previous logic does not work if the user provides `mutable=True`.

## Checklist
      checks if that's the case).
      link).
  - (No GitHub issue posted yet)
  - Covered by the existing tests on `BatchNorm`

### DIFF
--- a/flax/linen/normalization.py
+++ b/flax/linen/normalization.py
@@ -224,8 +224,8 @@ class BatchNorm(Module):
     """Normalizes the input using batch statistics.
 
     NOTE:
-    During initialization (when parameters are mutable) the running average
-    of the batch statistics will not be updated. Therefore, the inputs
+    During initialization (when `self.is_initializing()` is `True`) the running
+    average of the batch statistics will not be updated. Therefore, the inputs
     fed during initialization don't need to match that of the actual input
     distribution and the reduction axis (set with `axis_name`) does not have
     to exist.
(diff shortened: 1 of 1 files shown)


cgarciae:
@levskaya was there a reason not to do the change proposed in this PR when you introduced `is_initializing`?

cgarciae:
@yotarok can you merge/rebase with `main` to fix the tests?

yotarok:
@cgarciae Thanks for the comment. Rebased and force-pushed.

levskaya:
> @levskaya was there a reason not to do the change proposed in this PR when you introduced `is_initializing`?

Mostly I was overwhelmed and dropped it - Since it is a subtle change in semantics we need to test this against existing users to make sure we're not introducing a potentially invisible breaking change...  and probably announce the behavior change.

yotarok:
I agree, once this PR is approved, we (or Flax core devs) should announce the behavior change.
FYI, the issue I described above in PR description is also invisible.  When a user provides `mutable=True` to `apply`, `BatchNorm` silently stops updating the statistics, and it's hard to notice because the return value still contains those non-updated BatchNorm stats, and there's no warning message.

yotarok:
Ping.
Or, should I work on improving coverage for getting this reviewed?

jheek:
@yotarok I'm going to run a large internal test suite to check if this breaks in surprising way.

jheek (approved):

Issue #2493, opened 2022-09-29
nn.Sequential causes persistent GPU arrays across loop iterations


zhangqiaorjc:
A JAX GPU user reported seeing GPU OOM with a simple program that uses nn.Sequential in https://github.com/google/jax/issues/12575

If the repro doesn't use `model = nn.Sequential([nn.Dense(50)]` and does `model = nn.Dense(50)`, then the OOM goes away, and debugging info shows that persistent GPU arrays are gone as well.

nn.Sequential is quite simple, it's unclear what's holding onto the python references

Repro is in the linked JAX issue

zhangqiaorjc:
@levskaya

zhangqiaorjc:
@mattjj and I discovered that this `_caches` is leaking references.

https://github.com/google/flax/blob/main/flax/linen/module.py#L510-L511

cc @jheek as well

levskaya:
closing for now - but we can re-open if this hasn't fixed the original memory issue.

PR #2495, opened 2022-10-02
Adoption cache should use WeakValueDictionary.

levskaya:
We use a cache to preserve sharing-by-reference relationships when "adopting" outside modules into a new top-level functional module context (adoption occurs frequently when using advanced config systems like gin or fiddle, or during interactive development).

Previously we were careful not to hold references to Scope objects by having the cache itself be a WeakKeyDictionary, however the scope-keyed entries were just classic dicts, and this was leading to leaks of tracers and arrays ultimately due to long-lived references to submodule objects from the cache.  By using a WeakValueDictionary to refer to these we avoid these lifetime issues.


levskaya:
Fixes #2493

jheek (approved):
Thanks! Makes sense

PR #2496, opened 2022-10-03
Adds CallSetupUnboundModuleError

### LINKED ISSUE #667: Directly calling `module.setup()` should raise an exception
`flax.linen.Module.setup()` should not be called directly because it needs a `flax.linen.Scope` to be set up properly.

Since #653 there is no more exception risen when a user inadvertently calls `flax.linen.Module.setup()` (though there probably will be error messages if the user tries to access the scope...)

cgarciae:
# What does this PR do?

Fixes #667. Raises a new `CallSetupUnboundModuleError` when trying to call `.setup()` while `scope` is None.

### DIFF
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -538,6 +538,37 @@ class CallCompactUnboundModuleError(FlaxError):
   def __init__(self):
     super().__init__('Can\'t call compact methods on unbound modules')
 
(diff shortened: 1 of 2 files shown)


jheek (approved):
Nice

PR #2499, opened 2022-10-04
Don't create reference cycles among Modules.

levskaya:
Previously, the `parent` attribute of modules used hard references, this frequently leads to reference cycles among modules.  Instead, use a descriptor that automatically wraps the `parent` reference to Modules in a weak reference wrapper to avoid cycles.

This ensures faster refcount-driven memory deallocation of device memory in eager / non-jitted mode, whereas before large sequential allocations could lead to OOM before a gc sweep occurred.

It is also slightly more hygienic in not transiently leaking JAX tracers held by ref-cycles.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -523,6 +523,27 @@ capture_call_intermediates = lambda _, method_name: method_name == '__call__'
 _ParentType = Union[Type['Module'], Type[Scope], Type[_Sentinel], None]
 
 
+class ParentDescriptor:
+  """Wraps parent module references in weak refs.
+
+  This prevents reference cycles from forming via parent links which can lead
+  to accidental OOMs in eager mode due to slow garbage collection as well as
+  spurious tracer leaks during jit compilation.
+
+  Note: "descriptors" are the underlying python mechanism for implementing
+  dynamic @property decorators.  We need to use a raw descriptor instead of the
+  more common decorator in order to force that the appropriate getter/setter
+  logic applies in subclasses even after various dataclass transforms.
+  """
+  def __get__(self, obj, objtype=None):
+    parent = object.__getattribute__(obj, "_parent_ref")
+    return parent() if isinstance(parent, weakref.ReferenceType) else parent
+
+  def __set__(self, obj, value):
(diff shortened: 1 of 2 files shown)


levskaya:
Should fix #2493 even more!

tests/linen/linen_module_test.py
```diff
@@ -1755,5 +1757,31 @@ def __call__(self):
     self.assertFalse(foo.apply({}))
 
 
+class LeakTests(absltest.TestCase):
+
+  def test_tracer_leaks(self):
```
jheek, line 1762: nice test :)

flax/linen/module.py
```diff
@@ -519,6 +519,22 @@ def reimport(self, other: '_ModuleInternalState') -> None:
...
 
+class ParentDescriptor:
+  """Wraps parent module references in weak refs.
+
+  This prevents reference cycles from forming via parent links which can lead
+  to accidental OOMs in eager mode due to slow garbage collection as well as
+  spurious tracer leaks during jit compilation.
+  """
+  def __get__(self, obj, objtype=None):
+    parent = object.__getattribute__(obj, "_parent_ref")
```
jheek, line 530: Isn't this the same as:

```
@property
def parent(self):
  parent = self.__getattribute__(obj, "_parent_ref")
  return parent() if isinstance(parent, weakref.ReferenceType) else parent
 
@parent.setter
def parent(self, value):
   maybe_weak = weakref.ref(value) if isinstance(value, Module) else value
    object.__setattr__(self, "_parent_ref", maybe_weak)
```

This seems simpler than summoning the dark power of Python descriptors?

levskaya: Ha, it would be, but at least for now I have to keep `cls.parent` set to a `dataclasses.field` for the dataclass transform, and _then_ replace the `cls.parent` with a descriptor during `__init_subclass__`.   There's a separate PR that's going to refactor our custom dataclass transform machinery, and I might be able to use `property` instead once that's in.

jheek: Makes sense, let's go with the descriptor for now then

jheek (approved):

PR #2505, opened 2022-10-04
no-op when double wrapping with struct.dataclass

cgarciae:
# What does this PR do?

Makes double wrapping with `struct.dataclass` a no-op when wrapping an existing Flax dataclass. Also raises an error if you try to wrap an existing non-Flax dataclass.

### DIFF
--- a/flax/struct.py
+++ b/flax/struct.py
@@ -95,6 +95,11 @@ def dataclass(clz: _T) -> _T:
   Returns:
     The new class.
   """
+  # check if the class is already a dataclass
+  
+  if '_flax_dataclass' in clz.__dict__:
+    return clz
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -86,6 +86,7 @@ jobs:
         pip install .
         pip install .[testing]
         pip install -r docs/requirements.txt
+        pip install .
```
jheek, line 89: why are there now 2 pip install . lines?

cgarciae: Sorry I added this as I was trying to fix CI. Removed.

jheek (approved):

Issue #2341, opened 2022-07-27
Pip install fails on macOS with Apple M1 and Python >= 3.9


AlexIoannides:
### System information

- **OS Platform and Distribution**: macOS 12.4 (Monterey) running on Apple M1 Max chipset.
- **Flax, jax, jaxlib versions**: Flax version 0.5.3
- **Python version**: `>= 3.9`
- **GPU/TPU model and memory**: NA
- **CUDA version (if applicable)**: NA

### Problem you have encountered:

I cannot install Flax into a virtual environment using Python 3.9 or greater. This problem is **not** encountered when using Python 3.8.

### What you expected to happen:

Flax to be successfully installed in a virtual environment 

### Steps to reproduce:

```text
$ python3.9 -m venv .venv
$ source .venv/bin/activate
$ pip install -U pip
$ pip install flax
```

### Logs, error messages, etc:

```text
Building wheels for collected packages: jax, tensorstore
  Building wheel for jax (setup.py) ... done
  Created wheel for jax: filename=jax-0.3.15-py3-none-any.whl size=1201920 sha256=3b2e1ffc02c289747e698e9f1ad4b6d189e291a1498d4f41f4af16c80c9ba07f
  Stored in directory: /Users/alexioannides/Library/Caches/pip/wheels/59/9d/cb/21506ed546ffa6e339e75a0bb7500d363d68349a89f9c5a080
  Building wheel for tensorstore (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for tensorstore (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [173 lines of output]
      running bdist_wheel
      running build
      running build_py
      creating /var/folders/xk/fqw466sd3l7fztkl08x8hnk80000gn/T/tmpt8sfan0e/lib.macosx-10.9-universal2-cpython-39
      creating /var/folders/xk/fqw466sd3l7fztkl08x8hnk80000gn/T/tmpt8sfan0e/lib.macosx-10.9-universal2-cpython-39/tensorstore
      copying python/tensorstore/cc_test_driver_main.py -> /var/folders/xk/fqw466sd3l7fztkl08x8hnk80000gn/T/tmpt8sfan0e/lib.macosx-10.9-universal2-cpython-39/tensorstore
[...]

PR #2520, opened 2022-10-11
Add Tensorstore back to required dependencies.

IvyZX:
It was reversed back when user reported https://github.com/google/flax/issues/2341, but it was now fixed and verified.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,7 @@ install_requires = [
     "matplotlib",  # only needed for tensorboard export
     "msgpack",
     "optax",
+    "tensorstore",
     "rich>=11.1",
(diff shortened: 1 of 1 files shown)


cgarciae (approved):

wookayin:
tensorstore build fails google/tensorstore#63 for me -- on python 3.10 and 3.11 (Linux). I think this is an optional dependency [for an optional feature](https://github.com/google/flax/blob/5661dca74a84b42f1466b8c0533dd1489fab660d/flax/training/checkpoints.py#L43-L51); can we mark this dep as optional?

wookayin:
tensorstore should be dropped from flax's dependency. Submitted a PR #2692

Issue #2528, opened 2022-10-12
Improve error message when calling `init` or `apply` on a module class instead of a module object.


levskaya:
A common mistake I see (and do myself frequently) is:
```python
class B(nn.Module):
  @nn.compact
  def __call__(self, x):
    return x

k = random.PRNGKey(0)
x = random.uniform(random.PRNGKey(1), (2,))
B.init(k, x)   # B is module class, not B() a module instance
...
B.apply(vs, x)   # similar issue w. apply called on class instead of instance.
```

We could improve the error message here, instead of allowing the first argument to be treated as `self` accidentally, leading to a pretty mysterious error currently.

PR #2529, opened 2022-10-13
Added an error for when we call init, init_with_output and apply on Module class.

chiamp:
Improves error message for this misuse:
```python
class B(nn.Module):
  @nn.compact
  def __call__(self, x):
    return x

k = random.PRNGKey(0)
x = random.uniform(random.PRNGKey(1), (2,))
B.init(k, x)   # B is module class, not B() a module instance
...
B.apply(vs, x)   # similar issue w. apply called on class instead of instance.
```

Addresses #2528

### DIFF
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -569,6 +569,25 @@ class CallSetupUnboundModuleError(FlaxError):
   def __init__(self):
     super().__init__('Can\'t call compact methods on unbound modules')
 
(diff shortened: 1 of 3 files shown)


cgarciae:
Hey @chiamp run:

```
pip install pre-commit
pre-commit install
pre-commit run --all-files
```
To fix pre-commit issues

PR #2540, opened 2022-10-17
add rng_collection argument to Dropout

### LINKED ISSUE #2194: Make RNG name configurable in Dropout
Currently `'dropout'` is hard-coded as an argument of `make_rng` inside `Dropout`. However, when implementing support for "recurrent dropout"  in an LSTMCell or similar you need two kinds of dropout:
1. A regular dropout which is applied to the inputs with a different mask at each step.
2. A "recurrent dropout" that is applied to the state with the same mask at each step.

To implement `2` a possibility is to set the RNG name to `'recurrent_dropout'` on the Dropout layer applied to the state and guarantee that each step uses the same random state. From `nn.scan`'s perspective the would look like this:

```python
nn.scan(..., split_rngs={'dropout': True, 'recurrent_dropout': False})
```
The proposal is to add an `rng_name` (or similar) attribute to `Dropout` so we are able support these kind of use-cases. The alternative would be to create a separate `RecurrentDropout` layer with the same code but different hard-coded value.

cgarciae:
# What does this PR do?

Fixes #2194. Parametrizes the rng collection name, defaults to `dropout` but it can now be specified via a constructor argument.

### DIFF
--- a/flax/linen/stochastic.py
+++ b/flax/linen/stochastic.py
@@ -38,10 +38,12 @@ class Dropout(Module):
       deterministic: if false the inputs are scaled by `1 / (1 - rate)` and
         masked, whereas if true, no mask is applied and the inputs are returned
         as is.
+      rng_collection: the rng collection name to use when requesting an rng key.
   """
   rate: float
   broadcast_dims: Sequence[int] = ()
   deterministic: Optional[bool] = None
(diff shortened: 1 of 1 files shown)


jheek (approved):
Looks good!

8bitmp3:
👍  @cgarciae

PR #2563, opened 2022-10-28
Update getting_started.ipynb and getting_started.md

chiamp:
Replaced jax.value_and_grad with jax.grad in section 7 since the outputted loss value wasn't being used.


cgarciae (approved):
Looks good!

Issue #2577, opened 2022-11-02
Transpose convolution matching convolution with given kernel parameters


mathisgerdes:
In the current implementation, it appears to be quite challenging to compute the transpose convolution that actually corresponds to the *transpose* of a convolution *with given kernel weights*, at least when `padding='CIRCULAR'`.

In most cases, one can already achieve this by manually changing the kernel weights (flipping spacial dimensions and swapping channel axes for the kernel weights). This would arguably solve the issue. However, if `padding='CIRCULAR'`, this doesn't work since incompatible alignment conventions for the kernel are used.

I suppose this is a somewhat unusual situation since it only applies if parameters are specified externally.

## Motivation
Consider n-dimensional convolution. Omitting channel and batch dimensions, convolution is a linear operation and can generally be written as $A = WB$. (For simplicity, $B$ can be thought of as the flattened input, in which case $W$ is a matrix.) 
The transpose operation should then satisfy $\langle A, WB \rangle = \langle W^T A, B \rangle$.

For `jax.lax.conv_transpose` this is true if the parameter `transpose_kernel` is set to `True`.
In code, that corresponds to:
```python
IC, OC = 2, 3  # channels
k1, k2, k3 = jax.random.split(jax.random.PRNGKey(0), 3)

b = jax.random.normal(k1, (1, 30, 30, IC))
w = jax.random.normal(k2, (5, 5, IC, OC))
a = jax.random.normal(k3, (1, 26, 26, OC))

dim_num = ('NHWC', 'HWIO', 'NHWC')
w_b = jax.lax.conv_general_dilated(b, w, (1, 1), 'VALID', dimension_numbers=dim_num)

wt_a = jax.lax.conv_transpose(a, w, (1, 1), 'VALID', dimension_numbers=dim_num,
                              transpose_kernel=True)
# this corresponds to the scalar product
np.isclose(np.sum(a * w_b), np.sum(wt_a * b))  # True
```
This should ideally be reproducible with the `ConvTranspose` layer.
In this case, it is possible to manually modify the kernel weights:
```python
conv = nn.Conv(OC, (5, 5), (1, 1), 'VALID', use_bias=False)
conv_tr = nn.ConvTranspose(IC, (5, 5), (1, 1), 'VALID', use_bias=False)

params = conv.init(k2, b)
# manually transpose kernel
kernel = params['params']['kernel']
kernel = jnp.flip(kernel, (0, 1)).swapaxes(-2, -1)
[...]

PR #2578, opened 2022-11-02
Add transpose_kernel argument to ConvTranspose.

mathisgerdes:
This is a proposal for #2577, adding a `transpose_kernel` option to `ConvTranspose` such that the previous behavior is maintained by default.

- Add an additional `transpose_kernel` parameter to `ConvTranspose` which is passed to `jax.lax.conv_transpose`
- For  `padding='CIRCULAR'`: To maintain the previous convention of `ConvTranspose`, the alignment of the kernel is changed depending on `transpose_kernel`.
   ```python
   # Divide the padding equally between left and right. The choice to put
   # "+1" on the left (or on the right) represents a convention for
   # aligning even-sized kernels.
   if self.transpose_kernel:
     # match convention in Conv to achieve the proper transpose
     total_pad = [
        (size_diff // 2, (size_diff + 1) // 2) for size_diff in size_diffs
     ]
   else:
     # keep the previous convention for backward compatibility
     total_pad = [
        ((size_diff + 1) // 2, size_diff // 2) for size_diff in size_diffs
     ]
   ```

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -583,6 +583,8 @@ class ConvTranspose(Module):
       for details.
     kernel_init: initializer for the convolutional kernel.
     bias_init: initializer for the bias.
+    transpose_kernel: if True flips spatial axes and swaps the input/output
+      channel axes of the kernel.
   """
   features: int
   kernel_size: Union[int, Tuple[int, ...]]
@@ -596,6 +598,7 @@ class ConvTranspose(Module):
   precision: PrecisionLike = None
   kernel_init: Callable[[PRNGKey, Shape, Dtype], Array] = default_kernel_init
   bias_init: Callable[[PRNGKey, Shape, Dtype], Array] = zeros
+  transpose_kernel: bool = False
 
   @compact
   def __call__(self, inputs: Array) -> Array:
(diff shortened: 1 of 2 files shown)


flax/linen/linear.py
```diff
@@ -688,12 +695,17 @@ def __call__(self, inputs: Array) -> Array:
           -(y_dim - x_dim) % (2 * x_dim)
           for y_dim, x_dim in zip(y.shape[1:-1], scaled_x_dims)
       ]
-      # Divide the padding equaly between left and right. The choice to put
+      # Divide the padding equally between left and right. The choice to put
```
cgarciae, line 698: Maybe we could move this comment to the `else` branch and have a similar one for the positive branch?

mathisgerdes: I agree, the comments there could be clearer. How about this modified version, with clearer comments for both cases?

cgarciae (approved):
Thanks @mathisgerdes for doing this! Exposing `transpose_kernel` is a great idea.

marcvanzee:
@mathisgerdes @cgarciae this PR has build failures so I have removed the "pull ready" label. Could you please fix those first?

mathisgerdes:
> @mathisgerdes @cgarciae this PR has build failures so I have removed the "pull ready" label. Could you please fix those first?

I believe those were problems with the previous commit, not any changes here. I've rebased to the latest commit.

mathisgerdes:
Is there still something to be done here, @cgarciae, @andsteing?

cgarciae (approved):
Sorry for the delay. LGTM. 
Thanks for doing this!

PR #2586, opened 2022-11-04
Add gfile api shim to remove tensorflow dependency for basic IO.

chiamp:
Taking over #2073
Added a shim for tensorflow.io.gfile with io.py to remove tensorflow dependency.


levskaya (changes requested):

levskaya (approved):
lgtm!

flax/io.py
```diff
@@ -0,0 +1,176 @@
...
+
+io_mode = None
+gfile = None
+
+if importlib.util.find_spec('tensorflow'):
+  from tensorflow.io import gfile  # pytype: disable=import-error
+  io_mode = BackendMode.TF
+else:
+  logging.warning("Tensorflow library not found, tensorflow.io.gfile "
+                  "operations will use slower native shim calls.")
```
andsteing, line 45: Are native operations really slower than tensorflow?
Maybe worth pointing out that any access to GCS (i.e. "gs://" paths) will fail. Being explicit about this failure mode might also be a good idea inside below functions (e.g. when a program runs fine in one environment, but then crashes in another environment, it might not be obvious to the user that the 2nd invocation has crashed because it was trying to access a "gs://" path in an environment that did not have Tensorflow installed).

chiamp: I'm unsure on how the speed of native operations compare to tensorflow; maybe @levskaya can jump in.
As for GCS, yes good point, I'll add a disclaimer in the warning message.

chiamp: Created a new issue #2624

PR #2593, opened 2022-11-08
Add new 🔪 Flax - The Sharp Bits 🔪 Dropout and randomness

8bitmp3:


IvyZX (commented):
Thanks for putting this together!

In the example code, we also show that user should split and feed a different RNG for each step to randomize the dropout?

IvyZX (approved):

AnirudhDagar:
This is a great addition, thanks!

PR #2594, opened 2022-11-08
Remove Mypy type errors

### LINKED ISSUE #688: Run mypy in run_all_tests.sh
Ideally we want both type checkers to work correctly at least for the library. Also mypy should offer a speedup

_Originally posted by @jheek in https://github.com/google/flax/discussions/685#discussioncomment-140141_

zaxtax:
# What does this PR do?

This PR makes the annotation changes necessary so that mypy checks would all pass under CI

Fixes #688

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -28,7 +28,7 @@ from .frozen_dict import unfreeze
 import jax
 from jax import random
 from .scope import (CollectionFilter, DenyList, PRNGSequenceFilter,  # pylint: disable=g-multiple-import
-                    Scope, group_collections, in_filter,
+                    Filter, Scope, group_collections, in_filter,
                     intersect_filters, is_filter_empty, subtract_filters,
                     union_filters)
 
@@ -148,7 +148,7 @@ def pack(fn: Callable[..., Any],
       for inner_scope in inner_scopes:
         inner_scope.invalidate()
       inner_scopes = []
-      mutable = False
+      mutable: Filter = False
       for out_filter in out_variable_filters:
         mutable = union_filters(mutable, out_filter)
       # could be () in the edge case where no rngs or variable_groups are lifted
@@ -1264,8 +1264,8 @@ def jit(fn: Callable[..., Any],
   # Close over scope_fn & repack_fn to avoid recompilation
   # this is impure but we use the fingerprint arg to differentiate between cases
   # where scope_fn or repack_fn actually produce non-identical results.
-  scope_fn = None  # type: Callable
-  repack_fn = None  # type: Callable
+  scope_fn = None  # type: Optional[Callable]
+  repack_fn = None  # type: Optional[Callable]
   @functools.partial(jax.jit,
                      static_argnums=static_argnums,
                      donate_argnums=donate_argnums,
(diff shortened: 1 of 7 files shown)


flax/core/scope.py
```diff
@@ -619,8 +619,9 @@ def is_collection_empty(self, col: str) -> bool:
   def _mutable_collection(self, col: str) -> MutableCollection:
     """Returns the collection `col` as a mutable object."""
     assert self.is_mutable_collection(col), f'Collection {col} is not mutable'
+
     if col not in self._variables:
-      if self.parent:
+      if self.parent and self.name is not None:
```
cgarciae, line 624: I think its safer to use an assert over `name` so the runtime behaviour remains the same:
```python
if self.parent:
  assert self.name is not None
  ...
```

flax/core/scope.py
```diff
@@ -837,7 +838,7 @@ def bind(variables: VariableDict,
 
 def apply(fn: Callable[..., Any],
           mutable: CollectionFilter = False,
-          flags: Optional[Dict] = None) -> Callable[..., Any]:
+          flags: Optional[FrozenDict] = None) -> Callable[..., Any]:
```
cgarciae, line 841: We seem to accept both `FrozenDict` and `Dict`, probably the correct type is either `Union[Dict, FrozenDict]` or even `Mapping`.

zaxtax: I think changing to `Mapping` is best

flax/core/scope.py
```diff
@@ -632,7 +634,7 @@ def _mutable_collection(self, col: str) -> MutableCollection:
   def _collection(self, col: str) -> Collection:
     """Returns a collection of variables of collection `col`."""
     if col not in self._variables:
-      if self.parent:
+      if self.parent and self.name is not None:
```
cgarciae, line 637: Same issue as before, lets use an assert here.

levskaya:
I'm seeing a ton of failures internally that all seem like duplicates of:
```
Base signature: 'def builtins.object.__dir__(self) -> List[str]'.
  Subclass signature: 'def flax.linen.module.Module.__dir__(self) -> Iterable[str]'.
  Return type mismatch.
```

cgarciae:
Thanks @levskaya! Makes sense, we'll revert that.

zaxtax:
> I'm seeing a ton of failures internally that all seem like duplicates of:
> 
> ```
> Base signature: 'def builtins.object.__dir__(self) -> List[str]'.
>   Subclass signature: 'def flax.linen.module.Module.__dir__(self) -> Iterable[str]'.
>   Return type mismatch.
> ```

Thanks for the heads-up. If it's at all possible, @levskaya is there a minimal example that triggers the error you can share? @cgarciae does that seem like a reasonable request? Since, I think if that's the only issue, we might be able to not revert too much and still get something modest merged in.

cgarciae:
@zaxtax oh, I meant just revert the signature for `Module.__dir__`'s output from `Iterable` to `List`.

flax/core/lift.py
```diff
@@ -1264,8 +1264,8 @@ def jit(fn: Callable[..., Any],
   # Close over scope_fn & repack_fn to avoid recompilation
   # this is impure but we use the fingerprint arg to differentiate between cases
   # where scope_fn or repack_fn actually produce non-identical results.
-  scope_fn = None  # type: Callable
-  repack_fn = None  # type: Callable
+  scope_fn = None  # type: Optional[Callable]
```
levskaya, line 1267: Curious: why aren't we using inline actual type annotations here?

Issue #2412, opened 2022-08-23
Test TransformedMethodReturnValueError for a returned Variable


melissatan:
The linen_transforms_test file contains a unit test for TransformedMethodReturnValueError when a Module is returned (`test_returned_module_warning`), but doesn't yet cover the scenario where a Variable is returned.

PR #2610, opened 2022-11-14
Added test to check for Variable warning.

chiamp:
Addresses #2412

Added a test to check for jitting a function that returns a Variable type.


IvyZX (approved):
Thanks for adding this!

Issue #2624, opened 2022-11-17
Add explicit warning if flax.io is using default Python I/O operations


chiamp:
Are native operations really slower than tensorflow?
Maybe worth pointing out that any access to GCS (i.e. "gs://" paths) will fail. Being explicit about this failure mode might also be a good idea inside below functions (e.g. when a program runs fine in one environment, but then crashes in another environment, it might not be obvious to the user that the 2nd invocation has crashed because it was trying to access a "gs://" path in an environment that did not have Tensorflow installed).

_Originally posted by @andsteing in https://github.com/google/flax/pull/2586#discussion_r1023606713_

PR #2625, opened 2022-11-17
Added explicit warning if flax.io is using default Python I/O operations

chiamp:
Addresses #2624

Added explicit warning if flax.io is using default Python I/O operations.

### DIFF
--- a/flax/io.py
+++ b/flax/io.py
@@ -42,7 +42,8 @@ if importlib.util.find_spec('tensorflow'):
   io_mode = BackendMode.TF
 else:
   logging.warning("Tensorflow library not found, tensorflow.io.gfile "
-                  "operations will use slower native shim calls.")
+                  "operations will use native shim calls. "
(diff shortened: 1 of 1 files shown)


flax/io.py
```diff
@@ -42,7 +42,8 @@ class BackendMode(Enum):
   io_mode = BackendMode.TF
 else:
   logging.warning("Tensorflow library not found, tensorflow.io.gfile "
-                  "operations will use slower native shim calls.")
+                  "operations will use slower native shim calls. "
```
andsteing, line 45: So, did you verify that native calls are actually *slower* ?

levskaya: It's more that native calls can't support GCS / other networking file systems out of the box, so IF tensorflow is installed, we'd like to just use them via the tf.io.gfile support.

chiamp: I've removed the mention of native shim calls being slower in the warning message

PR #2631, opened 2022-11-18
Add Flax Dropout tutorial

8bitmp3:
@levskaya @IvyZX @chiamp @marcvanzee

### DIFF
--- a/docs/tutorials/dropout_tutorial.md
+++ b/docs/tutorials/dropout_tutorial.md
@@ -12,8 +12,6 @@ kernelspec:
   name: python3
 ---
 
-+++ {"id": "P9WN5UZzRqkE"}
-
 # Dropout in image classification
 
 [![Open in Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/google/flax/blob/main/docs/tutorials/dropout_tutorial.ipynb)
@@ -22,8 +20,6 @@ This tutorial provides an end-to-end example of a simple image classification mo
 
 The example below uses a lot of fundamental concepts covered in [Getting started](https://flax.readthedocs.io/en/latest/getting_started.html). If you're new to Flax, start there.
 
-+++ {"id": "NpnTy21qSWYV"}
-
 ## Setup
 
 - Install/upgrade Flax, which will also set up [Optax](https://optax.readthedocs.io/) (for common optimizers and loss functions), and JAX.
@@ -31,14 +27,10 @@ The example below uses a lot of fundamental concepts covered in [Getting started
 - Import the necessary libraries.
 
 ```{code-cell}
-:id: 4Q1sL4cQTbAt
-
 !pip install --upgrade -q flax tensorflow_datasets 
 ```
 
 ```{code-cell}
-:id: 1kYKJYxxXUiL
-
 import jax
 import jax.numpy as jnp                # JAX NumPy
 
@@ -50,19 +42,13 @@ import optax                           # The Optax library
 import tensorflow_datasets as tfds     # TFDS for the dataset
 ```
 
-+++ {"id": "xTf7YtdgWWNY"}
-
 Use a JAX PRNG key and split it to get one key for parameter initialization, and another one for dropout randomness:
 
 ```{code-cell}
-:id: G9lWcsK1WUzL
-
 seed = 0
 root_key = jax.random.PRNGKey(seed=seed)
 main_key, params_key, dropout_key = jax.random.split(key=root_key, num=3)
 ```
-
-+++ {"id": "P9pSTAIOSU1X"}
 
 Create a simple Flax model, subclassed from [Flax `Module`](https://flax.readthedocs.io/en/latest/guides/flax_basics.html#module-basics). Note that:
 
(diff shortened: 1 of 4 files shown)


IvyZX (changes requested):
For the `dropout_key` in `train_loop`, I believe you need to split a different key on each step so that different activation happens. So you should split `dropout_key` into two keys, use one for this step, and return the other so that the next step can take it as an input (and do more splitting). 

Other than this the doc looks good! I like the part that explains every piece of code cleanly.

andsteing:
(removed myself as a reviewer because I don't have time to properly review this ... maybe @marcvanzee can have a look instead? )

docs/tutorials/dropout_tutorial.md
```diff
@@ -0,0 +1,234 @@
...
+
+Create a simple Flax model, subclassed from [Flax `Module`](https://flax.readthedocs.io/en/latest/guides/flax_basics.html#module-basics). Note that:
+
+- To add a dropout layer in Flax, use [`flax.linen.Dropout`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.Dropout.html#flax.linen.Dropout).
+- In `flax.linen.Dropout`, the `deterministic` argument is `None` by default. If `false`, the inputs are scaled by `1 / (1 - dropout_rate)` and masked. When it's `true`, no mask is applied (the dropout is turned off).
+
+```{code-cell}
+# A simple convolutional network with a dropout layer.
+class CNN(nn.Module):
+  training: bool
```
cgarciae, line 61: `training`  (or `train` as its most commonly used) should be an argument to `__call__`, else you cannot modify the training state via `TrainState.apply_fn`.

docs/tutorials/dropout_tutorial.md
```diff
@@ -0,0 +1,234 @@
...
+- The model constructor argument should be set to `training=True`.
+
+```{code-cell}
+@jax.jit
+def train_step(state, batch):
+  """Train for a single step."""
+  def loss_fn(params):
+    # Perform the forward pass with `flax.linen.apply()`.
+    # Use the `dropout_key` for the `rngs` argument.
+    logits = CNN(training=True).apply({'params': params}, batch['image'], rngs={'dropout': dropout_key})
```
cgarciae, line 142: If the previous is addressed this could be written as:
```python
logits = state.apply_fn(
  {'params': params}, batch['image'], train=True, rngs={'dropout': dropout_key})
```

docs/tutorials/dropout_tutorial.md
```diff
@@ -0,0 +1,234 @@
...
+  metrics = compute_metrics(logits=logits, labels=batch['label'])
+  return state, metrics
+```
+
+Write the evaluation step function. Remember to set the model constructor argument to `training=false`.
+
+```{code-cell}
+@jax.jit
+def eval_step(params, batch):
+  logits = CNN(training=False).apply({'params': params}, batch['image'])
```
cgarciae, line 159: Same situation but `train=False`.

marcvanzee:
> (removed myself as a reviewer because I don't have time to properly review this ... maybe @marcvanzee can have a look instead? )

Sure, added myself as a reviewer.

marcvanzee:
Thanks for adding this @8bitmp3!

I have some questions / remarks:
* It seems RTD does not build due to a time out, which is because you are doing a full training run in the tutorial. The code blocks are ran as a part of our CI on CPU, and we try to keep them short.

* More generally, I am a bit worried that our documentation for Dropout is getting somewhat unwieldy now:
    * This guides explains Dropout in an end-to-end example
    * The "sharp bits" also explains Dropout in detail (but still contains some mistakes)

I was actually thinking that we would create a single "Using Dropout" guide, similar to @cgarciae 's ["Using BatchNorm"](https://flax--2536.org.readthedocs.build/en/2536/guides/batch_norm.html). In this guide we would show how one goes about adding Dropout to their model, with side-by-side diffs showing which parts have to be changed.

So therefore, before we start writing any new documentation for Dropout, I think we should replace the "sharp bits" content with a guide where we explain how to do Dropout. Once we are happy with that we can start thinking of more content for Dropout.

It is of course great you are writing these end-to-end tutorials proactively, similar to the one you are proposing in [Variational autoencoder with a learning rate schedule](https://github.com/google/flax/pull/2650/), but I suggest you first discuss what content you are planning to put in these tutorials, otherwise you might loose a lot of time writing them if it turns out we want to have other content.

Anyway, let's chat more offline!

docs/tutorials/dropout_tutorial.md
8bitmp3: Thanks @cgarciae 👍 

Just to confirm, do you mean?

```
  def __call__(self, x, train = True):
```

cgarciae: Yes! Usually just `train: bool` with no default value to be very explicit.

cgarciae: Yes! Usually just `train: bool` with no default value to be very explicit.

8bitmp3:
> The "sharp bits" also explains Dropout in detail (but still contains some mistakes)

Reviewed by @levskaya but we can revise it again.

> This guides explains Dropout in an end-to-end example

It's an end-to-end tutorial that uses dropout with a particular task in mind, like classification.

> I was actually thinking that we would create a single "Using Dropout" guide,

Let's do this! 👍 

> suggest you first discuss what content you are planning to put in these tutorials, otherwise you might loose a lot of time writing them if it turns out we want to have other content.

This doc was presented to the team 2 weeks ago.

8bitmp3:
Publishing a dropout _guide_, closing this PR. Discussed with marcvanzee 👍

PR #2623, opened 2022-11-17
WIP: Add path tracking for better deserialization error messages.

levskaya:
Quick hack to show better path hints in checkpoint deserialization error messages.

PR #2632, opened 2022-11-18
Added path discrepancy details in serialization errors

chiamp:
Taking over #2623
If there's a discrepancy between the structure of the target and state dict when serializing, a ValueError will be raised. This PR adds information on what the path is where this discrepancy occurred when the ValueError is raised.


flax/serialization.py
```diff
@@ -124,7 +153,12 @@ def _dict_state_dict(xs: Dict[str, Any]) -> Dict[str, Any]:
 
 
 def _restore_dict(xs, states: Dict[str, Any]) -> Dict[str, Any]:
-  return {key: from_state_dict(value, states[str(key)])
+  diff = set(map(str, xs.keys())).difference(states.keys())
```
levskaya, line 156: we might need to be careful about being too strict with dict restores in cases of partial restorations

flax/core/frozen_dict.py
```diff
@@ -220,8 +220,13 @@ def _frozen_dict_state_dict(xs):
 
 
 def _restore_frozen_dict(xs, states):
+  diff = set(map(str, xs.keys())).difference(states.keys())
```
levskaya, line 223: we might need to be careful about being too strict with dict restores in cases of partial restorations

levskaya (approved):
Two comments - we just need to make sure we're not accidentally making checkpoint imports too strict.

Issue #2620, opened 2022-11-15
Use new CI runners


cgarciae:
Use larger test runners provided by Google. Figure out how to config `build.yaml` to use them.
@levskaya says:
> I think it's a one-line change
something like "runs on" directive
it's in the docs somewhere but I've forgotten the line

cgarciae:
Additionally, run `pytype` tests in parallel via bash sub-processes to leverage the additional cores.

marcvanzee:
Fixed by #2635

PR #2635, opened 2022-11-19
Speed up Github Actions CI

levskaya:
Also enable more parallelism in CI tests.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -28,7 +28,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python 3.8
-        uses: actions/setup-python@v3
+        uses: actions/setup-python@v4
         with:
           python-version: 3.7
       - uses: pre-commit/action@v2.0.3
@@ -36,7 +36,7 @@ jobs:
     name: Check commit count
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     # We allow at most 5 commits in a branch to ensure our CI doesn't break.
     - name: Check commit count in PR
       if: always()
@@ -65,9 +65,9 @@ jobs:
       matrix:
         python-version: [3.7, 3.8, 3.9]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install standalone dependencies only
@@ -79,7 +79,7 @@ jobs:
   tests:
     name: Run Tests
     needs: [cancel-previous, pre-commit, commit-count, test-import]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04-16core
     strategy:
       matrix:
         python-version: [3.7, 3.8, 3.9]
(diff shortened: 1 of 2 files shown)


levskaya:
So this shaves off some time (maybe 3-5 min on average?) from our CI tests by using a larger 16-core test-runner, esp. due to faster pytype runs.

However we spend about half the time for all the pytest/doctest/pytype runs on just re-installing dependencies.  I've tried caching the downloaded packages, but I feel like we might save more time by caching the installed packages - that would save another 2-3 min per run.  See this [interesting comment](https://github.com/actions/setup-python/issues/330#issuecomment-1126638275) about caching the _installed_ packages... maybe we should think about doing something like that to speed this up further.

Side note: We can also soon drop the 3.7 tests in favor of 3.10 tests.  We may also want to add a mypy typecheck job as well.

levskaya:
Addresses #2620

levskaya:
NB: for caching e.g. a full virtualenv directory, the usual recommendation is to hash on the requirements.txt file, and to pin versions so as to be specific...  however that's annoying - maybe we can invalidate caches every week using the date instead.

levskaya:
OK I updated this to directly cache the virtual-environment directory _post-installation_ so we don't waste any time installing all of our _extensive_ set of deps for every single test.  As part of the cache key I added the week-number and year, so that the caches will automatically invalidate every week, so that we're testing our unpinned dependencies at their newest versions.  (Note that caches can be manually deleted under actions -> management -> caches)

Github supports 10G of caches per repo, our installed venv comes in at ~2.2Gb currently.

Note to self: I just need to make sure that the flax source itself isn't being cached the venv without updating / egg-linking.

So far the faster runners and caching drop our CI testing time: 17-19min  --> 6m 15s

levskaya:
The last major issue is that the codecov uploader can often take up to 1 minute to upload... this is horrible, and I don't think delaying every single CI test by ~30-60s is really worth it for a simple coverage indicator that none of us even probably look at.

.github/workflows/build.yml
```diff
@@ -90,24 +90,40 @@ jobs:
...
-    - name: Install dependencies
+    - name: Get week and year
+      id: week_year
+      run: echo "DATE=$(date +%U-%Y)" >> $GITHUB_OUTPUT
+    - name: Cached virtual environment
+      id: venv_cache
+      uses: actions/cache@v3
+      with:
+        path: venv
+        key: pip-${{ steps.setup_python.outputs.python-version }}-${{ steps.week_year.outputs.DATE }}-${{ hashFiles('**/requirements.txt', 'setup.py') }}
```
andsteing, line 107: What's the recommended way to launch a Github build that forces an uncached environment?
(This might be useful to test dependency interplay at HEAD).

levskaya: You can just delete the caches from the Actions submenu.  We could also probably figure out some way to signal against using caches, but in practice just deleting them is probably enough.

marcvanzee (approved):
Nice, thanks for adding this! Out of curiosity: do you have an idea how much of the speed improvement is due to the caching and how much is due to the improved pytype testing parallelism?

levskaya:
> Nice, thanks for adding this! Out of curiosity: do you have an idea how much of the speed improvement is due to the caching and how much is due to the improved pytype testing parallelism?

Kinda 50% / 50% - pytype is slow and really benefits from a larger machine (as do the rest of the tests) - but we were also wasting >3min or more for every test runner re-installing everything from scratch.

Issue #2656, opened 2022-11-24
Mac M1 chip does not support 'float128', 'f16', 'complex256' and 'c32' dtypes


chiamp:
Mac M1 chip does not support `float128`, `f16`, `complex256` and `c32` dtypes, which are used by the test function `test_numpy_serialization` in `tests/serialization_test.py`

PR #2657, opened 2022-11-24
Added check for Mac M1 chip, when conducting serialization test.

chiamp:
Addresses #2656

### DIFF
--- a/tests/serialization_test.py
+++ b/tests/serialization_test.py
@@ -14,8 +14,10 @@
 
 """Tests for flax.struct and flax.serialization."""
 
+import platform
 import collections
-from typing import Any, NamedTuple
+from typing import NamedTuple, Any
+import pytest
 
 from absl.testing import absltest
 from absl.testing import parameterized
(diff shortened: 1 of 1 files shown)


levskaya (approved):

tests/serialization_test.py
```diff
@@ -173,6 +174,9 @@ def test_numpy_serialization(self):
                      'complex', 'bool']
     np.random.seed(0)
     for dtype in normal_dtypes:
+      if (dtype in {'float128', 'f16', 'complex256', 'c32'}) and (platform.system() == 'Darwin') and (platform.machine() == 'arm64'):
```
marcvanzee, line 177: Perhaps we should print a warning so the user is aware that these tests are being skipped?

marcvanzee (changes requested):

tests/serialization_test.py
cgarciae: Maybe we should use the `pytest.mark.skipif` decorator over the test?

marcvanzee (approved):
Looks good, thanks!

PR #2665, opened 2022-11-28
Use a different rng key for each batch element in DenseGeneral init

j-towns:
This fixes a very minor issue: different elements of the batch of params in a batched DenseGeneral were initializing to the same value, because a JAX rng was being reused, without doing `jax.random.split`.

After this fix Dense and DenseGeneral necessarily initialize with different parameters, so I had to weaken `test_dense_is_dense_general` to only check that the result of _applying_ is identical between Dense and DenseGeneral, i.e. the init params are allowed to differ.

## Checklist
- [x] This PR fixes a minor issue (e.g.: typo or small bug) or improves the docs.

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -27,6 +27,7 @@ from flax.linen.module import Module
 from flax.linen.dtypes import promote_dtype
 from jax import eval_shape
 from jax import lax
+from jax import random
 from jax import ShapedArray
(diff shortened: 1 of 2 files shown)


cgarciae:
Thanks @j-towns for looking into this. I don't have enough context on the current implementation, @levskaya should give a proper review.

levskaya (approved):
thanks!

PR #2668, opened 2022-11-29
Improve and migrate model_surgery guide

IvyZX:
This PR does the following:

- Add a bunch more texts explaining why and how to do param and opt state tweaking to readers with little context of Flax terminologies.
- Explained the `FrozenDict` a bit more.
- Make the guide notebook-based, so that it's trivial to open a colab and test the code.


gnecula:
It is a good idea to transition docs from .rst to Colab+md files. Thanks!

docs/guides/model_surgery.ipynb
```diff
@@ -0,0 +1,412 @@
+{
```
levskaya, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    flat_params = traverse_util.flatten_dict(params)
</pre><p>I'd consider using the " sep='/' " keyword argument to make the flattened keys simple strings as opposed to tuples, which is usually nicer for most users</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/2668/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

levskaya: I'd consider doing this across the entire doc using the concatenated string "foo/bar/baz" flattened form.

8bitmp3:
> It is a good idea to transition docs from .rst to Colab+md files. Thanks!

+ 1 @gnecula We have some docs that use tables with before/after code. Unless we find a Markdown-way of implementing this without any (or too many) (external) Markdown "themes" for Sphinx, we can keep them for now. There may also be a way of doing this in HTML embedded in Markdown 🤔 (we do this in TensorFlow . org)

@IvyZX Thanks for the PR. Will review asap.

8bitmp3:
> It is a good idea to transition docs from .rst to Colab+md files. Thanks!

(Note: In some Markdown flavors, code diff highlighting can be done by adding "diff" (similar to adding "python" if you want the code to have python syntax colors). Will investigate if this works in Sphinx for ReadTheDocs. cc @cgarciae You can use it in GitHub comments that support Markdown (during reviews). Reference: https://stackoverflow.com/questions/40883421/diff-syntax-highlighting-in-github-markdown )

marcvanzee:
In my opinion, the Github diff highlighting is quite a lot worse than what we have now. We currently have a side-by-side view, which I think is very effective to demonstrate the effect of a change (see e.g. @cgarciae's [BatchNorm guide](https://flax.readthedocs.io/en/latest/guides/batch_norm.html)). The diff view with green and red is not so easy to read, I think.

I think these side-by-side views do have a benefit so I would be in favor of keeping them as .rst for now. I do agree that if we don't use these side-by-side views in a doc we should switch to notebooks.

docs/guides/model_surgery.ipynb
IvyZX: That's a good idea. Though it worth mentioning that people can use tuples as key - which is slightly more error-proof because it allows '/' as part of the dict keys. I will add it in this doc.

IvyZX:
I agree that the side-by-side views look pretty convenient and we should keep them as .rst, until we find a good alternative. The doc on this PR doesn't have the side-by-side views so it could be safely converted.

docs/guides/model_surgery.md
```diff
@@ -0,0 +1,140 @@
...
+Model Surgery
+==============================
+
+Usually, Flax modules and optimizers track and update the params for you. But there may be some time when you want to do some model surgery and tweak the param tensors yourself. This guide shows you how to do the trick.
+
+
+## Setup
+
+```python
+!pip install --upgrade -q pip jax jaxlib flax
```
cgarciae, line 25: This type of cells should be tagged with `skip-execution` to avoid installing dependencies on CI (could alter tests). See [myst-nb cell-tags](https://myst-nb.readthedocs.io/en/latest/configuration.html#cell-tags).

docs/guides/model_surgery.md
```diff
@@ -0,0 +1,140 @@
...
+      x = nn.relu(x)
+      x = nn.avg_pool(x, window_shape=(2, 2), strides=(2, 2))
+      x = x.reshape((x.shape[0], -1))
+      x = nn.Dense(features=256)(x)
+      x = nn.relu(x)
+      x = nn.Dense(features=10)(x)
+      x = nn.log_softmax(x)
+      return x
+
+def get_initial_params(rng):
```
cgarciae, line 65: Should we use the `key` variable name here as well?

8bitmp3:
After @cgarciae and @levskaya are finished with their reviews, I will add my final suggestions.

docs/guides/model_surgery.md
IvyZX: Done. This is such a useful feature - we should make sure to have it on every doc.

docs/guides/model_surgery.md
IvyZX: Done.

chiamp:
I think the FrozenDict explanation is very helpful, thanks for adding it!

levskaya (approved):

PR #2670, opened 2022-11-29
Include mypy in run_all_tests

zaxtax:
This continues work on #688 to include mypy in testing

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -108,7 +108,7 @@ class FrozenDict(Mapping[K, V]):
       self._hash = h
     return self._hash
 
-  def copy(self, add_or_replace: Mapping[K, V]) -> 'FrozenDict[K, V]':
+  def copy(self, add_or_replace: 'FrozenDict[K, V]') -> 'FrozenDict[K, V]':
     """Create a new FrozenDict with additional or replaced entries."""
     return type(self)({**self, **unfreeze(add_or_replace)})
 
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -108,9 +108,9 @@ class FrozenDict(Mapping[K, V]):
       self._hash = h
     return self._hash
 
-  def copy(self, add_or_replace: 'FrozenDict[K, V]') -> 'FrozenDict[K, V]':
+  def copy(self, add_or_replace: Mapping[K, V]) -> 'FrozenDict[K, V]':
     """Create a new FrozenDict with additional or replaced entries."""
-    return type(self)({**self, **unfreeze(add_or_replace)})
+    return type(self)({**self, **unfreeze(add_or_replace)}) # type: ignore[arg-type]
 
   def keys(self):
     return FrozenKeysView(self)
--- a/flax/core/meta.py
+++ b/flax/core/meta.py
@@ -32,7 +32,7 @@ from jax.experimental import maps
 from jax.experimental import pjit
 
 
-TAxisMetadata = TypeVar('TAxisMetadata', bound='AxisMetadata')
+TAxisMetadata = Any # TypeVar('TAxisMetadata', bound='AxisMetadata')
 
 
 class AxisMetadata(metaclass=abc.ABCMeta):
@@ -252,8 +252,8 @@ class Partitioned(struct.PyTreeNode, AxisMetadata):
     axis_name = self._get_partition_name(params)
     names = list(self.names)
     while len(names) < index:
-      names.append(None)
-    names.insert(index, axis_name)
+      names.append(None) # type: ignore
+    names.insert(index, axis_name) # type: ignore
     return self.replace(names=tuple(names))
 
   def remove_axis(self, index: int, params: Dict[Any, Any]) -> TAxisMetadata:
--- a/flax/core/nn/attention.py
+++ b/flax/core/nn/attention.py
@@ -299,8 +299,8 @@ def multi_head_dot_product_attention(
         indices[attn_dim] = i // attn_size
         i = i % attn_size
 
-      key = lax.dynamic_update_slice(cache_entry.key, key, indices)
-      value = lax.dynamic_update_slice(cache_entry.value, value, indices)
+      key = lax.dynamic_update_slice(cache_entry.key, key, indices) # type: ignore
+      value = lax.dynamic_update_slice(cache_entry.value, value, indices) # type: ignore
       one = jnp.array(1, jnp.uint32)
       cache_entry = cache_entry.replace(i=cache_entry.i + one,
                                         key=key,
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -112,9 +112,9 @@ def dot_product_attention_weights(query: Array,
     if broadcast_dropout:
       # dropout is broadcast across the batch + head dimensions
       dropout_shape = tuple([1] * (key.ndim - 2)) + attn_weights.shape[-2:]
-      keep = random.bernoulli(dropout_rng, keep_prob, dropout_shape)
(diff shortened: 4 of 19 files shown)


flax/core/frozen_dict.py
```diff
@@ -108,7 +108,7 @@ def __hash__(self):
       self._hash = h
     return self._hash
 
-  def copy(self, add_or_replace: Mapping[K, V]) -> 'FrozenDict[K, V]':
+  def copy(self, add_or_replace: 'FrozenDict[K, V]') -> 'FrozenDict[K, V]':
```
levskaya, line 111: Side Q: do you remember what python version lets us get rid of the quotes for forward-reference syntax?

flax/core/meta.py
```diff
@@ -32,7 +32,7 @@
 from jax.experimental import pjit
 
 
-TAxisMetadata = TypeVar('TAxisMetadata', bound='AxisMetadata')
+TAxisMetadata = Any # TypeVar('TAxisMetadata', bound='AxisMetadata')
```
levskaya, line 35: mypy doesn't support bound type-vars?!  is this intentional Any cast?

flax/core/nn/attention.py
```diff
@@ -299,8 +299,8 @@ def init_fn(shape, dtype=jnp.float32):
         indices[attn_dim] = i // attn_size
         i = i % attn_size
 
-      key = lax.dynamic_update_slice(cache_entry.key, key, indices)
-      value = lax.dynamic_update_slice(cache_entry.value, value, indices)
+      key = lax.dynamic_update_slice(cache_entry.key, key, indices) # type: ignore
```
levskaya, line 302: just curious what type error lax.dynamic_update_slice causes?

flax/linen/attention.py
```diff
@@ -112,9 +112,9 @@ def dot_product_attention_weights(query: Array,
     if broadcast_dropout:
       # dropout is broadcast across the batch + head dimensions
       dropout_shape = tuple([1] * (key.ndim - 2)) + attn_weights.shape[-2:]
-      keep = random.bernoulli(dropout_rng, keep_prob, dropout_shape)
+      keep = random.bernoulli(dropout_rng, keep_prob, dropout_shape) # type: ignore
```
levskaya, line 115: same Q here: how was jax breaking typechecks here?

flax/linen/experimental/layers_with_named_axes.py
```diff
@@ -30,10 +30,10 @@
 
 # Type annotations
 Array = jnp.ndarray
-Axes = Union[int, Iterable[int]]
-DType = jnp.dtype
+Axes = Union[int, Any]
```
levskaya, line 33: could we add a `IntSequence` (or something) to indicate what this `Any` is supposed to be for a human reader?

flax/linen/module.py
```diff
@@ -1559,7 +1559,7 @@ def loss(params, perturbations, inputs, targets):
                                   #      [-1.456924   -0.44332537  0.02422847]]
 
     """
-    value += self.variable(collection, name, lambda: jnp.zeros_like(value)).value
+    value += self.variable(collection, name, lambda: jnp.zeros_like(value)).value # type: ignore
```
levskaya, line 1562: are we messing up internal type annotations in flax module methods to require this opt-out?

flax/linen/recurrent.py
```diff
@@ -250,6 +250,7 @@ def _concat_dense(inputs: Array,
       inputs, kernel, bias = promote_dtype(inputs, kernel, bias, dtype=self.dtype)
       y = jnp.dot(inputs, kernel)
       if use_bias:
+        assert bias is not None
```
levskaya, line 253: I don't love having to add seemingly redundant asserts for mypy... but we can live with a few.

levskaya: maybe add a comment for these asserts that it's to satisfy typecheckers?

levskaya (approved):
Comments below.

Just curious - does modern mypy use mypyc compiler by default or is that a random side-effort?

levskaya:
when we add new open source config files we need to configure them to be sync'd by the internal sync system first - I just added your mypy.ini file to #2693 - maybe rebase on that after it goes in (hopefully soon).

flax/core/frozen_dict.py
zaxtax: It might be Python 3.10, but I am having trouble pulling up a Github issue to support this.

flax/core/meta.py
zaxtax: mypy has had lots of trouble dealing with type-vars in general. I've tried to preserve their use, and just define them as Any. That way, we can reintroduce stricter checking later where possible.

flax/core/nn/attention.py
zaxtax: `flax/core/nn/attention.py:302: error: Argument 1 to "dynamic_update_slice" has incompatible type "ndarray[Any, Any]"; expected "Array"  [arg-type]
`

To do this correctly, I think I'd need to update the types for `CacheEntry` and that felt like too intrusive of a step.

flax/linen/attention.py
zaxtax: `Argument 1 to "bernoulli" has incompatible type "Optional[Any]"; expected "Union[Array, PRNGKeyArray]"`

Basically `PRNGKey` is defined differently in flax and jax, and bringing the two in line would have involved much larger changes.

flax/linen/experimental/layers_with_named_axes.py
zaxtax: Sounds like a great idea! Just as an FYI, there are many places in the code where pytype wanted something to be `Iterabale` and mypy wanted something to be `Sequence`. I couldn't make both happy so they ended up with `Any`

flax/linen/recurrent.py
zaxtax: Will do.

flax/linen/experimental/layers_with_named_axes.py
levskaya: that's both sad and funny... I wonder if there's a way to harmonize these two in our code better or to agitate for change - for now just another Any alias indicating we intend a sequence of ints should be ok

zaxtax:
That and mypy not being able to reason about flax's dataclasses were the
two annoying bits of making this PR work. On the bright side, as things
like mypy improve, we can reintroduce more stringent types.

zaxtax:
> @zaxtax - seeing a small number of new internal pytype bugs of this sort:
> 
> ```
> File "rax/examples/flax_integration/main.py", line 143, in train_step: Function FrozenDict.copy was called with the wrong arguments [wrong-arg-types]
>          Expected: (self, add_or_replace: flax.core.frozen_dict.FrozenDict[str, Mapping[str, Any]])
>   Actually passed: (self, add_or_replace: Dict[str, Any])
> 
> File "jestimator/states.py", line 73, in variables: Function FrozenDict.copy was called with the wrong arguments [wrong-arg-types]
>          Expected: (self, add_or_replace: flax.core.frozen_dict.FrozenDict[str, Any])
>   Actually passed: (self, add_or_replace: Dict[str, Optional[flax.core.frozen_dict.FrozenDict[str, Any]]])
> 
> File "jestimator/states.py", line 143, in variables: Function FrozenDict.copy was called with the wrong arguments [wrong-arg-types]
>          Expected: (self, add_or_replace: flax.core.frozen_dict.FrozenDict[str, Any])
>   Actually passed: (self, add_or_replace: Dict[str, Optional[flax.core.frozen_dict.FrozenDict[str, Any]]])
> ```

@levskaya try now

PR #2671, opened 2022-11-29
updated flax docs front page

chiamp:
Updated the code example on the flax docs front page to include a simple regression training loop.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -92,31 +92,73 @@ Flax installs the vanilla CPU version of JAX, if you need a custom version pleas
 Basic usage
 ^^^^^^^^^^^^
 
-.. testsetup::
-
-   import jax
-   from jax.random import PRNGKey
-   import flax.linen as nn
-   import jax.numpy as jnp
-
 .. testcode::
 
+   import jax
+   import jax.numpy as jnp                  # JAX NumPy
+   from jax import random                   # JAX RNG
+   from flax import linen as nn             # The Linen API
+   from flax.training import train_state    # Useful dataclass to keep train state
+   import optax                             # Optimizers
+
    class MLP(nn.Module):                    # create a Flax Module dataclass
-     out_dims: int
+      out_dims: int
 
-     @nn.compact
-     def __call__(self, x):
-       x = x.reshape((x.shape[0], -1))
-       x = nn.Dense(128)(x)                 # create inline Flax Module submodules
-       x = nn.relu(x)
-       x = nn.Dense(self.out_dims)(x)       # shape inference
-       return x
+      @nn.compact
+      def __call__(self, x):
+         x = x.reshape((x.shape[0], -1))
+         x = nn.Dense(128)(x)               # create inline Flax Module submodules
+         x = nn.relu(x)
+         x = nn.Dense(self.out_dims)(x)     # shape inference
+         return x
 
-   model = MLP(out_dims=10)                 # instantiate the MLP model
+   n_samples = 1000
+   n_in_features = 10                       # number of features in data
+   n_out_features = 10                      # number of features in target
+   learning_rate = 0.01
+   momentum = 0.9
+   n_epochs = 10                            # number of training epochs
 
-   x = jnp.empty((4, 28, 28, 1))            # generate random data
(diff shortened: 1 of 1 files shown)


8bitmp3:
I think an extended sample is a good idea because it includes a simple training process.

Maybe we can include this in a README on GitHub (`google/flax`), instead of the Flax Landing Page (see Keras and NumPy examples further below)?

- If the code example gets much longer, some users may not realize we have other docs. such as  the first "tutorial" called _Getting Started_ (which covers end-to-end training). This especially applies to (new) users who are scrolling on a small device after coming across the Flax docs site for the first time.

- Currently, the user journey is:

  1 - Landing page: (easily) install with pip
  2 - A short code snippet to get familiarized with the syntax 
  3 - Go to _Learn More_ -> _Getting Started_ to train a simple example
  4 - ...
  N - Learn about the _Ecosystem_ with PaLM (!), DALL-E Mini (!), HF Flax models, etc 

- With a lengthened example on the landing page, some users may not be able to discover steps 3, 4, and the last part  🤷‍♂️  unless they keep scrolling down. Some users may be (re)visiting Flax for the first time while on a smaller screen (13" laptop, mobile).

image

At the moment, the landing page is short and simpler. It's similar to:

1. Keras https://www.keras.io - you click on **Guides** and then **Getting started**. Notice the code snippet on the landing page doesn't appear to demo how to train a Keras model (it'd probably be too long).

image

2. NumPy https://numpy.org/ - you click on **Learn** to get to the **Quickstart**, and more **Tutorials**:

image

LMKWYT!

chiamp:
Thanks for the feedback @8bitmp3! My intention for the extended example was to give a new user a quick snapshot of what imports, classes and functions are typically used in a training loop. Perhaps the front page is not the best place to put it for the reasons you mentioned. I wonder if there's another section in the docs we can put it in or if we can make a new section for it? The purpose would just be a "quick and dirty" example, but shouldn't take away from the more thorough examples we have in the documentation. I'd also be open to putting the extended example on the `google/flax` README on GitHub, but I'd also like the example to be somewhere in the documentation as well.

8bitmp3:
> I wonder if there's another section in the docs we can put it in or if we can make a new section for it? 

I am 💯 for including the regression example. Let's do a UI/UX research on other OSS project sites and check how they do it. I think pandas.org is another (Sphinx... RTD) site that has a great UI. Maybe @jakevdp also has experience on the OSS side when it comes to presenting useful (but long) code blocks on landing pages.

8bitmp3:
@chiamp Maybe we can include the full regression code in a 2nd column. I'll review the reStructured Text and Sphinx docs.

marcvanzee:
I like the example, but I agree with @8bitmp3 that maybe it is already too long for the front page. The front page should be a summary, mostly consisting of resources pointing to places where to look further.

We currently have already three "introduction" tutorials: [Flax Basics](https://flax--2671.org.readthedocs.build/en/2671/guides/flax_basics.html), [Getting Started](https://flax--2671.org.readthedocs.build/en/2671/getting_started.html), and [Linen intro](https://github.com/google/flax/blob/main/docs/notebooks/linen_intro.md). @8bitmp3 is planning on merging these into one tutorial.

Additionally, we have the [JAX for the impatient](https://flax--2671.org.readthedocs.build/en/2671/guides/jax_for_the_impatient.html) guide, which actually develops a linear regression example in JAX.

Actually, thinking about it now, maybe you could simply add a new guide "Linear Regression". This would be very simple, but as you say, people like to see this when they are trying out a new library for neural nets.

I am also curious what @andsteing and @jheek think.

8bitmp3:
Regression is a cool end-to-end example. We can experiment with adding a second column on the landing page, so that other important sections can still be visible without scrolling too much.

fyi A while back I added a Flax regression example to Optax's `/examples`: https://github.com/deepmind/optax/blob/master/examples/flax_example.py

andsteing:
+1 on not making the initial example more complicated

(one could consider adding a line for the imports like `import flax.linen as nn, jax, jax.numpy as jnp` or simply `import flax.linen as nn` maybe, but I would not add more functionality to avoid making it too length)

but we could have a prominent link to a more complete example (though maybe the link to the beginner's guide is arguably more useful)

PR #2680, opened 2022-12-01
Expose module.merge_param in RTD and fix docstring

marcvanzee:
This is a function that we use in e.g. Dropout. I propose we expose it in RTD so we can refer to it from the Dropout guide.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1657,9 +1657,9 @@ class Module:
 _ParentType = Union[Type[Module], Type[Scope], Type[_Sentinel], None]
 
 def merge_param(name: str, a: Optional[T], b: Optional[T]) -> T:
-  """Merges construction-and call-time argument.
+  """Merges construction- and call-time argument.
 
-  This is a utility for supporting the pattern where a Module hyperparameter
+  This is a utility for supporting a pattern where a Module hyperparameter
   can be passed either to ``__init__`` or ``__call__``, and the value that is
   not `None` will be used.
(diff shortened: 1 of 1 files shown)


flax/linen/module.py
```diff
@@ -1657,10 +1657,11 @@ def __call__(self, x):
 _ParentType = Union[Type[Module], Type[Scope], Type[_Sentinel], None]
 
 def merge_param(name: str, a: Optional[T], b: Optional[T]) -> T:
-  """Merges construction and call time argument.
+  """Merges construction-and call-time argument.
```
8bitmp3, line 1660: Nit:

```suggestion
  """Merges construction- and call-time argument.
```

flax/linen/module.py
```diff
@@ -1657,10 +1657,11 @@ def __call__(self, x):
...
 
-  This is a utility for supporting the pattern where a Module hyper parameter
-  can be passed to ``__init__`` or ``__call__``.
+  This is a utility for supporting the pattern where a Module hyperparameter
```
8bitmp3, line 1662: Nit: Perhaps we should say "a pattern" 🤷 

```suggestion
  This is a utility for supporting a pattern where a Module hyperparameter
```

8bitmp3 (commented):
A few minor suggestions. Thanks @marcvanzee

8bitmp3 (approved):
Thanks @marcvanzee 

LGTM

PR #2682, opened 2022-12-02
Update python version support

cgarciae:
# What does this PR do?

* Removes support for python 3.7
* Add support for python 3.10

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -63,7 +63,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.8', '3.9', '3.10']
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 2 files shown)


zaxtax:
Should 3.11 be included?

cgarciae:
@zaxtax good question, not sure what are current policy is. Maybe @levskaya or @marcvanzee can give some insight into how we are deciding which versions to support?

8bitmp3:
Thanks @cgarciae !

wookayin:
Of course python3.11 should be supported, and the only reason if we could miss py311 is due to tensorflow. Flax itself does work with python3.11 without TF installed (Note: TF doesn't support python 3.11 yet), but in tests there are some optional dependency. If we can add a route of testing that could skip everything that depends on TF, we can make py311 supported.

levskaya:
@zaxtax @cgarciae - JAX / jaxlib supports 3.11.   TF isn't expecting to release 3.11 support until 2023Q1 (so a month or two).

The main issue is trying to test anything without TF support... there are a ton of tests that rely on tf.data.  If y'all wanted to add a bunch of skip_test annotations we might be able to test against it sooner.

levskaya (approved):

wookayin:
I have been installing and using flax on python 3.11, and there's no significant problems so far (other than the tensorflow/tensorstore dependencies issues, like #2692). Flax itself should have no problems running on python 3.11 so py311 can be included if we workaround TF-related tests. I would like to take over it and send some PRs.

wookayin:
@cgarciae @levskaya Shouldn't we have `python_requires >= 3.7` in https://github.com/google/flax/blob/main/setup.py? Otherwise, once flax breaks on python 3.7, `flax` releases installed on python3.7 via `pip install flax` will be broken.

cgarciae:
Thought about this but while correct it seemed too harsh just in case someone is stuck in 3.7 and by chance we don't break it. I'll send a tentative PR with this change so we can discuss it.

PR #2701, opened 2022-12-08
Cleanup DenseGeneral

cgarciae:
# What does this PR do?

* Reimplements #2665 to use single rng key (avoids splitting).
* `Dense` and `DenseGeneral` are again equivalent for the basic cases. 
* `test_dense_is_dense_general` is reverted to its old behavior.


marcvanzee (approved):
Looks awesome!

PR #2716, opened 2022-12-12
Run actions/setup-python with Python 3.9 instead of 3.7

marcvanzee:
This is [the recommended practice](https://github.com/actions/setup-python#basic-usage). I suppose we updated the `setup-python` version but forgot to bump the Python version.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -5,20 +5,12 @@ filterwarnings =
     error
 # Jax warning when no gpu/tpu found.
     ignore:No GPU/TPU found, falling back to CPU.*:UserWarning
-# Jax warns about XLA not being able to use donated buffers.
-    ignore:Some donated buffers were not usable.*:UserWarning
 # We should remove our dependence on jtu.
     ignore:jax.test_util.check_eq is deprecated and will soon be removed.*:FutureWarning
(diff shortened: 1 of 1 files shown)


cgarciae:
@marcvanzee maybe we keep using 3.7 as its used internally? Python version for `pre-commit` should rarely matter, but when it does I'd prefer it to agree with Google's systems.

marcvanzee:
> @marcvanzee maybe we keep using 3.7 as its used internally? Python version for `pre-commit` should rarely matter, but when it does I'd prefer it to agree with Google's systems.

Good point. Internally we use 3.9 (I just checked), so i have changed it to 3.9.

.github/workflows/build.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Set up Python 3.8
         uses: actions/setup-python@v4
         with:
-          python-version: 3.7
+          python-version: '3.10'
```
marcvanzee, line 33:
```suggestion
          python-version: '3.9'
```

cgarciae:
@marcvanzee sorry yeah I meant 3.9

cgarciae (approved):

cgarciae:
Would be good to squash history before merging.

PR #2725, opened 2022-12-14
Call _try_setup in bind

cgarciae:
# What does this PR do?

Currently modules that take submdules as inputs such as `Sequential` have issues when using `bind` e.g:

```python
module = nn.Sequential([
    nn.Dense(8),
    nn.relu,
    nn.Dense(4),
])
x = jnp.ones((1, 5))

bound_module = module.bind(variables)

bound_module.layers[0](x) # Error: layer has no scope.
```
The proposed solution is to call `_try_setup` during `bind` to automatically bind all submodules.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1061,6 +1061,12 @@ class Module:
         self._state.in_setup = False
         self._state.setup_called = SetupState.DONE
 
+      # Recursively try to setup children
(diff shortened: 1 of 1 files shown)


cgarciae:
Closing this in favor of a more complete solution, possibly through `__getattribute__`.

PR #2728, opened 2022-12-15
Implement perturb no-op behaviour

cgarciae:
# What does this PR do?

* If `perturbations` are not passed, `perturb` behaves like a no-op. When the user is done inspecting the gradients it can omit passing `perturbations` without having to change the model.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1690,8 +1690,22 @@ class Module:
       print(intm_grads['dense3']) # ==> [[-1.456924   -0.44332537  0.02422847]
                                   #      [-1.456924   -0.44332537  0.02422847]]
 
+    If perturbations are not passed to `apply`, `perturb` behaves like a no-op
+    so you can easily disable the behavior when not needed::
+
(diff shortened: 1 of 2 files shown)


flax/linen/module.py
```diff
@@ -1655,8 +1655,18 @@ def loss(params, perturbations, inputs, targets):
...
+
+      model.apply({'params': params, 'perturbations': perturbations}, x) # works as expected
+      model.apply({'params': params}, x) # behaves like a no-op
+
     """
-    value += self.variable(collection, name, lambda: jnp.zeros_like(value)).value # type: ignore
+    def _root_has_collection():
+      assert self.scope is not None
+      return collection in self.scope.root._variables
+    if self.is_mutable_collection(collection) or _root_has_collection():
```
marcvanzee, line 1668: Maybe add a comment explaining this. I suppose the first clause handles initialization (since everything is mutable by default) and the second one apply (it will be ignored if you don't pass it).

cgarciae: Done!

marcvanzee (approved):
Looks good!

PR #2730, opened 2022-12-16
Guide for Flax partitioning (auto SPMD) API

IvyZX:
Original code PR: https://github.com/google/flax/pull/2704

### DIFF
--- a/docs/guides/flax_on_pjit.md
+++ b/docs/guides/flax_on_pjit.md
@@ -1,0 +1,437 @@
+---
+jupyter:
+  jupytext:
+    formats: ipynb,md
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
+      jupytext_version: 1.13.8
+  kernelspec:
+    display_name: Python 3 (ipykernel)
+    language: python
+    name: python3
+---
+
+# Scale up Flax modules on multiple devices
+
+This guide shows how to scale up Flax Modules on multiple devices and hosts using JAX's [`pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html#module-jax.experimental.pjit) and [`flax.linen.spmd`](https://flax.readthedocs.io/en/latest/api_reference/flax.linen.html#module-flax.linen.spmd).
+
+## Flax and `pjit`
+
+[`jax.experimental.pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html) provides a way to automatically compile and scale up JAX computations. It provides the following benefits:
+
+* `pjit` has the similar interface of [`jax.jit`](https://jax.readthedocs.io/en/latest/jax-101/02-jitting.html) and works as a decorator on a function that needs to be compiled.
+
+* When using `pjit`, you can write code as if it runs on a single device, and `pjit` will automatically compile and run it on multiple devices using the [Single Program Multi Data (SPMD)](https://jax.readthedocs.io/en/latest/glossary.html#term-SPMD) paradigm. 
+
+* With `pjit` you can state how the input and output of your code is partitioned across devices, and the compiler will figure out how to partition everything inside, and how to compile inter-device communications.
+
+* For more information, refer to [JAX-101 pjit tutorial](https://jax.readthedocs.io/en/latest/jax-101/08-pjit.html) and [JAX in multi-process environments](https://jax.readthedocs.io/en/latest/multi_process.html).
+
+Flax provides a few API to help you use `pjit` on Flax Modules, including:
+
+* An interface to specify partitions of your data when defining [`flax.linen.Module`](https://flax.readthedocs.io/en/latest/api_reference/flax.linen.html#module).
+
+* Utility function to generate the partition information that `pjit` requires to run.
+
+* An interface to customize your axis names called "logical axis annotations" to decouple both your Module code and partition plan to experiment with different partition layouts more easily.
+
+
+
+# Setup
+
+Install Flax from HEAD.
+
+```python
+# Once Flax v0.6.4 is out, replace this with `pip3 install flax`.
+!pip3 install -qq "git+https://github.com/google/flax.git@main#egg=flax"
+```
+
+## Imports
+
+Import some necessary dependencies.
+
+**Note:** This guide uses `--xla_force_host_platform_device_count=8` to emulate multiple devices on a CPU environment in a Google Colab/Jupyter Notebook. You can also follow the [JAX-101 pjit tutorial](https://jax.readthedocs.io/en/latest/jax-101/08-pjit.html#setup) to learn how to emulate a multi-device TPU environment (in which case you should ignore the `os.environ` cell).
+
+```python
+import os
+os.environ["XLA_FLAGS"] = '--xla_force_host_platform_device_count=8'
+```
+
+```python
+import functools
+import numpy as np
+import jax
+
+from jax import lax, random, numpy as jnp
+
+import flax
+from flax import struct, traverse_util, linen as nn
+from flax.linen import spmd
+from flax.core import freeze, unfreeze
+from flax.training import train_state, checkpoints
+
+import optax
+```
+
+In the next step, import all the `pjit`-related libraries.
+
+> **Note:** [`jax.experimental.pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html) is still in the experimental package of JAX, so there may be changes in the API in future.
+
+Start a device mesh using the 8 devices available—in this example, set them as a 2x4 device mesh, which is the same as the layout of [TPU v3-8](https://cloud.google.com/tpu/docs/system-architecture-tpu-vm#single_tpu_board).
+
+Annotate each axis with a name. A typical way to annotate axis names is `('data', 'model')`, where:
+  * `'data'`: the mesh dimension used for data-parallel sharding of the batch dimension of inputs and activations.
+  * `'model'`: the mesh dimension used for sharding parameters of the model across devices.
+
+
+```python
+from jax.experimental.pjit import pjit, with_sharding_constraint, PartitionSpec
+from jax.experimental.maps import Mesh
+from jax.experimental import mesh_utils
+
+device_mesh = mesh_utils.create_device_mesh((2, 4))
+print(device_mesh)
+mesh = Mesh(devices=device_mesh, axis_names=('data', 'model'))
+mesh
+```
+
+# Define a model
+
+Define an example layer `DotReluDot` (by subclassing `flax.linen.Module`), which creates two parameters `W1` and `W2` for dot products, and uses the `jax.nn.relu` activation function in-between. 
+
+To use this layer in `pjit` efficiently, apply the following APIs to annotate the parameters and intermediate variables correctly:
+
+* Use [`flax.linen.with_partitioning`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.with_partitioning.html#flax.linen.with_partitioning) to decorate the initializer function when creating parameters `W1` and `W2`.
+
+* Apply [`pjit.with_sharding_constraint`](https://github.com/google/jax/blob/main/jax/_src/pjit.py#L1516) to annotate intermediate variables like `y` and `z` to force a particular sharding pattern under `pjit` when the ideal constraint is known.
+
+  * This step is optional, but can sometimes help auto-SPMD to partition efficiently. In the example below, the call is not required, because `pjit` will figure out the same sharding layout for `y` and `z` regardless.
+
+```python
+class DotReluDot(nn.Module):
+  depth: int
+  @nn.compact
+  def __call__(self, x):
+    W1 = self.param(
+        'W1', 
+        nn.with_partitioning(nn.initializers.xavier_normal(), (None, 'model')),
+        (x.shape[-1], self.depth))
+
+    y = jax.nn.relu(jnp.dot(x, W1))
+    # force a local sharding annotation.
+    y = with_sharding_constraint(y, PartitionSpec('data', 'model'))
+
+    W2 = self.param(
+        'W2', 
+        nn.with_partitioning(nn.initializers.xavier_normal(), ('model', None)),
+        (self.depth, x.shape[-1]))
+
+    z = jnp.dot(y, W2)
+    # force a local sharding annotation.
+    z = with_sharding_constraint(z, PartitionSpec('data', None))
+
+    # Return a tuple to conform with the API `nn.scan` as shown in the cell below.
+    return z, None
+
+```
+
+Note that values like `'data'`, `'model'` or `None` are passed into these API calls. This refers to how each dimension of this data should be sharded - either across one of the device mesh dimensions, or not sharded at all.
+
+For example:
+
+* When we define `W1` with shape `(x.shape[-1], self.depth)` and annotate as `(None, 'model')`):
+
+  * The first dimension (of length `x.shape[-1]`) will be replicated across all devices.
+ 
+  * The second dimension (of length `self.depth`) will be sharded over the `model` axis of the device mesh. This means `W1` will be sharded 4-way on devices `(0, 4)`, `(1, 5)`, `(2, 6)` and `(3, 7)`, on this dimension.
+  
+* When we annotate the output `z` as `('data', None)`:
+
+  * The first dimension (aka. the batch dimension) will be sharded over the `data` axis, which  means half of the batch will be processed on devices `0-3` and another half on devices `4-7`.
+  
+  * The second dimension (aka. the data depth dimension) will be replicated across all devices.
+
+
+## Lifted transforms
+
+Now define the `MLP` model as multiple layers of `DotReluDot`.
+
+To replicate identical layers, you can either use `flax.linen.scan` or a for-loop: `flax.linen.scan` can offer faster compilation times, whereas the for-loop can be faster on runtime. This guide uses `flax.linen.scan` simply to show that [Flax lifted transforms](https://flax.readthedocs.io/en/latest/advanced_topics/lift.html#supported-transformations) work together with [JAX `pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html).
+
+**Note:** `flax.linen.scan` will introduce another dimension for the params (the dimension over which `scan` is applied), and you need to use the `metadata_params` argument to annotate the partition of this dimension. Since the parameters inside your `DotReluDot` (a sub-`Module`) are already sharded along the `model` axis, you don't need to partition multiple layers across the `model` dimension here, and therefore you should denote it as `None`.
+
+```python
+class MLP(nn.Module):
+  num_layers: int
+  depth: int
+  use_scan: bool
+  @nn.compact
+  def __call__(self, x):
+    if self.use_scan:
+      x, _ = nn.scan(DotReluDot, length=self.num_layers, 
+                     variable_axes={"params": 0},
+                     split_rngs={"params": True},
+                     metadata_params={nn.PARTITION_NAME: None}
+                     )(self.depth)(x)
+    else:
+      for i in range(self.num_layers):
+        x, _ = DotReluDot(self.depth)(x)
+    return x
+```
+
+# Sharding specification
+
+Next, you need to generate the [`jax.experimental.pjit.PartitionSpec`](html) that `pjit` should receive as annotations of _input_ and _output_ data.
+
+For data parallelism, you can shard the batched _input_ `x` across the `data` axis, by denoting the batch axis as `data`:
+
+```python
+x_spec = PartitionSpec('data', None)  # dimensions: (batch, length)
+x_spec
+```
+
+However, to generate `PartitionSpec` for the _output_, you need to use some actual output as reference. One solution is to create a model, and then evaluate `model.init` abstractly using `jax.eval_shape`, and then use `nn.get_partition_spec` to automatically generate the `PartitionSpec`.
+
+The code below shows how to get the output spec if you use `flax.training.train_state` to carry out your initialization and training steps, in which case your `pjit`ted function will output a `TrainState`. 
+
+(In a simpler case, people might choose the variable dict as in `variables = model.init(k, x)` as their `pjit`ted function's output. That works too.)
+
+> A side note: Here we define our `init_fn` as purely functional and takes `model` and `optimizer` as arguments. This is not necessary - you can simply define with `def init_fn(k, x):` and all will work fine here.
+> 
+> This guide doesn't do it because later we will show you another way to define your model and will run `init_fn` with another model instance. However, this is problematic because `jax.eval_shape` only takes pytrees as arguments, so we have to create an abstract closure before feeding the function in.
+
+```python
+# MLP hyperparams
+BATCH, LAYERS, DEPTH, USE_SCAN = 8, 4, 1024, True
+# fake inputs
+x = jnp.ones((BATCH, DEPTH))
+k = random.PRNGKey(0)
+
+optimizer = optax.adam(learning_rate=0.001)
+model = MLP(LAYERS, DEPTH, USE_SCAN)
+
+def init_fn(k, x, model, optimizer):
+  variables = model.init(k, x)
+  state = train_state.TrainState.create(
+    apply_fn=model.apply,
+    params=variables['params'],
+    tx=optimizer)
+  return state
+
+with mesh:
+  abstract_variables = jax.eval_shape(
+      functools.partial(init_fn, model=model, optimizer=optimizer), k, x)
+# we need a rule on every leaf of the input variables, in this case a nested dict.
+state_spec = nn.get_partition_spec(abstract_variables)
+state_spec
+```
+
+# `pjit` the initialization and train step
+
+Now you can `pjit` the `init_fn` in a similar fashion as `jit`, with two extra args `in_axis_resources` and `out_axis_resources`.
+
+You need to add `with mesh:` context when running a `pjit`ted function, so that it can refer to `mesh` to allocate data on devices correctly.
+
+```python
+pjit_init_fn = pjit(init_fn,
+                    static_argnums=(2, 3),
+                    in_axis_resources=(PartitionSpec(None), x_spec),  # RNG key and x
+                    out_axis_resources=state_spec
+                    )
+with mesh:
+  initialized_state = pjit_init_fn(k, x, model, optimizer)
+jax.tree_map(jnp.shape, initialized_state)
+```
+
+## Looking into the module output
+
+Note that in the output `initialized_state`, the params `W1` and `W2` are of type [`Partitioned`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.Partitioned.html). This is a wrapper around the actual JAX array that allows Flax to record metadata associated with it. You can access the raw JAX array by adding `.value` or running `.unbox()`.
+
+You can also check the underlying `.sharding` of the JAX array, which gives a hint on the way it is partitioned.
+
+```python
+print(type(initialized_state.params['ScanDotReluDot_0']['W1']))
+print(type(initialized_state.params['ScanDotReluDot_0']['W1'].value))
+print(initialized_state.params['ScanDotReluDot_0']['W1'].value.shape)
+```
+
+```python
+print(initialized_state.params['ScanDotReluDot_0']['W1'].value.sharding)
+```
+
+You can use `jax.tree_map` to perform mass computation on a dict of boxed params, in the same way as on a dict of JAX arrays.
+
+```python
+diff = jax.tree_map(
+    lambda a, b: a - b, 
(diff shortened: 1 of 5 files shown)


marcvanzee:
@IvyZX the doc build is failing, some problem with the sharding code. Could you please fix it?

docs/guides/spmd_api.ipynb
```diff
@@ -0,0 +1,1135 @@
+{
```
levskaya, line 1: <p>"scaled module output"  what does "scaled" mean here? do you mean "sharded" or am I missing something?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/2730/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='17'/>

levskaya, line 1: <p>"Profilling" --&gt; "Profiling"  (one "L")</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/2730/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='26'/>

levskaya, line 1: <p>"replaced"-&gt;"replaces"</p><p> </p><p>We should perhaps note that there are more complicated/advanced sharding patterns that require annotating -activation- dimension names differently from -parameter- dimension names.  For really advanced stuff, people may wish to do fine-grained manual mesh assignments using the simpler system above, but the logical-naming helpers are useful for getting started in exploring different sharding layouts.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/google/flax/pull/2730/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='28'/>

IvyZX:
> @IvyZX the doc build is failing, some problem with the sharding code. Could you please fix it?

This seems to be due to JAX dropping Python 3.7. I switched readthedocs build to 3.8 and the test passes now.

docs/guides/spmd_api.ipynb
IvyZX: Removed "scaled" - hopefully it's clearer now.

IvyZX: Done.

IvyZX: Added a new section at the bottom of logic axis introduction. Let me know if there could be a better phrasing of the idea!

docs/guides/index.rst
```diff
@@ -14,6 +14,7 @@ Guides
    extracting_intermediates
    lr_schedule
    use_checkpointing
+   spmd_api
```
marcvanzee, line 17: Maybe this is an advanced guide? It looks quite involved and you do a number of advanced things (lifted transformations, scan over layers, ...)

IvyZX: Lifted transformations is the only advanced topic here, and it is not really explored in detail. Since JAX team may want to upstream `pjit` to `jit` in the future, it may make sense to put it on the later section of guides as an introduction on how to scale up a model. (JAX also put their [pjit doc](https://jax.readthedocs.io/en/latest/jax-101/08-pjit.html) as the last piece of their 101 tutorial.)

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,432 @@
...
+    W2 = self.param(
+        'W2', 
+        meta.with_partitioning(
+            nn.initializers.xavier_normal(),
+            ('model', None)),
+        (self.depth, x.shape[-1]))
+
+    z = jnp.dot(y, W2)
+    # force a local sharding annotation, fairly redundant in this case.
+    z = with_sharding_constraint(z, PartitionSpec('data', None))
```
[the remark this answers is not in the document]
IvyZX: Because the output should be sharded as ('data', None), in which 'data' refers to the batch dimension of input.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,432 @@
...
+  def __call__(self, x):
+    W1 = self.param(
+        'W1', 
+        meta.with_partitioning(
+            nn.initializers.xavier_normal(),
+            (None, 'model')),
+        (x.shape[-1], self.depth))
+
+    y = jax.nn.relu(jnp.dot(x, W1))
+    # force a local sharding annotation, fairly redundant in this case.
```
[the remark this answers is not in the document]
IvyZX: The local sharding annotation here. I removed the comment here and added more explanation on the documentation above. Basically `with_sharding_constraint` is only to ensure that `pjit` shards this variable in this way, and it's redundant here because `pjit` will by default shard it this way too.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,432 @@
...
+
+* Use `meta.with_partitioning` to decorate the initializer function when creating params `W1` and `W2`.
+
+* Use `pjit.with_sharding_constraint` to annotate intermediate variables like `y` and `z` to force a particular sharding pattern under pjit when the ideal constraint is known.
+
+ * This is optional, but sometimes can help auto SPMD to partition efficiently.
+
+
+
+Each value passed into a sharding annotates a dimension, and its value should be among `data`, `model` or `None`. This refers to whether the data's axis should be sharded across one of the device mesh dimensions, or not sharded at all.
```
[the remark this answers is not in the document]
IvyZX: Yes, this is correct. I rephrased it to:

Calling these APIs annotates each dimension of your parameter and intermediate variables with an axis name, namely `'data'`, `'model'` or `None`. This refers to how this dimension should be sharded - across one of the device mesh dimensions, or not sharded at all.

For example, if we define `W1`, which has shape `(x.shape[-1], self.depth)`, with `(None, 'model')`, the first dimension (of length `x.shape[-1]`) will be replicated across all devices, while the second dimension (of length `self.depth`) will be sharded over the `model` axis of the device mesh.

IvyZX:
Thank you Marc! I address all your comments, either fixing them directly or leaving some comments beneath yours. Please take a look and let me know what you think.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+      format_version: '1.3'
+      jupytext_version: 1.13.8
+  kernelspec:
+    display_name: Python 3 (ipykernel)
+    language: python
+    name: python3
+---
+
+<!-- #region id="2-JYevEklHfB" -->
+# Scale up Flax modules with `pjit`
```
8bitmp3, line 17: Suggestion:

Maybe mention multi-host/multi-devices in the title so that users know what this scaling up with `pjit` means without indepth JAX knowledge. Similar to "Using JAX in multi-host and multi-process environments" https://jax.readthedocs.io/en/latest/multi_process.html

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+
+In this guide, you will learn how to use JAX's `pjit` and `flax.linen.spmd` to scale up your Flax modules on multiple devices and hosts.
+
+Modern JAX provides [`jax.experimental.pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html) as a way to automatically compile and scale up JAX computations using the [SPMD](https://jax.readthedocs.io/en/latest/glossary.html?highlight=spmd#term-SPMD) model. You can think of `pjit` as [`jit`](https://jax.readthedocs.io/en/latest/jax-101/02-jitting.html) where one explicitly specificies how the input and output data should be partitioned across devices. See the [JAX guide](https://jax.readthedocs.io/en/latest/jax-101/08-pjit.html) for more information on pjit.
```
8bitmp3, line 21: Suggestions:

- Do we need "modern"? (Currently, there's just one JAX 0.4, isn't it?)
- Spell out "Single Program Multi Data" before abbreviating it.
- Use the SPMD link without the `highlight=...` part like `https://jax.readthedocs.io/en/latest/glossary.html#term-SPMD)`
- Last sentence:
  - "You can think of `pjit` as [`jax.jit`]..." (with `jax.jit`).
  - Use "Refer to" instead of "See" (more accessibility-friendly language), "JAX guide" -> "JAX-101 `pjit` tutorial" (more precise).
  - `pjit` (formatting). 

```suggestion
JAX provides [`jax.experimental.pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html) as a way to automatically compile and scale up JAX computations using the [Single Program Multiple Device (SPMD)](https://jax.readthedocs.io/en/latest/glossary.html#term-SPMD) model. You can think of JAX's `pjit` as [`jax.jit`](https://jax.readthedocs.io/en/latest/jax-101/02-jitting.html) where you can explicitly specify how the input and output data should be partitioned across devices. Refer to the [JAX-101 `pjit` tutorial](https://jax.readthedocs.io/en/latest/jax-101/08-pjit.html) for more information on `pjit`.
```

8bitmp3 (commented):
@IvyZX A few suggestions for the first part of the guide. Thanks for putting this together! Amazing doc

docs/guides/spmd_api.md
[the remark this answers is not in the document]
marcvanzee: Ah makes sense, thanks for updating it!

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+from flax.core import freeze, unfreeze
+from flax.training import train_state, checkpoints
+
+import optax
+```
+
+<!-- #region id="XfuDTqt1mO6g" -->
+Import all the `pjit` related libraries. Note that they are still in the experimental package of JAX, so some API may change. In the near future, the plan is to move the `pjit` API out of the experimental status and merge it with the current `jax.jit` API.
+
+Start a device mesh using the 8 devices available. In this example, we set them as a `2x4` device mesh, same as the layout of TPU v3-8, and annotate each axis with a name. A typical way to annotate axis names is `('data', 'model')`: `'data'` being the mesh dimension used for data-parallel sharding of the batch dimension of inputs and activations, and the `'model'` being the mesh dimension used for sharding parameters of the model across devices.
```
marcvanzee, line 63: Thinking about this some more, if you have a `2x4` device and you annotate axis names with `('data', 'model')`, then this mean that if you partition over the `data` dimension you are effectively doing a two-way partition over the first dimension of the devices, and if you partition over the `model` dimension you partition it four way over the second dimension, right? It might be good to explain this explicitly, so the relation between the devices and the axis names becomes clearer.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+<!-- #endregion -->
+
+```python id="63TwLe6vNtID"
+# Input partition rule - data-parallel only
+x_spec = PartitionSpec('data', None)  # dimensions: (batch, length)
+x_spec
+```
+
+<!-- #region id="3d6b15d08980" -->
+However, to generate `PartitionSpec` for the output, we need to use some actual output as reference. One solution is to create a model and evaluate `model.init` abstractly using `jax.eval_shape`, and then use `nn.get_partition_spec` to automatically generate the `PartitionSpec`.
```
marcvanzee, line 164: "we need to use some actual output as reference" --> don't we need to generate the variable dict structure rather than the model outputs?

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+```
+
+<!-- #region id="3d6b15d08980" -->
+However, to generate `PartitionSpec` for the output, we need to use some actual output as reference. One solution is to create a model and evaluate `model.init` abstractly using `jax.eval_shape`, and then use `nn.get_partition_spec` to automatically generate the `PartitionSpec`. 
+
+The code below shows how to get the output spec if you use `flax.training.train_state` to carry out your initialization and training steps, in which case your `pjit`ted function will output a `TrainState`.
+
+> A side note: Here we define our `init_fn` as purely functional and takes `model` and `optimizer` as arguments. This is not necessary - you can simply define with `def init_fn(k, x):` and all will work fine here.
+> 
+> This guide doesn't do it because later we will show you another way to define your model and will run `init_fn` with another model instance. But this does a bit of trouble because `jax.eval_shape` only takes numeric inputs, and we have to create an abstract closure before feeding the function in.
```
marcvanzee, line 170:
```suggestion
> This guide doesn't do it because later we will show you another way to define your model and will run `init_fn` with another model instance. However, this is problematic because `jax.eval_shape` only takes numeric inputs, so we have to create an abstract closure before feeding the function in.
```

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+
+    # Return a tuple to conform with the API `nn.scan` as shown in the cell below.
+    return z, None
+
+```
+
+<!-- #region id="1gvf6_a_OeU3" -->
+Now define the `MLP` model as multiple layers of `DotReluDot`.
+
+Note that you can use `nn.scan` and other [lifted transforms](https://flax.readthedocs.io/en/latest/advanced_topics/lift.html) on the module. This will introduce another dimension on the params, and this dimension should be annotated in `metadata_params`. Since the parameters inside `DotReluDot` are already sharded along `model` axis, we don't need to partition multiple layers across `model` dimension here, and therefore denote it as `None`.
```
marcvanzee, line 124:
```suggestion
Note that you can use `nn.scan` and other [lifted transforms](https://flax.readthedocs.io/en/latest/advanced_topics/lift.html) on the module. This will introduce another dimension for the params (the dimension over which `scan` is applied), and this dimension should be speficied using the `metadata_params` argument. Since the parameters inside `DotReluDot` are already sharded along `model` axis, we don't need to partition multiple layers across `model` dimension here, and therefore denote it as `None`.
```

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+  return state
+
+# make an abstract closure here because jax.eval_shape only takes numeric values
+init_fn_abstract = functools.partial(init_fn, model=model, optimizer=optimizer)
+with mesh:
+  abstract_variables = jax.eval_shape(init_fn_abstract, k, x)
+# we need a rule on every leaf of the input variables, in this case a nested dict.
+state_spec = nn.get_partition_spec(abstract_variables)
+state_spec
+```
```
marcvanzee, line 198: The output shows now three dimensions for the `PartitionSpec`, while you only provide 2 as input. I suppose this is because of the scan, but it is a but surprising so it might be worth clarifying this.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+
+<!-- #region id="KAyPfNP9pTR4" -->
+# `pjit` the initialization and train step
```
marcvanzee, line 201: Initialization with pjit is complex! Quite a number of steps, maybe we can simplify this with a simple utility function (not now, but maybe as a follow-up?)

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+    lambda a, b: a - b, 
+    initialized_state.params['ScanDotReluDot_0'], initialized_state.params['ScanDotReluDot_0'])
+print(jax.tree_map(jnp.shape, diff))
+diff_array = diff['W1'].unbox()
+print(type(diff_array))
+print(diff_array.shape)
+```
+
+<!-- #region id="dcLFwH1dhr3v" -->
+## `pjit` the train step and inference
```
marcvanzee, line 252: It isn't clear which parts of these code blocks refer to "train step" and which refer to "inference". Maybe add some text before the inference code block to clarify that.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+  y = pjit_apply_fn(new_state, x)
+print(type(y))
+print(y.dtype)
+print(y.shape)
+```
+
+<!-- #region id="vXY2mx7WjOyv" -->
+## Profiling
+
+If you are running on a TPU pod or pod slice, you can use the `block_all` util below to profile the cell below to observe a performance increase.
```
marcvanzee, line 295:
```suggestion
If you are running on a TPU pod or pod slice, you can use the `block_all` utility function below to observe a performance increase.
```

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+  return xs
+
+with mesh:
+  new_state = block_all(pjit_step_fn(initialized_state, x))
+```
+
+<!-- #region id="GUXkJiQkotfm" -->
+# Logical axis annotation
+
+JAX auto SPMD encourages user to explore different sharding layouts to find the optimal one. To this end, in Flax you actually can annotate the dimensions of any array with more descriptive axis names, instead of only the device mesh axis names (i.e., `data` and `model`).
```
marcvanzee, line 313:
```suggestion
JAX auto SPMD encourages users to explore different sharding layouts to find the optimal one. To this end, in Flax you actually can annotate the dimensions of any array with more descriptive axis names, instead of only the device mesh axis names (i.e., `data` and `model`). 
```

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+```
+
+<!-- #region id="GUXkJiQkotfm" -->
+# Logical axis annotation
+
+JAX auto SPMD encourages user to explore different sharding layouts to find the optimal one. To this end, in Flax you actually can annotate the dimensions of any array with more descriptive axis names, instead of only the device mesh axis names (i.e., `data` and `model`). 
+
+Check out the `Logical-` model definitions below. It's exactly the same with the model above, except for two differences:
+
+1. Each axis are annotated with more concrete, meaningful names - like `embed`, `hidden`, `batch` and `layer`. These names are referred as "logical axis names" in Flax. They make the dimensional changes inside model definition more readable.
```
marcvanzee, line 317:
```suggestion
1. All axes are annotated with more concrete, meaningful names - like `embed`, `hidden`, `batch` and `layer`. These names are referred as "logical axis names" in Flax. They make the dimensional changes inside model definitions more readable.
```

marcvanzee (approved):
Awesome tutorial @IvyZX, great work and thanks a lot! Just some minor comments but overall LGTM.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+  return state
+
+pjit_step_fn = pjit(train_step,
+                    (state_spec, x_spec),  # input annotations
+                    state_spec,            # output annotations
+                    )
+```
+
+```python id="F5gy_gb_mOqR"
+with mesh:
```
andsteing, line 275: nit: in the other examples, the `with mesh:` code is in the same cell as the function being pjitted

docs/guides/spmd_api.md
andsteing, line 295: how do I see the "performance increase" ?

andsteing (commented):
Great guide! This is very helpful to make use of `nn.with_partitioning` and friends.

I'm fairly new to `pjit`, so I also marked some things where I had genuine understanding questions.

cgarciae:
Thanks Ivy! This material is super useful, I've been reading about this so personally its great to have this. Just a few remarks:
1. Given that JAX 0.4 is out, should we consider using the new APIs (e.g. `device_put` + `jit`)? I am a bit biased because this is how I am learning.
2. On one hand, showing the `scan` trick is nice as its used in real life, on the other hand its and additional complexity that might not be ideal as its yet another concept in an already packed guide.
3. Some references or explanation as to why the parameters in `DotReluDot` are sharded the way there are would be nice. I just happened learn about this before reading this guide so it made perfect sense but without it I'd have some questions.

marcvanzee:
> On one hand, showing the scan trick is nice as its used in real life, on the other hand its an additional complexity that might not be ideal as its yet another concept in an already packed guide.

+1, I think scan over layers deserves it own guide and right now it is a bit much to take in here.

But to avoid stalling this guide, maybe we can keep it in but @IvyZX  can file an issue saying we want to add a guide for scan over layers, and then link to it from here?

cgarciae:
@marcvanzee agreed. Lets merge this PR and improve the guide in the near future. Maybe just collect some of the points in this comments in an issue.

docs/guides/spmd_api.md
IvyZX: That works too. I am just trying to show a more complicated example here, in which the pjitted function output is another dataclass instead of the simple variable dict. Added the explanation in doc.

docs/guides/spmd_api.md
IvyZX: The PartitionSpec shows that the params W1 and W2 has three dimensions, which WAI. The output of both pjitted functions are `TrainState`, which contains the params.

docs/guides/spmd_api.md
IvyZX: You can use decorator `@functools.partial(pjit, ...)` like `jit`. I added the explanation in the doc.

docs/guides/spmd_api.md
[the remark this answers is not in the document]
IvyZX: Renamed to `flax_on_pjit.md`.

docs/guides/spmd_api.md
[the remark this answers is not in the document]
IvyZX: Interesting. I rewrote the introductory paragraph with more basic explanation.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+<!-- #region id="eTGIfmDclxJ2" -->
+# Setup
+
+Install Flax from head and make imports as necessary. 
+
+Note that this Colab uses `--xla_force_host_platform_device_count=8` to emulate multiple devices on a CPU environment. You may also choose to run on a multi-device TPU environment following [this guide](https://jax.readthedocs.io/en/latest/jax-101/08-pjit.html#setup), in which case you should ignore the `os.environ` cell.
+<!-- #endregion -->
+
+```python id="e3d005e672d6"
+# TODO: replace this with `pip3 install flax` when 0.6.4 release rolls out.
```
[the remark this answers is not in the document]
IvyZX: Since this is a public doc, it's probably better to not expose our names. I can open an issue to remind myself of that.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+from flax import struct, traverse_util, linen as nn
+from flax.linen import spmd
+from flax.core import freeze, unfreeze
+from flax.training import train_state, checkpoints
+
+import optax
+```
+
+<!-- #region id="XfuDTqt1mO6g" -->
+Import all the `pjit` related libraries. Note that they are still in the experimental package of JAX, so some API may change. In the near future, the plan is to move the `pjit` API out of the experimental status and merge it with the current `jax.jit` API.
```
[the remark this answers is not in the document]
IvyZX: Moving extra comments in note.

docs/guides/spmd_api.md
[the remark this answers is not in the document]
IvyZX: I put some more concrete explanation in `metadata_params`.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+  num_layers: int
+  depth: int
+  use_scan: bool
+  @nn.compact
+  def __call__(self, x):
+    if self.use_scan:
+      x, _ = nn.scan(DotReluDot, length=self.num_layers, 
+                    variable_axes={"params": 0},
+                    split_rngs={"params": True},
+                    metadata_params={nn.PARTITION_NAME: None}
```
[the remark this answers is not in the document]
IvyZX: This is just a constant to denote the partition name that annotates the `nn.scan` layer.

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+  initialized_state = pjit_init_fn(k, x, model, optimizer)
+jax.tree_map(jnp.shape, initialized_state)
+```
+
+<!-- #region id="qgVa3lqLgPLz" -->
+## Looking into the module output
+
+Note that in the output `initialized_state`, the params `W1` and `W2` are of type [`Partitioned`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.Partitioned.html). This is a wrapper around the actual JAX array that allows Flax to record metadata associated with it. You can access the raw JAX array by adding `.value` or running `.unbox()`.
+
+You can also check the underlying `.sharding` of the JAX array, which gives a hint on the way it is partitioned.
```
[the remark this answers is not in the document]
IvyZX: This is a fairly new concept in JAX and is not very well documented as for now. I only find one mentioning of OpSharding here: https://jax.readthedocs.io/en/latest/jax_array_migration.html#why-create-jax-array

docs/guides/spmd_api.md
```diff
@@ -0,0 +1,429 @@
...
+Now you can `pjit` the `init_fn` in a similar fashion as `jit`, with two extra args `in_axis_resources` and `out_axis_resources`.
+
+You need to add `with mesh:` context when running a `pjit`ted function, so that it can refer to `mesh` to allocate data on devices correctly.
+<!-- #endregion -->
+
+```python id="RFxvZjwPS5Za"
+pjit_init_fn = pjit(init_fn,
+                    static_argnums=(2, 3),
+                    in_axis_resources=(PartitionSpec(None), x_spec),  # RNG key and x
+                    out_axis_resources=state_spec
```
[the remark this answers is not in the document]
IvyZX: JAX team is making `pjit` more automatic so that user don't have to specify input and output axis. But as far as Anselm and I know, this approach is not very efficient on complex model for now, so we decided to still present the version that explicitly states the axis.

docs/guides/spmd_api.md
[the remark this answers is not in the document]
8bitmp3: fyi agreed with @IvyZX 

```
# Once Flax v0.6.4 is out, replace this with `pip3 install flax`.
```

This way, the docs can be more or less up-to-date after 0.6.4.

IvyZX:
Thank you all for the reviews! I have addressed all the comments listed here, either by a reply comment or by directly modifying the doc content. PTAL and let me know if there's any questions/concerns.

docs/guides/spmd_api.md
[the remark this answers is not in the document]
andsteing: `flax_on_pjit.md` is better, but it still does not really match the title `Scale up Flax modules on multiple devices` ...

docs/guides/spmd_api.md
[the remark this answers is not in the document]
andsteing: You could use Github username (which is already known) ... Or create an issue and assign it to yourself.

The main motivation here is to avoid accumulating TODOs in public docs.

docs/guides/flax_on_pjit.md
```diff
@@ -0,0 +1,437 @@
...
+  * The second dimension (aka. the data depth dimension) will be replicated across all devices.
+
+
+## Lifted transforms
+
+Now define the `MLP` model as multiple layers of `DotReluDot`.
+
+To replicate identical layers, you can either use `flax.linen.scan` or a for-loop: `flax.linen.scan` can offer faster compilation times, whereas the for-loop can be faster on runtime. This guide uses `flax.linen.scan` simply to show that [Flax lifted transforms](https://flax.readthedocs.io/en/latest/advanced_topics/lift.html#supported-transformations) work together with [JAX `pjit`](https://jax.readthedocs.io/en/latest/jax.experimental.pjit.html).
+
+**Note:** `flax.linen.scan` will introduce another dimension for the params (the dimension over which `scan` is applied), and you need to use the `metadata_params` argument to annotate the partition of this dimension. Since the parameters inside your `DotReluDot` (a sub-`Module`) are already sharded along the `model` axis, you don't need to partition multiple layers across the `model` dimension here, and therefore you should denote it as `None`.
```
andsteing, line 163: this "note" is formatted differently from other "notes" - is this on purpose?
image
vs
image

docs/guides/spmd_api.md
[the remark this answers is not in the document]
andsteing: it's clearer now with the added "Note:"

docs/guides/spmd_api.md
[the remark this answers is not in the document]
andsteing: ah ... so it's still a bit rough around the edges, I see.

I think I would still mention this because in many examples users will see examples of `pjit()` where the `out_axis_resources` is omitted (e.g. our very own `flax.core.meta.Partitioned`).

andsteing (approved):
I had another go at the updated guide. Nice improvements, thanks!

I left some more comments, but will approve here, since it's mostly nits and nothing is blocking.

docs/guides/flax_on_pjit.md
8bitmp3: @andsteing We are doing a 2nd (3rd?) review with @IvyZX asap, going through the second half. Will make sure "Note:" is formatted consistently :+1: Thanks for the feedback!!

8bitmp3:
Last review done https://github.com/google/flax/pull/2730/commits/dcf5517e662f925f28327755fbdd225741e99ac4. Thanks @IvyZX for the amazing work :+1:

8bitmp3 (approved):

PR #2735, opened 2022-12-17
added deep mutation contextmanager for dataclasses

chiamp:
Inspired by the work done by @brentyi in [jax_dataclasses](https://github.com/brentyi/jax_dataclasses).

Added the function `struct.create_mutable_copy` that generates a copy of a dataclass object and is mutable while inside the contextmanager.

```
from flax import struct
import dataclasses

@struct.dataclass
class Point:
  x: float
  y: float
  meta: Any = struct.field(pytree_node=False)

@struct.dataclass
class GenericDataClass:
  value: Any
  meta: Any = struct.field(pytree_node=False)

p = Point(x=1, y=2, meta={})
gdc1 = GenericDataClass(value=p, meta={})
gdc2 = GenericDataClass(value=gdc1, meta={})
gdc3 = GenericDataClass(value=gdc2, meta={})

with struct.create_mutable_copy(gdc3) as gdc3_copy:
  gdc3_copy.value.value.value.y = 3

assert gdc3_copy.value.value.value.y == 3 # confirm value was changed
assert gdc3.value.value.value.y == 2 # confirm original object is unchanged

# confirm that copied object is not mutable outside contextmanager
try:
  gdc3_copy.value.value.value.y = 4
except dataclasses.FrozenInstanceError:
  print('caught FrozenInstanceError')
```


flax/struct.py
```diff
@@ -162,11 +164,50 @@ def from_state_dict(x, state):
...
+  data_clz.__mutable__ = False # denotes whether the object's currently mutable
+  data_clz.__setattr__ = new_setattr # type: ignore
+
   # add a _flax_dataclass flag to distinguish from regular dataclasses
   data_clz._flax_dataclass = True # type: ignore[attr-defined]
 
   return data_clz # type: ignore
 
+@contextlib.contextmanager
+def create_mutable_copy(obj):
```
brentyi, line 182: Hi @chiamp!

Would you be open to adding a type annotation here, something like:
```
from typing import Generator

@contextlib.contextmanager
def create_mutable_copy(obj: _T) -> Generator[_T, None, None]:
  ...
```

It'd be nice for tab completion on the yielded pytree :slightly_smiling_face: 

I added an inner function in `jax_dataclasses` for a similar purpose, but it looks like that's not actually necessary (probably the `contextlib` stubs have improved a lot since then?).

chiamp: Thanks for the suggestion @brentyi! I added it in

flax/struct.py
```diff
@@ -162,11 +163,50 @@ def from_state_dict(x, state):
...
+    visited.add(id(obj))
+
+    if dataclasses.is_dataclass(obj) and hasattr(obj, '__mutable__'):
+      object.__setattr__(obj, '__mutable__', value)
+
+    for child_obj in get_children(obj):
+      set_mutable_attribute(child_obj, value, visited)
+
+  def get_children(obj):
+    registry_entry = _registry.get(type(obj))
```
levskaya, line 195: do we have to refer to the jax pytree private registry API, or is there a way to achieve the same with public API calls?

levskaya (commented):
looks great, just one comment

PR #2743, opened 2022-12-20
Adds some more comments to scope.py

marcvanzee:
Some of the code in `scope._mutable_collection` and `scope.put_variable` is actually quite intricate and wasn't explain well.


levskaya (approved):
This is great we - should comment the hell out of our central core/linen files.

PR #2748, opened 2022-12-20
Bind submodules using __getattribute__

### LINKED ISSUE #2744: Properly bind all submodules
Currently submodules are lazily bounded when `__getattr__` that calls `_try_setup` , however, this only works for submodules defined in `setup`, submodules that are passed as constructor arguments don't trigger `__getattr__` (since their fields always exist) and are never bounded e.g:

```python
import flax.linen as nn
import jax
import jax.numpy as jnp

module = nn.Sequential([
    nn.Dense(8),
    nn.relu,
    nn.Dense(4),
])
x = jnp.ones((1, 5))
bound_module = module.bind(variables)

print(bound_module.scope) # <flax.core.scope.Scope object at 0x7f54724fc610>
print(bound_module.layers[0].scope) # None  <== !!!
```
This adds some rough edge cases when using `bind` as some submodules become unavailable.

cgarciae:
Fixes #2744. Uses `__getattribute__` to call `_try_setup` when accessing a dataclass field. This properly binds submodules that are passed as constructor arguments instead of being defined in `setup`. This PR is based on #2028 with some minor modifications.

**NOTE**: #2028 was originally rolled back as it broke some internal tests, this PR will likely do the same. The solution would be to fix forward the internal code / tests.

### DIFF
--- a/docs/guides/convert_pytorch_to_flax.rst
+++ b/docs/guides/convert_pytorch_to_flax.rst
@@ -42,7 +42,7 @@ and the Flax kernel has shape [inC, outC]. Transposing the kernel will do the tr
   t_out = t_fc(t_x)
   t_out = t_out.detach().cpu().numpy()
 
-  np.testing.assert_almost_equal(j_out, t_out)
+  np.testing.assert_almost_equal(j_out, t_out, decimal=5)
 
 
 Convolutions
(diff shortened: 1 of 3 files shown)


marcvanzee:
We should probably also run some benchmarks to test whether this causes any slowdowns.

Issue #2733, opened 2022-12-17
Wrap `__post_init__` to check or force that `Module.__post_init__` is called


levskaya:
We added a runtime check in #2535 to emit a "IncorrectPostInitOverrideError" when users forget to call `super().__post_init__()` from a custom `__post_init__` method.  That implementation only runs the check at the top-level though, so it's not going to catch internal Modules that override `__post_init__`.

e.g.

```
import jax.numpy as jnp
from flax import linen as nn
import jax

class Foo(nn.Module):
  def __post_init__(self):
    pass
  @nn.compact
  def __call__(self, x):
    return nn.Dense(12)(x)

class Bar(nn.Module):
  @nn.compact
  def __call__(self, x):
    return Foo()(x)

b = Bar()
b.init(jax.random.PRNGKey(0), x=jnp.ones((10, 3)))
```

What I think we should do instead is to just wrap `__post_init__` on subclass instantiation to check (and possible even force) that `Module.__post_init__` is called.   

For forcing the only complex case is when users use inheritance and call their own parent-class `__post_init__` in some random order inside their child `__post_init__` (I've seen this.), but I think there's a way for a wrapper to detect this and to only run in the outermost child-defined `__post_init__`.

cgarciae:
> I think there's a way for a wrapper to detect this and to only run in the outermost child-defined __post_init__.

Maybe a context manager that contains that stores a references to the object calling `__post_init__` so we don't call it again for the same object in case there are multiple wrapped methods?

chiamp:
> For forcing the only complex case is when users use inheritance and call their own parent-class __post_init__ in some random order inside their child __post_init__ (I've seen this.), but I think there's a way for a wrapper to detect this and to only run in the outermost child-defined __post_init__.

Is there a short code example of this @levskaya?

PR #2761, opened 2022-12-30
forced post init to call super post init

chiamp:
Addresses #2733.

Wrapped `__post_init__` to force `super().__post_init__` call in Module sub-classes.


flax/linen/module.py
```diff
@@ -705,6 +705,17 @@ def __init_subclass__(cls, **kwargs: Any) -> None:
...
+    # Force all sub-classes to call `super(cls, self).__post_init__()`
+    def force_super_post_init(post_init_method):
+      @functools.wraps(post_init_method)
+      def new_post_init_method(self):
+        post_init_method(self)
+        if not hasattr(self, '_id') and hasattr(super(cls, self), '__post_init__'):
+          super(cls, self).__post_init__()
+      return new_post_init_method
+
+    cls.__post_init__ = force_super_post_init(cls.__post_init__) # type: ignore
```
levskaya, line 717: isn't this presuming that a cls.__post_init__ exists on this particular subclass?  I think the logic of your wrapper will do the right thing still (due to this recursive looking up the method resolution order)...

chiamp: I added an `if` condition to check if it has a `post_init`

PR #2767, opened 2023-01-03
rename ConvLSTM to ConvLSTMCell

cgarciae:
# What does this PR do?

Renames `ConvLSTM` to `ConvLSTMCell` for consistency. Since we are going to add full layers like `RNN` and `CudnnLSTM` in the near future, its imporant to be able to properly distinguish between cells and layers.


bastings (approved):
Thanks, makes sense! Are we ok with a breaking change?

cgarciae:
@bastings Marc was looking through the internal code base and it seems the issue is not that hard to solve, just changing a couple names. Wondering if you could help me do this in the CL? :pray: 

For OSS users we can bump the Flax version to `0.7.0`.

carlosgmartin:
@cgarciae Is it just me or is ConvLSTMCell missing from [the docs](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/layers.html)?

PR #2770, opened 2023-01-04
Add Bidirectional

cgarciae:
# What does this PR do?

* Adds a `Bidirectional` Module that takes two `RNN` instances and applies the bidirectional logic.
* Adds the `reverse` and `keep_order` arguments to `RNN` to allow the implementation of the bidirectional logic.
* Changes the type of `Array` from `Any` to `jax.Array` and fixes some type errors.

### DIFF
--- a/flax/linen/recurrent.py
+++ b/flax/linen/recurrent.py
@@ -18,10 +18,8 @@ THe RNNCell modules can be scanned using lifted transforms. For more information
 see: https://flax.readthedocs.io/en/latest/advanced_topics/lift.html.
 """
 
-from abc import abstractmethod
-import abc
 from functools import partial   # pylint: disable=g-importing-member
-from typing import Any, Callable, Mapping, Optional, Sequence, Tuple, Union, TypeVar
+from typing import Any, Callable, Dict, List, Mapping, Optional, Sequence, Tuple, Union, TypeVar, cast
 from typing_extensions import Protocol
 from absl import logging
 
@@ -47,7 +45,7 @@ A = TypeVar('A')
 PRNGKey = Any
 Shape = Tuple[int, ...]
 Dtype = Any  # this could be a real type?
-Array = Any
+Array = jax.Array
 Carry = Any
 CarryHistory = Any
 Output = Any
@@ -174,7 +172,7 @@ class DenseParams(Module):
   bias_init: Callable[[PRNGKey, Shape, Dtype], Array] = initializers.zeros_init()
 
   @compact
-  def __call__(self, inputs: Array) -> Tuple[Array, Array]:
+  def __call__(self, inputs: Array) -> Tuple[Array, Optional[Array]]:
     k = self.param(
         'kernel', self.kernel_init, (inputs.shape[-1], self.features),
         self.param_dtype)
@@ -246,14 +244,19 @@ class OptimizedLSTMCell(RNNCellBase):
     hidden_features = h.shape[-1]
 
     def _concat_dense(inputs: Array,
-                      params: Mapping[str, Tuple[Array, Array]],
-                      use_bias: bool = True) -> Array:
+                      params: Mapping[str, Tuple[Array, Optional[Array]]],
+                      use_bias: bool = True) -> Dict[str, Array]:
       # Concatenates the individual kernels and biases, given in params, into a
       # single kernel and single bias for efficiency before applying them using
       # dot_general.
-      kernels, biases = zip(*params.values())
+      kernels = [kernel for kernel, _ in params.values()]
       kernel = jnp.concatenate(kernels, axis=-1)
       if use_bias:
+        biases = []
+        for _, bias in params.values():
+          if bias is None:
+            raise ValueError('bias is None but use_bias is True.')
+          biases.append(bias)
         bias = jnp.concatenate(biases, axis=-1)
       else:
         bias = None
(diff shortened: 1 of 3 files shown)


flax/linen/recurrent.py
```diff
@@ -246,14 +244,19 @@ def __call__(self, carry: Tuple[Array, Array],
...
+                      params: Mapping[str, Tuple[Array, Optional[Array]]],
+                      use_bias: bool = True) -> Dict[str, Array]:
       # Concatenates the individual kernels and biases, given in params, into a
       # single kernel and single bias for efficiency before applying them using
       # dot_general.
-      kernels, biases = zip(*params.values())
+      kernels = [kernel for kernel, _ in params.values()]
       kernel = jnp.concatenate(kernels, axis=-1)
       if use_bias:
+        biases = []
```
cgarciae, line 255: This code is created to make `mypy` happy but also improves the correctness of the program.

flax/linen/recurrent.py
```diff
@@ -717,6 +744,12 @@ def scan_fn(
     else:
       carry, outputs = scan_output
 
+    if reverse and keep_order:
+      outputs = jax.tree_map(
```
marcvanzee, line 748: maybe factor this out into a separate function since you use it twice?

flax/linen/recurrent.py
```diff
@@ -726,12 +759,143 @@ def _select_last(sequence: A, segmentation_mask: jnp.ndarray, axis: int) -> A:
...
-    _last_idx = _expand_dims_for(last_idx, to_target=x)
+    _last_idx = _expand_dims_like(last_idx, target=x)
     x = jnp.take_along_axis(x, _last_idx, axis=axis)
     return x.squeeze(axis=axis)
 
   return jax.tree_map(_slice_array, sequence)
 
-def _expand_dims_for(x, *, to_target):
+def _expand_dims_like(x, *, target):
   """Expands the shape of 'x' to match those of 'to_target' by adding singleton dimensions."""
```
marcvanzee, line 769:
```suggestion
  """Expands the shape of `x` to match `target`'s shape by adding singleton dimensions."""
```

flax/linen/recurrent.py
```diff
@@ -726,12 +759,143 @@ def _select_last(sequence: A, segmentation_mask: jnp.ndarray, axis: int) -> A:
...
+  if segmentation_mask is None:
+    # reverse inputs and return
+    inputs = jnp.flip(inputs, axis=time_axis)
+    return inputs
+
+  lengths = jnp.sum(segmentation_mask, axis=time_axis, keepdims=True) # [*batch, 1]
+  # create indexes
+  idxs = jnp.arange(max_steps - 1, -1, -1) # [max_steps]
+  if time_major:
+    idxs = jnp.reshape(idxs, [max_steps] + [1] * num_batch_dims)
```
marcvanzee, line 816: Nit: Above you use `(1,)*dims` (inside `_expand_dims_like`), while here you use `[1]*dims` instead. I suggest using either that one or this one everywhere consistently.

flax/linen/recurrent.py
```diff
@@ -726,12 +759,143 @@ def _select_last(sequence: A, segmentation_mask: jnp.ndarray, axis: int) -> A:
...
+  # Compute the indices to put the inputs in flipped order as per above example.
+  time_axis = 0 if time_major else num_batch_dims
+  max_steps = inputs.shape[time_axis]
+
+  if segmentation_mask is None:
+    # reverse inputs and return
+    inputs = jnp.flip(inputs, axis=time_axis)
+    return inputs
+
+  lengths = jnp.sum(segmentation_mask, axis=time_axis, keepdims=True) # [*batch, 1]
```
marcvanzee, line 812: Thanks for adding the shapes as comments! I always find that very helpful.

flax/linen/recurrent.py
```diff
@@ -726,12 +759,143 @@ def _select_last(sequence: A, segmentation_mask: jnp.ndarray, axis: int) -> A:
...
   def _slice_array(x: jnp.ndarray):
-    _last_idx = _expand_dims_for(last_idx, to_target=x)
+    _last_idx = _expand_dims_like(last_idx, target=x)
     x = jnp.take_along_axis(x, _last_idx, axis=axis)
     return x.squeeze(axis=axis)
 
   return jax.tree_map(_slice_array, sequence)
 
-def _expand_dims_for(x, *, to_target):
+def _expand_dims_like(x, *, target):
```
marcvanzee, line 768: Nit: I am not sure why `target` is required to be a kwarg. I think it is also fine as a positional argument.

marcvanzee (approved):
Thank you!!

flax/linen/recurrent.py
cgarciae: changed to regular arg

flax/linen/recurrent.py
cgarciae: :)

flax/linen/recurrent.py
cgarciae: Changed everything to use list concatenations.

flax/linen/recurrent.py
cgarciae: Are we [DRY or WET](https://en.wikipedia.org/wiki/Don't_repeat_yourself)? For this use case I think abstracting away a tree_map leads to unnecessary indirection.

flax/linen/recurrent.py
```diff
@@ -651,16 +664,27 @@ def __call__(
         else it will return a tuple of the final carry and the output sequence.
       time_major: if ``time_major=False`` (default) it will expect inputs with shape
         ``(*batch, time, *features)``, else it will expect inputs with shape ``(time, *batch, *features)``.
+      reverse: overrides the ``reverse`` attribute, if ``reverse=False`` (default) the sequence
+        is processed from left to right, else it will be processed from right to left, output order
+        will be the same as input order. If ``segmentation_mask`` is passed, padding elements will
```
bastings, line 669: I think `keep_order` is still a bit difficult to understand, so good documentation is important. This might be too unclear:

```
reverse: overrides the ``reverse`` attribute, if ``reverse=False`` (default) the sequence
        is processed from left to right, else it will be processed from right to left, output order
        will be the same as input order.
```
what about:

reverse: overrides the ``reverse`` attribute, if ``reverse=False`` (default) the sequence
        is processed from left to right and returned in the original order, else it will be processed from right to left, and returned in reverse order (but with the padding values at the end).
```

maybe even a simple example could be good. e.g. 
`[1, 2, 3, 0, 0]` vs `[3, 2, 1, 0, 0]`?

bastings (commented):
Thanks for coding this up!

chiamp (approved):

Issue #1977, opened 2022-03-09
Improve nn.scan arguments documentation


marcvanzee:
(After feedback from @zhangqiaorjc)

Some `nn.scan` arguments are under-documented / make undocumented assumptions. In the example below, the scanning actually happens over the third argument of `body_fn` (`sub` and `layer_in` are ignored):

```py
def body_fn(sub, layer_in, layer_states):
  # some code

scan_fn = nn.scan(
    body_fn,
    in_axes=0,  # scan over axis 0 for layer_states only
    variable_axes=SCAN_VARIABLE_AXES,
    split_rngs=SCAN_SPLIT_RNGS)
```

However, this is not made clear in the [current documentation](https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.scan.html):

**in_axes**: Specifies the axis to scan over for the arguments. Should be a prefix tree of the arguments. Use `flax.core.broadcast` to feed an entire input to each iteration of the scan body.

We should go over all arguments of `nn.scan.` and check whethey they make any assumptions that should be mentioned.

marcvanzee:
#1980 improves some docstrings (also adds a variation of the example above in the `nn.scan` docstring). I added placeholders with `#1977` in that PR where docstrings should further be improved.

marcvanzee:
We should also replace `self.is_mutable_collection("params")` in the docstrings with `self.is_initializing()`.

PR #2774, opened 2023-01-06
Improve lifted transform docstring

zaxtax:
This improves the docstrings in some of the lifted transform functions. Fixing the examples and clarifying some of the arguments. This addresses some of the concerns in #1977

### DIFF
--- a/flax/linen/transforms.py
+++ b/flax/linen/transforms.py
@@ -1070,7 +1070,7 @@ def while_loop(
           y = nn.Dense(c.shape[-1])(c)
           return y
         c = x
-        if self.is_mutable_collection('params'):
+        if self.is_initializing():
           return body_fn(self, c)
         else:
           return nn.while_loop(cond_fn, body_fn, self, c,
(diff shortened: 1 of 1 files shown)


flax/linen/transforms.py
```diff
@@ -780,7 +780,8 @@ def body_fn(cls, carry, xs, *args, **kwargs):
 
   Args:
     target: a ``Module`` or a function taking a ``Module``
-      as its first argument.
+      as its first argument. Scanning happens from the third and all
```
jheek, line 783: Perhaps something like: The scanned function receives the module (self) and carry as the first two arguments. The remainder of the arguments are scanned over

Or any other phrasing that makes clear what the first two args are

jheek (changes requested):
Thanks, these docs could use some improvement indeed,  I left a few comments

zaxtax:
Yes I should be more clear about stating what are the first arguments
expected to be for the function

Issue #2749, opened 2022-12-21
Common Initializers Does Not Work with Bias


daskol:
It is expected that bias vector can be initialized not only with zeros but the following code fails.

```python
import jax, jax.numpy as jnp
import flax.linen as nn

layer = nn.Conv(features=32,
                kernel_size=(3, 3),
                use_bias=True,
                bias_init=nn.initializers.lecun_normal())

jax.jit(layer.init)(jax.random.PRNGKey(42), jnp.empty((5, 28, 28, 1)))

# File jax/core.py:1969, in NamedShape.__getitem__(self, idx)
#    1967 try:
#    1968   idx = operator.index(idx)
# -> 1969   return self.__positional[idx]
#    1970 except TypeError:
#    1971   pass
#
# IndexError: tuple index out of range
```

marcvanzee:
Assigning this to @zaxtax who is looking into improving our initializer documentaiton.

daskol:
Thanks. It is not only about documentation. Is is about universality and interfacing of common initialize routines as well.

marcvanzee:
Yes you are right!

zaxtax:
@daskol I agree! Part of making good documentation is writing code that's easy to explain :)

cgarciae:
The problem is that all initializes that are implemented on top of `variance_scaling` (like `lecun_normal`) require the shape to have at least 2 dimensions, so you can't pass these to `bias_init`. This should be documented, also maybe we should improve the error message.

daskol:
Agreed. Documentation should be updated.

From my perspective, `variance_scaling` initializer can be extrapolated on bias term. Bias term can be considered as a edge case of `variance_scaling`-based initializer.

Issue #1386, opened 2021-06-24
nn.Dense bias_init is poorly documented and can


billmark:
Provide as much information as possible. At least, this should include a description of your issue and steps to reproduce the problem. If possible also provide a summary of what steps or workarounds you have already tried.

### Steps to reproduce:
Whenever possible, please provide a *minimal example*. Please consider submitting it as a Colab link.

marcvanzee:
Hi @billmark , thanks for your issue! Could you tell me in a bit more detail what you were missing from the description of `kernel_init` and `bias_init` arguments?

billmark:
As I said in the original bug, "In particular, it's not clear what the parameters to the callback need to be."..  That is, what are the args passed to the kernel_init and bias_init callbacks?  How many of these args?  What types/shapes?  What do they mean?  etc.  There is *no* such information in the documentation.

FelipeMartins96:
I found myself having the same problem reported when trying to reproduce the DDPG algorithm and an error was raised when trying to use the `lecun_uniform()` on bias initialization of `nn.Dense` params.

Is there a reason why the on the kernel initialization it sends inputs and outputs `(inputs.shape[-1], self.features)` and on bias only inputs `(self.features,)`?
https://github.com/google/flax/blob/c63d9c69205629fca00ad82135213151d656ce49/flax/linen/linear.py#L171-L173
vs
https://github.com/google/flax/blob/c63d9c69205629fca00ad82135213151d656ce49/flax/linen/linear.py#L179

From what I understood, this argument is passed on to the initializers as shape parameter.

About documentation: it seems to be missing on the docs a page about the initializers available at linen.initializers.
And for the [linen.Dense docs](https://flax.readthedocs.io/en/latest/_autosummary/flax.linen.Dense.html#flax.linen.Dense.kernel_init) I think it could be included at the kernel init, and bias init documentation that the function will receive the arguments `key, shape, dtype`, where shape is  `(inputs.shape[-1], self.features)` for kernel and  `(self.features,)` for bias

FelipeMartins96:
Had the same issue. I found this issue on jax repo which I think is relevant, the initializer functions are intentionally made to not work for 1D shapes. 
https://github.com/google/jax/issues/2075#issuecomment-578465814

I think maybe a page on the docs explaining the initializers would be interesting, or maybe link to the jax docs which lists the available functions. An example notebook explaining the module weights and biases initializations could be useful.

marcvanzee:
[...]

PR #2776, opened 2023-01-07
Added initializers doc

chiamp:
Addresses #2749 and #1386.

Created initializers documentation.


docs/guides/initializers.md
```diff
@@ -0,0 +1,123 @@
...
+---
+layer = nn.Dense(features=5)
+default_params = layer.init(jax.random.PRNGKey(42), jnp.empty((1,4)))['params']
+
+layer = nn.Dense(features=5, kernel_init=lecun_normal(), bias_init=zeros)
+manual_params = layer.init(jax.random.PRNGKey(42), jnp.empty((1,4)))['params']
+
+# should get the same param values
+assert (default_params['kernel'] == manual_params['kernel']).all()
+assert (default_params['bias'] == manual_params['bias']).all()
```
marcvanzee, line 63: Currently all this shows is what the defaults of the init functions are, which one can also just see by looking at the code, so maybe it is not worth adding a code snippet for it.

Maybe it is more interesting to demonstrate how Flax uses the initial rng to generate the params? Also I would always use `np.testing.assert_allclose` when comparing JAX arrays.

```suggestion
x = jnp.empty((1,4))
rng = random.PRNGKey(0)
params = nn.Dense(features=5).init(rng, x)['params']

# Internally, Flax splits off a new RNG from the original one for each param
# using a counter which starts at 1.
kernel_rng = scope.LazyRng.create(rng, 1).as_jax_rng()

np.testing.assert_allclose(params['kernel'], lecun_normal()(kernel_rng, (4, 5)))
np.testing.assert_allclose(params['bias'], jnp.empty((5,)))
```

docs/guides/initializers.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
++++ {"id": "jPW_E1ZqHdrQ"}
+
+## `Initializer` function signature
+
++++ {"id": "3Kglqd9vuxTG"}
+
+To maintain consistency, all `Initializer` functions that are passed to the `kernel_init` and `bias_init` arguments **must follow the function signature: `[PRNGKey, Shape, Dtype] -> Array`**. Most functions in the [Flax initializer list](https://flax.readthedocs.io/en/latest/api_reference/flax.linen.html#module-flax.linen.initializers) are **builder functions** and build an `Initializer` function that follows this function signature. The two exceptions are [`flax.linen.initializers.zeros`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.zeros.html) and [`flax.linen.initializers.ones`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.ones.html), which are already `Initializer` functions that follow the function signature. This is why in the above example, we must call `lecun_normal()` to build an `Initializer` function, whereas we can directly use `zeros` since it's already an `Initializer` function.
```
marcvanzee, line 72: Thinking of "prefer simple APIs over more documentation", I am wondering whether it wouldn't be easier to just add two fake builder functions for `zeros` and `ones` as follows:

```py
def zeros(dtype: DTypeLikeInexact = jnp.float_) -> Array:
  def init(key: KeyArray,
           shape: core.Shape,
           dtype: DTypeLikeInexact = dtype) -> Array:
    del key
    return jnp.zeros(shape, dtype)
```

Then we don't need this complicated explanation and people can simply use all initializers consistently. WDYT?

docs/guides/initializers.md
```diff
@@ -0,0 +1,123 @@
...
+layer.init(jax.random.PRNGKey(42), jnp.empty((1,4)))['params'] # <-- successfully generate params
+```
+
++++ {"id": "S4X_xHHk-b4V"}
+
+## `Initializer` restrictions for `bias_init`
+
++++ {"id": "jLUFaliMy_DT"}
+
+Only the `Initializer` functions built by [`constant`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.constant.html), [`normal`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.normal.html) and [`uniform`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.uniform.html) and the `Initializer` functions [`zeros`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.zeros.html) and [`ones`](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.ones.html) can be used as valid arguments for `bias_init`.
```
marcvanzee, line 121: In principle we could use any function we want, but we just cannot use the ones you listed right? So I would remove the word "Only" here.

docs/guides/initializers.md
```diff
@@ -0,0 +1,123 @@
...
+id: RlOIOCj68p86
+outputId: 5871f462-1d24-47f9-cc20-6fa7f358ea99
+---
+layer = nn.Dense(features=5, kernel_init=zeros) # use flax.linen.initializers.zeros instead
+layer.init(jax.random.PRNGKey(42), jnp.empty((1,4)))['params'] # <-- successfully generate params
+```
+
++++ {"id": "S4X_xHHk-b4V"}
+
+## `Initializer` restrictions for `bias_init`
```
marcvanzee, line 117: Nice! Maybe we should link to this from our `bias_init` docstrings as well?

marcvanzee (changes requested):
Thanks for adding this guide, really nice! I suppose the next step is to improve docstrings? For instance, linking to this guide from docstrings seems useful.

docs/guides/initializers.md
chiamp: I think this is a great idea! It would make things much more consistent. The question is what we should do with the original [flax.linen.initializers.zeros](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.zeros.html) and [flax.linen.initializers.zeros](https://flax.readthedocs.io/en/latest/api_reference/_autosummary/flax.linen.initializers.ones.html). If we replace them with the fake builder functions you suggested, then wouldn't this break code using these initializers? On the other hand if we leave them, then I think users may get confused on what the difference is; i.e. which they should use, which ones to call versus which ones to use explicitly, etc.

docs/guides/initializers.md
chiamp: Do you mean all docstrings that contain a description for `bias_init` (like [here](https://github.com/google/flax/blob/main/flax/linen/linear.py#L167))? Something like: `bias_init: initializer function for the bias. To see restrictions on valid initializers, refer to our [guide](https://flax.readthedocs.io/en/latest/guides/initializers.html#initializer-restrictions-for-bias-init)`

PR #2783, opened 2023-01-10
Add metadata helper transform for adding axis metadata.

levskaya:
Some users have custom combinatory/transform layers that effectively add an axis to their layer parameters, but that aren't simple vmaps or scans.  This adds a metadata helper transform for adding an axis annotation "on the way out" for inits and mutations, and removing that axis "on the way in" during application, exactly the same as the behavior used in scan/vmap - but only touching the metadata of boxed variables, and not modifying any traced variables.

### DIFF
--- a/flax/linen/__init__.py
+++ b/flax/linen/__init__.py
@@ -135,6 +135,7 @@ from .transforms import (
   while_loop as while_loop,
   cond as cond,
   switch as switch,
+  add_metadata_axis,
 )
 from .summary import tabulate
 # pylint: enable=g-multiple-import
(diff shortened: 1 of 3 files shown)


flax/core/lift.py
```diff
@@ -1394,3 +1394,63 @@ def inner_loop(scope, carry):
 def _unzip2(xs):
   ys = tuple(zip(*xs))
   return ys if ys else ((), ())
+
+
+def add_metadata_axis(
```
jheek, line 1399: Shouldn't this just be a thin wrapper around map_variables?

levskaya: Yeah - I rewrote it.  Not sure why I didn't do that the first time, I was kinda psycho with sleep deprivation at the time.  Just a small helper in transforms.py now.

jheek (approved):
Looks good to me

PR #2792, opened 2023-01-12
improve Module.init docs

### LINKED ISSUE #2653: Improve `Module.init()` docstring
Right now the docstring of [`Module.init()`](https://github.com/google/flax/blob/df7fa7eed9965ab5117a50eee2813b8dc7c90426/flax/linen/module.py#L1351) is very short. It should explain much more and give more examples, including:

* Explain that users can pass both `{"params": rng}` as well as `rng` in case the only RNG used is the one for variable collection `params` (also link to the [Variables](https://flax.readthedocs.io/en/latest/api_reference/flax.linen.html#module-flax.core.variables) documentation).
* Give a few examples of how to use the various arguments.

We may have to change the docstring once #2428 is in as well.

cgarciae:
# What does this PR do?

Fixes #2653. Improves `Module.init`'s docs.

* Adds basic example
* Adds example of how to use rngs as a dict.
* Improves jitted init example.
* Test examples with doctest

Preview: https://flax--2792.org.readthedocs.build/en/2792/api_reference/flax.linen.html#flax.linen.Module.init

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1488,8 +1488,7 @@ class Module:
            **kwargs) -> FrozenVariableDict:
     """Initializes a module method with variables and returns modified variables.
 
-    ``init`` takes in a ``PRNGKey`` or dictionary of ``PRNGKey`` as its first argument and all
-    the ``*args`` and ``**kwargs`` accepted by ``method`` (``__call__`` by default), and returns
+    ``init`` takes as first argument either a single ``PRNGKey``, or a dictionary mapping variable collections names to their ``PRNGKeys``, and will call ``method`` (which is the module's ``__call__`` function by default) passing ``*args`` and ``**kwargs``, and returns
     a dictionary of initialized variables.
 
     Example::
@@ -1510,7 +1509,7 @@ class Module:
       >>> key = jax.random.PRNGKey(0)
       >>> variables = module.init(key, jnp.empty((1, 7)), train=False)
 
-    If you pass a single ``PRNGKey`` Flax will use it to feed the ``'params'`` RNG stream.
+    If you pass a single ``PRNGKey``, Flax will use it to feed the ``'params'`` RNG stream.
     If you want to use a different RNG stream or need to use multiple streams, you must pass a
     dictionary mapping each RNG stream name to its corresponding ``PRNGKey`` to ``init``.
 
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1488,12 +1488,61 @@ class Module:
            **kwargs) -> FrozenVariableDict:
     """Initializes a module method with variables and returns modified variables.
 
+    ``init`` takes in a ``PRNGKey`` or dictionary of ``PRNGKey`` as its first argument and all
+    the ``*args`` and ``**kwargs`` accepted by ``method`` (``__call__`` by default), and returns
+    a dictionary of initialized variables.
+
+    Example::
+
+      >>> import flax.linen as nn
+      >>> import jax.numpy as jnp
+      >>> import jax
+      ...
+      >>> class Foo(nn.Module):
+      ...   @nn.compact
+      ...   def __call__(self, x, train):
+      ...     x = nn.Dense(16)(x)
+      ...     x = nn.BatchNorm(use_running_average=not train)(x)
+      ...     x = nn.relu(x)
+      ...     return nn.Dense(1)(x)
+      ...
+      >>> module = Foo()
+      >>> key = jax.random.PRNGKey(0)
+      >>> variables = module.init(key, jnp.empty((1, 7)), train=False)
+
+    If you pass a single ``PRNGKey`` Flax will use it to feed the ``'params'`` RNG stream.
+    If you want to use a different RNG stream or need to use multiple streams, you must pass a
+    dictionary mapping each RNG stream name to its corresponding ``PRNGKey`` to ``init``.
+
+    Example::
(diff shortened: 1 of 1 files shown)


flax/linen/module.py
```diff
@@ -1488,12 +1488,61 @@ def init(self,
            **kwargs) -> FrozenVariableDict:
     """Initializes a module method with variables and returns modified variables.
 
+    ``init`` takes a ``PRNGKey`` as its first argument and all the ``*args`` and ``**kwargs``
+    that the module's ``__call__`` method takes, and returns a dictionary of initialized
```
8bitmp3, line 1492: Nit: `that the Module's ...`

marcvanzee, line 1492: Doesn't have to be `__call__`, it `__call__` by default but if argument `method` is specified it will use that method.

flax/linen/module.py
```diff
@@ -1488,12 +1488,61 @@ def init(self,
            **kwargs) -> FrozenVariableDict:
     """Initializes a module method with variables and returns modified variables.
 
+    ``init`` takes a ``PRNGKey`` as its first argument and all the ``*args`` and ``**kwargs``
```
marcvanzee, line 1491: It takes either a single rng key or a dict mapping variable collections to rngs as the first argument, so maybe clarify that immediately? The current statement is incorrect.

flax/linen/module.py
```diff
@@ -1488,12 +1488,61 @@ def init(self,
...
+      ...     x = nn.relu(x)
+      ...     return nn.Dense(1)(x)
+      ...
+      >>> module = Foo()
+      >>> key = jax.random.PRNGKey(0)
+      >>> variables = module.init(key, jnp.empty((1, 7)), train=False)
+
+    By default the ``PRNGKey`` will only be used to feed the ``'params'`` RNG stream. If you
+    want to use a different RNG stream or need to use multiple streams, you must pass a
+    dictionary of ``PRNGKey`` to ``init``.
```
marcvanzee, line 1515:
```suggestion
    dictionary mapping variable collections to rngs to ``init``.
```

flax/linen/module.py
```diff
@@ -1488,12 +1488,61 @@ def init(self,
...
+      ...     # Add gaussian noise
+      ...     noise_key = self.make_rng('noise')
+      ...     x = x + jax.random.normal(noise_key, x.shape)
+      ...
+      ...     return nn.Dense(1)(x)
+      ...
+      >>> module = Foo()
+      >>> rngs = {'params': jax.random.PRNGKey(0), 'noise': jax.random.PRNGKey(1)}
+      >>> variables = module.init(rngs, jnp.empty((1, 7)), train=False)
+
```
marcvanzee, line 1535: Perhaps say that we strongly recommend always jitting init. Actually maybe we should even throw a warning if someone calls init without jitting?

marcvanzee (approved):

flax/linen/module.py
cgarciae: Not sure we should strongly recommend it if we are going to implement JAXPR based init in the future.

flax/linen/module.py
cgarciae: I originally did it for narrative purposes, you are right, fixed.

flax/linen/module.py
cgarciae: done

flax/linen/module.py
cgarciae: fixed

flax/linen/module.py
marcvanzee: Hmm yes that is true. Then I think we should refer to the jaxpr-based init function from here when it is there.

marcvanzee (approved):

flax/linen/module.py
```diff
@@ -1488,12 +1488,61 @@ def init(self,
            **kwargs) -> FrozenVariableDict:
     """Initializes a module method with variables and returns modified variables.
 
+    ``init`` takes in a ``PRNGKey`` or dictionary of ``PRNGKey`` as its first argument and all
+    the ``*args`` and ``**kwargs`` accepted by ``method`` (``__call__`` by default), and returns
+    a dictionary of initialized variables.
```
marcvanzee, line 1493:
```suggestion
    ``init`` takes as first argument either a single ``PRNGKey``, or a dictionary mapping variable collections names to their ``PRNGKeys``, and will call ``method`` (which is the module's ``__call__`` function by default) passing ``*args`` and ``**kwargs``, and returns
    a dictionary of initialized variables.
```

flax/linen/module.py
```diff
@@ -1488,12 +1488,61 @@ def init(self,
...
+      ...     x = nn.Dense(16)(x)
+      ...     x = nn.BatchNorm(use_running_average=not train)(x)
+      ...     x = nn.relu(x)
+      ...     return nn.Dense(1)(x)
+      ...
+      >>> module = Foo()
+      >>> key = jax.random.PRNGKey(0)
+      >>> variables = module.init(key, jnp.empty((1, 7)), train=False)
+
+    If you pass a single ``PRNGKey`` Flax will use it to feed the ``'params'`` RNG stream.
```
marcvanzee, line 1513:
```suggestion
    If you pass a single ``PRNGKey``, Flax will use it to feed the ``'params'`` RNG stream.
```

marcvanzee (approved):

PR #2809, opened 2023-01-18
Allow specifying method as a string

cgarciae:
# What does this PR do?

Allows passing `method` as a string to `apply`. This will simply get the specified attribute from `self`.

## Rationale

Being able to specify the method without having direct access to the instance is important to improve the ergonomics of abstractions like `TrainState` where you only have access to `apply` via `TrainState.apply_fn`. 

Currently we promote the use of `state` like this:

```python
@jax.jit
def train_step(state, batch):
  ...
  logits = state.apply_fn({'params': state.params, ...)
  ...
```
However, if you happen to need to use a method you now have to pass it somehow e.g:

```python
@partial(jax.jit, static_argnums=(2,))
def train_step(state, batch, method):
  ...
  logits = state.apply_fn({'params': state.params, ..., method=method)
  ...
```
This involves a couple of changes throughout the code base for a relatively simple operation. With the proposed change you can now specify it very simply as:

```python
@jax.jit
def train_step(state, batch):
  ...
  logits = state.apply_fn({'params': state.params, ..., method='some_method')
  ...
```

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -1406,7 +1406,8 @@ class Module:
         The "params" PRNG sequence is used to initialize parameters.
       method: A function to call apply on. This is generally a function in the
         module. If provided, applies this method. If not provided, applies the
-        ``__call__`` method of the module.
+        ``__call__`` method of the module. A string can also be provided to
+        specify a method by name.
       mutable: Can be bool, str, or list. Specifies which collections should be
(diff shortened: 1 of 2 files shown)


jheek (approved):
LGTM

cgarciae:
@jheek added the same behavior for `init`

PR #2816, opened 2023-01-25
Implement lazy init

### LINKED ISSUE #2428: Re-introduce pre-Linen JAXPR based lazy init into Linen
Various internal users have been hit by slowdown due to Linen's initialization logic: in Linen, we initialize a model and then require the user to `jit` it, relying one XLA's Dead Code Elimination logic to optimize this. However, this does not always work and it seem useful to bring back the pre-Linen way of initializing.

jheek:
Re-implement lazy init, Fixes #2428 

With this PR you can call ``module.lazy_init(rng, jax.ShapeDtypeStruct(input_shape, input_dtype)`` to init without computing the forward pass.

### DIFF
--- a/flax/core/partial_eval.py
+++ b/flax/core/partial_eval.py
@@ -17,20 +17,20 @@ def _maybe_unknown(x: Any) -> pe.PartialVal:
 
 
 def lazy_init(fn):
-  """Lazily evaluate a function by using the shapes of the inputs.
+  """Lazily evaluates a function by using the shapes of the inputs.
 
   The returned function accepts a combination of JAX values and
   ``jax.ShapeDtypeStruct`` instances for the inputs for which we
   don't need concrete values (only the shape and dtype).
 
   This API is used by ``core.lazy_init`` or ``Module.lazy_init``
-  to initialize variables without doing any actual compute on the
+  to initialize variables without doing any actual computation on the
   inputs.
 
   Args:
     fn: the function to be lazily evaluated.
   Returns:
-    A new function that accepts a mix of concrete valeus and
+    A new function that accepts a mix of concrete values and
     ``jax.ShapeDtypeStruct`` instances.
   """
   @functools.wraps(fn)
@@ -40,7 +40,7 @@ def lazy_init(fn):
     inputs_flat, in_tree = jax.tree_util.tree_flatten((args, kwargs))
     f_flat, out_tree = jax.api_util.flatten_fun(lu.wrap_init(fn), in_tree)
     # map inputs to PartialVal known/unknown
-    # only compute depending on knowns will be computed
+    # only the computations depending on knowns will be executed
     in_pvals = [_maybe_unknown(x) for x in inputs_flat]
     _, out_pvals, _ = pe.trace_to_jaxpr_nounits(f_flat, in_pvals)
     # all outputs should be knowns. If this fails
--- a/flax/core/partial_eval.py
+++ b/flax/core/partial_eval.py
@@ -1,0 +1,58 @@
+from typing import Any
+
+import functools
+
+import jax
+from jax import linear_util as lu
+from jax.interpreters import partial_eval as pe
+
+from flax import errors
+
+
+def _maybe_unknown(x: Any) -> pe.PartialVal:
+  if isinstance(x, jax.ShapeDtypeStruct):
+    return pe.PartialVal.unknown(jax.ShapedArray(x.shape, x.dtype))
+  else:
+    return pe.PartialVal.known(x)
+
+
+def lazy_init(fn):
+  """Lazily evaluate a function by using the shapes of the inputs.
+
+  The returned function accepts a combination of JAX values and
+  ``jax.ShapeDtypeStruct`` instances for the inputs for which we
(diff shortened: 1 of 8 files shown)


flax/linen/module.py
```diff
@@ -1526,6 +1527,43 @@ def init(self,
...
+  def lazy_init(self,
+           rngs: Union[PRNGKey, RNGSequences],
+           *args,
+           method: Optional[Callable[..., Any]] = None,
+           mutable: CollectionFilter = DenyList('intermediates'),
+           **kwargs) -> FrozenVariableDict:
+    """Initializes a module without computing on an actual input.
+
+    lazy_init will initialize the variables without doing unnecessary compute.
+    The input data should be passed as a ``jax.ShapeDtypeStruct`` which specifies
```
cgarciae, line 1540: Can we accept any type that has `.shape` + `.dtype` and convert it to `ShapeDtypeStruct` under the hood? It would be nice if users could just switch from `init` to `lazy_init` without further changes.

jheek: Previously we did something magic like that but sometimes you need a real value instead a ShapeDtypeStruct for this to work

jheek: done

flax/core/scope.py
```diff
@@ -943,6 +944,38 @@ def wrapper(rngs, *args, **kwargs) -> Tuple[Any, VariableDict]:
   return wrapper
 
 
+def lazy_init(fn: Callable[..., Any],
+         mutable: CollectionFilter = True,
+         flags: Optional[Mapping] = None) -> Callable[..., Any]:
+  """Functionalize a `Scope` function for lazy initialization.
```
marcvanzee, line 950:
```suggestion
  """Functionalizes a `Scope` function for lazy initialization.
```

flax/errors.py
```diff
@@ -59,6 +59,37 @@ def __init__(self, message):
...
+  Example::
+
+    class Foo(nn.Module):
+      @compact
+      def __call__(self, x):
+        # This parameter depends on the input x
+        # this causes an error when using lazy_init.
+        k = self.param("kernel", lambda _: x)
+        return x * k
+    Foo().lazy_init(random.PRNGKey(0), jax.ShapeDtypeStruct((8, 4), jnp.float32))
```
marcvanzee, line 84: Maybe add that it does work if you use actual values (right?):

`Foo().lazy_init(random.PRNGKey(0), jnp.zeros((8, 4)))`

flax/core/partial_eval.py
```diff
@@ -0,0 +1,58 @@
...
+
+def lazy_init(fn):
+  """Lazily evaluate a function by using the shapes of the inputs.
+
+  The returned function accepts a combination of JAX values and
+  ``jax.ShapeDtypeStruct`` instances for the inputs for which we
+  don't need concrete values (only the shape and dtype).
+
+  This API is used by ``core.lazy_init`` or ``Module.lazy_init``
+  to initialize variables without doing any actual compute on the
```
marcvanzee, line 27: What happens if there exists a Module that initializes variables dependent on some input, and we provide a concrete JAX array for this input when calling lazy init? In that case I think we *do* execute a computation on the inputs (otherwise we cannot create the variables). So maybe it is more correct to say that we don't do any compute on the inputs if the variables do not depend on them?

marcvanzee (changes requested):
Nice!

flax/core/partial_eval.py
```diff
@@ -0,0 +1,58 @@
...
+  @functools.wraps(fn)
+  def wrapper(*args, **kwargs):
+    # TODO(mattjj,jheek): use a public JAX API
+    # flatten fn and prepare for internal JAX transform
+    inputs_flat, in_tree = jax.tree_util.tree_flatten((args, kwargs))
+    f_flat, out_tree = jax.api_util.flatten_fun(lu.wrap_init(fn), in_tree)
+    # map inputs to PartialVal known/unknown
+    # only compute depending on knowns will be computed
+    in_pvals = [_maybe_unknown(x) for x in inputs_flat]
+    _, out_pvals, _ = pe.trace_to_jaxpr_nounits(f_flat, in_pvals)
```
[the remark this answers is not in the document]
jheek: This is adapted from a snippet provided by the JAX team a long time ago as a workaround for partial eval not being available as a public API. I don't recall the precise differences between the variants and indeed it doesn't seem documented either

flax/errors.py
jheek: You shouldn't do that though because you risk unnecessary compute again.

flax/core/partial_eval.py
jheek: I refer to inputs as the things provided with a ShapeDtypeStruct so doing any computations on them would cause an error (because the value isn't actually available)

marcvanzee: Why we would allow non-abstract inputs then in the first place? Why not only always ShapeDtypeStructs?

marcvanzee (approved):
Approved it so you can submit it before you leave but I do have one clarification question.

PR #2817, opened 2023-01-26
feat: create pyproject.toml

SauravMaheshkar:
# What does this PR do?

This PR aims to minimize the number of configuration files for the project. Instead of a separate `.style.yapf`, `mypy.ini`, `pytest.ini`, `pytype.cfg` and `setup.cfg`, we now would have all the configurations within the `pyproject.toml` file leading to a overall minimal code structure.

#### Some Thoughts

* I'd recommend some updates to the `.pre-commit-config.yaml` such as the addition of a "check-toml" hook.
* If this change is acceptable we could also move the `pylint` configuration within the `pyproject.toml` file.

### DIFF
--- a/flax/training/dynamic_scale.py
+++ b/flax/training/dynamic_scale.py
@@ -18,7 +18,7 @@
 import functools
 from typing import Any, Callable, NamedTuple, Optional, Sequence, Union
 
-from .. import struct
+from flax import struct
 
 import jax
 from jax import lax


SauravMaheshkar:
I don't know if I should squash these commits or not. What if in the future we'd like to rollback one particular config file.

cgarciae:
Hey @SauravMaheshkar, yes please squash your commits, we don't squash commits on merge contributors have to do it.

SauravMaheshkar:
@cgarciae, I opened a new PR #2828

PR #2833, opened 2023-02-01
Introduce relaxed naming policy behind a feature flag.

levskaya:
Historically Flax had strong opinions about how variables and submodules could be named, in the spirit of maintaining a single coherent namespace amongst python attributes, variable, and submodule names.

It turns out we never really enforced that "single namespace" concept correctly, and furthermore, that the motivations for such a thing were very weak at best and any possible advantages left unrealized.

In practice, these name constraints are just really annoying to our users and mostly just get in the way, especially with advanced configuration systems.

This PR removes the pointless constraints:

 - "Adopted" modules brought in from the outside, as commonly occurs with configuration systems like fiddle, gin, etc. are now able to carry "scope names" which are not reset on adoption during module construction.
 - Variables in different collections in the same scope/module can have the same "scope name" in the variable nested dict.  Conflicts between submodule names and variable names are still enforced.
 - We no longer force non-interference of variable "scope names" with python attributes on the module.  We only enforce non-collision within the backing scope itself.

Since some of these changes could potentially change checkpoints in certain situations, this new behavior is released behind a flag `flax_relaxed_naming` for the time being.

### DIFF
--- a/flax/configurations.py
+++ b/flax/configurations.py
@@ -85,3 +85,8 @@ flax_use_orbax_checkpointing = define_bool_state(
     name='use_orbax_checkpointing',
     default=False,
     help=('Whether to use Orbax to save checkpoints.'))
+
+flax_relaxed_naming = define_bool_state(
+    name='relaxed_naming',
+    default=False,
+    help=('Whether to relax naming constraints.'))
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -403,7 +403,7 @@ class Scope:
   <https://github.com/google/flax/tree/main/tests/core/design>`_
   for a number of examples using ``Scopes``.
   """
-  reservations: Set[str]
+  reservations: Dict[str, Optional[str]]
 
   def __init__(self,
                variables: MutableVariableDict,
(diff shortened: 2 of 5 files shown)


jheek (approved):
LGTM

flax/configurations.py
```diff
@@ -85,3 +85,8 @@ def static_bool_env(varname: str, default: bool) -> bool:
     name='use_orbax_checkpointing',
     default=False,
     help=('Whether to use Orbax to save checkpoints.'))
+
+flax_relaxed_naming = define_bool_state(
```
jheek, line 89: Should we just set the default to True here?

cgarciae: +1

levskaya: I'll give it a try - if it breaks too much I might have to do a staged off by default, we'll see soon.

levskaya: Flipping it on immediately is definitely going to break some code <-> checkpoint agreement in the wild, particularly for people using fancy config systems.  So we'll have to start with it off and alert some of our users first before defaulting to the new mode.

PR #2839, opened 2023-02-02
Improve nn.scan docs

### LINKED ISSUE #2754: Initialization of Submodules Lifted with `flax.nn.scan`
One more issue 😄 . Promise this is the last one. There are a lot of question about `flax.nn.scan` and RTD and existing GitHub issues do not solve them.

With very deep model compilation times become insane and it takes about 1 hour to compile model for Nvidia runtime. So, I decided to prevent loop unrolling with `jax.lax.scan` and its lifting counterpart `flax.nn.scan`. However, I faced multiple issues. Incomplete list of issues follows.

1. There is no clear way to initialize scanned submodules. I came up with solution to pass everything as `args` and `kwargs` to `__call__` of submodule (in this case `MLP`).
2. There is no keyword argument of `flax.nn.scan` as RTD says.
3. Func `flax.nn.scan` always returns `(carry, args)` even if there is only `carry` and no `args`.
4. RTD says that `target` should be either a type of `nn.Module` or a function which accepts `nn.Module` (type?) as its first position argument.
5. If one specified `name` of modules in `MLP` then an exception is thrown. It is a bit strange because all parameter trees merged to a single parameter tree.

```python
import flax.linen as nn
import jax
import jax.numpy as jnp


def initializer(val):
    def init(key, shape, dtype):
        return jnp.full(shape, val, dtype)

    return init


class MLP(nn.Module):

    @nn.compact
    def __call__(self, xs, var):
        h = nn.Dense(features=2, kernel_init=initializer(var))(xs)
        h = nn.relu(h)
        h = nn.Dense(features=2, kernel_init=initializer(var))(xs)
        return xs + h, None


class Transformer(nn.Module):

    length: int = 3

    def setup(self):
        def fn(self, *args, **kwargs):
            return MLP(self, *args, **kwargs)

        # FAIL: Function instead of derived type from nn.Module does not work.
        #
        #   ScanMLP = nn.scan(target=fn, ...)
        #
        #   jax._src.traceback_util.UnfilteredStackTrace: TypeError:
        #   Transformer.setup.<locals>.fn() missing 1 required positional
        #   argument: 'self'

        # OK: No problems.
        ScanMLP = nn.scan(target=fn,
                          variable_axes={'params': 0},
                          variable_broadcast=False,
                          split_rngs={'params': True},
                          length=self.length)

        self.vars = jnp.arange(self.length)  # e.g. [0, 1, 2]
        self.mlp = ScanMLP()  # FAIL: ScanMLP(self.vars)

    @nn.compact  # OK: This decorator does nothing. Why?
    def __call__(self, xs):
        carry, out = self.mlp(xs, self.vars)  # OK: Axis 0 (implicitely).
        assert out is None
        return carry


model = Transformer(length=1250)
ys, state = jax.jit(model.init_with_output)(jax.random.PRNGKey(42),
                                            jnp.ones((3, 2)))
kernel = state['params']['mlp']['Dense_0']['kernel']
assert (kernel[0, ...] == jnp.zeros((2, 2))).all()
assert (kernel[1, ...] == jnp.ones((2, 2))).all()
```

In this experiments flax v0.6.3 and jax v0.4.1 are used.

cgarciae:
# What does this PR do?

Fixes #2754. Improves the scan documentation in the following ways:
* Improves the current LSTM example to make it more readable.
* Adds a new example showing how to scan over layers.
* Adds a reference to `remat_scan`

### DIFF
--- a/flax/linen/transforms.py
+++ b/flax/linen/transforms.py
@@ -738,47 +738,85 @@ def scan(target: Target,
 
   Example::
 
-    import flax
-    import flax.linen as nn
-    from jax import random
-
-    class SimpleScan(nn.Module):
-      @nn.compact
-      def __call__(self, c, xs):
-        LSTM = nn.scan(nn.LSTMCell,
-                       variable_broadcast="params",
-                       split_rngs={"params": False},
-                       in_axes=1,
-                       out_axes=1)
-        return LSTM()(c, xs)
-
-    seq_len, batch_size, in_feat, out_feat = 20, 16, 3, 5
-    key_1, key_2, key_3 = random.split(random.PRNGKey(0), 3)
-
-    xs = random.uniform(key_1, (batch_size, seq_len, in_feat))
-    init_carry = nn.LSTMCell.initialize_carry(key_2, (batch_size,), out_feat)
-
-    model = SimpleScan()
-    variables = model.init(key_3, init_carry, xs)
-    out_carry, out_val = model.apply(variables, init_carry, xs)
-
-    assert out_val.shape == (batch_size, seq_len, out_feat)
-
+    >>> import flax.linen as nn
+    >>> import jax
+    >>> import jax.numpy as jnp
+    ...
+    >>> class LSTM(nn.Module):
+    ...   features: int
+    ...
+    ...   @nn.compact
+    ...   def __call__(self, x):
+    ...     batch_size = x.shape[0]
+    ...     ScanLSTMCell = nn.scan(
+    ...       nn.LSTMCell, variable_broadcast="params",
+    ...       split_rngs={"params": False}, in_axes=1, out_axes=1)
+    ...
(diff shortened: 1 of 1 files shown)


flax/linen/transforms.py
```diff
@@ -780,6 +777,36 @@ def body_fn(cls, carry, xs, *args, **kwargs):
...
+  to an input. For example::
+
+    >>> class ResidualMLPBlock(nn.Module):
+    ...   @nn.compact
+    ...   def __call__(self, x, _):
+    ...     h = nn.Dense(features=2)(x)
+    ...     h = nn.relu(h)
+    ...     return x + h, None
+    ...
+    >>> class RecidualMLP(nn.Module):
```
daskol, line 792: It seems that there is a typo in `RecidualMLP`.

IvyZX (approved):

PR #2881, opened 2023-02-16
Added RMS normalization layer

### LINKED ISSUE #2849: Incorporate RMSNorm
Is there interest to incorporate RMSNorm (https://arxiv.org/pdf/1910.07467.pdf) either as a seperate normalization module or to make LayerNorm configurable to switch off the centering? It is becoming quite popular as a faster alternative to LayerNorm that maintains performance and is used for example in Palm.

chiamp:
Resolves #2849.

Added an optional argument `use_mean` in the `_compute_stats` function in `flax/linen/normalization.py`, which will compute the mean and variance if set to `True`, and will set the mean to 0 and compute the variance without subtracting the mean if set to `False`. The latter mode is useful as square rooting this "variance" value (which is done in the `_normalize` function) will give you the RMS.

### DIFF
--- a/flax/linen/normalization.py
+++ b/flax/linen/normalization.py
@@ -49,7 +49,8 @@ def _abs_sq(x):
 def _compute_stats(x: Array, axes: Optional[Axes],
                    dtype: Optional[Dtype],
                    axis_name: Optional[str] = None,
-                   axis_index_groups: Any = None):
+                   axis_index_groups: Any = None,
+                   use_mean: bool = True):
   """Computes mean and variance statistics.
 
   This implementation takes care of a few important details:
@@ -68,6 +69,9 @@ def _compute_stats(x: Array, axes: Optional[Axes],
       are always at least float32 for stability (default: dtype of x).
     axis_name: Optional name for the pmapped axis to compute mean over.
     axis_index_groups: Optional axis indices.
+    use_mean: If true, calculate the mean from the input and use it when
+      computing the variance. If false, set the mean to zero and compute
+      the variance without subtracting the mean.
 
   Returns:
     A pair ``(mean, var)``.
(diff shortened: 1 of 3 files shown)


flax/linen/normalization.py
```diff
@@ -335,6 +343,70 @@ def __call__(self, x):
         self.bias_init, self.scale_init)
 
 
+class RMSNorm(Module):
```
marcvanzee, line 346: Could you please add an example of how to use the layer? I think we should starting doing this for every layer, like @cgarciae does in his RNN PR: https://github.com/google/flax/pull/2604/files#r1107264719.

levskaya (approved):
Looks good!  Thanks!! -- I second Marc's ask to add a small usage example in the docstring.

chiamp:
@marcvanzee @levskaya I added a [docstring](https://github.com/google/flax/pull/2881/files#diff-f025750faae1b20976622f833c327e7c3c9bf25f48a5de676eb9744febd0d8a5R356), let me know if this works!

levskaya (approved):

levskaya:
@chiamp - I added a exception for the deprecation warning, your tests all seem to pass now!

chiamp:
> @chiamp - I added a exception for the deprecation warning, your tests all seem to pass now!

Thanks @levskaya!

flax/linen/normalization.py
```diff
@@ -335,6 +343,80 @@ def __call__(self, x):
...
+    axis_index_groups: groups of axis indices within that named axis
+      representing subsets of devices to reduce over (default: None). For
+      example, `[[0, 1], [2, 3]]` would independently batch-normalize over
+      the examples on the first two and last two devices. See `jax.lax.psum`
+      for more details.
+  """
+  epsilon: float = 1e-6
+  dtype: Optional[Dtype] = None
+  param_dtype: Dtype = jnp.float32
+  use_bias: bool = True
```
levskaya, line 390: sorry I just noticed this - we probably don't want `use_bias` and `bias_init` here since we're never adjusting the offset?

Issue #2707, opened 2022-12-10
0.6.3 introduced a circular dependency with orbax


mattsoulanille:
Provide as much information as possible. At least, this should include a description of your issue and steps to reproduce the problem. If possible also provide a summary of what steps or workarounds you have already tried.

### System information
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04):
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`:
- Python version:
- GPU/TPU model and memory:
- CUDA version (if applicable):

### Problem you have encountered:
Flax 0.6.3 added a [dependency on orbax](https://github.com/google/flax/blob/main/setup.py#L33), which [has a dependency on flax](https://github.com/google/orbax/blob/main/pyproject.toml#L28). This is causing https://github.com/tensorflow/tfjs/issues/7159 in the TensorFlow.js repository. TFJS resolves pypi packages using Bazel, which does not support circular dependencies.

Was this change intentional? If so, I can file a bug with rules_python instead, although [last time this kind of circular dependency issue arose, it was determined to be a bug in the downstream package](https://github.com/bazelbuild/rules_python/issues/439#issuecomment-810736253). I'm not sure if that true in this case, though.
### What you expected to happen:
No circular dependency.

### Steps to reproduce:
Whenever possible, please provide a *minimal example*. Please consider submitting it as a Colab link.

mattsoulanille:
Thanks! TFJS has pinned flax to 0.6.2 for now, so we are unblocked for the moment.

IvyZX:
In internal equivalents of Bazel, the BUILD rules are specified so that there is no actual code-level circular dependency between Flax and Orbax. Is it possible to specify BUILD rules in the open-source codebase so that Bazel knows how to run it?

IvyZX:
Discussed with @mattsoulanille and @levskaya offline. TFJS seems to be okay with pinning `flax==0.6.2` for a while, since it seems a bit tricky to figure out a Bazel-based solution for TFJS that recognizes Flax and Orbax dependency graph at code-level.

But in the long term this circular dependency of Flax and Orbax may be a blocker for other installations as well. Generally the right direction is probably to cut Orbax's dependency on Flax, since Orbax is designated for general (JAX/TF) checkpointing.

Right now Orbax is dependent on Flax for `serialization.py` and `traverse_util.py` packages. 
[...]

PR #2882, opened 2023-02-16
Make orbax an optional dependency

daskol:
The issue is `orbax` itself depends on `flax`. So, we have a cycle dependency between packages. Despite that the cycle is not that harmful (pip install flax and `orbax` without any issues), it is better to untie cycle in order to avoid any possible edge cases in package dependency resolution.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,11 +30,14 @@ install_requires = [
     "matplotlib",  # only needed for tensorboard export
     "msgpack",
     "optax",
-    "orbax",
     "tensorstore",
     "rich>=11.1",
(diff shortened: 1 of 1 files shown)


samuela:
Is there a reason not to merge this PR? IIUC it would resolve #2707

daskol:
@samuela I guess, it's stale. Orbax maintainers manages dependencies and split `orbax` on two namespace packages. There is not circular dependency any more but there is an issue with namespaces (see https://github.com/google/orbax/issues/397#issuecomment-1630455413).

PR #2885, opened 2023-02-18
Add Orbax migration guide

IvyZX:
Adds a documentation page and a colab to provide one-to-one mapping of `flax.training.checkpoints` save/restore calls to their equivalents in Orbax.

Page and colab content is the same. Only difference is that the page has codediff view.


IvyZX:
Thank you @8bitmp3 for the review! All changes LGTM. Feel free to make the approval whenever you are ready.

8bitmp3 (approved):

PR #2901, opened 2023-02-23
Bind submodules using __getattribute__

cgarciae:
# What does this PR do?

* Implements `Module.__getattribute__` to automatically register submodules that are passed as inputs (i.e. dataclass fields) upon access. 
* Add a `_deep_clone` flag to `Module.clone` such that it correctly clones external submodules from dataclass fields when entering a new context (e.g. `init` / `apply` / `bind`) while preserving shared identities.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -490,6 +490,18 @@ def _get_unbound_fn(method_or_fn: Callable[..., Any]) -> Callable[..., Any]:
 
   return method_or_fn
 
+def _tree_has_modules(x):
+  if isinstance(x, Module):
+    return True
+  elif isinstance(x, (int, float, bool, str, np.ndarray, jnp.ndarray)):
+    return False
+  else:
+    return any(isinstance(v, Module) for v in jax.tree_util.tree_leaves(x))
+
+def _map_submodules(fn: Callable[['Module'], Any], tree):
+  """Map a function over all submodules in a tree."""
+  g = lambda _, x: fn(x) if isinstance(x, Module) else x
+  return _freeze_attr(_map_over_modules_in_tree(g, tree))
 
 class SetupState(enum.IntEnum):
   # setup() has not been called.
@@ -515,6 +527,7 @@ class _ModuleInternalState:
   autoname_cursor: Dict[str, int] = dataclasses.field(default_factory=dict)
   children: Dict[str, Union[str, 'Module']] = dataclasses.field(
       default_factory=dict)
+  registered_submodules: Set[str] = dataclasses.field(default_factory=set)
 
   def reset(self) -> None:
     """Resets transient state.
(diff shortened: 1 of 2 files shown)


flax/linen/module.py
```diff
@@ -897,6 +910,20 @@ def __getattr__(self, name: str) -> Any:
...
       raise AttributeError(msg)
 
+  def __getattribute__(self, name):
+    """Call setup() before accessing any submodule attributes."""
+    # NB: all code here is very "hot" and will be run very frequently.
+    if (object.__getattribute__(self, 'scope') is not None
+        and '_submodule_dataclass_fields' in object.__getattribute__(self, '__dict__')
+        and name in object.__getattribute__(self, '_submodule_dataclass_fields')
+        and name not in self._state.registered_submodules):
+      # object.__getattribute__(self, '_try_setup')()
```
levskaya, line 920: this comment is old / removable?

flax/linen/module.py
```diff
@@ -897,6 +910,20 @@ def __getattr__(self, name: str) -> Any:
...
+      value = object.__getattribute__(self, name)
+      object.__getattribute__(self, '_register_submodules')(name, value)
```
levskaya, line 922: I do worry about adding a codepath outside of try_setup here for doing "setup" stuff - just in terms of future readability.

flax/linen/module.py
```diff
@@ -96,7 +96,7 @@ def _module_repr(module: 'Module', num_spaces: int = 4):
     rep += '# attributes\n'
     for attr in attributes.keys():
       # TODO(jheek): can we get a nice string representation of attribute types?
-      value = getattr(module, attr, None)
+      value = module.__dict__.get(attr, None)
```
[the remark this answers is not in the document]
cgarciae: Will probably depend on the context, in this case it's probably the right thing to do as we are getting the local attributes, and more than that, it is necessary to avoid certain issues when getting the representation for bounded modules, see https://github.com/google/flax/commit/a648605bd708a19c9ce8bc332fa359cb98585837 which @chiamp recently merged.

flax/linen/module.py
```diff
@@ -482,6 +482,18 @@ def _get_unbound_fn(method_or_fn: Callable[..., Any]) -> Callable[..., Any]:
 
   return method_or_fn
 
+def _tree_has_modules(x):
+  if isinstance(x, Module):
+    return True
+  elif isinstance(x, (int, float, bool, str, np.ndarray, jnp.ndarray)):
+    return False
+  else:
+    return any(isinstance(v, Module) for v in jax.tree_util.tree_leaves(x))
```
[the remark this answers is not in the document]
cgarciae: So the first 2 cases are optimization so you don't need to call `tree_leaves` for basic types which we know how to handle, the last case is indeed the general solution. Maybe I'll just use the general case if that is preferred.

flax/linen/module.py
cgarciae: Maybe we can just call `_try_setup`, it would be better for determinism.

PR #2939, opened 2023-03-08
permit passing mesh or rules to spmd primitives

### LINKED ISSUE #2938: Incompatibility between `jax.jit` in jax>=4.3 and `nn.spmd.with_logical_constraint`.
With the pjit-jit merge, functions partitioned with `jax.jit` are not run within mesh contexts - instead their `in_shardings`, `out_shardings` and sharding_constraints use pytrees of Shardings like `NamedSharding(mesh, pspec)` (not direct `PartitionSpecs` - see https://github.com/google/jax/blob/main/jax/_src/pjit.py#L753). Currently, `nn.spmd.with_logical_constraint` is a no-op when not run within a mesh context. This is fine for older versions of JAX where pjit needs to be run in a mesh context, but now I think the logic would need to be similar to:

```
def with_logical_constraint(x, logical_axes, mesh):
    sharding = jax.tree_util.tree_map(
        lambda pspec: NamedSharding(mesh, pspec),
        nn.logical_to_mesh(logical_axes),
    )
    return jax.lax.with_sharding_constraint(x, sharding)
```

to work with partitioned functions not run in mesh contexts. 

Right now, it seems that any calls to `nn.spmd.with_logical_constraint` will be no-ops with the new `jax.jit` style of partitioning a function when not run in mesh contexts.

cgarciae:
# What does this PR do?

Fixes #2938. Optionally lets users explicitly pass a mesh or rules into the various smpd primitives instead or relying exclusively on context managers.

### DIFF
--- a/docs/examples_google_research_examples.rst
+++ b/docs/examples_google_research_examples.rst
@@ -29,7 +29,7 @@ Self-attention Does Not Need O(n^2) Memory
 - Research paper:
 
   - `Self-attention Does Not Need O(n^2) Memory <https://arxiv.org/abs/2112.05682>`__ (Rabe and Staats, 2021)
-  
+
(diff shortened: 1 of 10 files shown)


IvyZX (commented):
Thanks for adding this! Can you add an example in the unit test to test when user pass non-`None` mesh?

IvyZX (approved):

PR #2944, opened 2023-03-09
release 0.6.7

IvyZX:


flax/version.py
```diff
@@ -13,5 +13,5 @@
 # limitations under the License.
 
 """Current Flax version at head on Github."""
-__version__ = "0.6.7"
+__version__ = "0.6.8"
```
cgarciae, line 16: This will result in users getting `0.6.8` if they do:

```python
import flax
print(flax.__version__)
```
I think we have to update this _after_ release to keep it on ahead.

IvyZX: Ah, this makes sense!

chiamp (approved):

PR #2962, opened 2023-03-16
Fix checkpointing docs for Orbax 0.1.4

IvyZX:
Also fixed a trivial (but surprisingly there for a long time) bug when step=0.


cgarciae (approved):
Looks good!

PR #3034, opened 2023-04-17
Use seq_lengths representation for RNN

cgarciae:
# What does this PR do?

Replaces the `segmentation_mask` representation in favor of the `seq_lengths` representation for a couple of reasons:
* its the representation used by cudnn
* internally `segmentation_mask` is transformed to the `seq_lengths` representation anyway so just using it is more performant

Note: this is a breaking change, internal tests need to be run. However, since this interface is really new and RNNs are not that popular now a days its unlikely to require lots of effort to fix (if at all).

### DIFF
--- a/examples/seq2seq/models.py
+++ b/examples/seq2seq/models.py
@@ -27,6 +27,7 @@ import numpy as np
 
 Array = jax.Array
 PRNGKey = jax.random.KeyArray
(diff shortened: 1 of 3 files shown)


bastings (approved):
Thanks, LGTM!

PR #3077, opened 2023-05-04
Eagerly bind submodules

### LINKED ISSUE #2864: Unexpected ScopeParamNotFoundError when sharing submodules
### System information
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Linux Ubuntu 20.04
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`: flax 0.6.5, jax 0.4.3, jaxlib 0.4.3+cuda11.cudnn82
- Python version: 3.10.8
- GPU/TPU model and memory: NVIDIA GeForce GTX 1080ti
- CUDA version (if applicable): 11.7

I encounter a `ScopeParamNotFoundError` when a shared submodule is nested under an unshared submodule. Here is a minimal working example:
```python
class Shared(nn.Module):
    @nn.compact
    def __call__(self, x):
        return nn.Dense(1)(x)

class Unshared(nn.Module):
    shared: nn.Module

    def __call__(self, x):
        return self.shared(x)

class Super(nn.Module):
    a: nn.Module
    b: nn.Module

    def run_a(self, x):
        return self.a(x)

    def run_b(self, x):
        return self.b(x)
    
    def __call__(self, x):
        return self.a(x) + self.b(x)


sh = Shared()
a = Unshared(shared=sh)
b = Unshared(shared=sh)
super = Super(a=a, b=b)

rng = jax.random.PRNGKey(0)
params = super.init(rng, jnp.ones(1))["params"]

print(params)
# FrozenDict({
#   a: {
#       shared: {
#            Dense_0: {
#                kernel: Array([[0.13082343]], dtype=float32),
#                bias: Array([0.], dtype=float32),
#            },
#        },
#    },
# })
# 
# as expected, the weights for `shared` only appear under the `a` key, since by my understanding flax caches the
# `shared` object and re-uses the same scope when it's encountered for the second time

print(super.apply({"params": params}, jnp.ones(1)))  # works as expected
print(super.apply({"params": params}, jnp.ones(1), method="run_a"))  # works as expected
print(super.apply({"params": params}, jnp.ones(1), method="run_b"))  # ScopeParamNotFoundError: Could not find parameter named "kernel" in scope "/b/shared/Dense_0"
```

When I call `apply` with `method="run_b"`, I expect flax to realize that the weights for `/b/shared` are under the path `/a/shared`, just like it does when `__call__` is used, but clearly it is looking in the wrong scope.

Edit: After reading a bit more, I'm thinking this has something to do with lazy binding? When `__call__` is used, first `a.shared` is bound, and then when `b.shared` is bound, flax finds the scope `a/shared` in the cache and so re-uses it for `b.shared`. However, when `run_b` is used, `a.shared` is never bound so the scope `a/shared` is never created.

Is there any way around this, perhaps using some sort of tricks in `setup`? I think there's a pretty strong argument for this use-case: it's nice to have a wrapper around multiple submodules where the submodules all get initialized at once (using `__call__`) but then individual subsets of modules can be used later on, potentially in a more complex way. And I don't see any reason why these submodules shouldn't be able to share parameters.

cgarciae:
# What does this PR do?

Alternative to #2901. Fixes #2864.

* Binds submodules during `__post_init__` if a `scope` is available using `_register_submodule`.
* Add a `_deep_clone` flag to `Module.clone` such that it correctly clones external submodules from dataclass fields when entering a new context (e.g. `init` / `apply` / `bind`) while preserving shared identities.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -490,6 +490,10 @@ def _get_unbound_fn(method_or_fn: Callable[..., Any]) -> Callable[..., Any]:
 
   return method_or_fn
 
+def _map_submodules(fn: Callable[['Module'], Any], tree):
+  """Map a function over all submodules in a tree."""
+  g = lambda _, x: fn(x) if isinstance(x, Module) else x
+  return _freeze_attr(_map_over_modules_in_tree(g, tree))
 
 class SetupState(enum.IntEnum):
   # setup() has not been called.
@@ -1000,6 +1004,12 @@ class Module(ModuleBase):
     else:
       raise ValueError('parent must be None, Module or Scope')
 
+    # eagerly bind submodules if scope is available
+    if self.scope is not None:
+        for field in dataclasses.fields(self):
+          if field.name not in ('parent', 'name') and field.init:
+            self._register_submodules(field.name, getattr(self, field.name))
+
     self._state.is_initialized = True
 
   def __repr__(self) -> str:
@@ -1048,8 +1058,8 @@ class Module(ModuleBase):
     _caches[root] = cache
     queue = []
     preserve_adopted_names = config.flax_preserve_adopted_names
-    if hasattr(self, 'preserve_adopted_names'):
-      preserve_adopted_names = self.preserve_adopted_names
+    if hasattr(type(self), 'preserve_adopted_names'):
+      preserve_adopted_names = type(self).preserve_adopted_names
     def adopt_attr_modules(cache, queue, suffix, subvalue):
       if isinstance(subvalue, Module):
         adopted_name = None
@@ -1092,7 +1102,7 @@ class Module(ModuleBase):
         # not call the user's setup. This avoids running before a
         # transformation.
         for field in dataclasses.fields(self):
-          if field.name != 'parent' and field.init:
+          if field.name not in ('parent', 'name') and field.init:
             self._register_submodules(field.name, getattr(self, field.name))
         if not shallow:
           self.setup()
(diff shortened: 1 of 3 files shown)


flax/linen/module.py
```diff
@@ -1000,6 +1007,17 @@ def __post_init__(self) -> None:
     else:
       raise ValueError('parent must be None, Module or Scope')
 
+    # eagerly bind submodules if scope is available
+    if self.scope is not None:
+      for field in dataclasses.fields(self):
+        if (
+          field.name not in ('parent', 'name')
+          and field.name in self.__dict__ # ignore fields that have not been set
```
levskaya, line 1015: why not `and hasattr(self, field.name)` ?

flax/linen/module.py
```diff
@@ -1000,6 +1007,17 @@ def __post_init__(self) -> None:
...
+        ):
+          value = getattr(self, field.name)
+          if _tree_has_modules(value):
```
levskaya, line 1018: Could you tell me again why we're doing `_tree_has_modules` when `_register_submodules` implements a recursion over submodules?  is that missing something?

levskaya:
also the test `test_repr_should_not_cause_setup` has a too-brittle testing strategy for `setup()` being called that's breaking with the updated PR

levskaya (approved):
thanks so much for the slog, lol.

kvablack:
Original author of #2864 here, just wanted to pop in and say thanks so much @cgarciae! I've actually been following your series of PRs for the past few months lol, and I can't wait for this to get merged and released so I can finally implement my design that's been blocked by #2864 this whole time. It will simplify huge swaths of code :relieved:

levskaya:
@kvablack - so very sorry that this took so long to merge - this is a subtle change that required checking a lot of internal code to make sure that we didn't break anyone.  Everything seems good so far.  Hopefully I won't be forced to roll back anything.  

@cgarciae extra thanks for helping plow through this rather annoyingly intricate problem.

PR #3114, opened 2023-05-24
Allow manually specifying the rng key for Dropout

### LINKED ISSUE #3115: Force no split in `make_rng`
### Discussed in https://github.com/google/flax/discussions/3113

<sup>Originally posted by **zaccharieramzi** May 24, 2023</sup>
I have the following situation: I am using a `Dropout` layer multiple times without a [`nn.scan` or `nn.while_loop`](https://github.com/google/flax/discussions/2920#discussioncomment-5180446), therefore I cannot use `split_rngs={"dropout": False}`. 
However, I would still like to use the same dropout mask twice.

Is it possible to specify "no split" to make rng for certain collections?

If I just take the original dropout example I would like to do something like:

```python
# Setup.
import jax
import jax.numpy as jnp
import flax.linen as nn

# Randomness.
seed = 0
root_key = jax.random.PRNGKey(seed=seed)
main_key, params_key, dropout_key = jax.random.split(key=root_key, num=3)

# A simple network.
class MyModel(nn.Module):
  num_neurons: int
  training: bool
  @nn.compact
  def __call__(self, x):
    x = nn.Dense(self.num_neurons)(x)
    # Set the dropout layer with a rate of 50% .
    # When the `deterministic` flag is `True`, dropout is turned off.
    x = nn.Dropout(rate=0.5, deterministic=not self.training)(x)
    x = nn.Dropout(rate=0.5, deterministic=not self.training)(x)
    return x

# Instantiate `MyModel` (you don't need to set `training=True` to
# avoid performing the forward pass computation).
my_model = MyModel(num_neurons=3, training=False)

x = jax.random.uniform(key=main_key, shape=(3, 4, 4))

# Initialize with `flax.linen.init()`.
# The `params_key` is equivalent to a dictionary of PRNGs.
# (Here, you are providing only one PRNG key.) 
variables = my_model.init(params_key, x)

# Perform the forward pass with `flax.linen.apply()`.
my_model.training = True
y = my_model.apply(variables, x, rngs={'dropout': dropout_key})
```

and still have `jnp.sum(y == 0.) / (3*4*3) == 0.5` approx.

For more context I am actually trying to implement [Deep Equilibrium Models](https://arxiv.org/abs/1909.01377) using [`jaxopt`](https://jaxopt.github.io/stable/) and `flax`, where the fixed point defining function uses dropout. 
I also [tried to see](https://github.com/google/jaxopt/issues/432) if the `split_rngs` functionality could be extended to `jaxopt` but I think it's going to be difficult.</div>

cgarciae:
# What does this PR do?

Fixes #3115. 

* Adds an optional `rng` argument to `Dropout.__call__` so users can manually specify a `PRNGKey` if more fine grained control is required.
* Improves typing for `PRNGKey` type alias (renamed to `KeyArray` to match JAX naming).

### DIFF
--- a/flax/configurations.py
+++ b/flax/configurations.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Global configuration options for Flax.
+r"""Global configuration options for Flax.
 
 Now a wrapper over jax.config, in which all config vars have a 'flax\_' prefix.
 
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -49,8 +49,8 @@ from flax.linen import kw_only_dataclasses
 
 traceback_util.register_exclusion(__file__)
 
-PRNGKey = Any  # pylint: disable=invalid-name
-RNGSequences = Dict[str, PRNGKey]
+KeyArray = Union[jax.Array, jax.random.KeyArray]  # pylint: disable=invalid-name
+RNGSequences = Dict[str, KeyArray]
 Array = Any    # pylint: disable=invalid-name
 
 
@@ -1270,7 +1270,7 @@ class Module(ModuleBase):
       raise ValueError("Can't query for RNGs on unbound modules")
     return self.scope.has_rng(name)
 
-  def make_rng(self, name: str) -> PRNGKey:
+  def make_rng(self, name: str) -> KeyArray:
     """Returns a new RNG key from a given RNG sequence for this Module.
 
     The new RNG key is split from the previous one. Thus, every call to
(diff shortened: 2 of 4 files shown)


flax/configurations.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Global configuration options for Flax.
+r"""Global configuration options for Flax.
```
cgarciae, line 15: Not sure why I needed to add this but Python was refusing to run this line due to a strange character.

IvyZX (approved):

PR #3139, opened 2023-06-07
Add haiku_migration_guide

cgarciae:
# What does this PR do?

WIP. Adds the Migrating from Haiku to Flax guide.
Only code for now, text will be added once we agree on the examples.

Live Preview: https://flax--3139.org.readthedocs.build/en/3139/guides/haiku_migration_guide.html

### DIFF
--- a/docs/guides/haiku_migration_guide.rst
+++ b/docs/guides/haiku_migration_guide.rst
@@ -1,0 +1,561 @@
+
+Migrating from Haiku to Flax
+==========
+
+This guide will walk through the process of migrating Haiku models to Flax,
+and highlight the differences between the two libraries.
+
+.. testsetup::
+
+  import jax
+  import jax.numpy as jnp
+  from jax.random import PRNGKey
+  import optax
+  import flax.linen as nn
+
+Basic Example
+-----------------
+
+To create custom Modules you subclass from a ``Module`` base class in
+both Haiku and Flax. However, Haiku classes use a regular ``__init__`` method
+whereas Flax classes are ``dataclasses``, meaning you define some class
+attributes that are used to automatically generate a constructor. Also,
+all Flax Modules accept a ``name`` argument without needing to define it,
+whereas in Haiku ``name`` must be explicitly defined in the constructor
+signature and passed to the superclass constructor.
+
+.. codediff::
+  :title_left: Haiku
+  :title_right: Flax
+  :sync:
+
+  import haiku as hk
+
+  class Block(hk.Module):
+    def __init__(self, features: int, name=None):
+      super().__init__(name=name)
+      self.features = features
+
+    def __call__(self, x, training: bool):
+      x = hk.Linear(self.features)(x)
+      x = hk.dropout(hk.next_rng_key(), 0.5 if training else 0, x)
+      x = jax.nn.relu(x)
+      return x
+
+  class Model(hk.Module):
+    def __init__(self, dmid: int, dout: int, name=None):
+      super().__init__(name=name)
+      self.dmid = dmid
+      self.dout = dout
+
+    def __call__(self, x, training: bool):
+      x = Block(self.dmid)(x, training)
+      x = hk.Linear(self.dout)(x)
+      return x
+
+  ---
+
+  import flax.linen as nn
+
+  class Block(nn.Module):
+    features: int
+
+
+    @nn.compact
+    def __call__(self, x, training: bool):
+      x = nn.Dense(self.features)(x)
+      x = nn.Dropout(0.5, deterministic=not training)(x)
+      x = jax.nn.relu(x)
+      return x
+
+  class Model(nn.Module):
+    dmid: int
+    dout: int
+
+
+    @nn.compact
+    def __call__(self, x, training: bool):
+      x = Block(self.dmid)(x, training)
+      x = nn.Dense(self.dout)(x)
+      return x
+
+The ``__call__`` method looks very similar in both libraries, however, in Flax
+you have to use the ``@nn.compact`` decorator in order to be able to define
+submodule inline. In Haiku, this is the default behavior.
+
+Now, a place where Haiku and Flax differ substantially is in how you construct
+the model. In Haiku, you use ``hk.transform`` over a function
+that calls your Module, ``transform`` will return an object with ``init``
+and ``apply`` methods. In Flax, you simply instantiate your Module.
+
+.. codediff::
+  :title_left: Haiku
+  :title_right: Flax
+  :sync:
+
+  def forward(x, training: bool):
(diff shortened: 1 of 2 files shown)


docs/guides/haiku_migration_guide.rst
```diff
@@ -0,0 +1,192 @@
...
+
+.. codediff::
+  :title_left: Haiku
+  :title_right: Flax
+  :sync:
+
+  sample_x = jax.numpy.ones((1, 784))
+  params, state = model.init(
+      PRNGKey(0),
+      sample_x, training=False # <== inputs
```
saeta, line 98: Why is `training=False` here, but `training=True` for Flax?

docs/guides/haiku_migration_guide.rst
```diff
@@ -0,0 +1,192 @@
...
+  ---
+
+  sample_x = jax.numpy.ones((1, 784))
+  variables = model.init(
+      PRNGKey(0),
+      sample_x, training=True # <== inputs
+  )
+  params, batch_stats = variables["params"], variables["batch_stats"]
+
+.. codediff::
```
saeta, line 111: I think we'll definitely need some text here to explain things. :-)

docs/guides/haiku_migration_guide.rst
```diff
@@ -0,0 +1,192 @@
...
+    params = jax.tree_map(lambda p, g: p - 0.1 * g, params, grads)
+
+    return params, state
+
+  ---
+
+  def train_step(key, params, batch_stats, inputs, labels):
+    def loss_fn(params):
+        logits, updates = model.apply(
+            {'params': params, 'batch_stats': batch_stats},
```
saeta, line 181: Silly idea: if we're in a world where Transformers are all you need, and transformers don't use batch norm, would it make sense to have the simple example be simpler and just have `params` and no `batch_stats` and have a second subsection that can go in more depth about state and its myriad uses (including batch_norm) with this worked example?

Alternatively, can we simplify this code by passing in `variables` instead of `params` and `batch_stats` to `train_step` separately, and switching to `mutable='batch_stats'` to make this code have a lot less syntactical overhead?

saeta (changes requested):
Thanks for putting this together! When I look at the preview, I end up having to toggle between the Flax-vs-Haiku. It'd be much nicer to have them both displayed side-by-side; can we be sure to do that?

Further, we'll a few additional sections for advanced Flax and/or Haiku patterns. If nothing else, a mapping between [common Haiku](https://dm-haiku.readthedocs.io/en/latest/api.html#common-modules) & Flax modules, but possibly also what to do if your code uses [`hk.multi_transform`](https://dm-haiku.readthedocs.io/en/latest/api.html#haiku.multi_transform), etc. Fine to leave this to a subsequent PR, IMO.

cgarciae:
Hey @saeta, thanks for the feedback! Just want to clarify that text/explanations will be added once we decide the examples are on the right direction 😅

> When I look at the preview, I end up having to toggle between the Flax-vs-Haiku. It'd be much nicer to have them both displayed side-by-side; can we be sure to do that?

We had this before (side-by-side) but switched our custom `codediff` view to tabs because it looked better, we had the problem that it overflowed outside of the page.

docs/guides/haiku_migration_guide.rst
cgarciae: In Flax we commonly initialize with `training=False` so Dropout behaves deterministically and doesn't request a PRNGKey. However, Haiku's BatchNorm layer doesn't seem to like to be initialize with `training=False` so I had to turn it to `True`. I'll be sure to mention this in the text.

docs/guides/haiku_migration_guide.rst
cgarciae: Yes! Sorry, as I mentioned, I put this together very quickly so I could get feedback on the material. I'll do the writing once I know the examples are what we want.

docs/guides/haiku_migration_guide.rst
cgarciae: So this is what I had in mind: using `Dropout` and `BatchNorm` usually reveals most hidden knobs in most JAX frameworks, even if the model if very simple. So this example, while simple, shows 90% of what most users struggle with (managing state and random state).

I think this example is very worthwhile because even in transformers you will have the `cache` collection, or use the `intermediates` collection if the `sow` stuff, so dealing with the `batch_stats` collection serves as a good starting point even if the layer is not used in transformers.

That said, if we feel starting with something even simpler is better I can just remove the `BatchNorm` layer.

levskaya: in the transformer world you have to deal w. cache - I think in a lot of haiku codebases they just thread it through? (not sure I'd have to look) but they could in theory handle it with their state mechanism.

cgarciae: Should we progressively increase the complexity of the examples in this guide? e.g. start with a regular model, then add Dropout, then add BatchNorm (or any other state). Might be friendlier but longer, this is more concise, straight to the crux of the matter.

cgarciae:
cc @lenamartens

docs/guides/haiku_migration_guide.rst
```diff
@@ -0,0 +1,551 @@
...
+    grads = jax.grad(loss_fn)(params)
+    params = jax.tree_map(lambda p, g: p - 0.1 * g, params, grads)
+
+    return params
+
+  ---
+
+  def train_step(key, params, inputs, labels):
+    def loss_fn(params):
+        logits, updates = model.apply(
```
levskaya, line 207: don't want an updates here as no mutables are set

docs/guides/haiku_migration_guide.rst
```diff
@@ -0,0 +1,551 @@
...
+
+    grads = jax.grad(loss_fn)(params)
+    params = jax.tree_map(lambda p, g: p - 0.1 * g, params, grads)
+
+    return params
+
+The most notable differences is that in Flax you have to
+pass the parameters inside a dictionary with a ``params`` key, and the
+PRNGKey inside a dictionary with a ``dropout`` key. This is because in Flax
+you can have many times of model state and random state. In Haiku, you
```
levskaya, line 222: "times" --> "types"

levskaya (approved):

docs/guides/haiku_migration_guide.rst
cgarciae: thanks for catching this! added a `testcode` block to catch this on CI.

docs/guides/haiku_migration_guide.rst
cgarciae: fixed

PR #3223, opened 2023-07-25
Fix checkpointing guide error at head

IvyZX:
Fix the Orbax error (https://readthedocs.org/projects/flax/builds/21393546/) that broke the main CI.


chiamp (approved):

cgarciae (approved):
Thanks @IvyZX !

PR #3242, opened 2023-07-31
Improve Typing support

cgarciae:
# What does this PR do?

* Makes `AxisMetadata` generic over the `unbox` type and improves some method signatures.
* Adds `overload` signatures for `Module.param`, `Module.variable`, `Module.apply`, `Scope.param`, and `Scope.variable`.


levskaya (approved):
Looks good - but I think we need to do the same thing for `scope.variable` and `self.variable` ?

PR #3236, opened 2023-07-29
Cursor API

cgarciae:
# What does this PR do?

WIP

PR #3246, opened 2023-08-01
added cursor api

chiamp:
Continuing from #3236.

Added the [Cursor API](https://flax--3246.org.readthedocs.build/en/3246/api_reference/flax.cursor.html), which allows for mutability of pytrees. This API provides a more ergonomic solution to making partial-updates of deeply nested immutable data structures, compared to making many nested `dataclasses.replace` calls.

To illustrate, consider the example below:
```
from flax.cursor import cursor
import dataclasses
from typing import Any

@dataclasses.dataclass
class A:
  x: Any

a = A(A(A(A(A(A(A(0)))))))
```
To replace the int 0 using ``dataclasses.replace``, we would have to write many nested calls:
```
a2 = dataclasses.replace(
  a,
  x=dataclasses.replace(
    a.x,
    x=dataclasses.replace(
      a.x.x,
      x=dataclasses.replace(
        a.x.x.x,
        x=dataclasses.replace(
          a.x.x.x.x,
          x=dataclasses.replace(
            a.x.x.x.x.x,
            x=dataclasses.replace(a.x.x.x.x.x.x, x=1),
          ),
        ),
      ),
    ),
  ),
)
```
The equivalent can be achieved much more simply using the Cursor API:
```
a3 = cursor(a).x.x.x.x.x.x.x.set(1)
assert a2 == a3
```
The Cursor object keeps tracks of changes made to it and when `.build` is called, generates a new object with the accumulated changes. Basic usage involves wrapping the object in a Cursor, making changes to the Cursor object and generating a new copy of the original object with the accumulated changes.

There are three ways to use the Cursor API:
* multiple changes using the `.build` method
* single line change using the `.set` method
* multiple conditional changes using the `.apply_update` method

Once you wrap the object in a Cursor and make changes to it, calling the `.build` method will generate a new object with the accumulated changes. For example:
```
from flax.cursor import cursor
from flax.training import train_state
import optax

dict_obj = {'a': 1, 'b': (2, 3), 'c': [4, 5]}
c = cursor(dict_obj)          # wrap the object in a Cursor
c['b'][0] = 10                # change 1
c['a'] = (100, 200)           # change 2
modified_dict_obj = c.build() # generate new object with the above changes
assert modified_dict_obj == {'a': (100, 200), 'b': (10, 3), 'c': [4, 5]}

state = train_state.TrainState.create(
    apply_fn=lambda x: x,
    params=dict_obj,
    tx=optax.adam(1e-3),
)
new_fn = lambda x: x + 1
c = cursor(state)          # wrap the object in a Cursor
c.params['b'][1] = 10      # change 1
c.apply_fn = new_fn        # change 2
modified_state = c.build() # generate new object with the above changes
assert modified_state.params == {'a': 1, 'b': (2, 10), 'c': [4, 5]}
assert modified_state.apply_fn == new_fn
```
Calling the `.set` method will apply a single-line change and call `.build` immediately after. Therefore, you don't need to manually call `.build` after `.set`. For example:
```
from flax.cursor import cursor
from flax.training import train_state
import optax

dict_obj = {'a': 1, 'b': (2, 3), 'c': [4, 5]}
modified_dict_obj = cursor(dict_obj)['b'][0].set(10) # apply change and generate new object with the change
assert modified_dict_obj == {'a': 1, 'b': (10, 3), 'c': [4, 5]}

state = train_state.TrainState.create(
    apply_fn=lambda x: x,
    params=dict_obj,
    tx=optax.adam(1e-3),
)
modified_state = cursor(state).params['b'][1].set(10) # apply change and generate new object with the change
assert modified_state.params == {'a': 1, 'b': (2, 10), 'c': [4, 5]}
```
Calling `.apply_update` will traverse the Cursor object and apply conditional changes recursively via a filter function. 
The filter function has a function signature of `(str, Any) -> Any`:
- The input arguments are the current key path (in the form of a string delimited by ‘/’) and value at that current key path
- The output is the new value (either modified by the `update_fn` or same as the input value if the condition wasn’t fulfilled)

To generate a copy of the original object with the accumulated changes, call the `.build` method. For example:
```
import flax.linen as nn
from flax.cursor import cursor
import jax
import jax.numpy as jnp

class Model(nn.Module):
  @nn.compact
  def __call__(self, x):
    x = nn.Dense(3)(x)
    x = nn.relu(x)
    x = nn.Dense(3)(x)
    x = nn.relu(x)
    x = nn.Dense(3)(x)
    x = nn.relu(x)
    return x

params = Model().init(jax.random.PRNGKey(0), jnp.empty((1, 2)))['params']

def update_fn(path, value):
  '''Multiply all dense kernel params by 2 and add 1.
  Subtract the Dense_1 bias param by 1.'''
  if 'kernel' in path:
    return value * 2 + 1
  elif 'Dense_1' in path and 'bias' in path:
    return value - 1
  return value

c = cursor(params)
new_params = c.apply_update(update_fn).build()
for layer in ('Dense_0', 'Dense_1', 'Dense_2'):
  assert (new_params[layer]['kernel'] == 2 * params[layer]['kernel'] + 1).all()
  if layer == 'Dense_1':
    assert (new_params[layer]['bias'] == jnp.array([-1, -1, -1])).all()
  else:
    assert (new_params[layer]['bias'] == params[layer]['bias']).all()
```

### DIFF
--- a/flax/cursor.py
+++ b/flax/cursor.py
@@ -1,0 +1,392 @@
+import enum
+from typing import Any, Callable, Dict, Generator, Generic, Mapping, Optional, Protocol, Tuple, TypeVar, Union, runtime_checkable
+from flax.core import FrozenDict
+import dataclasses
+
+
+A = TypeVar('A')
+Key = Any
+
+
+@runtime_checkable
+class Indexable(Protocol):
+
+  def __getitem__(self, key) -> Any:
+    ...
+
+
+@dataclasses.dataclass
+class ParentKey(Generic[A]):
+  parent: 'Cursor[A]'
+  key: Any
+
+
+class AccessType(enum.Enum):
+  GETITEM = enum.auto()
+  GETATTR = enum.auto()
+
+
+def is_named_tuple(obj):
+  return (
+      isinstance(obj, tuple)
+      and hasattr(obj, '_fields')
+      and hasattr(obj, '_asdict')
+      and hasattr(obj, '_replace')
+  )
+
+
+def _get_changes(path, obj, update_fn):
+  """Helper function for ``Cursor.apply_update``. Returns a generator of
+  Tuple[Tuple[Union[str, int], Any], ...], where the first element is a
+  tuple key path where the change was applied from the ``update_fn``, and
+  the second element is the newly modified value. If the generator is
+  non-empty, then the tuple key path will always be non-empty as well."""
+  if path:
+    str_path = '/'.join(str(key) for key, _ in path)
+    new_obj = update_fn(str_path, obj)
+    if new_obj is not obj:
+      yield path, new_obj
+      return
+
+  if isinstance(obj, (FrozenDict, dict)):
+    items = obj.items()
+    access_type = AccessType.GETITEM
+  elif is_named_tuple(obj):
+    items = ((name, getattr(obj, name)) for name in obj._fields)  # type: ignore
+    access_type = AccessType.GETATTR
+  elif isinstance(obj, (list, tuple)):
+    items = enumerate(obj)
+    access_type = AccessType.GETITEM
+  elif dataclasses.is_dataclass(obj):
+    items = (
+        (f.name, getattr(obj, f.name))
+        for f in dataclasses.fields(obj)
+        if f.init
+    )
+    access_type = AccessType.GETATTR
+  else:
+    yield from ()  # empty generator
+    return
+
+  for key, value in items:
+    yield from _get_changes(path + ((key, access_type),), value, update_fn)
+
+
+class Cursor(Generic[A]):
+  obj: A
+  parent_key: Optional[ParentKey[A]]
+  changes: Dict[Any, Union[Any, 'Cursor[A]']]
+
+  def __init__(self, obj: A, parent_key: Optional[ParentKey[A]]):
+    # NOTE: we use `vars` here to avoid calling `__setattr__`
+    # vars(self) = self.__dict__
+    vars(self)['obj'] = obj
+    vars(self)['parent_key'] = parent_key
+    vars(self)['changes'] = {}
+
+  @property
+  def root(self) -> 'Cursor[A]':
+    if self.parent_key is None:
+      return self
+    else:
+      return self.parent_key.parent.root  # type: ignore
+
+  def __getitem__(self, key) -> 'Cursor[A]':
+    if key in self.changes:
+      return self.changes[key]
+
+    if not isinstance(self.obj, Indexable):
+      raise TypeError(f'Cannot index into {self.obj}')
+
+    if isinstance(self.obj, Mapping) and key not in self.obj:
+      raise KeyError(f'Key {key} not found in {self.obj}')
+
+    if is_named_tuple(self.obj):
+      return getattr(self, self.obj._fields[key])  # type: ignore
+
+    child = Cursor(self.obj[key], ParentKey(self, key))
+    self.changes[key] = child
+    return child
+
+  def __getattr__(self, name) -> 'Cursor[A]':
+    if name in self.changes:
+      return self.changes[name]
+
+    if not hasattr(self.obj, name):
+      raise AttributeError(f'Attribute {name} not found in {self.obj}')
+
+    child = Cursor(getattr(self.obj, name), ParentKey(self, name))
+    self.changes[name] = child
+    return child
+
+  def __setitem__(self, key, value):
+    if is_named_tuple(self.obj):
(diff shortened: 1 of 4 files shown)


ASEM000:
Nice API, would you take a look at https://github.com/ASEM000/PyTreeClass, I attempted a slightly different API with the same objective as this API.

flax/cursor.py
```diff
@@ -0,0 +1,304 @@
...
+  parent_key: Optional[ParentKey[A]]
+  changes: dict[Any, Union[Any, 'Cursor[A]']]
+
+  def __init__(self, obj: A, parent_key: Optional[ParentKey[A]]):
+    if isinstance(obj, (FrozenDict, dict, list, tuple)):
+      obj_copy = jax.tree_util.tree_map(lambda x: x, obj)
+    elif dataclasses.is_dataclass(obj):
+      obj_copy = dataclasses.replace(obj)
+    else:
+      obj_copy = obj
```
cgarciae, line 42: Why do we need this deep copy?

flax/cursor.py
```diff
@@ -0,0 +1,304 @@
...
+    self.changes[name] = child
+    return child
+
+  def __setitem__(self, key, value):
+    self.changes[key] = Cursor(value, ParentKey(self, key))
+
+  def __setattr__(self, name, value):
+    self.changes[name] = Cursor(value, ParentKey(self, name))
+
+  def apply_filter(
```
cgarciae, line 88: `apply_filter` makes me think this method would remove/filter out values from the cursor. Maybe `apply_update` and `update_fn` could be and alternative names?

flax/cursor.py
```diff
@@ -0,0 +1,304 @@
...
+      self,
+      filter_fn: Callable[[Sequence[str], Any], Tuple[bool, Any]],
+      _path: List[str] = [],
+  ):
```
cgarciae, line 92: What if this method returns self (`Cursor[A]`) so it can be chained with build? e.g.

```python
obj2 = cursor(obj).apply_filter(f).build()
```

flax/cursor.py
```diff
@@ -0,0 +1,304 @@
...
+  def __setitem__(self, key, value):
+    self.changes[key] = Cursor(value, ParentKey(self, key))
+
+  def __setattr__(self, name, value):
+    self.changes[name] = Cursor(value, ParentKey(self, name))
+
+  def apply_filter(
+      self,
+      filter_fn: Callable[[Sequence[str], Any], Tuple[bool, Any]],
+      _path: List[str] = [],
```
cgarciae, line 91: Make `_path` a `tuple[str, ...] = ()`. Its usually not a good idea to have lists as a default value because its not immutable.

flax/cursor.py
```diff
@@ -0,0 +1,341 @@
+# from sqlite3 import Cursor
```
levskaya, line 1: `# from sqlite3 import Cursor` <-- huh?  delete

flax/cursor.py
```diff
@@ -0,0 +1,341 @@
...
+        key: child.build() if isinstance(child, Cursor) else child
+        for key, child in self.changes.items()
+    }
+    if isinstance(self.obj, FrozenDict):
+      obj = self.obj.copy(changes)  # type: ignore
+    elif isinstance(self.obj, (dict, list)):
+      obj = self.obj.copy()  # type: ignore
+      for key, value in changes.items():
+        obj[key] = value
+    elif isinstance(self.obj, tuple):
```
levskaya, line 153: we should also support setting attributes by name on a NamedTuple - these can only be ducktyped by looking for a `_fields` attribute.  A lot of libraries still use namedtuples over dataclasses.

flax/cursor.py
```diff
@@ -0,0 +1,341 @@
...
+          x = nn.relu(x)
+          x = nn.Dense(3)(x)
+          x = nn.relu(x)
+          x = nn.Dense(3)(x)
+          x = nn.relu(x)
+          return x
+
+      params = Model().init(jax.random.PRNGKey(0), jnp.empty((1, 2)))['params']
+
+      def update_fn(path, value):
```
levskaya, line 219: why return both boolean and possibly updated value?  what's the optimization here?

levskaya (commented):
just a few questions

flax/cursor.py
cgarciae: My bad, if we use `id` to detect changes it should work.

levskaya (approved):
Looks like a good start, we might iterate on a few details/extensions soon.

PR #3268, opened 2023-08-08
Add Sharding Annotations to Flax Modules

terrykong:
Adds sharding annotations to Flax's `Dense`, `DenseGeneral`, `MultiHeadDotProductAttention` and normalization modules. This change makes Flax's modules compatible with [T5X](https://github.com/google-research/t5x).

### DIFF
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -27,6 +27,7 @@ from flax.linen.module import compact
 from flax.linen.module import merge_param
 from flax.linen.module import Module
 from flax.linen.normalization import LayerNorm
+from flax.linen.partitioning import variable_with_axes
 import jax
 from jax import lax
 from jax import random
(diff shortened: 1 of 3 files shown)


levskaya:
We don't want to add explicit sharding annotations to our base layers.  In fact the older `partitioning.py` API is deprecated in favor of the newer "boxed metadata" approach using wrappers of the initializers (see `core/meta.py` and `linen/spmd.py`).  

This approach doesn't usually require editing the base layers at all.  An example of using this can be found in https://github.com/google/maxtext/blob/main/MaxText/layers.py - note that this example uses the T5X style "logical named axis  indirection" via `nn.with_logical_constraint` but can instead be used with direct mesh annotations via `nn.with_partitioning`

levskaya:
I should add that we generally want framework authors / downstream library writers to handle the particular partitioning handling rather than baking it in to our core example layers, as these are mainly pedagogical starting points.

I'll close the issue for now but feel free to comment/reopen if desired.

cgarciae:
I think we should expose the initializers for `cached_key`, `cached_value`, and `cache_index` so users can pass the partition metadata through the `nn.with_partitioning` decorator.

terrykong:
Closing this in favor of https://github.com/google/flax/pull/3340

PR #3277, opened 2023-08-12
Add method-to-model section to Haiku migration guide

IvyZX:
This sections showcases method-to-model and handling of raw hk data API like `get_parameter` and `get_state`.

### DIFF
--- a/docs/guides/haiku_migration_guide.rst
+++ b/docs/guides/haiku_migration_guide.rst
@@ -1,6 +1,6 @@
 
 Migrating from Haiku to Flax
-==========
+============================
 
 This guide will walk through the process of migrating Haiku models to Flax,
 and highlight the differences between the two libraries.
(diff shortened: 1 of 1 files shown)


docs/guides/haiku_migration_guide.rst
```diff
@@ -861,4 +861,47 @@ as we are using ``5`` layers.
...
-      })
\ No newline at end of file
+      })
+
+Raw Haiku functions vs Flax modules
+----------------------
+
+In Haiku, `hk.Module` is not required, and writing a forward function to be passed into `hk.transform` is enough. In this process, people use raw `hk.<get|set>_<parameter|state>` functions to access model data and states.
+
+Flax recommends a more module-centric approach that uses `__call__` to define the forward function. The corresponding accessor will be `nn.module.param` and `nn.module.variable` (see section [Handling State](#handling-state) for explantaion on collections).
```
cgarciae, line 871: Fix link from markdown to rst
```suggestion
Flax recommends a more module-centric approach that uses `__call__` to define the forward function. The corresponding accessor will be `nn.module.param` and `nn.module.variable` (see section `Handling State <#handling-state>`__ for explantaion on collections).
```

docs/guides/haiku_migration_guide.rst
```diff
@@ -861,4 +861,47 @@ as we are using ``5`` layers.
...
               },
           },
-      })
\ No newline at end of file
+      })
+
+Raw Haiku functions vs Flax modules
+----------------------
+
+In Haiku, `hk.Module` is not required, and writing a forward function to be passed into `hk.transform` is enough. In this process, people use raw `hk.<get|set>_<parameter|state>` functions to access model data and states.
```
cgarciae, line 869: Here is a suggestion for the wording 😅:
```suggestion
In Haiku, it is possible to write the entire model as a single function by using ``hk.{get,set}_{parameter,state}`` to define/access model parameters and states. It very common to write the top-level "Module" as a function instead:
```
If you prefer the current wording, you do need to fix the "inline code" syntax which in .rst is using double back tick: "\`\`"

docs/guides/haiku_migration_guide.rst
```diff
@@ -861,4 +861,47 @@ as we are using ``5`` layers.
                   kernel: (5, 64, 64),
               },
           },
-      })
\ No newline at end of file
+      })
+
+Raw Haiku functions vs Flax modules
```
cgarciae, line 866: Maybe "top-level" is a better description for what we are trying to express here? I think in practice in Haiku people only code the "main module" as a function, submodules are proper `hk.Module`s.

```suggestion
Top-level Haiku functions vs top-level Flax modules
```

docs/guides/haiku_migration_guide.rst
8bitmp3, line 869: Suggestion (as discussed with @IvyZX):

```suggestion
In Haiku, using `hk.Module` is not required - you can just write a forward function and pass it into `hk.transform`. Inside the transformed function, you would typically use the "raw" `hk.<get|set>_<parameter|state>` functions to access model data and states.
```

docs/guides/haiku_migration_guide.rst
8bitmp3, line 871: Suggestion:

```suggestion
The Flax team recommends a more Module-centric approach that uses `__call__` to define the forward function. The corresponding accessor will be `nn.module.param` and `nn.module.variable` (go to [Handling State](#handling-state) for an explanation on collections).
```

8bitmp3 (commented):
Thank you! A few small suggestions, including:

```
In Haiku, using `hk.Module` is not required - you can just write a forward function and pass it into `hk.transform`. Inside the transformed function, you would typically use the "raw" `hk.<get|set>_<parameter|state>` functions to access model data and states.
```

and

```
The Flax team recommends a more Module-centric approach that uses `__call__` to define the forward function. The corresponding accessor will be `nn.module.param` and `nn.module.variable` (go to [Handling State](#handling-state) for an explanation on collections).
```

IvyZX:
Thanks for all the suggestions! All are applied.

cgarciae (approved):

PR #3290, opened 2023-08-22
added logical partitioning to pjit guide

chiamp:
added logical partitioning to pjit guide


IvyZX:
> > 
> 
> > Seems we are missing the rule for `embed`, we currently only have:
> > ```python
> > rules = (('batch', 'data'),
> >          ('hidden', 'model'))
> > ```
> 
> I think the `'embed'` axis is intended to be replicated instead of being sharded. Can you confirm @IvyZX?

Yes - because it needs to match the partitioning layout above (on the non-logical example). You can think that `rules` implies the mapping of `('embed', None)` here.

IvyZX (approved):

PR #3302, opened 2023-08-31
updated lm1b example with jit

chiamp:
Updated lm1b example to use jit instead of pmap.
lm1b now uses both data and model parallelism instead of just data parallelism.

### DIFF
--- a/examples/lm1b/configs/default.py
+++ b/examples/lm1b/configs/default.py
@@ -30,11 +30,11 @@ def get_config():
   config.max_corpus_chars = 10**7
 
   # Name of TFDS translation dataset to use.
-  config.dataset_name = "lm1b"
+  config.dataset_name = 'lm1b'
 
   # Optional name of TFDS translation dataset to use for evaluation.
-  config.eval_dataset_name = "lm1b"
-  config.eval_split = "test"
+  config.eval_dataset_name = 'lm1b'
+  config.eval_split = 'test'
 
   # Per device batch size for training.
   config.per_device_batch_size = 32
@@ -114,7 +114,44 @@ def get_config():
   # Integer for PRNG random seed.
   config.seed = 0
 
-  # Prompt for language model sampling.
-  config.prompts = "I love to "
+  # Prompt for language model sampling,
+  # taken from MaxText (https://github.com/google/maxtext/blob/main/MaxText/configs/base.yml).
+  config.prompts = 'I love to '
+
+  # Parallelism
+  config.mesh_axes = ['data', 'fsdp', 'tensor']
+  config.logical_axis_rules = [
+      ['activation_batch', ['data', 'fsdp']],
+      ['activation_length', ['data', 'fsdp']],
+      ['activation_embed', 'tensor'],
+      ['activation_mlp', 'tensor'],
+      ['activation_heads', 'tensor'],
+      ['activation_kv', 'tensor'],
+      ['activation_vocab', 'tensor'],
+      ['mlp', 'tensor'],
+      ['vocab', 'tensor'],
+      ['embed', 'fsdp'],
+      ['heads', 'tensor'],
+  ]
+  config.full_sharding = ['data', 'fsdp', 'tensor']
+  config.data_sharding = ['data']
+
+  # One axis for each parallelism type may hold a placeholder (-1)
+  # value to auto-shard based on available slices and devices.
+  # By default, product of the DCN axes should equal number of slices
+  # and product of the ICI axes should equal number of devices per slice.
+  # ICI (Inter-Chip Interconnection): A high-speed connection between
+  # sets of TPU chips, which form the TPU network.
+  # DCN (Data Center Network): A connection between the TPU networks;
+  # not as fast as ICI.
+  # ICI has around 100x the bandwidth of DCN, but it is not a general
+  # purpose connection, which is why DCN is necessary for scaling to
+  # extremely large ML models.
+  config.dcn_data_parallelism = -1  # recommended DCN axis to be auto-sharded
+  config.dcn_fsdp_parallelism = 1
+  config.dcn_tensor_parallelism = 1
+  config.ici_data_parallelism = 1
+  config.ici_fsdp_parallelism = -1  # recommended ICI axis to be auto-sharded
+  config.ici_tensor_parallelism = 1
 
   return config
--- a/examples/lm1b/train.py
+++ b/examples/lm1b/train.py
@@ -21,20 +21,18 @@ This script trains a Transformer on a LM1B dataset.
 # pytype: disable=attribute-error
 
 import collections
-import functools
 import os
 
 from absl import logging
 from clu import metric_writers
 from clu import periodic_actions
-from flax import jax_utils
 from flax import linen as nn
 from flax.training import checkpoints
 from flax.training import common_utils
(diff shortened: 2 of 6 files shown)


examples/lm1b/configs/default.py
```diff
@@ -115,6 +115,35 @@ def get_config():
   config.seed = 0
 
   # Prompt for language model sampling.
-  config.prompts = "I love to "
+  config.prompts = 'I love to '
+
+  # Parallelism
```
marcvanzee, line 120: Maybe mention you got these for maxtext and link to it.

examples/lm1b/configs/default.py
```diff
@@ -115,6 +115,35 @@ def get_config():
...
+      ['activation_mlp', 'tensor'],
+      ['activation_heads', 'tensor'],
+      ['activation_kv', 'tensor'],
+      ['activation_vocab', 'tensor'],
+      ['mlp', 'tensor'],
+      ['vocab', 'tensor'],
+      ['embed', 'fsdp'],
+      ['heads', 'tensor'],
+  ]
+  config.full_sharding = [['data', 'fsdp', 'tensor']]
```
marcvanzee, line 135: Why is this a list of lists and later used with * expansion? Why not simple a list and pass it to `NamedSharding` without expanding it? Same for `data_sharding` below.

examples/lm1b/configs/default.py
```diff
@@ -115,6 +115,35 @@ def get_config():
...
+  config.data_sharding = [['data']]
+
+  # One axis for each parallelism type may hold a placeholder (-1)
+  # value to auto-shard based on available slices and devices.
+  # By default, product of the DCN axes should equal number of slices
```
marcvanzee, line 140: Maybe explain what DCN and ICI are, I suppose not everyone knows this.

examples/lm1b/train.py
```diff
@@ -527,10 +556,11 @@ def encode_strings(strs, max_len):
 
       # Shard data to devices and do a training step.
       with jax.profiler.StepTraceAnnotation("train", step_num=step):
-        batch = common_utils.shard(
-            jax.tree_util.tree_map(np.asarray, next(train_iter))
+        batch = next(train_iter)
+        batch = jax.tree_map(lambda x: jnp.array(x), batch)
```
marcvanzee, line 560: here you cast to jax array where the original seems to use numpy arrays, why did you do this?

examples/lm1b/train.py
```diff
@@ -527,10 +556,11 @@ def encode_strings(strs, max_len):
...
+        state, metrics = p_train_step(
+            state, batch, train_config, learning_rate_fn, 0.0, dropout_rngs
```
marcvanzee, line 562: why do you pass 0.0 here? isn't this the default value? also I would always use kwarg here for a number.

examples/lm1b/train.py
```diff
@@ -542,7 +572,7 @@ def encode_strings(strs, max_len):
       if step % config.eval_every_steps == 0 or is_last_step:
         with report_progress.timed("training_metrics"):
           logging.info("Gathering training metrics.")
-          train_metrics = common_utils.get_metrics(train_metrics)
+          train_metrics = common_utils.stack_forest(train_metrics)
```
marcvanzee, line 575: why was this necessary?

examples/lm1b/utils.py
```diff
@@ -0,0 +1,166 @@
...
+import jax.numpy as jnp
+from jax.experimental import mesh_utils
+
+
+# Mesh utils.
+# -----------------------------------------------------------------------------
+
+
+def create_device_mesh(config):
+  """Creates a device mesh with each slice in its own data parallel group. If there is only one slice, uses two replicas"""
```
marcvanzee, line 18:
```suggestion
  """Creates a device mesh with each slice in its own data parallel group. If there is only one slice, uses two replicas."""
```

examples/lm1b/utils.py
```diff
@@ -0,0 +1,166 @@
+import functools
```
marcvanzee, line 1: Is this file copied from maxtext? Please add a comment if so.

marcvanzee (changes requested):

examples/lm1b/train.py
chiamp: I get this error without casting:
```
File "/home/marcuschiam/flax/examples/lm1b/train.py", line 565, in train_and_evaluate
    state, metrics = jit_train_step(
TypeError: dtype tf.int32 not understood
```
The original element type is `EagerTensor` I think, with dtype `tf.int32`.

examples/lm1b/train.py
chiamp: If I try:
```
state, metrics = jit_train_step(
            state, batch, train_config, learning_rate_fn, dropout_rng=dropout_rngs
        )
```
I get:
```
File "/home/marcuschiam/flax/examples/lm1b/train.py", line 565, in train_and_evaluate
    state, metrics = jit_train_step(
ValueError: pjit does not support kwargs when in_shardings is specified.
```

examples/lm1b/train.py
chiamp: [`get_metrics`](https://github.com/google/flax/blob/main/flax/training/common_utils.py#L72) is a utility function for pmap to get a single copy of a device-replicated metric. It calls `stack_forest` on the metrics.
If I try using `get_metrics` with `jit`, I get:
```
File "/home/marcuschiam/.local/lib/python3.8/site-packages/flax/training/common_utils.py", line 83, in <lambda>
    device_metrics = jax.tree_util.tree_map(lambda x: x[0], device_metrics)
  File "/home/marcuschiam/.local/lib/python3.8/site-packages/jax/_src/array.py", line 317, in __getitem__
    return lax_numpy._rewriting_take(self, idx)
  File "/home/marcuschiam/.local/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 4143, in _rewriting_take
    return _gather(arr, treedef, static_idx, dynamic_idx, indices_are_sorted,
  File "/home/marcuschiam/.local/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 4152, in _gather
    indexer = _index_to_gather(shape(arr), idx)  # shared with _scatter_update
  File "/home/marcuschiam/.local/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 4253, in _index_to_gather
    idx = _canonicalize_tuple_index(len(x_shape), idx)
  File "/home/marcuschiam/.local/lib/python3.8/site-packages/jax/_src/numpy/lax_numpy.py", line 4586, in _canonicalize_tuple_index
    raise IndexError(
IndexError: Too many indices for array: 1 non-None/Ellipsis indices for dim 0.
```

examples/lm1b/train.py
marcvanzee: Ok makes sense!

examples/lm1b/train.py
marcvanzee: Cool, thanks!

examples/lm1b/train.py
marcvanzee: Hmm ok, thanks for letting me know.

marcvanzee (approved):

PR #3303, opened 2023-08-31
Removed `with mesh` context manager from Flax jit guide, since it doesn't do anything with jax.jit

chiamp:
Removed `with mesh` context manager from [Flax jit guide](https://flax--3303.org.readthedocs.build/en/3303/guides/flax_on_pjit.html#compile-the-train-step-and-inference), since it doesn't do anything with jax.jit

### DIFF
--- a/docs/guides/flax_on_pjit.ipynb
+++ b/docs/guides/flax_on_pjit.ipynb
@@ -808,8 +808,7 @@
     "  state = state.apply_gradients(grads=grads)\n",
     "  return state\n",
     "\n",
-    "with mesh:\n",
-    "  new_state = train_step(initialized_state, x)"
+    "new_state = train_step(initialized_state, x)"
    ]
   },
   {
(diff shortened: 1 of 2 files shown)


IvyZX:
Sorry for the late reply - I am actually debating whether to use `with mesh` or the `mesh_sharding()` util in the guide. It's quite common for larger model library to know nothing about the mesh, so maybe it will actually be cleaner if we annotate with the `PartitionSpec` and add mesh only on the top-level with `with mesh`. 

Originally I made the guide with sharding passed everywhere because there was plan to deprecate jit/pjit API support to `PartitionSpec`, but this doesn't seem to be the case any more. Please let me know if you have any thoughts.

chiamp:
Ya I think removing the `mesh_sharding` function and just adding the `mesh` to `NamedSharding(mesh, PartitionSpec(...))` would be more clear.

> add mesh only on the top-level with `with mesh`.

I spoke with @yashk2810 and he said that `with mesh` doesn't do anything with `jax.jit`

andsteing:
Is the `with mesh:` required with `nn.with_logical_constraints()` ?

(Under the hood this is calling `lax.with_sharding_constraint()`)

PR #3306, opened 2023-09-02
Add `find` methods and magic methods for Cursor API

### LINKED ISSUE #3309: Add `.find` and `.find_all` to Cursor API
Given a function `f(path, value) -> bool`, `Cursor.find_all(f)` would return a generator of cursor objects that fulfill the condition function `f` (or an empty generator if the condition was never fulfilled). Example code usage:
```
c = cursor(config)
for layer_cursor in c.find_all(lambda path, x: isinstance(x, LayerConfig)):
  layer_cursor.attention.dropout = 0.5
  layer_cursor.attention.abc = 1
mutated_obj = c.build()
```
`Cursor.find(f)` would do something similar, except it'll return a single cursor object. Example code usage:
```
cursor(tree).find(lambda x: isinstance(x, Foo)).some_property_of_foo.set(4)
```
If no object or more than one object is found, we could have the following behavior:
* do nothing, and simply return `None` or the first object found that fulfills the condition of function `f`
* raise an error if no object or more than one object is found that fulfills the condition of function `f`
* add arguments to `Cursor.find(f, raise_if_empty=True, raise_if_more_than_one=True)` so that the user can choose which behavior they want

### LINKED ISSUE #3311: Should we override magic/dunder methods for the Cursor API?
For example, if we override `__add__`:
```
class Cursor(Generic[A]):
  obj: A
  parent_key: Optional[ParentKey[A]]
  ...
  def __add__(self, other):
    return self.obj + other
```
It would allow us to do something like this:
```
c = cursor({'a': 1, 'b': {'c': 2, 'd': 3}})
c['b']['d'] += 1
c.build()
```
Rather than:
```
c = cursor({'a': 1, 'b': {'c': 2, 'd': 3}})
c['b']['d'] = c['b']['d'].obj + 1
c.build()
```

chiamp:
Initial Cursor PR: #3246
Resolves #3309 and resolves #3311.

The PR adds the following features to the [Cursor API](https://flax--3306.org.readthedocs.build/en/3306/api_reference/flax.cursor.html):
* Moved all Cursor attributes and properties to private (e.g. `Cursor.obj -> Cursor._obj`)
* Added magic methods to support more intuitive usage and updates of Cursor objects
```
c = cursor(((0, 1), (1, 2)))
for child_c in c: # iterating over child Cursors is now possible if the underlying object is a tuple or list
  child_c[1] *= -1 # Cursor objects now support in-place updates
assert c.build() == (((0, -1), (1, -2)))
assert len(c) == 2
```
* Added a `.find_all` method which finds all child Cursor objects that fulfill a `cond_fn` that is passed into the method
```
c = cursor({'a1': {'b': 1}, 'a2': {'b': 2}, 'c': 3})
for child_c in c.find_all(lambda path, value: path.endswith('a1') or path.endswith('a2')):
  child_c['b'] += 5
assert c.build() == {'a1': {'b': 6}, 'a2': {'b': 7}, 'c': 3}
```
* Added a `.find` method which finds a single child Cursor object that fulfills a `cond_fn` passed into the method. An error is raised if no object or more than one object is found.
```
c = cursor({'a1': {'b': 1}, 'a2': {'b': 2}, 'c': 3})
child_c = c.find(lambda path, value: path.endswith('a1'))
child_c['b'] += 5
assert c.build() == {'a1': {'b': 6}, 'a2': {'b': 2}, 'c': 3}
```
* Added `._path` property for Cursor objects
```
@dataclasses.dataclass
class GenericDataClass:
    x: Any
    y: Any

c = cursor({'a': ['x', ('y', 'z')], 'b': (1, GenericDataClass(x=2, y=3))})
assert c.find(lambda path, value: value == 2)._path == "['b'][1].x"
```
* Added a more informative representation for Cursor objects
```
@dataclasses.dataclass
class GenericDataClass:
    x: Any
    y: Any

c = cursor({'a': ['x', ('y', 'z')], 'b': (1, GenericDataClass(x=2, y=3))})
print(repr(c))
# Cursor(
#   _obj={'a': ['x', ('y', 'z')], 'b': (1, GenericDataClass(x=2, y=3))},
#   _changes={}
# )

c['a'][1][0] += 'yy'
c['b'][0] = []
c['b'][1].x *= -1
print(repr(c))
# Cursor(
#   _obj={'a': ['x', ('y', 'z')], 'b': (1, GenericDataClass(x=2, y=3))},
#   _changes={
#     'a': Cursor(
#            _obj=['x', ('y', 'z')],
#            _changes={
#              1: Cursor(
#                   _obj=('y', 'z'),
#                   _changes={
#                     0: Cursor(
#                          _obj='yyy',
#                          _changes={}
#                        )
#                   }
#                 )
#            }
#          ),
#     'b': Cursor(
#            _obj=(1, GenericDataClass(x=2, y=3)),
#            _changes={
#              0: Cursor(
#                   _obj=[],
#                   _changes={}
#                 ),
#              1: Cursor(
#                   _obj=GenericDataClass(x=2, y=3),
#                   _changes={
#                     'x': Cursor(
#                            _obj=-2,
#                            _changes={}
#                          )
#                   }
#                 )
#            }
#          )
#   }
# )
```

### DIFF
--- a/flax/cursor.py
+++ b/flax/cursor.py
@@ -15,6 +15,7 @@
 import enum
 from typing import Any, Callable, Dict, Generator, Generic, Mapping, Optional, Protocol, Tuple, TypeVar, Union, runtime_checkable
 from flax.core import FrozenDict
+from flax.errors import CursorFindError, TraverseTreeError
 import dataclasses
 
 
@@ -29,15 +30,16 @@ class Indexable(Protocol):
     ...
 
 
+class AccessType(enum.Enum):
+  ITEM = enum.auto()
+  ATTR = enum.auto()
+
+
 @dataclasses.dataclass
 class ParentKey(Generic[A]):
   parent: 'Cursor[A]'
   key: Any
-
-
-class AccessType(enum.Enum):
-  GETITEM = enum.auto()
-  GETATTR = enum.auto()
+  access_type: AccessType
 
 
 def is_named_tuple(obj):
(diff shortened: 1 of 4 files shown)


flax/cursor.py
```diff
@@ -49,41 +51,72 @@ def is_named_tuple(obj):
...
+    change was applied from the ``update_fn``, and the second element is
+    the newly modified value. If the generator is non-empty, then the
+    tuple key path will always be non-empty as well.
+  - If ``cond_fn`` is not None, then ``Cursor.find_all`` is calling this
+    function and ``_traverse_tree`` will return a generator where each
+    generated element is of type Tuple[Union[str, int], AccessType]. The
+    tuple contains the key path and access type where the object was found
+    that fulfilled the conditions of the ``cond_fn``.
+  """
+  assert not (
```
cgarciae, line 71: I think these should be proper errors

flax/cursor.py
```diff
@@ -105,6 +138,19 @@ def root(self) -> 'Cursor[A]':
     else:
       return self.parent_key.parent.root  # type: ignore
 
+  @property
+  def path(self) -> str:
+    if self.parent_key is None:
+      return 'root'
```
cgarciae, line 144: This is not consistent with how the path is generated for `apply_update`

flax/cursor.py
```diff
@@ -49,97 +51,138 @@ def is_named_tuple(obj):
...
+    change was applied from the ``update_fn``, and the second element is
+    the newly modified value. If the generator is non-empty, then the
+    tuple key path will always be non-empty as well.
+  - If ``cond_fn`` is not None, then ``Cursor.find_all`` is calling this
+    function and ``_traverse_tree`` will return a generator where each
+    generated element is of type Tuple[Union[str, int], AccessType]. The
+    tuple contains the key path and access type where the object was found
+    that fulfilled the conditions of the ``cond_fn``.
+  """
+  if (update_fn and cond_fn) or (update_fn is None and cond_fn is None):
```
cgarciae, line 71:
```suggestion
  if update_fn ^ cond_fn: # xor
```

flax/cursor.py
```diff
@@ -49,97 +51,138 @@ def is_named_tuple(obj):
...
   def root(self) -> 'Cursor[A]':
-    if self.parent_key is None:
+    if self._parent_key is None:
       return self
     else:
-      return self.parent_key.parent.root  # type: ignore
+      return self._parent_key.parent.root  # type: ignore
+
+  @property
+  def path(self) -> str:
```
cgarciae, line 137: I think we should expose the minimum amount of attributes because they act as reserved names the user cannot use on their objects. We should make `root` and `path` private (`_root`, `_path`).

flax/cursor.py
```diff
@@ -327,33 +356,369 @@ def update_fn(path, value):
...
+          child = getattr(child, key)
+      yield child
+
+  def __str__(self):
+    return str(self._obj)
+
+  def __repr__(self):
+    return self.pretty_repr()
+
+  def pretty_repr(self, indent=0):
```
cgarciae, line 551: Rename to `_pretty_repr`

flax/cursor.py
```diff
@@ -49,97 +51,138 @@ def is_named_tuple(obj):
...
   elif dataclasses.is_dataclass(obj):
     items = (
         (f.name, getattr(obj, f.name))
         for f in dataclasses.fields(obj)
         if f.init
     )
-    access_type = AccessType.GETATTR
+    access_type = AccessType.ATTR
   else:
     yield from ()  # empty generator
```
cgarciae, line 102: I missed this in the last PR, empty generator is just `return`.

flax/cursor.py
```diff
@@ -327,33 +356,369 @@ def update_fn(path, value):
...
+    obj_str = (
+        "'" + self._obj + "'" if isinstance(self._obj, str) else str(self._obj)
+    )
+    s = 'Cursor(' + f'_obj={obj_str},\n'
+    s += ' ' * (indent + 7) + '_changes={'
+    if self._changes:
+      s += '\n'
+      for key in self._changes:
+        if isinstance(key, str):
+          str_key = "'" + key + "'"
```
cgarciae, line 561:
```suggestion
          str_key = repr(key)
```

flax/cursor.py
```diff
@@ -327,33 +356,369 @@ def update_fn(path, value):
...
+        else:
+          str_key = str(key)
+        prefix = ' ' * (indent + 7 + 2) + str_key + ': '
```
cgarciae, line 564: `+ 9`? Also, why so much indent? 😲 
I think `pretty_print` uses `4`?

flax/cursor.py
```diff
@@ -327,33 +356,369 @@ def update_fn(path, value):
...
+  def __str__(self):
+    return str(self._obj)
+
+  def __repr__(self):
+    return self.pretty_repr()
+
+  def pretty_repr(self, indent=0):
+    obj_str = (
+        "'" + self._obj + "'" if isinstance(self._obj, str) else str(self._obj)
+    )
```
cgarciae, line 554: Maybe this could be simplified to:
```suggestion
    obj_str = repr(self._obj)
```
However, if `obj_str` includes new-line characters, these wont respect the indent. You can fix by calling something like:

```python
obj_str = obj_str.replace('\n', '\n' + ' ' * indent)
```

flax/cursor.py
```diff
@@ -327,33 +356,369 @@ def update_fn(path, value):
...
+    s = 'Cursor(' + f'_obj={obj_str},\n'
+    s += ' ' * (indent + 7) + '_changes={'
+    if self._changes:
+      s += '\n'
+      for key in self._changes:
+        if isinstance(key, str):
+          str_key = "'" + key + "'"
+        else:
+          str_key = str(key)
```
cgarciae, line 563: This could be simplified to:
```suggestion
      for key in self._changes:
        str_key = repr(key)
```

marcvanzee (approved):

flax/cursor.py
chiamp: I think `^` only works with boolean operands, so I could do something like this:
```
if not (bool(update_fn) ^ bool(cond_fn)):
```
What do you think?

cgarciae: Yeah, I sorry missed. This looks good!

PR #3322, opened 2023-09-11
Add fp8 custom op and unit test

wenscarl:
In this PR, we add one custom op to utilize the fp8 gemms provided by the XLA to better utilize NVIDIA Hopper GPUs. Users can then code-inject this custom op like:
```
dense=nn.DenseGeneral(dot_general_cls=nn.Fp8EinsumOp)
```
The PR also provides unit tests for the injection and accuracy.

Reference:

[FP8 basics](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/fp8_primer.html).

[Basics of QDQ process used in our custom op](https://github.com/NVIDIA/atex/blob/experimental-xla-fp8-layers/experimental/xla-fp8-layers/fp8-tutorial.md).

cc. @pjannaty @reedwm @kaixih @levskaya

### DIFF
--- a/flax/linen/fp8_ops.py
+++ b/flax/linen/fp8_ops.py
@@ -1,8 +1,6 @@
-from typing import Callable, Iterable, Union, Tuple
+from typing import Callable
 from functools import partial
 
-from flax import core
-from flax.core import scope as flax_scope
 from flax.linen import initializers
 from flax.linen.module import Module
 from flax.linen import partitioning as nn_partitioning
@@ -11,25 +9,13 @@ from jax import lax
 from jax import numpy as jnp
 from jax import random
 
-from fp8layers.jax_common.fp8 import fp8_dot_general
 
-param_with_axes = nn_partitioning.param_with_axes
 variable_with_axes = nn_partitioning.variable_with_axes
 
 # Type annotations
 Array = jnp.ndarray
 Dtype = jnp.dtype
 PRNGKey = jnp.ndarray
-Shape = Iterable[int]
-ActivationFn = Callable[..., Array]
-DotGeneralT = Callable[..., Array]
-PrecisionLike = Union[
-    None,
-    str,
-    lax.Precision,
-    Tuple[str, str],
-    Tuple[lax.Precision, lax.Precision],
-]
 
 class FP8Helper:
   FP8_COLLECTION_NAME: str = "fp8_params"
--- a/flax/linen/fp8_ops.py
+++ b/flax/linen/fp8_ops.py
@@ -3,14 +3,11 @@ from functools import partial
 
 from flax.linen import initializers
 from flax.linen.module import Module
-from flax.linen import partitioning as nn_partitioning
 from jax import custom_vjp
 from jax import lax
 from jax import numpy as jnp
 from jax import random
 
-
-variable_with_axes = nn_partitioning.variable_with_axes
 
 # Type annotations
 Array = jnp.ndarray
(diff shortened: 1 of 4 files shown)


tests/linen/linen_linear_test.py
```diff
@@ -1016,6 +1017,66 @@ def __call__(self, x):
...
+
+      def _train(variables, x):
+        y = p.apply(variables, x)
+        loss = y * dy
+        return jnp.mean(loss)
+      train_fn = jax.jit(jax.value_and_grad(_train, argnums=[0, 1]))
+      outputs, grads = train_fn(initial_vars, x)
+      return outputs, grads
+     
+    #y, variables = Foo().init_with_output(random.PRNGKey(0), x)
```
kaixih, line 1057: Can we remove the commented lines?

flax/linen/fp8_ops.py
```diff
@@ -0,0 +1,207 @@
...
+variable_with_axes = nn_partitioning.variable_with_axes
+
+# Type annotations
+Array = jnp.ndarray
+Dtype = jnp.dtype
+PRNGKey = jnp.ndarray
+Shape = Iterable[int]
+ActivationFn = Callable[..., Array]
+DotGeneralT = Callable[..., Array]
+PrecisionLike = Union[
```
kaixih, line 26: Please clean up the unused alias and imports.

tests/linen/linen_linear_test.py
```diff
@@ -1016,6 +1017,66 @@ def __call__(self, x):
     )
     self.assertEqual(y.shape, (2, 8, 6))
 
+  def test_fp8_dot_general_cls_injection(self):
+    # Used to cast the inputs to be representable in FP8, so that the difference
+    # of the results from the original gemm and fp8 gemm is small.
+    cast_to_representable = functools.partial(nn.fp8_quantize_dequantize,
+                                              scale=jnp.ones((1,)),
+                                              compute_dtype=jnp.float32)
+    class Foo(nn.Module):
```
kaixih, line 1026: Why do we need a Foo class? Can we directly use the DenseGeneral and provide our custom op to it?

kaixih (commented):
Thanks. Just added some minor comments.

flax/linen/fp8_ops.py
```diff
@@ -0,0 +1,207 @@
...
+from flax.core import scope as flax_scope
+from flax.linen import initializers
+from flax.linen.module import Module
+from flax.linen import partitioning as nn_partitioning
+from jax import custom_vjp
+from jax import lax
+from jax import numpy as jnp
+from jax import random
+
+from fp8layers.jax_common.fp8 import fp8_dot_general
```
kaixih, line 14: This will cause module importing error and needs to be removed. Please clean up other unused imports.

flax/linen/fp8_ops.py
```diff
@@ -0,0 +1,193 @@
...
+        jnp.float32,
+    )
+    amax_history_args = (
+        initializers.zeros_init(),
+        random.PRNGKey(0),
+        (self.amax_history_length,),
+        jnp.float32,
+    )
+
+    self.input_amax_history = variable_with_axes(
```
levskaya, line 143: ...why are you using the deprecated paritioning helper `variable_with_axes` to seed a trivial piece of metadata here?  Why not just use `self.variable`?  I don't see anything in this code that's actually doing any non-trivial operation using this metadata.

flax/training/train_state.py
```diff
@@ -15,10 +15,23 @@
...
+  axis_names = nn.partitioning.get_axis_names(params_axes)
+  missing_params_axes = (
+      set(traverse_util.flatten_dict(params, sep='/')) -
+      set(traverse_util.flatten_dict(axis_names, sep='/')))
+  if missing_params_axes:
+    raise ValueError(
+        f'Missing axis names for parameters: {missing_params_axes}')
+
+
 class TrainState(struct.PyTreeNode):
```
levskaya, line 35: I might ask that we add this FP8 specific logic to a subclass of the general TrainState class? (it can be located in this file).

levskaya (commented):
I can help advise how to replace it, but definitely don't use the old `variable_with_axes`.  I'm not sure you even need it for what you're trying to do here.  e.g. If you just want to mark the fp8 use-places we can use a simple flax.struct.dataclass with a single-field as a kind of leaf type-marker.

flax/linen/fp8_ops.py
kaixih: Sure. I think we can simply use the `self.variable`. The reason why we used the `variable_with_axes`  was that we originally developed this in the context of t5x, where it requires all the variables have the axes information. Since this PR focuses on the FLAX, let's remove them. @wenscarl

wenscarl: `variable_with_axes` has been replaced by `self.params`.

tests/linen/linen_linear_test.py
```diff
@@ -1016,6 +1020,142 @@ def __call__(self, x):
...
+
+  def test_fp8_with_train_state(self):
+    x = random.uniform(random.PRNGKey(1), (16, 16), dtype=jnp.float32)
+    dense = nn.DenseGeneral(features=32, use_bias=True,
+                            dot_general_cls=nn.Fp8DenseGeneralOp)
+    key = random.PRNGKey(0)
+    variables = dense.init(key, x)
+
+    opt = optax.adam(learning_rate=.1)
+    state = train_state.TrainState.create(params=variables, tx=opt,
```
levskaya, line 1084: shouldn't this be `Fp8TrainState` now?

levskaya (approved):
Looks good - just one question inline about your test code:

levskaya:
bleh sorry it looks like pyink formatter is complaining again (we're about to disable it for a while as this is proving more annoying than helpful - sorry about that.)

levskaya:
We disabled the autoformatter stuff, I tried retriggering tests, but it's not capturing the update, so if you rebase and push a new commit hopefully it will clear.  Sorry for the trouble.

tests/linen/linen_linear_test.py
wenscarl: Good catch and fixed. Thanks!

PR #3330, opened 2023-09-13
BugFix 836e0e4 PR #3293 - TypeError: 'type' object is not subscriptable

DEKHTIARJonathan:
Fixing Bug introduced in 836e0e439bc3dc6f9fe91446a1e50bc3efe89e70 @cgarciae - PR #3293 (reviewer: @superbobry)

```bash
$ python --version
Python 3.8.10

$ pip freeze | grep jax
jax==0.4.13
jaxlib==0.4.13

$ pip freeze | grep flax
flax==0.7.3
```

```python
>>> from flax.core.frozen_dict import FrozenDict
  File "<stdin>", line 1, in <module>
  File "/home/user/.local/lib/python3.8/site-packages/flax/__init__.py", line 25, in <module>
    from . import linen
  File "/home/user/.local/lib/python3.8/site-packages/flax/linen/__init__.py", line 34, in <module>
    from .activation import (
  File "/home/user/.local/lib/python3.8/site-packages/flax/linen/activation.py", line 21, in <module>
    from flax.linen.module import compact
  File "/home/user/.local/lib/python3.8/site-packages/flax/linen/module.py", line 68, in <module>
    from flax.linen import kw_only_dataclasses
  File "/home/user/.local/lib/python3.8/site-packages/flax/linen/kw_only_dataclasses.py", line 126, in <module>
    def _process_class(cls: type[M], extra_fields=None, **kwargs):
TypeError: 'type' object is not subscriptable
```

Small bug though fairly wide implications, it's breaking a whole lot of code with the new release `0.7.3`

***Note:***
It might be linked to a minimum version of python 3.9 for JAX for the most recent releases. Though I would argue that it's an improper assumptions given that the dependency game might install FLAX `0.7.3` with an older version of JAX (which does support python 3.8). And nothing really prevents FLAX to be compatible with older versions of python. 
Given the super narrow impact of such a change, (and that other places of FLAX also use `typing.Type` -> example: https://github.com/google/flax/blob/main/flax/linen/transforms.py#L88) I am of the opinion that it's a zero impact change for FLAX and help FLAX customers/users and the ecosystem.

### DIFF
--- a/flax/linen/kw_only_dataclasses.py
+++ b/flax/linen/kw_only_dataclasses.py
@@ -54,7 +54,7 @@ import dataclasses
 import inspect
 import functools
 from types import MappingProxyType
-from typing import Any, TypeVar
+from typing import Any, Type, TypeVar
 from typing_extensions import dataclass_transform
 import flax
 
@@ -123,7 +123,7 @@ def dataclass(cls=None, extra_fields=None, **kwargs):
   return wrap if cls is None else wrap(cls)
 
 
-def _process_class(cls: type[M], extra_fields=None, **kwargs):
+def _process_class(cls: Type[M], extra_fields=None, **kwargs):
   """Transforms `cls` into a dataclass that supports kw_only fields."""
   if '__annotations__' not in cls.__dict__:
     cls.__annotations__ = {}
@@ -205,7 +205,7 @@ def _process_class(cls: type[M], extra_fields=None, **kwargs):
   create_init = '__init__' not in vars(cls) and kwargs.get('init', True)
 
   # Apply the dataclass transform.
-  transformed_cls: type[M] = dataclasses.dataclass(cls, **kwargs)
+  transformed_cls: Type[M] = dataclasses.dataclass(cls, **kwargs)
 
   # Restore the base classes' __dataclass_fields__.
   for _cls, fields in base_dataclass_fields.items():


levskaya:
We are intending to support only the python minimum version that JAX does - so 3.9 at this point - I think we need to be more clear about that fact.

DEKHTIARJonathan:
@levskaya quite possibly. Though it's kinda weird to on purpose breaks backward compatibility when it brings no advantage and actively cause problems for some users (who might get FLAX by transitive dependency).

levskaya:
I'm fixing the messed up package metadata now - we'll yank this release and redo the release to fix transient packaging issues hopefully.

Generally we follow JAX and [nep-0029](https://numpy.org/neps/nep-0029-deprecation_policy.html) We will be using features present in python3.9 soon, beyond just typing improvements.  We're just about to require the new version of JAX  which will force a 3.9 minimum.  The downstream package should pin to flax=0.7.2 if automatic resolution doesn't work.

levskaya:
We yanked 0.7.3 and pushed a new 0.7.4 with correct python version metadata - hopefully this helps.

DEKHTIARJonathan:
@levskaya it does
```
$ pip freeze | grep flax
flax==0.7.2
```

I'm closing

PR #3340, opened 2023-09-15
Add Sharding Annotations to Flax Modules

terrykong:
Adds sharding annotations to Flax's `Dense`, `DenseGeneral`, `MultiHeadDotProductAttention` and normalization modules. This change makes Flax's modules compatible with [T5X](https://github.com/google-research/t5x).

### DIFF
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -27,6 +27,7 @@ from flax.linen.module import compact
 from flax.linen.module import merge_param
 from flax.linen.module import Module
 from flax.linen.normalization import LayerNorm
+from flax.linen.partitioning import variable_with_axes
 import jax
 from jax import lax
 from jax import random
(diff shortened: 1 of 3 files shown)


flax/linen/attention.py
```diff
@@ -313,15 +332,15 @@ def __call__(
...
-      cached_key = self.variable(
-          'cache', 'cached_key', jnp.zeros, key.shape, key.dtype
-      )
-      cached_value = self.variable(
-          'cache', 'cached_value', jnp.zeros, value.shape, value.dtype
-      )
-      cache_index = self.variable(
-          'cache', 'cache_index', lambda: jnp.array(0, dtype=jnp.int32)
-      )
+      cached_key = variable_with_axes('cache', 'cached_key',
```
levskaya, line 335: `param_with_axes` and `variable_with_axes` is from an older, deprecated API for sharding (we need to delete it or just move it into the t5x repo which is the primary user).  The newer approach can be found in `core/meta.py` and `linen/spmd.py` and relies on using "boxed" variables that are created by wrapping array initializers.

levskaya (commented):
We aren't quite ready yet to add baked-in sharding to the default flax layers.  We'd recommend checking out the [maxtext](https://github.com/google/maxtext) repo for a highly tested example of a flax-based transformer library.

PR #3342, opened 2023-09-15
updated flax_basics

### LINKED ISSUE #3260: Documentation question on "Variables"
In the "Variables and collections of variables" section of [Flax basics](https://flax.readthedocs.io/en/latest/guides/flax_basics.html), there's a line of code which says ra_mean.value either initializes or get the value.

```
class BiasAdderWithRunningMean(nn.Module):
  decay: float = 0.99

  @nn.compact
  def __call__(self, x):
    # easy pattern to detect if we're initializing via empty variable tree
    is_initialized = self.has_variable('batch_stats', 'mean')
    ra_mean = self.variable('batch_stats', 'mean',
                            lambda s: jnp.zeros(s),
                            x.shape[1:])
    mean = ra_mean.value # This will either get the value or trigger init
    bias = self.param('bias', lambda rng, shape: jnp.zeros(shape), x.shape[1:])
    if is_initialized:
      ra_mean.value = self.decay * ra_mean.value + (1.0 - self.decay) * jnp.mean(x, axis=0, keepdims=True)

    return x - ra_mean.value + bias
```

But when reading the code, it looks to me the variable should have initialized already. This is part of Scope code:

```
  def variable(
      self,
      col: str,
      name: str,  # pylint: disable=keyword-arg-before-vararg
      init_fn: Optional[Callable[..., T]] = None,
      *init_args,
      unbox: bool = True,
  ) -> Variable[T]:
    """Creates a variable if it doesn't exist yet in this scope and returns it.

    Args:
      col: the collection of the variable.
      name: the name of the variable.
      init_fn: a function taking a PRNGKey plus any other number of positional
        arguments. If None, the variable must already be initialized otherwise
        an error is raised.
      *init_args: the arguments to evaluate init_fn on lazily.
      unbox: If True, ``AxisMetadata`` instances are replaced by their unboxed
        value, see ``flax.nn.meta.unbox`` (default: True).

    Returns:
      The variable.  Throws an error if the variable exists already.
    """
    self.reserve(name, col)
    if not self.has_variable(col, name):
      if not self.is_mutable_collection(col) or init_fn is None:
        if self.is_collection_empty(col):
          raise errors.ScopeCollectionNotFound(col, name, self.path_text)
        raise errors.ScopeVariableNotFoundError(name, col, self.path_text)
      init_value = init_fn(*init_args)
      self.put_variable(col, name, init_value)
    return Variable(self, col, name, unbox=unbox)
```

See that `init_fn(*init_args)` is already called.

So, it looks to me that init is already done without `ra_mean.value`.

Please remove the line of `mean = ra_mean.value # This will either get the value or trigger init`. It made me wonder if that pattern should always exist whenever I add a variable and why we don't need to do similar for bias.

### LINKED ISSUE #3261: Need to explain lax.dot_general in Flax basics
In the Module Parameters section of [Flab basics](https://flax.readthedocs.io/en/latest/guides/flax_basics.html), there's a line of code with TODO under `SimpleDense`.

```
    y = lax.dot_general(inputs, kernel,
                        (((inputs.ndim - 1,), (0,)), ((), ())),) # TODO Why not jnp.dot?
```

Could you pleaes add explanation why it's not jnp.dot? It's the first doc I read and this line made me feel wonder if I can rely on tutorials to learn flax.

Using jnp.dot on the line. Or some high level explanation.

chiamp:
Resolves #3260.
Removed `mean = ra_mean.value # This will either get the value or trigger init` in "Variables and collections of variables" because the code line is unnecessary and the comment is incorrect.

Resolves #3261.
Changed `lax.dot_general` to `jnp.dot`.

### DIFF
--- a/docs/guides/flax_basics.ipynb
+++ b/docs/guides/flax_basics.ipynb
@@ -68,7 +68,7 @@
    "source": [
     "import jax\n",
     "from typing import Any, Callable, Sequence\n",
-    "from jax import lax, random, numpy as jnp\n",
+    "from jax import random, numpy as jnp\n",
     "import flax\n",
     "from flax import linen as nn"
    ]
@@ -775,8 +775,7 @@
     "    kernel = self.param('kernel',\n",
     "                        self.kernel_init, # Initialization function\n",
     "                        (inputs.shape[-1], self.features))  # shape info.\n",
-    "    y = lax.dot_general(inputs, kernel,\n",
-    "                        (((inputs.ndim - 1,), (0,)), ((), ())),) # TODO Why not jnp.dot?\n",
+    "    y = jnp.dot(inputs, kernel)\n",
     "    bias = self.param('bias', self.bias_init, (self.features,))\n",
     "    y = y + bias\n",
     "    return y\n",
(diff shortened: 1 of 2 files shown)


8bitmp3:
:+1:

IvyZX (approved):

PR #3343, opened 2023-09-15
updated haiku guide with new JAX RNG api

chiamp:
updated haiku guide, after #3337 changed Flax to use the new Jax RNG api


IvyZX (approved):
Thanks for fixing it!

PR #3354, opened 2023-09-20
use jax.Array type for rng keys

chiamp:
for more details, see [the Jax PR](https://github.com/google/jax/pull/17297) and [JEP 9263](https://jax--17297.org.readthedocs.build/en/17297/jep/9263-typed-keys.html)


IvyZX (approved):
Thanks for making the change!

PR #3358, opened 2023-09-21
Make checkpoint guide path absolute

IvyZX:
This should unbreak head - if it's possible to modify "/tmp" directory in CI :)


chiamp (approved):
Thanks!

PR #3370, opened 2023-09-26
updated docs

chiamp:
Updated docs.

A couple highlights:
* [Removed](https://github.com/google/flax/pull/3370/files#diff-dce016d0e4f8a6b6c5bda025db3ab4fea095d2bf2843db45f1dab22a1cb226bdL59-L62) BlackJax example since the link is broken and it seems like the [current codebase](https://github.com/blackjax-devs/blackjax) doesn't use Flax (although it is listed as a [required dependency](https://github.com/blackjax-devs/blackjax/blob/main/requirements-doc.txt))
* Added a more detailed [example](https://github.com/google/flax/pull/3370/files#diff-9401b9f921c0646d40ec072c53d29fea7051e47410b0a30340e54c3012c86a0fR38-R47) in the docstring of `Dropout`, since a new user may find it unclear what the `train=True` argument is for in the original code snippet:
```
model.apply({'params': params}, inputs=inputs, train=True, rngs={'dropout': dropout_rng})
```

### DIFF
--- a/docs/api_reference/flax.cursor.rst
+++ b/docs/api_reference/flax.cursor.rst
@@ -12,7 +12,7 @@ To illustrate, consider the example below::
   import dataclasses
   from typing import Any
 
-  @dataclasses.dataclass
+  @dataclasses.dataclass(frozen=True)
   class A:
     x: Any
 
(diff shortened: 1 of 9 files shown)


IvyZX (approved):
Thank you so much!

8bitmp3 (approved):
:+1: :+1: :+1: Thank you

PR #3395, opened 2023-10-04
added warning filter for jax.random.KeyArray

chiamp:
Warning is due to a [`chex`](https://github.com/google/flax/actions/runs/6407312853/job/17393862864?pr=3394#step:9:570) import from `tests/jax_utils_test.py`


jakevdp:
It looks like the warning was coming from `chex`; this was fixed in https://github.com/google-deepmind/chex/pull/307, which is part of the 0.1.83 release, so you should be able to fix this without ignoring the warnings by pinning chex.

Not filtering the warning might be helpful to avoid inadvertently using deprecated JAX APIs.

PR #3396, opened 2023-10-04
add truncated_normal to initializers

levskaya:


cgarciae:
We should add the cur corresponding entry to the docs.

PR #3419, opened 2023-10-16
fixed GRU docstring

chiamp:
Fixed GRU docstring and [math equations](https://flax--3419.org.readthedocs.build/en/3419/api_reference/flax.linen/layers.html#flax.linen.GRUCell) so that it's accurately representing what's happening in the code.

Before the math equations were:
Screenshot 2023-10-16 at 3 51 42 PM

Now it's:
Screenshot 2023-10-16 at 3 51 59 PM


cgarciae (approved):
Thanks for catching this @chiamp!

PR #3425, opened 2023-10-19
fully deprecated old RNN api

### LINKED ISSUE #3399: Remove RNNCellBase refactor error messages
During the `RNNCellBase` refactor some tooling was created to produce error messages if users tried to leverage the old API. Now that a few months have passed, it would be a bit safer to assume that most active users have seen that error message and have fixed their code so we can remove this tooling.

https://github.com/google/flax/blob/0347dd2f34b1a581dd664698b7c842ca2f650745/flax/linen/recurrent.py#L60-L66

chiamp:
Resolves #3399.
Fully deprecated old RNN API.


cgarciae (approved):
Thanks @chiamp ! Looks good.

PR #3428, opened 2023-10-20
Fix warning with random number generation

NeilGirdhar:
This Flax function creates a dummy RNG key consisting of an empty uint32 array, and uses it as an RNG key.  Recent versions of Jax warn on using uint32 arrays as RNG keys.

To prevent the warning and simplify the code, just call jax.random.key(0).

# What does this PR do?

Fixes # (issue)

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -954,9 +954,8 @@ class Scope:
     """
     self.reserve(name, 'params')
     if self.has_variable('params', name):
-      abs_rng = jax.ShapeDtypeStruct(
-          random.default_prng_impl().key_shape, jnp.uint32
-      )
+      # The seed is irrelevant since we only want to evaluate the shape.
+      abs_rng = random.key(0)
       value = self.get_variable('params', name)
       # Validate that the shape of the init_fn output is the same as the shape
       # of the existing parameter. This is to make sure that the hparams set up


levskaya:
I think #3424 will fix this, I'm just waiting on github's CI runners to stop being busted to get them in.

NeilGirdhar:
@levskaya Unfortunately, I don't think it does:
```
In [2]: import jax.random as random
In [3]: import jax
In [6]: with jax.enable_custom_prng():
   ...:     abs_rng = jax.eval_shape(lambda s: random.key_data(random.key(s)), 0)
   ...: 
In [7]: abs_rng
Out[7]: ShapeDtypeStruct(shape=(2,), dtype=uint32)
```
The problem is you're still calling `key_data`, which gives you the `uint32` arrays.  I think you should just call `key`.

levskaya:
Thanks Neil!  

I tried getting a warning at jax HEAD but couldn't trigger it, but taking another look at this old code it is a bit daft.  I think we can just fold the PRNG key instantiation into the `eval_shape` (the goal is to avoid _running_ any actual code).  That should avoid any present or future issue.

I tried doing that at #3429 - I expect tests should pass, but I can submit that to avoid this altogether.

NeilGirdhar:
@levskaya Sounds great, thanks for solving these issues proactively :)

I'll close this PR in deference to yours.

PR #3431, opened 2023-10-23
Remove transformers dependency from docs

cgarciae:
# Changes

`transformers` has a `flax<=0.7.0` constraint that is currently making installation of the docs impossible. To get around this, this PR removes the `transformers`dependency from `docs/requirements.txt` and excludes the `transfer_learning.ipynb` notebook from running on CI.


chiamp (approved):
Thanks!

PR #3441, opened 2023-10-26
[NVIDIA] Update the algorithm to compute fp8 scales

kaixih:
This pull request implements updates to the algorithm for computing new FP8 scales, in line with the design outlined in the following documentation: https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/jax.html#transformer_engine.jax.update_fp8_metas.

cc. @wenscarl @reedwm @lukaszlew @levskaya

### DIFF
--- a/flax/linen/fp8_ops.py
+++ b/flax/linen/fp8_ops.py
@@ -155,7 +155,7 @@ class Fp8DotGeneralOp(module.Module):
         OVERWRITE_WITH_GRADIENT, 'output_grad_scale', *scale_args)
 
 
-  def __call__(self, *args, **kwargs) -> jnp.ndarray:
+  def __call__(self, *args, **kwargs):
 
     assert len(args) == 3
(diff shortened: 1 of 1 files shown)


levskaya:
The change looks OK, but mypy is now yielding a type error - would you mind double-checking if that's finding something real, or is it spurious?

kaixih:
@levskaya thank you for bringing this to my attention. I've removed the function return value annotation that was causing the previous mypy check to trigger. I don't believe it's necessary because this custom operation for FP8 is not configured to the layers, such as DenseGeneral, by default. There won't be any other type checking tracing back to this operation.

kaixih:
@levskaya Any updates? I see this PR has stayed in "pull ready" status for a couple of days.

levskaya:
@kaixih - super sorry for the delay, not sure why it didn't go through earlier!  It should have merged last night.

PR #3448, opened 2023-10-30
Forked a subset of JAX configuration APIs

superbobry:
These APIs are internal to JAX and should not be used by other projects for managing their configuration.

### DIFF
--- a/docs/api_reference/flax.config.rst
+++ b/docs/api_reference/flax.config.rst
@@ -4,4 +4,5 @@ flax.config package
 
 .. automodule:: flax.configurations
     :members:
-    :exclude-members: temp_flip_flag
+    :undoc-members:
+    :exclude-members: FlagHolder, bool_flag, temp_flip_flag, static_bool_env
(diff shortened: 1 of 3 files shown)


IvyZX:
Thanks for working on it! 

1. We only have bool flags now, but with this setup, does adding non-bool flags just requires forking a few more lines?

2. Since we no longer rely on an existing framework, can we have a simple unit test for this? To make sure the runtime update works, and that the flag value will not reset after re-importing the `configuration.py`. Thank you!

superbobry:
Thanks for having a look, Ivy! 

1. Yes, adding support for new flag types should be easy. E.g. here is how `int_flag` could look
 
    ```python
    def int_flag(name: str, default: bool, help: str) -> FlagHolder[int]:
      name = name.lower()
      if default_env := os.getenv(name.upper()):
        try:
          default = int(default_env)
        except ValueError:
          raise ValueError(f'Invalid value "{default_env}" for flag {name}')
      config._add_option(name, default)
      fh = FlagHolder[int](name, help)
      setattr(Config, name, property(lambda _: fh.value, doc=help))
      return fh
    ```
2. Will add a test, sure!

IvyZX (approved):

PR #3461, opened 2023-11-04
Add user-friendly module copy method.

levskaya:
This provides a version of the `Module.clone()` method with simpler arguments and better defaults for allowing copying of modules and the use of existing modules as "templates".  By default `copy` resets `parent` and `name` attributes of the given module, but effectively performs a deep copy of everything "inside".


cgarciae (approved):
looks good!

PR #3463, opened 2023-11-06
Add simple argument-only lifted nn.grad function.

levskaya:
This function only peforms a lifted value-and-grad operation with respect to the arguments of a function, and does not try to calculate gradients with respect to any variables.  This mirrors the behavior of the haiku grad function, and also easily works in the multi-scope setting (external modules passed in) while avoiding the complexities associated with that case for a more general vjp.

### DIFF
--- a/flax/linen/transforms.py
+++ b/flax/linen/transforms.py
@@ -1055,7 +1055,8 @@ def vjp(
   vjp_variables: lift.CollectionFilter = 'params',
   variables: lift.CollectionFilter = True,
   rngs: lift.PRNGSequenceFilter = True,
-) -> Tuple[Any, Any]:
+  multi_scope: bool = False,
+):
   """A lifted version of ``jax.vjp``.
 
   See ``jax.vjp`` for the unlifted vector-Jacobiam product (backward gradient).
@@ -1105,7 +1106,8 @@ def vjp(
     variables: other variables collections that are available inside `fn` but
       do not receive a cotangent.
     rngs: the prngs that are available inside `fn`.
-
+    multi_scope: for Modules containing multiple scopes from outside modules passed in,
+      allow for variable gradients to be returned for multiple scopes instead of erroring.
   Returns:
     If ``has_aux`` is ``False``, returns a ``(primals_out, vjpfun)`` pair, where
     ``primals_out`` is ``fn(*primals)``.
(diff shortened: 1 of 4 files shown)


flax/linen/transforms.py
```diff
@@ -1130,6 +1132,95 @@ def __call__(self, x, y):
...
+    If ``has_aux`` is ``False``, returns a ``primals_out, grads`` pair, where
+    ``primals_out`` is ``fn(*primals)``.  ``grads`` are the gradients for the
+    corresponding primals and do not include the gradients for module variables.
+    If ``has_aux`` is ``True``, returns a
+    ``(primals_out, aux), grads`` tuple where ``aux`` is the auxiliary data
+    returned by ``fn``.
+  """
+
+  @functools.wraps(fn)
+  def grad_fn(mdl, *primals):
```
cgarciae, line 1198: why do we create `grad_fn` and immediately use it below?

flax/linen/transforms.py
```diff
@@ -1130,6 +1132,95 @@ def __call__(self, x, y):
...
+def grad(
+  fn: Callable[..., Any],
+  mdl: Module,
+  *primals,
+  has_aux: bool = False,
+  reduce_axes=(),
+  variables: lift.CollectionFilter = True,
+  rngs: lift.PRNGSequenceFilter = True,
+):
+  """A limited, lifted equivalent of ``jax.value_and_grad``.
```
cgarciae, line 1144: Should we call this `value_and_grad` then for consistency with `jax` APIs?

flax/linen/transforms.py
```diff
@@ -1130,6 +1132,95 @@ def __call__(self, x, y):
...
+      reduce_axes=reduce_axes,
+      vjp_variables=False,
+      variables=variables,
+      rngs=rngs,
+      multi_scope=True,
+    )
+
+    if has_aux:
+      out, vjp_fun, aux = vjp_partial()
+      _, *argument_grads = vjp_fun(jax.numpy.ones_like(out))
```
cgarciae, line 1214: Don't we get an additional dimension compared to `grad` here? Maybe we need to `squeeze` the additional dim?

levskaya: huh? I don't think so?  why would it add dimensions?  you can double-check the tests

flax/linen/transforms.py
levskaya: this is how the "direct" vjp/jvp transforms work - they directly call the function.

levskaya: oh lol, I see what you mean - it's completely pointless here, yeah I'll fix it.

cgarciae (approved):

PR #3470, opened 2023-11-07
remove SelfAttention test and warning filter

chiamp:
Since `SelfAttention` is deprecated, remove `SelfAttention` from tests and remove the deprecation warning filter


cgarciae (approved):
looks good!

PR #3482, opened 2023-11-13
disabled ruff formatter

chiamp:
Disabled ruff formatter for now because hanging indent can't be configured separately from regular indent, whereas the internal Google formatter has 4 space hanging indent and 2 space regular indent.

Original PR: #3455
Feature request for `ruff`: https://github.com/astral-sh/ruff/issues/3907


cgarciae (approved):
hope we can re-enable it in the future

Issue #3508, opened 2023-11-24
Add missing `mask` arguments to `__call__` methods of normalization modules


carlosgmartin:
Currently, only [BatchNorm](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/layers.html#flax.linen.BatchNorm.__call__) receives a `mask` argument in its `__call__` method.

**Feature request:** Add the same `mask` argument to the remaining normalization modules:
* [LayerNorm](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/layers.html#flax.linen.LayerNorm.__call__)
* [GroupNorm](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/layers.html#flax.linen.GroupNorm.__call__)
* [RMSNorm](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/layers.html#flax.linen.RMSNorm.__call__)

I can submit a PR for this, if desired.

chiamp:
Closing this after #3510 landed

PR #3510, opened 2023-11-28
Add missing mask argument to LayerNorm, RMSNorm, and GroupNorm.

carlosgmartin:
https://github.com/google/flax/issues/3508


chiamp:
Thanks for this! Can you also add tests using the mask for the normalization layers in the [`NormalizationTest`](https://github.com/google/flax/blob/main/tests/linen/linen_test.py#L139) class in `tests/linen/linen_test.py`?

carlosgmartin:
@chiamp Done.

PR #3517, opened 2023-11-29
Make fp8 scaling factors scalar

wenscarl:
The original design of Fp8 scaling factors is of shape (1,) which could result in a fusion pattern of convert->bitcast->convert if bf16 is used as wider type. The fused kernel is run on GPU with block and grid of size <<<1,1,1,>>> which is inefficient. Since cublasLt Fp8 gemm only takes Fp32 typed scales, this change could avoid the above issue with the help of `SimplifyFPConversions` pass.


cgarciae (approved):

cgarciae:
Hey @wenscarl, sorry it took so long.
I approved it but you have some errors.

PR #3518, opened 2023-11-30
[nnx] Fix graph_utils bug

cgarciae:
# What does this PR do?

* Fixes a bug in flatten/unflatten logic in `graph_utils`, `Variable`s were not being copied which is a problem in new Reference semantics.

### DIFF
--- a/flax/experimental/nnx/nnx/graph_utils.py
+++ b/flax/experimental/nnx/nnx/graph_utils.py
@@ -147,6 +147,9 @@ class _HashableMapping(tp.Mapping[HA, HB], tp.Hashable):
   def __init__(self, mapping: tp.Mapping[HA, HB] | tp.Iterable[tuple[HA, HB]]):
     self._mapping = dict(mapping)
 
+  def __contains__(self, key: object) -> bool:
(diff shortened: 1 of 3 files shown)


flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -147,6 +147,9 @@ class _HashableMapping(tp.Mapping[HA, HB], tp.Hashable):
   def __init__(self, mapping: tp.Mapping[HA, HB] | tp.Iterable[tuple[HA, HB]]):
     self._mapping = dict(mapping)
 
+  def __contains__(self, key: object) -> bool:
```
superbobry, line 150: Is this unrelated? I think `Mapping` implements `__contains__` in terms of other dunder methods.

superbobry (approved):

flax/experimental/nnx/nnx/graph_utils.py
cgarciae: Yeah it does by calling getitem and catching a KeyError but that is super annoying for the debugger so I overwrote it.

PR #3523, opened 2023-11-30
Add StateVariablesMapping

cgarciae:
# What does this PR do?
* Renames `VariablesMapping` to `ModuleVariablesMapping`
* `State.variables` now returns a `StateVariablesMapping` with a similar behavior to `ModuleVariablesMapping`.
* Adds `State.raw_mapping` which exposes `._mapping`

### DIFF
--- a/flax/experimental/nnx/nnx/state.py
+++ b/flax/experimental/nnx/nnx/state.py
@@ -55,11 +55,14 @@ class StateVariablesMapping(
     else:
       object.__setattr__(self, '_mapping', mapping)
 
-  def __getitem__(self, name: str) -> Variable[tp.Any]:
-    value = self._mapping[name]
+  def __getitem__(self, key: str | int) -> Variable[tp.Any]:
+    if isinstance(key, int):
+      key = str(key)
+
+    value = self._mapping[key]
 
     if not isinstance(value, Variable):
-      raise KeyError(f"Variable '{name}' not found.")
+      raise KeyError(f"Variable '{key}' not found.")
 
     return value
 
--- a/flax/experimental/nnx/nnx/state.py
+++ b/flax/experimental/nnx/nnx/state.py
@@ -44,6 +44,55 @@ Key = str
 FlatState = dict[Path, Variable[Leaf]]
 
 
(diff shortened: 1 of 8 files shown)


flax/experimental/nnx/nnx/state.py
```diff
@@ -44,6 +44,58 @@
...
+  def __setitem__(self, name: str, value: Variable[tp.Any]) -> None:
+    self._mapping[name] = value
+
+  def __getattr__(self, name: str) -> Variable[tp.Any]:
+    value = self._mapping[name]
+    if not isinstance(value, Variable):
+      raise AttributeError(f"Variable '{name}' not found.")
+    return value
+
+  def __setattr__(self, name: str, value: Variable[tp.Any]) -> None:
```
superbobry, line 78: I would recommend to constrain the API and allow either `["x"]` or `.x` but not both (possibly in a separate PR).

flax/experimental/nnx/nnx/state.py
```diff
@@ -72,9 +124,13 @@ def __init__(
       super().__setattr__('_mapping', dict(mapping))
 
   @property
-  def variables(self) -> dict[Key, dict[str, tp.Any] | tp.Any]:
+  def raw_mapping(self) -> dict[Key, dict[str, tp.Any] | tp.Any]:
```
superbobry, line 127: Another naming idea: `storage`.

superbobry (approved):

flax/experimental/nnx/nnx/state.py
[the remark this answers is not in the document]
cgarciae: In this case we a embracing mutability in case the user wants to do some surgery.

flax/experimental/nnx/nnx/state.py
cgarciae: We have this dual API in a couple of place in NNX, the main idea is to make intuitive both hard-coded and programmatic access, e.g:

```python
# hard-coded
x = state.linear.kernel
# programmatic
x = state[module_name][param_name]
```

PR #3529, opened 2023-12-05
added sow attention weights

### LINKED ISSUE #3530: Feature request: optionally sow attention weight in `dot_product_attention`
### Discussed in https://github.com/google/flax/discussions/2869

<sup>Originally posted by **JyChang012** February 13, 2023</sup>
It seems that there is no good way to retrieve attention weight using `MultiHeadDotProductAttention`. Probably we can add an optional `mdl` parameter to  `dot_product_attention` and `dot_product_attention_weights`, and `sow` the weight inside. This should have no effect when `intermediates` is not mutable. Here is my [implementation](https://github.com/JyChang012/espnet/blob/jax/asr_train/espnex/models/transformer/attention.py). I can create a PR if you like.</div>

Also requested in #3496

chiamp:
Resolves #3530 using @JyChang012's implementation of sowing attention weights; this behavior can be configured at call time. An alternative option would be to return the weights as a tuple, as how [Pytorch](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention.forward) and [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/MultiHeadAttention) does it

### DIFF
--- a/flax/linen/attention.py
+++ b/flax/linen/attention.py
@@ -50,6 +50,7 @@ def dot_product_attention_weights(
   deterministic: bool = False,
   dtype: Optional[Dtype] = None,
   precision: PrecisionLike = None,
+  module: Optional[Module] = None,
 ):
   """Computes dot-product attention weights given query and key.
 
@@ -76,6 +77,10 @@ def dot_product_attention_weights(
     dtype: the dtype of the computation (default: infer from inputs and params)
     precision: numerical precision of the computation see `jax.lax.Precision`
       for details.
+    module: the Module that will sow the attention weights into the
+      'intermediates' collection. Remember to mark 'intermediates' as mutable via
+      `mutable=['intermediates'] in order to have that collection returned.
+      If `module` is None, the attention weights will not be sowed.
 
   Returns:
     Output of shape `[batch..., num_heads, q_length, kv_length]`.
@@ -107,6 +112,9 @@ def dot_product_attention_weights(
   # normalize the attention weights
   attn_weights = jax.nn.softmax(attn_weights).astype(dtype)
 
+  if module:
+    module.sow('intermediates', 'attention_weights', attn_weights)
+
   # apply attention dropout
   if not deterministic and dropout_rate > 0.0:
     keep_prob = 1.0 - dropout_rate
@@ -134,6 +142,7 @@ def dot_product_attention(
   deterministic: bool = False,
   dtype: Optional[Dtype] = None,
   precision: PrecisionLike = None,
+  module: Optional[Module] = None,
 ):
   """Computes dot-product attention given query, key, and value.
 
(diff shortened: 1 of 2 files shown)


chiamp:
cc: @cgarciae @JyChang012 @Xiaoming-Zhao

JyChang012:
👍

cgarciae:
Just a small comment, can we give a more descriptive name to the `mdl` argument?

chiamp:
> Just a small comment, can we give a more descriptive name to the `mdl` argument?

would `module` work instead

flax/linen/attention.py
```diff
@@ -76,6 +77,10 @@ def dot_product_attention_weights(
     dtype: the dtype of the computation (default: infer from inputs and params)
     precision: numerical precision of the computation see `jax.lax.Precision`
       for details.
+    mdl: if not None, the attention weights are sowed into the 'intermediates'
```
cgarciae, line 80: maybe explain that this is the `module` used to sow the attention weights (if given).

cgarciae (approved):

Xiaoming-Zhao:
Hi @chiamp , may I know whether we could change to the following alternative:
> An alternative option would be to return the weights as a tuple, as how [Pytorch](https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention.forward) and [Tensorflow](https://www.tensorflow.org/api_docs/python/tf/keras/layers/MultiHeadAttention) does it

I am asking because I just realized that intermediates in `sow` can only extracted at the outer-most call to `nn.Module`.

For example, if we have nested modules, which are all specified with `nn.compact` and `nn.MultiHeadDotProductAttention` is only called within some inner modules instead of the outer-most one. Then, it is not trivial to do any operation on the attentions within those inner modules in the forward pass.

However, if we can directly return a tuple, then any manipulation of the attention weights is straightforward.

I know that returning a tuple could be a break change. I am wondering whether there are some better solutions to this.

Or maybe I just misunderstand `sow` and the attention weights could actually be extracted within some inner modules in a nested case. I am happy to learn more about it. Thanks a lot in advance.

chiamp:
hi @Xiaoming-Zhao, would accessing the sowed intermediates via `self.variables` work for you:
```
class SubModel(nn.Module):
  attention_kwargs: dict

  @nn.compact
  def __call__(self, x, return_weights=False):
    x = nn.MultiHeadDotProductAttention(**self.attention_kwargs)(
      x, return_weights=return_weights
    )
    x = nn.MultiHeadDotProductAttention(**self.attention_kwargs)(x)
    x = nn.MultiHeadDotProductAttention(**self.attention_kwargs)(
      x, return_weights=return_weights
    )

    # access intermediates via self.variables
    if return_weights:
      attention_weights_0 = self.variables['intermediates']['MultiHeadDotProductAttention_0']['attention_weights']
      attention_weights_2 = self.variables['intermediates']['MultiHeadDotProductAttention_2']['attention_weights']

    return x


class Model(nn.Module):

  @nn.compact
  def __call__(self, x, return_weights=False):
    x = SubModel(
      dict(
        num_heads=8,
        qkv_features=16,
        kernel_init=initializers.ones,
        bias_init=initializers.zeros,
        deterministic=False,
      )
    )(x, return_weights=return_weights)

    # access intermediates via self.variables
    if return_weights:
      submodel_attention_weights_0 = self.variables['intermediates']['SubModel_0']['MultiHeadDotProductAttention_0']['attention_weights']
      submodel_attention_weights_2 = self.variables['intermediates']['SubModel_0']['MultiHeadDotProductAttention_2']['attention_weights']

    return x
```

PR #3537, opened 2023-12-06
add nnx conv support for int kernel size

### LINKED ISSUE #3534: NNX Conv documentation states that kernel size can be an integer for 1D convs.
https://github.com/google/flax/blob/829e004c6d5646f24bd048023a4407e3b4d5aa98/flax/experimental/nnx/nnx/nn/linear.py#L368

The NNX Conv documentation says that you can pass kernel size as an integer for 1D convolutions, however, the very first thing the constructor does is throw a TypeError if the kernel size is an integer.

chiamp:
Resolves #3534.

Added support for handling integer type `kernel_size` inputs for both NNX and Linen `Conv` layers.

Additional context: #1451

### DIFF
--- a/flax/experimental/nnx/nnx/nn/linear.py
+++ b/flax/experimental/nnx/nnx/nn/linear.py
@@ -371,8 +371,9 @@ class Conv(Module):
   Attributes:
     features: number of convolution filters.
     kernel_size: shape of the convolutional kernel. For 1D convolution,
-      the kernel size can be passed as an integer. For all other cases, it must
-      be a sequence of integers.
+      the kernel size can be passed as an integer, which will be interpreted
+      as a tuple of the single integer. For all other cases, it must be a
+      sequence of integers.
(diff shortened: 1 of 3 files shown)


flax/experimental/nnx/nnx/nn/linear.py
```diff
@@ -429,11 +429,7 @@ def __init__(
     rngs: rnglib.Rngs,
   ):
     if isinstance(kernel_size, int):
-      raise TypeError(
-        'Expected Conv kernel_size to be a'
-        ' tuple/list of integers (eg.: [3, 3]) but got'
-        f' {kernel_size}.'
-      )
+      kernel_size = (kernel_size,)
```
cgarciae, line 432: we need to update the signature to be:
```python
kernel_size: int | tp.Sequence[int],
```

chiamp: done!

cgarciae (approved):

PR #3547, opened 2023-12-11
implement Rng.__getattr__

cgarciae:
# What does this PR do?

Adds better error type to `Rngs._make_rng` depending on whether `__getitem__` or `__getattr__` is being called.

### DIFF
--- a/flax/experimental/nnx/nnx/rnglib.py
+++ b/flax/experimental/nnx/nnx/rnglib.py
@@ -130,14 +130,14 @@ class Rngs(tp.Mapping[str, tp.Callable[[], jax.Array]]):
     }
     self._trace_state = tracers.TraceState()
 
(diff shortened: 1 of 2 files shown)


flax/experimental/nnx/nnx/rnglib.py
```diff
@@ -130,24 +130,25 @@ def __init__(
     }
     self._trace_state = tracers.TraceState()
 
-  def _make_rng(self, name: str) -> jax.Array:
+  def _make_rng(self, name: str, error_type: Exception) -> jax.Array:
```
chiamp, line 133: I think presubmits failing because [this test](https://github.com/google/flax/blob/1721f4b31c07a80c73e1440175c2e8adfc16cb15/flax/experimental/nnx/tests/test_rngs.py#L40) is expecting a `ValueError` to be thrown instead of an `Exception`

PR #3548, opened 2023-12-12
added sow attention weights to NNX

chiamp:
Added sow attention weights feature to NNX, after adding the same functionality to Linen in #3529.


cgarciae (approved):
looks great! Just as a side comment because I know Linen uses the same name alread, `sow_weights` might have been a better name 😅

PR #3552, opened 2023-12-13
renamed attention layer to MultiHeadAttention

chiamp:
Renamed attention layer `nn.MultiHeadDotProductAttention` to `nn.MultiHeadAttention`. This is requested in [#1737](https://github.com/google/flax/discussions/1737#discussioncomment-7568103).

`MultiHeadDotProductAttention` can still be used, but a [`DeprecationWarning`](https://github.com/google/flax/pull/3552/files#diff-b7e399166c22f33b2518396c1cf200fd4c84b9ab3fca1ac754928b4f1794b3dfR157-R167) will be thrown.

Also added a [test](https://github.com/google/flax/pull/3552/files#diff-ffc73f4960d444a491d3b1ed16a242fb0359a42084c397bdc549dfacb4c5feccR395-R429) to check the alias and deprecation warning is working properly.

### DIFF
--- a/docs/api_reference/flax.linen/layers.rst
+++ b/docs/api_reference/flax.linen/layers.rst
@@ -88,6 +88,10 @@ Attention
 
 .. flax_module::
   :module: flax.linen
+  :class: MultiHeadAttention
+
+.. flax_module::
+  :module: flax.linen
   :class: MultiHeadDotProductAttention
 
 .. autofunction:: dot_product_attention_weights
@@ -148,6 +152,7 @@ Recurrent
   Sequential
   Dropout
   SelfAttention
+  MultiHeadAttention
   MultiHeadDotProductAttention
   RNNCellBase
   LSTMCell
--- a/flax/linen/__init__.py
+++ b/flax/linen/__init__.py
@@ -61,7 +61,7 @@ from .activation import (
     tanh as tanh,
 )
 from .attention import (
-    MultiHeadDotProductAttention as MultiHeadDotProductAttention,
+    MultiHeadAttention as MultiHeadAttention,
     SelfAttention as SelfAttention,
     combine_masks as combine_masks,
     dot_product_attention_weights as dot_product_attention_weights,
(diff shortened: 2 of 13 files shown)


flax/linen/__init__.py
```diff
@@ -151,3 +151,17 @@
...
+  if attr == 'MultiHeadDotProductAttention':
+    import warnings
+    warnings.warn(
+      'flax.linen.MultiHeadDotProductAttention is deprecated. Use flax.linen.MultiHeadAttention instead.',
+      DeprecationWarning,
+      stacklevel=2,
+    )
+    del warnings
+    return MultiHeadAttention
+  raise AttributeError(f"module '{__name__}' has no attribute '{attr}'")
```
cgarciae, line 167: To improve IDE usability its nice to use `typing.TYPE_CHECKING` to statically express what the API will look like on type checking, and the implement the actual behavior for runtime on the else branch e.g:

```python
import typing as _typing

if _typing.TYPE_CHECKING:
  from .attention import MultiHeadAttention as MultiHeadDotProductAttention
else:
  def __getattr__(attr: str):
    ...
```

docs/api_reference/flax.linen/layers.rst
```diff
@@ -148,7 +148,7 @@ Recurrent
   Sequential
   Dropout
   SelfAttention
-  MultiHeadDotProductAttention
+  MultiHeadAttention
```
cgarciae, line 151: maybe we should document both until we finally remove the `MultiHeadDotProductAttention` alias?

tests/linen/linen_test.py
```diff
@@ -442,20 +442,20 @@ def __call__(self, x):
...
-        'MultiHeadDotProductAttention_0/value/bias/u',
-        'MultiHeadDotProductAttention_0/value/kernel/u',
-        'MultiHeadDotProductAttention_0/key/bias/sigma',
-        'MultiHeadDotProductAttention_0/key/kernel/sigma',
-        'MultiHeadDotProductAttention_0/out/kernel/sigma',
-        'MultiHeadDotProductAttention_0/query/bias/sigma',
-        'MultiHeadDotProductAttention_0/query/kernel/sigma',
-        'MultiHeadDotProductAttention_0/value/bias/sigma',
-        'MultiHeadDotProductAttention_0/value/kernel/sigma',
+        'MultiHeadAttention_0/key/bias/u',
```
cgarciae, line 445: a bit worried that we might break a lot of checkpoints

chiamp: This could also break code that indexes into params, since the class name is changed:
```
>>> class A(nn.Module):
...   @nn.compact
...   def __call__(self, x):
...     x = nn.MultiHeadAttention(2)(x)
...     return x
... 
>>> x = jnp.ones((2, 3, 4))
>>> a = A()
>>> v = a.init(jax.random.key(0), x)
>>> v
{'params': {'MultiHeadAttention_0': ...}} # instead of MultiHeadDotProductAttention
```
To prevent breakage, we could force the name to still be `MultiHeadDotProductAttention`, but that may not be intuitive to users since it doesn't actually align with the class name. We could also keep `MultiHeadDotProductAttention` and just add `MultiHeadAttention` as an alias (rather than the other way around), but it could still be confusing for users using `MultiHeadAttention`.

chiamp:
Re-naming this layer could potentially break checkpoints and code that indexes into attention variables, since `MultiHeadAttention` will have a different name than `MultiHeadDotProductAttention`. Instead we will create a `MultiHeadAttention` alias that is a wrapper around `MultiHeadDotProductAttention`, and users are free to use either one. See #3572 for more details

PR #3554, opened 2023-12-14
re-factored features arg

chiamp:
Standardized feature args to `in_features` and `out_features`, since NNX uses both [`in/out_features`](https://github.com/google/flax/blob/main/flax/experimental/nnx/nnx/nn/linear.py#L306-L307) and [`features_in/out`](https://github.com/google/flax/blob/main/flax/experimental/nnx/nnx/nn/linear.py#L147-L148), while Linen only uses `in/out_features`.


cgarciae:
This is great, thanks!

cgarciae (approved):

PR #3555, opened 2023-12-14
re-factored nnx initializers

chiamp:
Re-naming `zeros` and `ones` initializers to `zeros_init` and `ones_init`, so that it's consistent with Linen initializers and users won't be confused between NNX `zeros/ones` and JAX `zeros/ones` since they have different function signatures.

See also #2002 and #2790.


cgarciae:
I believe we only used `zeros/ones_init` as a work around for historical reason. Maybe we should make this right in NNX and just use `zeros` and `ones` directly?

cgarciae (approved):
looks goods!

PR #3562, opened 2023-12-15
changed kwargs to actual key-word args

chiamp:
Changed kwargs to actual key-word args.

~~Changed NNX `Dtype` annotation to use `jax.typing.DTypeLike`.~~ (this has been updated with the [`flax.typing`](https://github.com/google/flax/pull/3624) PR)

### DIFF
--- a/flax/experimental/nnx/tests/nn/test_conv.py
+++ b/flax/experimental/nnx/tests/nn/test_conv.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from collections.abc import Sequence
+import typing as tp
 
 import jax
 from absl.testing import parameterized
@@ -22,6 +23,11 @@ from numpy.testing import assert_array_equal
 
 from flax import linen
 from flax.experimental import nnx
+from flax.typing import (
+  PaddingLike,
+  Dtype,
+  PrecisionLike
+)
 
 
 class TestConvLinenConsistency(parameterized.TestCase):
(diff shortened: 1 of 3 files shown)


flax/experimental/nnx/tests/nn/test_conv.py
```diff
@@ -25,36 +25,72 @@
...
+    feature_group_count=[3],
+    use_bias=[True, False],
+    dtype=[jnp.float32],
+    param_dtype=[jnp.float16],
+    precision=[Precision.HIGHEST],
   )
-  def test_nnx_linen_equivalence(self, **kwargs):
+  def test_nnx_linen_equivalence(
+    self,
+    strides,
```
cgarciae, line 41: can you also add the type hints?

chiamp: added!

flax/experimental/nnx/tests/nn/test_embed.py
```diff
@@ -12,31 +12,63 @@
...
+    input_dtype=[jnp.int16, jnp.int32],
+    dtype=[jnp.float32, jnp.float16],
+    param_dtype=[jnp.float32, jnp.float16],
   )
-  def test_nnx_linen_equivalence(self, input_dtype, **kwargs):
+  def test_nnx_linen_equivalence(
+    self,
+    input_dtype: tp.Optional[DTypeLike],
+    dtype: tp.Optional[Dtype],
+    param_dtype: Dtype,
```
chiamp, line 54: Note: in the `test_nnx_linen_equivalence` test in `test_embed.py`, the `input_dtype` arg is of `DType` typing (with an upper case T, that's derived from Jax), whereas the `dtype` and `param_dtype` arg is of `Dtype` typing (with a lower case t, that's derived from NNX). You can see their aliases at the top of the file [here](https://github.com/google/flax/pull/3562/files#diff-ba0465aa3c924f608942a9b6a40cb876fdcc50c4084d8080aa4569c783c4a413R26-R27).

Should we standardize the typing annotation (both in spelling as well as definition)?

cgarciae: standarization would be nice, can get reuse something from JAX? It feels like they should expose this so we can reuse it if possible.

cgarciae:
should we expose / share this file in / as `flax/typing.py`?

chiamp:
> should we expose / share this file in / as `flax/typing.py`?

done!

cgarciae (approved):

PR #3581, opened 2023-12-21
[nnx] add submodule iterator

cgarciae:
# What does this PR do?

Adds method to iterator over all unique submodules.

### DIFF
--- a/flax/experimental/nnx/nnx/graph_utils.py
+++ b/flax/experimental/nnx/nnx/graph_utils.py
@@ -458,6 +458,7 @@ def _graph_unflatten(
   if graphdef.index in index_to_node:
     raise RuntimeError(f'GraphDef index {graphdef.index} already used.')
 
+  state = state.copy()
   node_impl = get_node_impl(graphdef.type)
 
   def _get_children():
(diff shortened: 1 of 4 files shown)


flax/experimental/nnx/nnx/module.py
```diff
@@ -482,29 +482,10 @@ def sow(
       reduced_value = reduce_fn(init_fn(), value)
       setattr(self, name, variable_type(reduced_value))
 
-  def for_each(
```
superbobry, line 485: Should this be called `submodules` and not `module`? Or better yet `iter_submodules`?

I would also recommend adding a few tests.

superbobry (approved):
I would recommend adding a few tests for the new API.

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -720,6 +721,27 @@ def clone(node: Node) -> Node:
...
 
 
+def iter_nodes(node: tp.Any) -> tp.Iterator[tuple[Path, tp.Any]]:
+  visited: set[int] = set()
+  path_parts: PathParts = ()
+  yield from _iter_nodes(node, visited, path_parts)
+
+
+def _iter_nodes(
+  node: tp.Any, visited: set[int], path_parts: PathParts
```
chiamp, line 731: can the `node` arg have a more specific type annotation than `tp.Any`?

chiamp (approved):

flax/experimental/nnx/nnx/graph_utils.py
cgarciae: problem is that a node could be of any type that is registered.

flax/experimental/nnx/nnx/module.py
cgarciae: Called it [modules()](https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.modules) as it would be familiar to Pytorch users. Wondering if we should try to follow their conventions when possible?

cgarciae:
@superbobry added test.

PR #3582, opened 2023-12-21
[nnx] flaglib add get overloads

cgarciae:
# What does this PR do?

Adds typing overloads for `flagslib.get`.

### DIFF
--- a/flax/experimental/nnx/nnx/flaglib.py
+++ b/flax/experimental/nnx/nnx/flaglib.py
@@ -11,12 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from __future__ import annotations
 
 import dataclasses
 import threading
 import typing as tp
 from contextlib import contextmanager
 from types import MappingProxyType
+
+A = tp.TypeVar('A')
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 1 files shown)


flax/experimental/nnx/nnx/flaglib.py
```diff
@@ -60,7 +63,15 @@ def __call__(self, **kwargs: tp.Any):
     finally:
       FLAGS_CONTEXT.flags_stack.pop()
 
-  def get(self, name: str, default: tp.Any = None) -> tp.Optional[tp.Any]:
+  @tp.overload
+  def get(self, name: str) -> tp.Any:
+    ...
+
+  @tp.overload
+  def get(self, name: str, default: A) -> A:
```
superbobry, line 71: I think the right return type is `A | Any`, because the value stored in the context is not guaranteed to be of type `A`.

superbobry (approved):

flax/experimental/nnx/nnx/flaglib.py
```diff
@@ -60,7 +63,15 @@ def __call__(self, **kwargs: tp.Any):
...
+    ...
+
+  def get(self, name: str, default: A = None) -> A | None:
```
chiamp, line 74: do you need a separate `overload` if the method signatures are the same?

chiamp (approved):

flax/experimental/nnx/nnx/flaglib.py
superbobry: The last non-@overloaded definition is not used by type checkers, so this LGTM.

PR #3590, opened 2023-12-26
improve embed

jianyizh:
# What does this PR do?

This PR improve the performance of flax.linen.Embed when num_embeddings == 1. I find that in huggingface transformers, some models use embedding even for num_embedding == 1 (e.g. [token_type_embeddings here](https://github.com/huggingface/transformers/blob/main/src/transformers/models/roberta/modeling_flax_roberta.py#L170)). The atomic ops in scatter from it's backward with all indices equals to 0 will cause performance issues. For example, one bf16 train step of this roberta model with bs=128 costs A100 210ms. After this modification, it improves to 156ms.

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -422,7 +422,7 @@ class _Conv(Module):
       kernel_size = tuple(self.kernel_size)
 
     def maybe_broadcast(
-      x: Optional[Union[int, Sequence[int]]]
+      x: Optional[Union[int, Sequence[int]]],
     ) -> Tuple[int, ...]:
       if x is None:
         # backward compatibility with using None as sentinel for
@@ -941,7 +941,9 @@ class Embed(Module):
   matrix with shape ``(num_embeddings, features)``. When calling this layer,
   the input values will be used to 0-index into the ``embedding`` matrix.
   Indexing on a value greater than or equal to ``num_embeddings`` will result
-  in ``nan`` values.
+  in ``nan`` values. When ``num_embeddings`` equals to 1, it will
+  broadcast the ``embedding`` matrix to input shape with ``features``
+  dimension appended.
 
   Example usage::
 
@@ -1009,6 +1011,13 @@ class Embed(Module):
     (embedding,) = promote_dtype(
       self.embedding, dtype=self.dtype, inexact=False
     )
+    if self.num_embeddings == 1:
+      return jnp.where(
+        jnp.broadcast_to(inputs[..., None], inputs.shape + (self.features,))
+        == 0,
+        embedding,
+        jnp.nan,
+      )
     return jnp.take(embedding, inputs, axis=0)
 
   def attend(self, query: Array) -> Array:
--- a/flax/experimental/nnx/nnx/nn/linear.py
+++ b/flax/experimental/nnx/nnx/nn/linear.py
@@ -489,7 +489,7 @@ class Conv(Module):
     kernel_size = self.kernel_size
 
     def maybe_broadcast(
-      x: tp.Optional[tp.Union[int, tp.Sequence[int]]]
+      x: tp.Optional[tp.Union[int, tp.Sequence[int]]],
     ) -> tp.Tuple[int, ...]:
       if x is None:
         # backward compatibility with using None as sentinel for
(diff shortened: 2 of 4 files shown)


jianyizh:
Hi @chiamp, could you help review this PR please?

cgarciae:
Hey @jianyizh , sorry for the delay. Can you please add this implementation detail to the docstrings?

jianyizh:
> Hey @jianyizh , sorry for the delay. Can you please add this implementation detail to the docstrings?

I have updated the docstrings, thank you

cgarciae:
pre-commit hooks are failing on CI. To fix this run:

```bash
pip install pre-commit
pre-commit install # makes sure future commits pass
pre-commit run --all-files # fixes current issue
```

jianyizh:
> pre-commit hooks are failing on CI. To fix this run:
> 
> ```shell
> pip install pre-commit
> pre-commit install # makes sure future commits pass
> pre-commit run --all-files # fixes current issue
> ```

Thank you. pre-commit hooks can pass now.

flax/linen/linear.py
```diff
@@ -1009,6 +1011,8 @@ def __call__(self, inputs: Array) -> Array:
     (embedding,) = promote_dtype(
       self.embedding, dtype=self.dtype, inexact=False
     )
+    if (self.num_embeddings == 1):
+      return jnp.broadcast_to(embedding, inputs.shape + (self.features,))
```
chiamp, line 1015: If we pass in index values in `inputs` that are greater or equal to `self.num_embeddings`, the current expected behavior is that it returns a `nan` embedding. 

With this change, we would still be able to get the embedding matrix (e.g. if we pass in `inputs` that contains index value 2), since the `if` condition only uses `inputs.shape` and not `inputs`. 

This could introduce silent errors if the user isn't aware of this special case with `self.num_embeddings==1`, since no `nan` will be returned if they try to index index past `self.num_embeddings`.

We should either add a check in this `if` block to check for over-indexing and if so, return `nan` like the previous behavior, or we can keep this change as in and describe this special case in the docstring.

What do you think @cgarciae?

chiamp (changes requested):

flax/linen/linear.py
jianyizh: If we prefer `nan` in this case, It can be
```
    if (self.num_embeddings == 1):
      return jax.lax.cond((inputs==0).all(),
                          lambda x: jnp.broadcast_to(embedding,x.shape+(self.features,)),
                          lambda x: jnp.take(embedding, x, axis=0),
                          inputs)
```

chiamp: Sounds good to me! Can you also write a test in [`tests/linen/linen_linear_test.py`](https://github.com/google/flax/blob/main/tests/linen/linen_linear_test.py#L1046) and [`flax/experimental/nnx/tests/nn/test_linear.py`](https://github.com/google/flax/blob/main/flax/experimental/nnx/tests/nn/test_linear.py#L69) to confirm the `nan` behavior for both `self.num_embeddings == 1` and `self.num_embeddings > 1`

jianyizh: Thank you. I have updated the code for the `nan` case and added two tests.

chiamp: Great, thanks! Could you also run `ruff format flax/experimental/nnx/nnx/nn/linear.py flax/experimental/nnx/tests/nn/test_embed.py flax/linen/linear.py tests/linen/linen_linear_test.py`

jianyizh: Thanks! I have updated these files.

chiamp (approved):

flax/linen/linear.py
cgarciae: @jianyizh I think you should use `jnp.where` instead of `cond` for performance, also because here you are conditioning on the entire batch being within bounds.

jianyizh: I have updated to jnp.where. Actually, since the inputs are constants in the models when `num_embedding == 1`, there is no performance difference. XLA do some optimizations like constant folding, both `where` and `cond` are canceled. The two HLO graphs are both simplified to reduce_sum on backward.

cgarciae (approved):

cgarciae:
You need to squash the commits. Run

```
git reset --soft main
```
then create a new commit, and finally force push:
```
git push -f
```

jianyizh:
> You need to squash the commits. Run
> 
> ```
> git reset --soft main
> ```
> 
> then create a new commit, and finally force push:
> 
> ```
> git push -f
> ```

Thank you. I have squashed the commits.

PR #3606, opened 2024-01-08
[nnx] add more decode flags for MultiHeadAttention

cgarciae:
# What does this PR do?

Adds more `decode` flags for `MultiHeadAttention`:
* Current `decode` field is made `Optional` with a default of `None`.
* Adds a `decode` argument to `__call__` which can be used to override all other flags.
* `__call__` now check if the `decode` flag is set in  `nnx.flags` as a fallback in case neither the field or the `__call__` arguments are passed.


cgarciae:
# continued in #3607

PR #3607, opened 2024-01-08
[nnx] improve runtime flags

cgarciae:
# What does this PR do?

* Improves error messages for runtime flags.
* Makes MultiHeadAttention decode runtime flag behave like runtime flags for other layers.

### DIFF
--- a/flax/experimental/nnx/tests/nn/test_attention.py
+++ b/flax/experimental/nnx/tests/nn/test_attention.py
@@ -26,7 +26,7 @@ class TestMultiHeadAttention:
       out_features=6,
       rngs=nnx.Rngs(0),
     )
-    y = module(jnp.ones((1, 7, 3)))
+    y = module(jnp.ones((1, 7, 3)), decode=False)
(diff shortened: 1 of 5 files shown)


flax/experimental/nnx/tests/nn/test_attention.py
```diff
@@ -26,7 +26,7 @@ def test_basic(self):
       out_features=6,
       rngs=nnx.Rngs(0),
     )
-    y = module(jnp.ones((1, 7, 3)))
+    y = module(jnp.ones((1, 7, 3)), decode=False)
```
chiamp, line 29: Might be worth adding another check/test to test instantiating `MultiHeadAttention` with `decode=False` as an `init` arg. It's already tested in the [autoregressive test](https://github.com/google/flax/pull/3607/files#diff-1f2d4a5e6d0e93c2db09be241c0b131dde6ab0d2d1c807f314af65eebf06a06aR90), but the main point of that test was to test `jax.experimental.enable_x64`

chiamp (approved):

PR #3621, opened 2024-01-12
Add compact_name_scope decorator

cgarciae:
# What does this PR do?

Adds the `nn.compact_name_scope ` decorator that enables methods to act as compact name scopes as with regular Haiku methods, this makes porting Haiku code easier.

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -538,6 +538,84 @@ def nowrap(fun: _CallableT) -> _CallableT:
   return fun
 
 
+def compact_name_scope(fun: _CallableT) -> _CallableT:
+  """Creates compact submodules from a method.
+
+  This is a decorator that allows you to define compact submodules from a
+  method. It's intention is to make it easier to port code Haiku code to Flax
+  by providing the same functionality.
+
+  Example::
+
+    >>> import flax.linen as nn
+    >>> import jax
+    >>> import jax.numpy as jnp
+    >>> from flax.core import pretty_repr
+    ...
+    >>> class Foo(nn.Module):
+    ...   @nn.compact_name_scope
+    ...   def up(self, x):
+    ...     return nn.Dense(3)(x)
+    ...
+    ...   @nn.compact_name_scope
+    ...   def down(self, x):
+    ...     return nn.Dense(3)(x)
+    ...
+    ...   def __call__(self, x):
+    ...     return self.up(x) + self.down(x)
+    ...
+    >>> module = Foo()
+    >>> variables = module.init(jax.random.PRNGKey(0), jnp.ones((1, 2)))
+    >>> params = variables['params']
+    >>> print(pretty_repr(jax.tree_map(jnp.shape, params)))
+    {
+        down: {
+            Dense_0: {
+                bias: (3,),
+                kernel: (2, 3),
+            },
+        },
+        up: {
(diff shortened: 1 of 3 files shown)


flax/linen/module.py
```diff
@@ -538,6 +538,80 @@ def nowrap(fun: _CallableT) -> _CallableT:
   return fun
 
 
+def nontransparent(fun: _CallableT) -> _CallableT:
```
marcvanzee, line 541: Thinking about this a bit more high level: What happens actually if you do `module.init(..., method=Foo.up)`? Like, is it an actual compact method? If not, then this should probably be explained clearly. If so, then this beg the question: why don't we just extend `nn.compact` to allow multiple methods? I think there were good reasons to keep one method and it probably isn't wise to change this behavior, but the difference between the proposed decorator and nn.compact isn't really clear to me.

marcvanzee (changes requested):

flax/linen/module.py
cgarciae: `module.init(..., method=Foo.up)` should behave like a compact method BUT the name `up` would be part of the param structure which is the point here.

marcvanzee: Alright, thanks for clarifying. Maybe the name could then be something like `@nn.compact_name_scope`?

Also, would it still work if someone would only create methods with this decorator, or do we always need one method annotated with `@nn.compact`?

cgarciae: I like that name better!

> Also, would it still work if someone would only create methods with this decorator, or do we always need one method annotated with @nn.compact?

This implementation does NOT require you to have a main `compact` method 🙂

cgarciae: I'll update the PR soon with the new name.

flax/linen/module.py
```diff
@@ -538,6 +538,80 @@ def nowrap(fun: _CallableT) -> _CallableT:
...
+  by providing the same functionality.
+
+  Example::
+
+    >>> import flax.linen as nn
+    >>> import jax
+    >>> import jax.numpy as jnp
+    >>> from flax.core import pretty_repr
+    ...
+    >>> class Foo(nn.Module):
```
marcvanzee, line 555: Now that this is working so well (meaning we can call the method directly), this example is actually a bit confusing. It raises the question whether it perhaps is necessary to always have a `compact` method.

A simpler example could be:

```
class Foo(nn.Module):
  @nn.compact_name_scope
  def up(self, x):
    return nn.Dense(4)(x)

  @nn.compact_name_scope
  def down(self, x):
    return nn.Dense(3)(x)
```

This also makes it clear that we don't need any `compact` methods, and you could change the `init` call by adding `method=Foo.up`.

Then I think it would also be good to explain that `compact` and `compact_name_scope` can be mixed within a method, but there can only be one `compact` method.

marcvanzee (approved):
Really useful addition! Just a nit on the docstring but other than that looks great!

flax/linen/module.py
cgarciae: I see your point. I'll update the example a bit to make this distinction clearer, since the share the word `compact` in their name users might make that association based on this example.

marcvanzee: Awesome! Already approved it so feel free to submit :)

PR #3624, opened 2024-01-16
added flax.typing

chiamp:
Added [`flax/typing.py`](https://github.com/google/flax/pull/3624/files#diff-f06fb18c99233972846e0a2db5c4ee574ac4b6b632e7ee1fe2bdc57d8901277dR1-R132) file to unify typing annotations.

Few things to note:
* [`Filter`](https://github.com/google/flax/pull/3624/files#diff-f06fb18c99233972846e0a2db5c4ee574ac4b6b632e7ee1fe2bdc57d8901277dR59) is defined [differently in NNX](https://github.com/google/flax/blob/7b0f6537f48ec016ad021b963e53564836c2c041/flax/experimental/nnx/nnx/filterlib.py#L27), so I did not unify NNX's definition under the `flax.typing` definition
* [`Axis`](https://github.com/google/flax/pull/3624/files#diff-f06fb18c99233972846e0a2db5c4ee574ac4b6b632e7ee1fe2bdc57d8901277dR113) is defined [differently in NNX](https://github.com/google/flax/blob/7b0f6537f48ec016ad021b963e53564836c2c041/flax/experimental/nnx/nnx/nn/linear.py#L55), so I did not unify NNX's definition under the `flax.typing` definition
* In addition to regular JAX initializers, `Partitioned` objects can also be used as [`kernel_init` and `bias_init` arguments](https://github.com/google/flax/blob/7b0f6537f48ec016ad021b963e53564836c2c041/examples/lm1b/models.py#L189-L192). Currently to accommodate this, the `Initializer` type annotation is [`Union[jax.nn.initializers.Initializer, Callable[..., Any]]`](https://github.com/google/flax/pull/3624/files#diff-f06fb18c99233972846e0a2db5c4ee574ac4b6b632e7ee1fe2bdc57d8901277dR54). The more correct annotation would be `Union[jax.nn.initializers.Initializer, Callable[..., Partitioned[Any]]]`, but then the linter will complain about code that tries to call `Array` methods or access `Array` attributes like `.reshape` and `.shape` on the initialized `kernel` or `bias` since `Partitioned[Any]` does not have any of these methods or attributes. The linter will also complain about functions that take in the initialized `kernel` or `bias` as an input argument, because they expect an `Array` and not `Partitioned[Any]`.

### DIFF
--- a/flax/core/__init__.py
+++ b/flax/core/__init__.py
@@ -38,7 +38,6 @@ from .meta import (
     unbox as unbox,
 )
 from .scope import (
-    Array as Array,
     DenyList as DenyList,
     Scope as Scope,
     apply as apply,
@@ -50,4 +49,8 @@ from .tracers import (
     check_trace_level as check_trace_level,
     current_trace as current_trace,
     trace_level as trace_level,
+)
+
+from flax.typing import (
+    Array as Array,
 )
--- a/flax/core/lift.py
+++ b/flax/core/lift.py
@@ -15,13 +15,11 @@
 """Jax transform lifting."""
 
 import collections
-import dataclasses
 import functools
 from typing import (
     Any,
     Callable,
     Dict,
-    Generic,
     Iterable,
     List,
     Mapping,
@@ -34,6 +32,12 @@ from typing import (
 import warnings
 
 from flax import traceback_util
+from flax.typing import (
+  In,
+  Out,
+  InOutAxis,
+  InOutScanAxis,
+)
 import jax
 from jax import random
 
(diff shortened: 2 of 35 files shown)


flax/core/__init__.py
```diff
@@ -51,3 +49,8 @@
     current_trace as current_trace,
     trace_level as trace_level,
 )
+
+from ..typing import (
```
cgarciae, line 53: use absolute paths

cgarciae (approved):

Issue #3617, opened 2024-01-12
[struct.dataclass] Consider adding optional `kw_only` arguments


epignatelli:
I often run into the following issue:

```python
from flax import struct

class Foo(struct.PyTreeNode):
  bar: int = struct.field(pytree_node=False, default=1)

class Baz(Foo):
  qux: str
```

Since `qux` does not have a default value, I get:
```
Fields without default values cannot appear after fields with default values
```

Can we consider adding a simple wrapper to `dataclasses.dataclass(kw_only=True)`?

It should be easy for the `struct.dataclass`, we can maybe have another object for inheritance, like `PyTreeNodeKwOnly`?

chiamp:
#3645 allows you to pass in general kwargs to `struct.dataclass`, so you should be able to use `kw_only=True` now.

#3651 adds a test to test this `kw_only` functionality. Please note that `kw_only` is only available for python 3.10 or higher. Currently Flax supports, python 3.9 or higher, and so this test is commented out until Flax upgrades its minimum version requirement.

epignatelli:
Thanks @chiamp, would #3645 pass the test case in the issue above?

chiamp:
Yes, you could do something like this:
```
from flax import struct
import functools

@functools.partial(struct.dataclass, kw_only=True)
class Foo():
  bar: int = struct.field(pytree_node=False, default=1)

@struct.dataclass
class Baz(Foo):
  qux: str

b = Baz('qux_value', bar=123)
print(b)
# Baz(bar=123, qux='qux_value')
```

epignatelli:
Thanks, @chiamp, that is a different test case, though.
The original one still fails.

In the second test case, Pylance also does not recognise that `Baz` has a `bar` attribute: `No parameter named "bar"`.

Is there any plan to fully support type checking with this and propagate the change to the `PyTreeNode` class?

chiamp:
Sorry for the delay. After #3785, `PyTreeNode` can take additional kwargs just like `struct.dataclass`.

epignatelli:
That's great, thanks @chiamp !

PR #3651, opened 2024-01-24
Add `kw_only` struct.dataclass test

chiamp:
In #3617, A user requested to allow a `kw_only` argument to be passed into the `flax.struct.dataclass` decorator.
#3645 allows general `kwargs` to be passed into the `flax.struct.dataclass` decorator, originally with the intention of allowing the `slots` argument to be passed, but this will also solve #3617 as well.

This PR adds a test to make sure `kw_only` is working as intended. It is currently commented out because `kw_only` is only supported in python 3.10 or greater, but Flax currently supports Python 3.9 or greater. This test was tested locally and passed on python 3.10. This test will be commented back in once Flax upgrades its minimum python version to 3.10.

~This PR also [enforces `frozen=True`](https://github.com/google/flax/pull/3651/files#diff-99b09b7db6644aaccda9ec75198369c2e578cbed72b921c440ea972417ce3089R107-R112) for the `struct.dataclass` decorator, otherwise a `ValueError` will be raised.~
^ Removing frozen enforcement, because a lot of internal tests rely on mutable `struct.dataclass`.

### DIFF
--- a/tests/struct_test.py
+++ b/tests/struct_test.py
@@ -92,6 +92,25 @@ class StructTest(absltest.TestCase):
     class A(struct.PyTreeNode):
       a: int
 
+  # TODO(marcuschiam): Uncomment when Flax upgrades to Python 3.10.
+  # def test_kw_only(self):
+  #   @struct.dataclass
+  #   class A:
+  #     a: int = 1
+
+  #   with self.assertRaisesRegex(TypeError, "non-default argument 'b' follows default argument"):
+  #     @struct.dataclass
+  #     class B(A):
+  #       b: int
+
+  #   @functools.partial(struct.dataclass, kw_only=True)
+  #   class B(A):
+  #     b: int
+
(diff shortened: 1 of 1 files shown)


flax/struct.py
```diff
@@ -104,7 +104,9 @@ class method that provides the smart constructor.
   if '_flax_dataclass' in clz.__dict__:
     return clz
 
-  data_clz = dataclasses.dataclass(frozen=True)(clz)  # type: ignore
+  if 'frozen' not in kwargs.keys():
```
cgarciae, line 107: I have the feeling that we should raise an error if users passes `frozen = False` instead to keep `struct.dataclass` instances immutable.

chiamp: ok, I enforced `frozen=True` otherwise a `ValueError` is raised, and I added a test

cgarciae (approved):

PR #3652, opened 2024-01-24
added InstanceNorm

### LINKED ISSUE #3622: Instance Normalization
Could you implement Instance Normalization, please?

chiamp:
Added InstanceNorm. See the [API reference](https://flax--3652.org.readthedocs.build/en/3652/api_reference/flax.linen/layers.html#flax.linen.InstanceNorm) for more details.

Resolves #3622.

Alternative implementations:
* have `InstanceNorm` subclass `LayerNorm` and derive the `feature_axes` and `reduction_axes` from the `channel_axes` argument in `InstanceNorm`, and then pass them into `LayerNorm`
* create a base normalization layer, and have both `InstanceNorm` and `LayerNorm` subclass it

### DIFF
--- a/flax/linen/normalization.py
+++ b/flax/linen/normalization.py
@@ -730,6 +730,127 @@ class GroupNorm(Module):
     )
 
 
+class InstanceNorm(Module):
+  """Instance normalization (https://arxiv.org/abs/1607.08022v3).
+
+  InstanceNorm normalizes the activations of the layer for each channel (rather
+  than across all channels like Layer Normalization), and for each given example
+  in a batch independently (rather than across an entire batch like Batch
+  Normalization). i.e. applies a transformation that maintains the mean activation
+  within each channel within each example close to 0 and the activation standard
+  deviation close to 1.
+
+  NOTE: This normalization operation is identical to LayerNorm and GroupNorm; the
+  difference is simply which axes are reduced and the shape of the feature axes
+  (i.e. the shape of the learnable scale and bias parameters).
+
+  Example usage::
+
+    >>> import flax.linen as nn
+    >>> import jax
+    >>> import numpy as np
+
+    >>> # dimensions: (batch, height, width, channel)
+    >>> x = jax.random.normal(jax.random.key(0), (2, 3, 4, 5))
+    >>> layer = nn.InstanceNorm()
+    >>> variables = layer.init(jax.random.key(1), x)
+    >>> variables
+    {'params': {'scale': Array([1., 1., 1., 1., 1.], dtype=float32), 'bias': Array([0., 0., 0., 0., 0.], dtype=float32)}}
+    >>> y = layer.apply(variables, x)
+
+    >>> # having a channel_axis of -1 in InstanceNorm is identical to reducing all non-batch,
+    >>> # non-channel axes and using the channel_axes as the feature_axes in LayerNorm
+    >>> y2 = nn.LayerNorm(reduction_axes=[1, 2], feature_axes=-1).apply(variables, x)
+    >>> np.testing.assert_allclose(y, y2, atol=1e-7)
+    >>> y3 = nn.GroupNorm(num_groups=x.shape[-1]).apply(variables, x)
+    >>> np.testing.assert_allclose(y, y3, atol=1e-7)
+
+  Attributes:
+    epsilon: A small float added to variance to avoid dividing by zero.
+    dtype: the dtype of the result (default: infer from input and params).
+    param_dtype: the dtype passed to parameter initializers (default: float32).
+    use_bias:  If True, bias (beta) is added.
+    use_scale: If True, multiply by scale (gamma). When the next layer is linear
+      (also e.g. nn.relu), this can be disabled since the scaling will be done
+      by the next layer.
+    bias_init: Initializer for bias, by default, zero.
+    scale_init: Initializer for scale, by default, one.
+    channel_axes: Axes for channel. This is considered the feature axes for the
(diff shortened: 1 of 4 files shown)


flax/linen/normalization.py
```diff
@@ -730,6 +730,122 @@ def __call__(self, x, mask=None):
...
+  use_scale: bool = True
+  bias_init: Callable[[PRNGKey, Shape, Dtype], Array] = initializers.zeros
+  scale_init: Callable[[PRNGKey, Shape, Dtype], Array] = initializers.ones
+  channel_axes: Axes = -1
+  axis_name: Optional[str] = None
+  axis_index_groups: Any = None
+  use_fast_variance: bool = True
+
+  @compact
+  def __call__(self, x, mask=None):
```
cgarciae, line 808:
```suggestion
  def __call__(self, x, mask: jax.Array | None = None):
```

flax/linen/normalization.py
```diff
@@ -730,6 +730,122 @@ def __call__(self, x, mask=None):
...
+    Args:
+      x: the inputs
+      mask: Binary array of shape broadcastable to ``inputs`` tensor, indicating
+        the positions for which the mean and variance should be computed.
+
+    Returns:
+      Normalized inputs (the same shape as inputs).
+    """
+    channel_axes = _canonicalize_axes(x.ndim, self.channel_axes)
+    reduction_axes = [i for i in range(1, x.ndim) if i not in channel_axes]
```
cgarciae, line 820: how is this avoiding reducing the batch dimensions? I assume the method wants to avoid cross-instance reductions.

flax/linen/normalization.py
```diff
@@ -730,6 +730,122 @@ def __call__(self, x, mask=None):
     )
 
 
+class InstanceNorm(Module):
```
cgarciae, line 733: Keras implements [InstanceNormalization](https://www.tensorflow.org/addons/api_docs/python/tfa/layers/InstanceNormalization) as a special case of `GroupNorm`. Kinda makes sense, wondering if we could do the same?

flax/linen/normalization.py
chiamp: The assumption is the leading dimension is the batch dimension. With a 4D-input and `channel_axes=-1`, only the two middle dimensions will be reduced. If the input was a `(batch, height, width, channel)` image, then this would correspond to reducing only the `height` and `width` dimension. Alternatively, we could explicitly add a `batch_axis` argument to denote which dimension is the batch axis.

flax/linen/normalization.py
chiamp: should this kwarg enforcement and type annotation be added to all normalization call methods? If so, we'd also need to check if this breaks any internal tests

flax/linen/normalization.py
chiamp: I will look into this

flax/linen/normalization.py
cgarciae: I see, so you are saying `_canonicalize_axes`'s return sequence doesn't include `0`?

flax/linen/normalization.py
cgarciae: It would be a nice, we should strive to have a clean API :)

flax/linen/normalization.py
chiamp: Yes, `_canonicalize_axes` returns only the axes you input. So if `self.channel_axes=-1`, it will only return `(3,)`, in the case of a 4D input.

flax/linen/normalization.py
chiamp: I will add this in a separate PR, where I will enforce kwargs in all normalization call methods, and then test internally to see if there are any breakages

flax/linen/normalization.py
```diff
@@ -730,6 +730,127 @@ def __call__(self, x, mask=None):
...
+  """Instance normalization (https://arxiv.org/abs/1607.08022v3).
+
+  InstanceNorm normalizes the activations of the layer for each channel (rather
+  than across all channels like Layer Normalization), and for each given example
+  in a batch independently (rather than across an entire batch like Batch
+  Normalization). i.e. applies a transformation that maintains the mean activation
+  within each channel within each example close to 0 and the activation standard
+  deviation close to 1.
+
+  NOTE: This normalization operation is identical to LayerNorm; the difference is
```
cgarciae, line 743: NIT: maybe also mention GroupNorm

flax/linen/normalization.py
```diff
@@ -730,6 +730,127 @@ def __call__(self, x, mask=None):
...
+    >>> variables
+    {'params': {'scale': Array([1., 1., 1., 1., 1.], dtype=float32), 'bias': Array([0., 0., 0., 0., 0.], dtype=float32)}}
+    >>> y = layer.apply(variables, x)
+
+    >>> # having a channel_axis of -1 in InstanceNorm is identical to reducing all non-batch,
+    >>> # non-channel axes and using the channel_axes as the feature_axes in LayerNorm
+    >>> y2 = nn.LayerNorm(reduction_axes=[1, 2], feature_axes=-1).apply(variables, x)
+    >>> np.testing.assert_allclose(y, y2, atol=1e-7)
+    >>> y3 = nn.GroupNorm(num_groups=x.shape[-1], use_bias=False, use_scale=False).apply(variables, x)
+    >>> np.testing.assert_allclose(y, y3, atol=1e-7)
```
cgarciae, line 766: These comparisons are great!

cgarciae (approved):

PR #3654, opened 2024-01-24
add Module.module_paths

cgarciae:
# What does this PR do?

Adds the `Module.module_paths` method that returns a mapping between paths and copies of unbound Module instances.


flax/linen/module.py
```diff
@@ -2585,6 +2585,68 @@ def tabulate(
...
+        be treated as mutable: ``bool``: all/no collections are mutable.
+        ``str``: The name of a single mutable collection. ``list``: A list of
+        names of mutable collections. By default, all collections except
+        'intermediates' are mutable.
+      **kwargs: keyword arguments to pass to the forward computation.
+
+    Returns:
+      A dict`ionary mapping module paths to module instances.
+    """
+    from flax.linen import summary
```
chiamp, line 2638: Should this be moved to a global import? It's also imported locally within the [`tabulate`](https://github.com/google/flax/blob/a638b16f09e8c4e8bfbfe9e1076a666181c3257e/flax/linen/module.py#L2572) Module method.

chiamp (approved):

flax/linen/module.py
cgarciae: If you import it at the top level it creates a dependency cycle, so we delay the inside the methods. This is cached by Python and it only runs once.

PR #3663, opened 2024-01-27
enforce mask kwarg in norm layers

chiamp:
enforce mask kwarg in norm layers


cgarciae (approved):

cgarciae:
This PR is technically a breaking change, we could consider bumping version to `0.9.0` on next release.

PR #3669, opened 2024-01-30
add default make_rng

### LINKED ISSUE #3506: Full Module RNG Scope
Converting this to an issue since I'm repeatedly stumbling into the current non-streamlined implementation for module random-key management (see discussion below; it doesn't present a solution yet, but it writes down some ideas) .

The biggest issue that the current design has is that it prevents me from writing composing generic modules without having to awkardly pass a scope of RNGs (and likely make a mistake). 

This snippet shows a example use-case for sampling and applying random functions,
```python
from typing import Generic, TypeVar

import jax
import jax.numpy as jnp

from flax import struct
from flax import linen as nn


M = TypeVar('M', bound=nn.Module)

@struct.dataclass
class Composite(Generic[M]):
    module: M

    def do(self, key: jax.Array, ins: jax.Array):
        # Sample random module parameters, and apply the function.
        
        key_init, keys_sample = jax.random.split(key)
        
        # Problem: Throws an error if an arbitrary key in `M` is not covered.
        params = self.module.init(key, ins)

        # Problem: I do not know the RNG scope of self.module
        outs = self.module.apply(params, ins)
        
        return outs


c = Composite(nn.Dense(5))
c.do(jax.random.key(0), jnp.ones(5))
```

### Discussed in https://github.com/google/flax/discussions/3375

<sup>Originally posted by **joeryjoery** September 28, 2023</sup>
Hi, if I implement a Module that makes use of `make_rng`, e.g., Dropout, I now have to manually specify the `rng` when I use the Module's call function. Though I completely agree that the RNG streams should be split into distinct paths like it is now, it is quite annoying that I have to bookkeep this at every call to `Module.apply`.

```python
import jax
import flax.linen as nn


rng = jax.random.PRNGKey(0)


class A(nn.Module):

    def __call__(self):
        x = self.make_rng('x')
        y = self.make_rng('y')

        return jax.random.uniform(x, ()) + jax.random.uniform(y, ())

myA = A()

# Currently
keys = jax.random.split(rng, 3)
scope = {'params', 'x', 'y'}  # I don't want to do this manually

# Desired functionality (or something like this)
# scope = myA.get_rng_scope()

rngs = dict(zip(scope, keys))

v = myA.init(rngs)
_ = myA.apply(v, rngs=rngs)
```
Is there already something like this implemented? Or could this be done in the future?

I'd also argue that flax should allow me to just pass in one global RNG key to either `init` or `apply`.
```python
# Even better (just use one key)
_ = myA.apply(v, rngs=jax.random.PRNGKey(0))
```
This functionality would be great to have with the *option* to pass in the RNGS explicitly. This has no consequences for reproducibility since the rngs need to be split somewhere anyway. 

At the moment the RNG design forces me to write work-arounds in my code which requires me to pass around a namespace that tracks all registered scopes. But this is quite error-prone and ideally this would be handled internally.</div>

### LINKED ISSUE #3668: Make `self.make_rng()` callable by using a default RNG stream and have `.init()` and `.apply()` use the default RNG stream if no explicit RNG mapping is passed.
There has been some internal and external (#3506) discussion on making `self.make_rng()` callable by using a default RNG stream and having `.init()` and `.apply()` use the default RNG stream if no explicit RNG mapping is passed. 

In the current design:
* a string name must be passed into `self.make_rng(name)` to explicitly specify the RNG stream to use in the Module
* an explicit RNG mapping (e.g. `{'params': random.key(0), 'other_rng_stream': random.key(1)}`) must be passed to `.init()` and `.apply()` for each RNG stream that's used in the Module

A proposed change:
* Make `self.make_rng(name='default')` callable without explicitly passing in a string name, in which it would default to the `'default'` RNG stream
  * This would allow the user to pass in a single RNG mapping (e.g. `{'default': random.key(0)}`) to `.init()` and `.apply()` that `self.make_rng('other_rng_stream')` will use as the RNG seed if, for example, `{'other_rng_stream': random.key(1)}` wasn't in the RNG mapping that was passed into `.init()` or `.apply()`

chiamp:
Resolves #3668
Resolves #3506

Add default behavior for `self.make_rng`. See #3668 for more details.

### DIFF
--- a/flax/core/scope.py
+++ b/flax/core/scope.py
@@ -750,10 +750,13 @@ class Scope:
     """Returns true if a PRNGSequence with name `name` exists."""
     return name in self.rngs
 
-  def make_rng(self, name: str) -> PRNGKey:
+  def make_rng(self, name: str = 'default') -> PRNGKey:
     """Generates A PRNGKey from a PRNGSequence with name `name`."""
     if not self.has_rng(name):
-      raise errors.InvalidRngError(f'{self.name} needs PRNG for "{name}"')
+      if self.has_rng('default'):
+        name = 'default'
+      else:
+        raise errors.InvalidRngError(f'{self.name} needs PRNG for "{name}"')
     self._check_valid()
     self._validate_trace_level()
     self.rng_counters[name] += 1
--- a/tests/linen/linen_module_test.py
+++ b/tests/linen/linen_module_test.py
@@ -2522,6 +2522,145 @@ class ModuleTest(absltest.TestCase):
     self.assertIn('Bar_0/Dense_1', module_paths)
     self.assertEqual(type(module_paths['Bar_0/Dense_1']), nn.Dense)
 
+  def test_init_apply_default_rng(self):
+    class SubModel(nn.Module):
+      @nn.compact
+      def __call__(self, x, apply_dropout):
+        x = nn.Dense(8)(x)
+        x = nn.Dropout(0.8)(x, deterministic=not apply_dropout)
+        p = self.param(
+          'parameter', lambda key, shape: jax.random.normal(key, shape), x.shape
+        )
+        noise = jax.random.normal(self.make_rng('noise'), x.shape)
+        return x * p + noise
+
+    class Model(nn.Module):
+      @nn.compact
+      def __call__(self, x, apply_dropout):
+        x = nn.Dense(16)(x)
+        x = SubModel()(x, apply_dropout)
+        x = nn.Dropout(0.5)(x, deterministic=not apply_dropout)
+        v = self.variable(
+          'var_collection',
+          'variable',
+          lambda shape: jax.random.normal(self.make_rng('var_rng'), shape),
+          x.shape,
+        )
+        noise = jax.random.normal(self.make_rng('noise'), x.shape)
+        return x * v.value + noise
+
+    key0, key1, key2 = jax.random.split(jax.random.key(0), 3)
+    x = jax.random.normal(key0, (10, 4))
+    model = Model()
+
+    # test init equality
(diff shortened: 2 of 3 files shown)


flax/core/scope.py
```diff
@@ -750,10 +750,15 @@ def has_rng(self, name: str) -> bool:
...
+  def make_rng(self, name: str = 'default') -> PRNGKey:
     """Generates A PRNGKey from a PRNGSequence with name `name`."""
     if not self.has_rng(name):
-      raise errors.InvalidRngError(f'{self.name} needs PRNG for "{name}"')
+      if name != 'default':
+        if not self.has_rng('default'):
+          raise errors.InvalidRngError(f'{self.name} needs PRNG for "{name}"')
+        name = 'default'
+      else:
+        raise errors.InvalidRngError(f'{self.name} needs PRNG for "{name}"')
```
cgarciae, line 761:
```suggestion
      if self.has_rng('default'):
        name = 'default'
      else:
        raise errors.InvalidRngError(f'{self.name} needs PRNG for "{name}"')
```

tests/linen/linen_module_test.py
```diff
@@ -2522,6 +2522,122 @@ def __call__(self, x):
...
+          x.shape,
+        )
+        noise = jax.random.normal(self.make_rng('noise'), x.shape)
+        return x * v.value + noise
+
+    key0, key1, key2 = jax.random.split(jax.random.key(0), 3)
+    x = jax.random.normal(key0, (10, 4))
+    model = Model()
+
+    # test init equality
```
cgarciae, line 2556: Interesting! This doesn't hold in NNX because its entirely counter-based.

cgarciae (approved):

PR #3675, opened 2024-02-01
added kwargs to self.param and self.variable

chiamp:
added kwargs to self.param and self.variable

### DIFF
--- a/docs/guides/flax_fundamentals/flax_basics.ipynb
+++ b/docs/guides/flax_fundamentals/flax_basics.ipynb
@@ -739,11 +739,11 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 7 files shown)


marcvanzee (approved):
Looks great!

chiamp:
btw @marcvanzee, I was getting `mypy` errors until I removed `**init_kwargs` from the first overloaded function signature (e.g. [here](https://github.com/google/flax/pull/3675/files#diff-d53cff3509f30c883e4991afdf992b5b5ef97ae3070d28b50812b8fe1fb3f728R903)). Not sure if it's intended behavior with using `**kwargs`, but I filed a `mypy` bug [here](https://github.com/python/mypy/issues/16854)

PR #3680, opened 2024-02-06
[nnx] add demo.ipynb

cgarciae:
# What does this PR do?

Adds an NNX demo notebook.


chiamp (approved):
Maybe consider executing the code blocks, so users can see the expected output of the print statements.

For reference, this is the [rendered notebook](https://github.com/google/flax/blob/nnx-demo-2/flax/experimental/nnx/docs/demo.ipynb) that I see.

PR #3720, opened 2024-02-26
[nnx] explicit Variables

cgarciae:
# What does this PR do?

* The `Variable.value` attribute which contained the unprocessed values is now named `.raw_value`.
* Replaces `Variable.get_value` and `Variable.set_value` with a `.value` property.
* `Module` getattr and setattr no longer extract the inner `value` of Variables.
* Removes `Module.variables` and `State.variables` helpers.

Basic example now looks like this:

```python
class Linear(nnx.Module):
  def __init__(self, din: int, dout: int, *, rngs: nnx.Rngs):
    key = rngs.params()
    self.w: Param[Array] = nnx.Param(jax.random.uniform(key, (din, dout)))
    self.b: Param[Array] = nnx.Param(jnp.zeros((dout,)))

    return x @ self.w.value + self.b.value
 ```


levskaya (approved):

NeilGirdhar:
Woohoo!!!  Awesome!!

PR #3741, opened 2024-03-05
added Einsum layer to NNX

chiamp:
Unlike the Linen implementation, the NNX implementation of the `Einsum` layer requires an explicit [`bias_shape`](https://github.com/google/flax/pull/3741/files#diff-66832bf05a8eff953cecd238affae58842fefeb11741c6e11285c51ad3375cf0R395) to be passed at construction time since the bias param is instantiated during construction time and requires that information. During call time, the `broadcasted_bias_shape` is inferred by passing the `inputs`, `kernel` and `einsum_str` as arguments to [`opt_einsum.parser.parse_einsum_input`](https://github.com/dgasmith/opt_einsum/blob/master/opt_einsum/parser.py#L265) (similar to the [Linen implementation](https://github.com/google/flax/blob/main/flax/linen/linear.py#L391)).

~~The NNX implementation of the `Einsum` layer requires an explicit [`input_shape`](https://github.com/google/flax/pull/3741/files#diff-66832bf05a8eff953cecd238affae58842fefeb11741c6e11285c51ad3375cf0R395) to be passed at construction time since the bias param is instantiated during construction time and requires that information. A dummy array is constructed using [`jax.ShapeDtypeStruct`](https://jax.readthedocs.io/en/latest/_autosummary/jax.ShapeDtypeStruct.html#jax.ShapeDtypeStruct)) and is passed to [`opt_einsum.parser.parse_einsum_input`](https://github.com/dgasmith/opt_einsum/blob/master/opt_einsum/parser.py#L265) (in the case of the Linen implementation, [both arrays are provided during call time](https://github.com/google/flax/blob/main/flax/linen/linear.py#L391)).~~

While both implementations allow `einsum_str` to be passed during construction and call time, the bias shape can be determined either by the `einsum_str` passed in constructor or call time because of lazy init in Linen, while the bias shape is **_always_** determined by the `einsum_str` passed **_in construction time_** in NNX (meaning **_it must always be passed during construction time_**). NNX allows the option to pass in a different `einsum_str` during call time which will take precedence over the constructor `einsum_str`, but the bias shape is never determined at call time, since it's already been instantiated.

Another difference to note is that the Linen implementation has a construction signature of `Einsum(kernel_shape,  einsum_str=None, ...)` while the NNX implementation has a construction signature of `Einsum(einsum_str, kernel_shape, bias_shape=None, ...)`. `einsum_str` can be `None` in Linen, since it can also be passed as an argument during call time, whereas an `einsum_str` **_must be passed_** in the NNX implementation during construction time.

### DIFF
--- a/flax/experimental/nnx/nnx/nn/linear.py
+++ b/flax/experimental/nnx/nnx/nn/linear.py
@@ -34,14 +34,16 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 from jax import lax
+import opt_einsum
 
 from flax.experimental import nnx
 from flax.experimental.nnx.nnx import rnglib, variables
-from flax.experimental.nnx.nnx.module import Module
+from flax.experimental.nnx.nnx.module import Module, first_from
 from flax.experimental.nnx.nnx.nn import dtypes, initializers
 from flax.typing import (
   Array,
   Dtype,
+  Shape,
   Initializer,
   PrecisionLike,
   DotGeneralT,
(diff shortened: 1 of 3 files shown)


cgarciae:
I think it would be cleaner if the user provides `bias_shape` and that way we don't even need `_get_bias_shape` and the einsum parsing machinery.

chiamp:
> I think it would be cleaner if the user provides the `bias_shape` instead of the `input_shape`. This way we don't even need `_get_bias_shape` and the einsum parsing machinery.

I originally wrote it this way. The user would also have to provide `broadcasted_bias_shape` as well; my thought process was that it would be easier for the user to just provide the `input_shape`. What if we allow the user to pass in either `input_shape` (in which bias shape would be inferred), or `bias_shape` and `broadcasted_bias_shape` (in which bias shape is specifically passed)? The constructor signature could be something like:
```
def __init__(
    self,
    einsum_str: str,
    kernel_shape: Shape,
    input_shape: tp.Optional[Shape] = None,
    bias_shape: tp.Optional[Shape] = None,
    broadcasted_bias_shape: tp.Optional[Shape] = None,
    *,
    ...
```

flax/experimental/nnx/nnx/nn/linear.py
```diff
@@ -356,6 +358,164 @@ def __call__(self, inputs: Array) -> Array:
...
 
+class Einsum(Module):
+  """An einsum transformation with learnable kernel and bias.
+
+  Example usage::
+
+    >>> from flax.experimental import nnx
+    >>> import jax.numpy as jnp
+
+    >>> layer = nnx.Einsum((3, 4, 5), (5, 6, 7), 'abc,cde->abde', rngs=nnx.Rngs(0))
```
cgarciae, line 369:
```suggestion
    >>> layer = nnx.Einsum('abc,cde->abde', (3, 4, 5), (5, 6, 7), rngs=nnx.Rngs(0))
```

flax/experimental/nnx/nnx/nn/linear.py
```diff
@@ -356,6 +358,164 @@ def __call__(self, inputs: Array) -> Array:
...
+    rngs: rng key.
+  """
+
+  def __init__(
+    self,
+    einsum_str: str,
+    kernel_shape: Shape,
+    bias_shape: tp.Optional[Shape] = None,
+    *,
+    use_bias: bool = True,
```
cgarciae, line 399: Maybe `bias_shape` being present or not is enough and we don't need `use_bias`?

cgarciae (changes requested):

cgarciae (approved):

PR #3753, opened 2024-03-13
Fix doctest and unbreak head

IvyZX:
* Add functionality to suppress certain `absl.logging` messages during `doctest`. This will not break or unbreak anything (because logs will not break `doctest`), but can help clean up the printouts.
* Fix the Orbax upgrade guide to unbreak head.
* Add default required arguments for `ContextManager.__exit__` in tensorboard.py


chiamp (approved):
Thanks!

Issue #3718, opened 2024-02-25
Add `reset_gate` flag to `MGUCell`


carlosgmartin:
There is a small simplification of [`MGUCell`](https://flax.readthedocs.io/en/latest/api_reference/flax.linen/layers.html#flax.linen.MGUCell) in the literature that removes the extra elementwise multiplication inside the computation of the candidate new hidden state. That is, it changes

$$n = \tanh(W_{in} x + b_{in} + f * (W_{hn} h + b_{hn}))$$

to

$$n = \tanh(W_{in} x + b_{in} + W_{hn} h)$$

For context, according to [*Light Gated Recurrent Units for Speech Recognition*](https://arxiv.org/abs/1803.10225) by Yoshua Bengio *et al.*:

> A. Removing the reset gate
>
> We also argue that a certain redundancy in the activations of reset and update gates might occur when processing speech sequences. For instance, when it is necessary to give more importance to the current information, the GRU model can set small values of $r_t$. A similar effect can be achieved with the update gate only, if small values are assigned to $z_t$. The latter solution tends to weight more the candidate state $\tilde{h}_t$, that depends heavily on the current input. Similarly, a high value can be assigned either to $r_t$ or to $z_t$, in order to place more importance on past states. This redundancy is also highlighted in Fig. 1, where a temporal correlation in the average activations of update and reset gates can be readily appreciated for a GRU trained on TIMIT. [...]
>
> Based on these reasons, the first variation to standard GRUs thus concerns the removal of the reset gate $r_t$.

In my own experiments, I've found this to improve performance.

**Feature request:** Add to `MGUCell` a `reset_gate` flag (default `True`) that controls whether this extra multiplicative gating is performed.

I can submit a PR for this. It would require editing only a few lines of `MGUCell.__call__`:

```python3
        # add bias when the linear transformations aren't directly summed.
        x = dense_h(name="hn", use_bias=self.reset_gating)(h)
        if self.reset_gating:
            x *= f
        n = self.activation_fn(
            dense_i(name="in", bias_init=self.activation_bias_init)(inputs) + x
        )
        new_h = (1.0 - f) * n + f * h
        return new_h, new_h
```

cgarciae:
Hey @carlosgmartin! Are you interested in adding this feature? Happy to review the PR :)

carlosgmartin:
@cgarciae Certainly! Before doing that, can [this PR](https://github.com/google/flax/pull/3712) be merged?

[...]

PR #3760, opened 2024-03-14
Add reset_gate flag to MGUCell.

carlosgmartin:
#3718

### DIFF
--- a/flax/linen/recurrent.py
+++ b/flax/linen/recurrent.py
@@ -635,6 +635,16 @@ class MGUCell(RNNCellBase):
 
   where x is the input and h is the output of the previous time step.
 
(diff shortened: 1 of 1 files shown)


cgarciae (approved):

cgarciae:
Thanks @carlosgmartin !

flax/linen/recurrent.py
```diff
@@ -635,6 +635,16 @@ class MGUCell(RNNCellBase):
 
   where x is the input and h is the output of the previous time step.
 
+  If `reset_gate` is false, the above becomes
```
chiamp, line 638:
```suggestion
  If ``reset_gate`` is false, the above becomes
```

chiamp (approved):
Thanks!

PR #3772, opened 2024-03-20
fix tabulate on norm wrappers

chiamp:
fix tabulate on norm wrappers


flax/linen/summary.py
```diff
@@ -460,7 +460,7 @@ def _get_variables():
       call_depth = len(c.path)
       inputs = _process_inputs(c.args, c.kwargs)
 
-      if c.path in visited_paths:
+      if c.path in visited_paths or not hasattr(c.module, c.method):
```
cgarciae, line 463: Can you explain what is happening here? Why `c` doesn't have `module` and `method`?

PR #3782, opened 2024-03-22
Add `nnx.training`

chiamp:
Add `nnx.training`.

API reference for [`nnx.metrics`](https://flax--3782.org.readthedocs.build/en/3782/api_reference/flax.experimental.nnx/training/metrics.html) and [`nnx.optimizer`](https://flax--3782.org.readthedocs.build/en/3782/api_reference/flax.experimental.nnx/training/optimizer.html).

~Had to import [`chex.ArrayTree`](https://github.com/google/flax/pull/3782/files#diff-7bb758349c08d151354150d8d215d8b98a0ce1308134088bafc41c395c59074aR126) to make typing happy. Is there an alternative where we don't have to use this dependency? e.g. annotating with `# type: ignore`.~

Renamed `optimizer.update` method to `optimizer.apply_gradients` method, because `optimizer.update` was being called instead of the base class `nnx.Module.update` method when using `nnx.jit`.
Added `OptState` and `MetricState` `Variable` wrapper classes for `nnx.optimizer.Optimizer` and `nnx.metrics.Metric`, respectively.

Updated [`toy_examples/02_lifted_transforms.py`](https://github.com/chiamp/flax/blob/nnx_module_subclass/flax/experimental/nnx/examples/toy_examples/02_lifted_transforms.py) to use `nnx.optimizer.Optimizer`. Training loop runs slightly slower.
Updated [`docs/experimental/nnx/mnist_tutorial.ipynb`](https://flax--3782.org.readthedocs.build/en/3782/experimental/nnx/mnist_tutorial.html) to use `nnx.metrics.MultiMetric` and `nnx.optimizer.Optimizer`.

For some reason, I'm getting an `nbstripout` error on CI when I try to include the outputs of the mnist tutorial, so I have removed them for now.

PR #3791, opened 2024-03-27
[nnx] Arrays are state

cgarciae:
# What does this PR do?

`jax.Array` and `np.ndarray` are now treated as state same as `Variables`. 

### Discussion
Previously we just treated all non-Variable and non-node as static, however, since we support pytrees and certain pytrees such as `opt_state` in the `Optimizer` class from #3782 sometimes include arrays such as the training `step`  array, then we should also support Array. This sadly breaks from the simplicity of `State` being just a nested mapping with `Variables` as leaves, but on the other hand Array is a very natural state type to try to support and in fact the NNX already supported it in the past.


flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -455,16 +459,19 @@ def _graph_flatten(
     if is_node(value):
       graphdef = _graph_flatten((*path, key), ref_to_index, flat_state, value)
       subgraphs.append((key, graphdef))
-    elif isinstance(value, Variable):
+    elif is_state_leaf(value):
       str_path = '/'.join((*path, key))
-      if value in ref_to_index:
-        variables.append((key, ref_to_index[value]))
+      if isinstance(value, Variable):
```
superbobry, line 464: Nit: maybe move this into a separate branch before `is_state_leaf` to reduce the nesting here?

```
elif isinstance(value, Variable):
  ...
elif is_state_leaf(value):
  ...
```

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -746,23 +756,32 @@ def _graph_update_dynamic(
 
     # case 2: subgraph is being updated
     if is_node(current_value):
-      if isinstance(value, Variable):
+      if is_state_leaf(value):
         raise ValueError(
           f'Expected a subgraph for {key!r}, but got a Variable: {value!r}'
```
superbobry, line 761: The error message needs updating.

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -746,23 +756,32 @@ def _graph_update_dynamic(
...
         )
+      assert not isinstance(value, Variable)
```
superbobry, line 763: Is this `assert` only needed to please the type checker?

flax/experimental/nnx/nnx/state.py
```diff
@@ -74,29 +74,23 @@ def __init__(
...
     value = self._mapping[key]
-    if isinstance(value, Variable):
-      return value
-    return State(value)
+    if isinstance(value, dict):
+      return State(value)
+    return value
 
-  def __getattr__(self, key: Key) -> Variable | State:
+  def __getattr__(self, key: Key) -> tp.Any:
```
superbobry, line 85: Is the return type genuinely `Any` here? It looks it could be `StateLeaf | State` where

```
StateLeaf = Union[Variable, jax.Array, np.ndarray]
```

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -152,6 +153,9 @@ def is_node(x: tp.Any) -> bool:
 def is_node_type(x: type[tp.Any]) -> bool:
   return x in NODE_TYPES or x is PytreeType
 
+def is_state_leaf(x: tp.Any) -> bool:
```
superbobry, line 156: Can this be a `typing.TypeGuard`?

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -152,6 +153,9 @@ def is_node(x: tp.Any) -> bool:
...
+  return isinstance(x, (Variable, np.ndarray, jax.Array))
```
superbobry, line 157: Am I right that you're intentionally disallowing Python scalars here?

flax/experimental/nnx/nnx/state.py
```diff
@@ -229,25 +223,20 @@ def __sub__(self, other: 'State') -> 'State':
 def _state_flatten_with_keys(x: State):
   items = sorted(x._mapping.items(), key=lambda item: item[0])
   children = tuple((jtu.DictKey(key), value) for key, value in items)
-  return children, tuple(x._mapping.keys())
+  return children, tuple(key for key, _ in items)
 
 
 def _state_unflatten(
-  static: tp.Tuple[Path, ...] | None,
+  static: tp.Tuple[Path, ...],
```
superbobry, line 230: Use the built in `tuple` here and elsewhere.

superbobry (changes requested):
Could you motivate this PR a bit more in the description? Why not require users to wrap their arrays into variables?

flax/experimental/nnx/nnx/graph_utils.py
cgarciae: indeed

flax/experimental/nnx/nnx/graph_utils.py
cgarciae: great idea! I should start using `TypeGuard` more

flax/experimental/nnx/nnx/graph_utils.py
cgarciae: Main problem here is that `Variable` is registered as a pytree so it would get accepted but we explicitly want to forbid it.

flax/experimental/nnx/nnx/state.py
cgarciae: wow, coded this exact solution after implementing the TypeGuard :)

cgarciae: ...but hadn't seen you suggested exactly this

flax/experimental/nnx/nnx/state.py
cgarciae: wdym?

superbobry (approved):

PR #3807, opened 2024-03-31
[nnx] Rng Variable tags

cgarciae:
# What does this PR do?

This PR adds the necessary changes so we can filter `RngState` Variables based on the stream name using `Filter`s, this allows you to manage RNG state a regular graph state but using the previous name-based filter convention.

### Changes
* Adds `RngKey` and `RngCount` Variable types that inherit from `RngState`. `RngKey` is expected to have  a`tag: str` metadata attribute.
* `RngStream.key` is now a `RngKey` and `RngStream.count` is now a `RngCount`. `Rngs` sets `RngKey.tag` as the stream's name.
* `str` filters now create a `WithTag` filter that selects Variables with a `tag` attribute that matches that string.
* Removes `Rngs._rngs`, `RngStreams` are now stored as attributes of `Rngs`.
* Refactors `RngStream.make_rng` to `__call__`.
* `Rngs.__getattr__` and `Rngs.__getitem__` now returns a `RngStream`.

### DIFF
--- a/flax/experimental/nnx/nnx/filterlib.py
+++ b/flax/experimental/nnx/nnx/filterlib.py
@@ -26,10 +26,14 @@ Predicate = tp.Callable[[PathParts, tp.Any], bool]
 FilterLiteral = tp.Union[type, str, Predicate, bool, ellipsis, None]
 Filter = tp.Union[FilterLiteral, tuple[FilterLiteral, ...], list[FilterLiteral]]
 
+@tp.runtime_checkable
+class _HasTag(tp.Protocol):
+  tag: str
+
 
 def to_predicate(filter: Filter) -> Predicate:
   if isinstance(filter, str):
-    return AtPath(filter)
+    return WithTag(filter)
   elif isinstance(filter, type):
     return OfType(filter)
   elif isinstance(filter, bool):
@@ -47,11 +51,11 @@ def to_predicate(filter: Filter) -> Predicate:
 
 
 @dataclasses.dataclass
-class AtPath:
-  str_key: str
+class WithTag:
+  tag: str
 
   def __call__(self, path: PathParts, x: tp.Any):
-    return self.str_key in path
+    return isinstance(x, _HasTag) and x.tag == self.tag
 
 
 @dataclasses.dataclass
--- a/flax/experimental/nnx/nnx/rnglib.py
+++ b/flax/experimental/nnx/nnx/rnglib.py
@@ -49,27 +49,37 @@ class Missing:
 
 MISSING = Missing()
 
+
 class RngState(Variable[jax.Array]):
   pass
+
+
+class RngCount(RngState):
+  pass
+
+
+class RngKey(RngState):
+  tag: str
 
 
 @dataclasses.dataclass
 class RngStream(GraphNode):
   def __init__(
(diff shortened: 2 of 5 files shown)


flax/experimental/nnx/nnx/filterlib.py
```diff
@@ -47,11 +51,11 @@ def to_predicate(filter: Filter) -> Predicate:
...
 
 @dataclasses.dataclass
-class AtPath:
-  str_key: str
+class WithTag:
+  tag: str
 
   def __call__(self, path: PathParts, x: tp.Any):
-    return self.str_key in path
+    return isinstance(x, _HasTag) and x.tag == self.tag
```
superbobry, line 58: I think you can simply use `hasattr()` here?

flax/experimental/nnx/nnx/filterlib.py
```diff
@@ -47,11 +51,11 @@ def to_predicate(filter: Filter) -> Predicate:
 
 
 @dataclasses.dataclass
-class AtPath:
-  str_key: str
+class WithTag:
+  tag: str
 
   def __call__(self, path: PathParts, x: tp.Any):
```
superbobry, line 57: mypy will not type check this unless you add `-> bool` IIRC.

flax/experimental/nnx/nnx/rnglib.py
```diff
@@ -110,44 +120,48 @@ def __init__(
...
+      if name != '_graph_node__state':
+        yield name
 
   def __len__(self) -> int:
-    return len(self._rngs)
+    return len(vars(self)) - 1
 
   def __contains__(self, name: tp.Any) -> bool:
-    return name in self._rngs
+    return name in vars(self)
```
superbobry, line 160: Do you want to special-case `"_graph_node__state"` here for consistency with other methods?

flax/experimental/nnx/nnx/rnglib.py
```diff
@@ -110,44 +120,48 @@ def __init__(
...
 
   def __iter__(self) -> tp.Iterator[str]:
-    return iter(self._rngs)
+    for name in vars(self):
+      if name != '_graph_node__state':
+        yield name
 
   def __len__(self) -> int:
-    return len(self._rngs)
+    return len(vars(self)) - 1
```
superbobry, line 157: Is `"_graph_node__state"` always present? If not, you should probably do `- ('_graph_node__state' in vars(self))`.

flax/experimental/nnx/nnx/rnglib.py
```diff
@@ -110,44 +120,48 @@ def __init__(
...
 
   def __contains__(self, name: tp.Any) -> bool:
-    return name in self._rngs
+    return name in vars(self)
 
   def replace(self, **kwargs: tp.Union[int, jax.Array, RngStream]) -> 'Rngs':
-    rngs: dict[str, tp.Any] = self._rngs.copy()
+    rngs: dict[str, tp.Any] = vars(self).copy()
+    del rngs['_graph_node__state']
```
superbobry, line 164: It's unfortunate that the implementation detail of GraphNode leaks into the subclasses. Can we do better?

PR #3808, opened 2024-04-01
[linen] allows multiple compact methods

cgarciae:
# What does this PR do?

Allows multiple compact methods to be used in a single Module. However, its up to the user to guarantee that auto-naming collisions don't occur between different `apply` calls using different call paths.


ZacCranko:
Hi there @cgarciae, how far off is this from getting merged & released. I'd love to be able to mark multiple methods as compact. Cheers!

PR #3809, opened 2024-04-01
[nnx] allow all jit kwargs in nnx.jit

cgarciae:
# What does this PR do?

Allows all `nnx.jit` kwargs to be passed to the underlying `jax.jit` transform.

### DIFF
--- a/flax/experimental/nnx/nnx/transforms.py
+++ b/flax/experimental/nnx/nnx/transforms.py
@@ -198,16 +198,8 @@ class JITOptions:
     del kwargs['donate_object_state']
     if kwargs['in_shardings'] is UNSPECIFIED:
       kwargs.pop('in_shardings')
-    else:
-      raise ValueError('in_shardings is not supported yet')
     if kwargs['out_shardings'] is UNSPECIFIED:
       kwargs.pop('out_shardings')
-    else:
-      raise ValueError('out_shardings is not supported yet')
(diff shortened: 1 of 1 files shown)


chiamp (approved):

flax/experimental/nnx/nnx/transforms.py
```diff
@@ -198,16 +198,8 @@ def get_jit_kwargs(self) -> dict[str, tp.Any]:
...
-    else:
-      raise ValueError('in_shardings is not supported yet')
     if kwargs['out_shardings'] is UNSPECIFIED:
       kwargs.pop('out_shardings')
-    else:
-      raise ValueError('out_shardings is not supported yet')
-    if kwargs['donate_argnums'] != ():
-      raise ValueError('donate_argnums is not supported yet')
-    if kwargs['abstracted_axes'] is not None:
-      raise ValueError('abstracted_axes is not supported yet')
```
chiamp, line 210: Was there some change in another PR that allowed kwargs to be passed to `jax.jit`, or this behavior wasn't tested yet and we just erred on the side of caution by raising these errors?

cgarciae: I initially introduced these restrictions as I didn't know if I could support all arguments, but then I figured out a way and forgot to remove them.

PR #3813, opened 2024-04-02
[nnx] improve toy examples

cgarciae:
# What does this PR do?

Improves toy examples, makes them more idiomatic according to current standards.

### DIFF
--- a/flax/experimental/nnx/examples/toy_examples/01_functional_api.py
+++ b/flax/experimental/nnx/examples/toy_examples/01_functional_api.py
@@ -67,10 +67,11 @@ def train_step(params, counts, batch):
   x, y = batch
 
   def loss_fn(params):
-    y_pred, (_, updates) = static.apply(params, counts)(x)
-    counts_ = updates.extract(Count)
+    model = static.merge(params, counts)
+    y_pred = model(x)
+    new_counts = model.extract(Count)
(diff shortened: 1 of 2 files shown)


flax/experimental/nnx/examples/toy_examples/01_functional_api.py
```diff
@@ -67,10 +67,11 @@ def train_step(params, counts, batch):
   x, y = batch
 
   def loss_fn(params):
-    y_pred, (_, updates) = static.apply(params, counts)(x)
-    counts_ = updates.extract(Count)
+    model = static.merge(params, counts)
+    y_pred = model(x)
+    _counts = model.extract(Count)
```
superbobry, line 72: Why the `_` prefix? It is normally used (by convention) to indicate that a variable is unused, but here this isn't the case.

cgarciae: Should we keep `counts_` then?  Here we need a new variable name, cannot reuse `counts`.

superbobry: extracted_counts or new_counts?

cgarciae: `new_counts` it is

superbobry (approved):

PR #3826, opened 2024-04-04
[nnx] refactor graph_utils

cgarciae:
# What does this PR do?

Adds the `UpdateContex` API that contains special versions of `split`, `merge`, and `update` which are able to perform full graph updates.

```python
with nnx.UpdateContext() as ctx:
  m = Foo()
  graphdef, state = ctx.split(m)

  @jax.jit
  def f_pure(graphdef, state):
    m = ctx.merge(graphdef, state)
    f(m)
    graphdef, state = ctx.split(m)
    return graphdef, state

  graphdef, state = f(graphdef, state)
  
  ctx.update(graphdef, state)
```

To make this possible `GraphDef` is renamed to `NodeDef` and a new `GraphDef` dataclass is defined which contains both the `NodeDef` and a `index_mapping` which is populated by the second `ctx.split` and used by `ctx.update`.

```python
@dataclasses.dataclass(frozen=True)
class GraphDef(tp.Generic[Node], reprlib.Representable):
  nodedef: NodeDef[Node]
  index_mapping: dict[Index, Index] | None
```

### DIFF
--- a/flax/experimental/nnx/nnx/graph_utils.py
+++ b/flax/experimental/nnx/nnx/graph_utils.py
@@ -37,8 +37,13 @@ from flax.experimental.nnx.nnx.proxy_caller import (
   CallableProxy,
   DelayedAccessor,
 )
-from flax.experimental.nnx.nnx.state import State, StateLeaf, is_state_leaf
-from flax.experimental.nnx.nnx.variables import EMPTY, Empty, Variable
+from flax.experimental.nnx.nnx.state import (
+  FlatState,
+  State,
+  StateLeaf,
+  is_state_leaf,
+)
+from flax.experimental.nnx.nnx.variables import EMPTY, Variable
 from flax.typing import PathParts, Key
 
 A = tp.TypeVar('A')
@@ -99,8 +104,11 @@ class _HashById(tp.Hashable, tp.Generic[A]):
 class RefMap(tp.MutableMapping[A, B], reprlib.MappingReprMixin[A, B]):
   """A mapping that uses object id as the hash for the keys."""
 
-  def __init__(self):
+  def __init__(
+    self, mapping: tp.Mapping[A, B] | tp.Iterable[tuple[A, B]] = (), /
+  ):
     self._mapping: dict[_HashById[A], B] = {}
+    self.update(mapping)
 
   def __getitem__(self, key: A) -> B:
     return self._mapping[_HashById(key)]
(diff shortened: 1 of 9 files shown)


flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -336,60 +344,51 @@ def __init__(
...
+  def __deepcopy__(self, memo=None):
+    nodedef = deepcopy(self.nodedef, memo)
+    index_mapping = deepcopy(self.index_mapping, memo)
+    return GraphDef(nodedef, index_mapping)
 
-  @property
-  def variables(self):
-    return self._variables
+  def __hash__(self):
+    # refmap is opaque
```
superbobry, line 386: What is `refmap` here and below?

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -903,33 +894,49 @@ def _graph_update_static(
...
+  RefMap[tp.Any, Index], GraphDef[A], State, tpe.Unpack[tuple[State, ...]]
+]:
   ...
 
 
-def split(
-  graph_node: A, *filters: filterlib.Filter
-) -> tuple[GraphDef[A], State, tpe.Unpack[tuple[State, ...]]]:
-  graphdef, state, _ = graph_flatten(graph_node)
+def full_split(
```
superbobry, line 932: Please ad a docstring for this and the other new APIs.

superbobry (changes requested):
I cannot really evaluate these new APIs properly. I would recommend adding another Flax contributor as a reviewer.

flax/experimental/nnx/nnx/graph_utils.py
cgarciae: removed this, no longer relevant

flax/experimental/nnx/nnx/graph_utils.py
```diff
@@ -902,15 +891,116 @@ def _graph_update_static(
...
+  def __enter__(self):
+    return self
+
+  def __exit__(self, *args, **kwargs):
+    self.refmap = None
+    self.idxmap = None
+
+  # define hash and eq to make this an opaque object
+  def __hash__(self):
+    return 0
```
superbobry, line 910: This will make all instance hash to the same value. Why is it safe?

flax/experimental/nnx/nnx/graph_utils.py
superbobry: The docstrings are still missing in the latest version. Do you think they are unnecessary and the signatures are sufficient?

superbobry (changes requested):

PR #3827, opened 2024-04-04
[NVIDIA] Extend the custom fp8 accumulate dtype in non-jit scenarios

kaixih:
This PR extends the custom fp8 dtype in the non-jit scenarios. The custom fp8 dtype is for the grad accumulation where we can do a max op rather than the default add op to accumulate the grads. This is designed for the fp8 parameters whose grads are their new values and we only want to pick the max among the grads of the sub-tensors.

Previously, this custom dtype has to be in the jit scope and this PR relaxes that restriction.

Note, this PR depends on https://github.com/google/jax/pull/20266

cc. @nluehr @mingxu1067

### DIFF
--- a/flax/linen/fp8_ops.py
+++ b/flax/linen/fp8_ops.py
@@ -21,6 +21,8 @@ from jax import custom_jvp, custom_vjp, lax, random
 from jax import numpy as jnp
 from jax._src import core
 from jax._src import dtypes
+from jax._src import earray
+from jax._src.interpreters import pxla
 
 from flax.linen import initializers, module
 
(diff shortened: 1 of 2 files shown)


kaixih:
This [JAX PR](https://github.com/google/jax/pull/20266/files) has been merged and so I rebased my change to import the newly-introduced `earray`. @mattjj Can you take a look?

IvyZX:
Hi thanks for adding this change. 
Looks like JAX hasn't released `earray`, so you would need to add a check when importing it, if you want this PR to be merged in early.
Otherwise looks good to me.

kaixih:
@IvyZX Thanks for the review. I made one more change to make sure it compatible with the current JAX version that has no earray. PTAL.

IvyZX (approved):

PR #3834, opened 2024-04-09
[linen] fix DenseGeneral init

cgarciae:
# What does this PR do?

Removes the flattening / unflattening logic from `DenseGeneral` init wrappers. Original motivation for this was to better preserve the statistics for arrays with more than 2 feature dimension, however this can also be achieve via the `in_axis` and `out_axis` parameters of most initializers making this flattening process unnecesary.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -146,7 +146,7 @@ filterwarnings = [
     "ignore:.*Deprecated call to.*pkg_resources.declare_namespace.*:DeprecationWarning",
     # DeprecationWarning: jax.tree_map is deprecated: use jax.tree.map (jax v0.4.25 or newer) or jax.tree_util.tree_map (any JAX version).
     "ignore:.*jax.tree_map is deprecated.*:DeprecationWarning",
-]
+    ]
 
 [tool.coverage.report]
 exclude_lines = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -146,6 +146,8 @@ filterwarnings = [
     "ignore:.*Deprecated call to.*pkg_resources.declare_namespace.*:DeprecationWarning",
     # DeprecationWarning: jax.tree_map is deprecated: use jax.tree.map (jax v0.4.25 or newer) or jax.tree_util.tree_map (any JAX version).
     "ignore:.*jax.tree_map is deprecated.*:DeprecationWarning",
+    # DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
+    "ignore:.*elementwise comparison failed; this will raise an error in the future.*:DeprecationWarning",
```
chiamp, line 150: Where is this error originating from? Can it be fixed by `jnp.broadcast_to` or jnp.expand_dims`? If this will throw an error in the future, we should probably try to fix it.

chiamp (approved):

pyproject.toml
cgarciae: I removed as it appears this is only happening locally

PR #3835, opened 2024-04-09
[linen] enable separate initializers for out layer in MultiHeadDotProductAttention

cgarciae:
# What does this PR do?

Adds `out_kernel_init` and `out_bias_init` Initializer arguments so users can specify the initializers for the `out` layer independently. This is useful main to be able to define a different sharding scheme for that layer.

### DIFF
--- a/flax/experimental/nnx/nnx/nn/attention.py
+++ b/flax/experimental/nnx/nnx/nn/attention.py
@@ -281,7 +281,11 @@ class MultiHeadAttention(Module):
     precision: numerical precision of the computation see `jax.lax.Precision`
       for details.
     kernel_init: initializer for the kernel of the Dense layers.
+    out_kernel_init: optional initializer for the kernel of the output Dense layer,
+      if None, the kernel_init is used.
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -146,6 +146,8 @@ filterwarnings = [
     "ignore:.*Deprecated call to.*pkg_resources.declare_namespace.*:DeprecationWarning",
     # DeprecationWarning: jax.tree_map is deprecated: use jax.tree.map (jax v0.4.25 or newer) or jax.tree_util.tree_map (any JAX version).
     "ignore:.*jax.tree_map is deprecated.*:DeprecationWarning",
+    # np.testing.assert_array_equal DeprecationWarning
+    "ignore:.*elementwise comparison failed; this will raise an error in the future.*:DeprecationWarning",
```
chiamp, line 150: Where is this error originating from? Can it be fixed by `jnp.broadcast_to` or jnp.expand_dims`? If this will throw an error in the future, we should probably try to fix it.

chiamp (approved):

PR #3841, opened 2024-04-10
updated nnx dropout

chiamp:
Follow-up from #3815.

Updated NNX `DropOut` with the following:
* added an error message when `deterministic=False` but nothing was passed to the constructor or call `rngs` arg
* added test checking equivalence between constructor and call `rngs` args
* added test checking `rngs` override functionality
* added test checking for raised errors if `deterministic` or `rngs` are not passed in either the constructor or call arg

### DIFF
--- a/flax/experimental/nnx/tests/nn/test_stochastic.py
+++ b/flax/experimental/nnx/tests/nn/test_stochastic.py
@@ -14,28 +14,81 @@
 
 
 import jax.numpy as jnp
+import numpy as np
 
 from flax.experimental import nnx
+
+import pytest
 
 
 class TestStochastic:
   def test_dropout_internal_rngs(self):
     n = 0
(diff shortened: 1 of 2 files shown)


flax/experimental/nnx/tests/nn/test_stochastic.py
```diff
@@ -17,25 +17,78 @@
...
+    np.testing.assert_raises_regex(
+      ValueError, 'No `deterministic` argument was provided to Dropout', m, x
+    )
+    # deterministic call arg provided
+    m(x, deterministic=True)
+    # deterministic constructor arg provided
+    m.set_attributes(deterministic=True)
+    y = m(x)
+    # both deterministic call and constructor arg provided
+    np.testing.assert_raises(
```
cgarciae, line 81: interesting, we usually use `self.assertRaisesRegex`

cgarciae (approved):
this is great, thanks!

PR #3847, opened 2024-04-11
Added three tab option to sphinx directive `codediff` and added testing for first tab

chiamp:
This PR does the following:
* updated sphinx directive codediff to have the option of including an arbitrary amount of tabs
* added testing for the first tab, since it wasn't tested before by doctest
* added `skip_test` to skip doctesting for any number of tabs. This is useful for legacy code snippets that no longer work in migration guides
* added `groups` to allow user to denote which code snippets should be grouped and executed together
See the [docstring](https://github.com/google/flax/pull/3847/files#diff-8fb46ffa43cafe72f26b7d5b91232315f539776e081cc1302ea24746dc625b11R50-R81) for more details on how to use the `codediff` directive.
* updated and added more tests to test the codediff directive (and enabled CI testing)

This is in preparation for adding an NNX tab in the haiku migration guide, for users who want to compare the differences between Haiku, Linen and NNX.

---

~This PR does the following:~
* ~updated sphinx directive codediff to have the option of including a third tab (title_middle) for code diff~
* ~added testing for the first tab, since it wasn't tested before by doctest~
* ~added `skip_left_doctest` to skip doctesting of legacy code snippets (in the left tab) that no longer work in migration guides~
* ~added user option to denote whether a code snippet should executed in the left (`testcode_left`) or right (`testcode_right`) tab, or both~

~This is in preparation for adding an NNX tab in the haiku migration guide, for users who want to compare the differences between Haiku, Linen and NNX.~

### DIFF
--- a/docs/guides/training_techniques/batch_norm.rst
+++ b/docs/guides/training_techniques/batch_norm.rst
@@ -10,7 +10,7 @@ of non-differentiable state that must be handled appropriately.
 
 Throughout the guide, you will be able to compare code examples with and without Flax ``BatchNorm``.
 
-.. testsetup::
+.. testsetup:: No BatchNorm, With BatchNorm
 
   import flax.linen as nn
   import jax.numpy as jnp
@@ -36,8 +36,7 @@ or ``tf.keras.Model`` by setting the
 `training <https://www.tensorflow.org/api_docs/python/tf/keras/Model#call>`__ flag).
 
 .. codediff::
-  :title_left: No BatchNorm
-  :title_right: With BatchNorm
+  :title: No BatchNorm, With BatchNorm
   :sync:
 
   class MLP(nn.Module):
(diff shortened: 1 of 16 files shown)


docs/guides/training_techniques/batch_norm.rst
```diff
@@ -79,17 +85,17 @@ The ``batch_stats`` collection must be extracted from the ``variables`` for late
...
   x = jnp.ones((1, 3))
-  variables = mlp.init(jax.random.key(0), x)
+  variables = model.init(jax.random.key(0), x)
   params = variables['params']
 
 
   jax.tree_util.tree_map(jnp.shape, variables)
   ---
-  mlp = MLP()
+  bn_model = BatchNormModel()
```
cgarciae, line 96: can we keep names the same as much as possible? I think it helps visually when you change tabs that only the actual differences in code change.

chiamp: The problem is that `model` is referenced in later code snippets, e.g.: https://flax--3847.org.readthedocs.build/en/3847/guides/training_techniques/batch_norm.html#training-and-evaluation
So if we use `model` for both Haiku and Flax, Flax would overwrite the `model` and the next Haiku code snippet would be using the Flax `model` instead of the Haiku `model`.

cgarciae (approved):

PR #3876, opened 2024-04-22
[nnx] v0.1

cgarciae:
# What does this PR do?

* `nnx.split`, `nnx.merge`, `nnx.update`, `nnx.pop`, and `nnx.state` (previously called `extract`) are now the only way to use the Functional API, in particular there methods where removed from `Module` and `GraphDef`.
* Upgraded lifted transforms: they are now supersets of JAX transforms and accept nodes in any input and output (previously it was limited to only the first input).
* RNG state is now regular state and can be lifted using `nnx.fork`, the `Rngs.fork` method was removed.
* `graph_utils` was renamed to `graph`
* Many other small changes and fixes


ASEM000:
One minor suggestion

What do you think about renaming graph_utils to graph and graph flatten to flatten? 
To have 1) shorter names 2) avoid confusing users between garph_utils/graph_util 3) conform with jax new convention like renaming of tree_util.tree_flatten to tree.flatten

cgarciae:
Hey @ASEM000, you read my mind! Been planning to do this for a while. Will do this for this refactor.

levskaya (approved):

Issue #3879, opened 2024-04-23
No way to call nnx.State.from_flat_path


NeilGirdhar:
Consider:
```python
        filtered_flat: dict[PathParts, nnx.Variable[Any]] | Array] = {}
        for parameter_path, variable in weights_bar.flat_state().items():
            assert is_string_tuple(parameter_path)
            assert isinstance(variable, nnx.Variable)
            filtered_flat[parameter_path] = (variable
                                             if segment_filter.filter_(parameter_path)
                                             else jnp.zeros_like(variable.get_value()))
        return some_state.from_flat_path(filtered_flat)  # Type error
```
There is no nice and easy way to call `flax.nnx.State.from_flat_path` since its parameter type is `FlatState = dict[PathParts, StateLeaf]` since:
* `dict` is invariant in both key and value, so these types must match exactly, and
* `StateLeaf` is not exposed.

Suggestion:
* Make `FlatState` a `Mapping`, which is covariant in the value, or else
* Add a `FlatStateMapping` and use it in places where you accept a mapping.

For example `unflatten_dict` should just accept a `Mapping`—not a `dict`.

NeilGirdhar:
@cgarciae

cgarciae:
Hey @NeilGirdhar! This makes sense, I've added this change to #3876.

cgarciae:
Let me take a look.

NeilGirdhar:
Okay, I updated the issue comment with a rundown of all the the changes and their motivations.  The main issue is that nnx calls that function, so it should support Mappings as well, and I think it should have a type annotation to prevent other developers from changing this function in the future and breaking nnx.

NeilGirdhar:
@tkoeppe Did this pull request break something?  Was there a reason why this was backed out?  Without it, there are problems in nnx (see top of the issue).

PR #3880, opened 2024-04-23
Make FlatState a Mapping instead of a dict

NeilGirdhar:
Other changes:
* Since `nnx.State` calls `flatten_dict`, make the `traverse_util.flatten_dict` accept mappings.
* Add a type annotation to `flatten_dict` so that any future changes to the function (e.g., changing it back to work with dicts only) triggers a type error.
* Add the appropriate type overloads to the signature to satisfy MyPy errors.
* Minor tweaks to imports:
  * import from `collections.abc` instead of `typing` since the latter imports are deprecated.
  * Import from `flax.typing` instead of `flax.core.scope` since the latter has been moved.
* Add `from __future__ import annotations` so that the annotations work on Python 3.9.
* Add `pytype: skip-file` to work around google/pytype#1619.
* Annotate some private functions to make them easier to understand.
* When printing a type, print its `__qualname__` since that's a bit easier to read (`str` instead of `<class 'str'>`).

Fixes https://github.com/google/flax/issues/3879

### DIFF
--- a/flax/traverse_util.py
+++ b/flax/traverse_util.py
@@ -52,18 +52,19 @@ Traversals can also be used to make changes using the ``update`` method::
 Traversals never mutate the original data. Therefore, an update essentially
 returns a copy of the data including the provided updates.
 """
+from __future__ import annotations
 
 import abc
+from collections.abc import Callable, Mapping
 import copy
 import dataclasses
 import warnings
-from typing import Any, Callable
+from typing import Any, Union, overload
 
 import jax
 
 import flax
-from flax.core.scope import VariableDict
-from flax.typing import PathParts
+from flax.typing import PathParts, VariableDict
 
 from . import struct
 
@@ -77,7 +78,37 @@ class _EmptyNode:
 empty_node = _EmptyNode()
 
 
-def flatten_dict(xs, keep_empty_nodes=False, is_leaf=None, sep=None):
+# TODO: In Python 3.10, use TypeAlias.
+IsLeafCallable = Callable[[tuple[Any, ...], Mapping[Any, Any]], bool]
+
+
+@overload
+def flatten_dict(xs: Mapping[Any, Any],
+                 /,
+                 *,
+                 keep_empty_nodes: bool = False,
+                 is_leaf: Union[None, IsLeafCallable] = None,
+                 sep: None = None
+                 ) -> dict[tuple[Any, ...], Any]:
+  ...
+
+@overload
+def flatten_dict(xs: Mapping[Any, Any],
+                 /,
+                 *,
+                 keep_empty_nodes: bool = False,
+                 is_leaf: Union[None, IsLeafCallable] = None,
+                 sep: str,
+                 ) -> dict[str, Any]:
+  ...
+
+def flatten_dict(xs: Mapping[Any, Any],
+                 /,
+                 *,
+                 keep_empty_nodes: bool = False,
+                 is_leaf: Union[None, IsLeafCallable] = None,
+                 sep: Union[None, str] = None
+                 ) -> dict[Any, Any]:
   """Flatten a nested dictionary.
 
   The nested keys are flattened to a tuple.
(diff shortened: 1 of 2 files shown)


NeilGirdhar:
Note that this may change the behavior of calling `flatten_dict` on a structure with mappings that are neither `dict` nor `FrozenDict`.  In practice, that's probably extremely rare though?

cgarciae:
Let me take a look.

flax/traverse_util.py
```diff
@@ -77,7 +80,40 @@ class _EmptyNode:
...
+                 ) -> dict[str, Any]:
+    ...
+
+@overload
+def flatten_dict(xs: Mapping[Any, Any],
+                 keep_empty_nodes: bool = False,
+                 is_leaf: None | Callable[[tuple[Any, ...],
+                                           Mapping[Any, Any]], bool] = None,
+                 *,
+                 sep: str
```
cgarciae, line 107: why is this overload needed?

cgarciae (approved):

cgarciae:
Thanks @NeilGirdhar ! I've been wanting to do this for a while.
I've approved but left a comment.

flax/traverse_util.py
NeilGirdhar: I couldn't figure out how to support both:
`flatten_dict(xs, sep='/')` and `flatten_dict(xs, True, None, '/')`.  Do you know a way to do it one line?  Can I just get rid of the defaults when writing overloads?

NeilGirdhar: Another option would be to make keep_empty_nodes, is_leaf, and sep keyword-only parameters since it appears that they are always called this way.

NeilGirdhar:
@tkoeppe Did this pull request break something? Was there a reason why this was backed out? Without it, there are problems in nnx (see top of the issue).

tkoeppe:
@NeilGirdhar Yes, I think this broke something, but I don't have details. Please ask the owners about that. We just saw a bunch of breakages and reverted the change before as a matter of course, and that got propagated to GitHub automatically.

NeilGirdhar:
@tkoeppe  That's okay, but in the future, would you mind adding a message as to what broke?  Making this change to quite a bit of time.

@cgarciae Can  you look into this for me?  I imagine that the problem was the narrowing of the parameter list to positional-only/keyword-only.  I can revert that and we'll have to live with the more clumsy annotation.

tkoeppe:
@NeilGirdhar: I'm sorry, but I have nothing to do with this GitHub interaction. I'm maintaining unrelated, internal code and reverted a breaking change internally. The GitHub interaction is managed by those code's owners. I understand that the GitHub push happened automatically, but please take it up with the owners if you think they are making inappropriate or inadequate commits. If this is a problem, perhaps we need to have stronger review requirements or not automate these changes, but I'll leave that for them to decide.

cgarciae:
@NeilGirdhar I think you change is good but since this is used at many other code bases at Google it was very hard to try to fix it internally (I tried for a few hours). The main problem is that some users where passing some structures that behaved like Mappings that where now being traversed when before they where leaves.

I have a simpler suggestion, what if we create `nnx/traversals.py` and copy those over with the desired signature?

NeilGirdhar:
@tkoeppe I understand, and I don't want to create more work for you.  It's just that I was waiting for this change in the next release to fix my code, and I was waiting for nothing.  Just a note on the external pull request would help a lot 😄 and a copy of the error would help even more.

> I have a simpler suggestion, what if we create `nnx/traversals.py` and copy these over with the desired signature?

> Sounds great!  Would you like me to do that, or do you want to do it?

> The main problem is that some users where passing some structures that behaved like Mappings that where now being traversed when before they where leaves.

That should be fixable with:
```python
flatten_dict(..., is_leaf=lambda xs: not isinstance(xs, (flax.core.FrozenDict, dict)))
```

cgarciae:
> Sounds great! Would you like me to do that, or do you want to do it?

If you wanna do it that would be great! Happy to review it.

> That should be fixable with:

Maybe, sadly that was only some of the issues, the bigger issue is that this broke pytype in many places.

PR #3889, opened 2024-04-28
[nnx] fix iter_nodes

cgarciae:
# What does this PR do?

* `iter_nodes` now returns leaves first, this makes surgery less error prone.
* exposes `iter_nodes` as `nnx.iter_nodes`

### DIFF
--- a/flax/experimental/nnx/nnx/graph.py
+++ b/flax/experimental/nnx/nnx/graph.py
@@ -1072,11 +1072,11 @@ def _iter_nodes(
   if id(node) in visited:
     return
   visited.add(id(node))
-  yield path_parts, node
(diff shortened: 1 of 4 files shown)


flax/experimental/nnx/nnx/graph.py
```diff
@@ -1072,11 +1072,11 @@ def _iter_nodes(
   if id(node) in visited:
```
superbobry, line 1072: Please add a docstring to `iter_nodes` explaining the iteration order, since that function is now public.

superbobry (approved):
I would recommend adding a few more sentences to the PR description for posterity, explaining how this change makes surgery less error-prone.

PR #3921, opened 2024-05-13
[nnx] add compat

cgarciae:
# What does this PR do?

Adds `nnx.compat` module with the goal of making it possible to port Linen codebases to an NNX system with as few changes as possible. It would contain the following functionality:
* `Module`: inherits from `nnx.Module` and adds methods from `linen.Module`.
* `compact`: allows defining submodules inline
* `wrappers`: some types that simplify NNX <-> Linen interop.


flax/experimental/nnx/tests/compat/test_module.py
```diff
@@ -0,0 +1,134 @@
...
+          assert self.scope is not None
+          rngs = self.scope.rngs
+          self.w = nnx.Param(
+            jax.random.uniform(rngs(), (x.shape[-1], self.dout))
+          )
+          self.b = nnx.Param(jnp.zeros((self.dout,)))
+        return x @ self.w + self.b[None]
+
+    @dataclasses.dataclass
+    class Foo(nnx.compat.Module):
```
chiamp, line 44:
```suggestion
    class Foo(compat.Module):
```
Change all `nnx.compat` references to `compat` for consistency?

flax/experimental/nnx/nnx/compat/module.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Example::
+
+      >>> from flax.experimental import nnx
+      >>> from flax.experimental.nnx import compat as nnc
+      >>> import jax
+      >>> import jax.numpy as jnp
+      ...
+      >>> class Linear(nnc.Module):
+      ...   def __init__(self, dout, rngs: nnx.Rngs):
```
chiamp, line 150: For clarification:
* if we define an `__init__` method, we will be able to instantiate module parameters using the `.init` method?
* if we define a [`setup`](https://github.com/google/flax/pull/3921/files#diff-6b8d092686d8c9b95f179640fa4095d53562e065a929b93c0010d7631ab38f2dR29) method or wrap `__call__` with [`compact`](https://github.com/google/flax/pull/3921/files#diff-6b8d092686d8c9b95f179640fa4095d53562e065a929b93c0010d7631ab38f2dR47), we will be able to instantiate the module parameters by [calling the module](https://github.com/google/flax/pull/3921/files#diff-6b8d092686d8c9b95f179640fa4095d53562e065a929b93c0010d7631ab38f2dR59) on a sample input and invoking shape inference?
* the module parameters that are instantiated are bound to the module so they can be dot-accessed, which is different from Linen where they are returned separately as a variable dict?
* Instead of defining an `__init__` method, can we define a `setup` method or wrap `__call__` with `compact` to use the `.init` method as well?

chiamp (approved):

PR #3922, opened 2024-05-14
Support direct quantization for FP8 matmul

wenscarl:
Historically, FP8 matmul quantization followed the pattern of fake quantization, which involved a sequence of operations: quantization -> dequantization -> dot. Here, (de)quantization refers to type casting and the application of scaling factors. The XLA GemmWriter pass was designed to transform this pattern into a custom cublasLt call.

This PR proposes a departure from the historical approach by adopting direct quantization, which is quantization -> dot -> dequantization. This adjustment aligns better with mainstream quantization implementations for other data types. However, the success of this PR hinges on another PR in JAX ([PR-21211](https://github.com/google/jax/pull/21211)) because of the mixed fp8 type matmul. 
cc @lukaszlew @kaixih

### DIFF
--- a/flax/linen/fp8_ops.py
+++ b/flax/linen/fp8_ops.py
@@ -26,7 +26,6 @@ from jax import custom_jvp, custom_vjp, lax, random
 from jax import numpy as jnp
 from jax._src import core
 from jax._src import dtypes
-from jax._src.lax import lax
 from jax._src.typing import DTypeLike
 
 try:
--- a/flax/linen/fp8_ops.py
+++ b/flax/linen/fp8_ops.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import dataclasses
+import itertools
 import numpy as np
 import warnings
 from functools import partial
@@ -25,6 +26,8 @@ from jax import custom_jvp, custom_vjp, lax, random
 from jax import numpy as jnp
 from jax._src import core
 from jax._src import dtypes
+from jax._src.lax import lax
+from jax._src.typing import DTypeLike
 
 try:
   from jax._src import earray
@@ -163,39 +166,91 @@ def compute_amax_history(x, amax_history):
   return new_history
 
 
-def qdq_and_return(x, q_dtype, scale, amax_history, compute_dtype):
-  is_fmax32 = (scale.dtype == fp32_max_grad and
-               amax_history.dtype == fp32_max_grad)
-  # convert fmax32->f32 so we can do math
+def quantize_and_update(
+    x, q_dtype, scale, amax_history, compute_dtype, use_direct_quant=False
+):
+  is_fmax32 = (scale.dtype == fm32 and amax_history.dtype == fm32)
+  # convert fm32->f32 so we can do math
   if is_fmax32:
     amax_history = lax.convert_element_type(amax_history, jnp.float32)
     scale = lax.convert_element_type(scale, jnp.float32)
 
+  # Update the fp8 meta
   dtype_max = get_fp8_max(q_dtype, jnp.float32)
   amax_from_history = jnp.max(amax_history, axis=0)
+
   new_scale = compute_scale(amax_from_history, scale, dtype_max)
-
-  qx = quantize_dequantize(x, q_dtype, new_scale, compute_dtype)
-
   new_history = compute_amax_history(x, amax_history)
 
   # convert f32->fmax32 so the autodiff system accumulates fp8 meta correctly
   if is_fmax32:
     new_history = lax.convert_element_type(new_history, fp32_max_grad)
     new_scale = lax.convert_element_type(new_scale, fp32_max_grad)
+
+  # Quantize the input
+  if not use_direct_quant:
+    qx = quantize_dequantize(x, q_dtype, new_scale, compute_dtype)
+    return qx, new_scale, new_history
+
+  return new_scale, new_history
+
+
   return qx, new_scale, new_history
 
+def dot_general_transpose_lhs(g, x, y, *, dimension_numbers, precision,
+                              preferred_element_type: DTypeLike | None,
+                              swap_ans=False):
+  def _remaining(original, *removed_lists):
+    removed = set(itertools.chain(*removed_lists))
+    return [i for i in original if i not in removed]
+
+  def _ranges_like(*xs):
+    start = 0
+    for x in xs:
+      x_len = len(x)
+      yield range(start, start + x_len)
+      start += x_len
+
+  (x_contract, y_contract), (x_batch, y_batch) = dimension_numbers
+  x_ndim = x.aval.ndim
+  x_kept = _remaining(range(x_ndim), x_contract, x_batch)
+  y_kept = _remaining(range(np.ndim(y)), y_contract, y_batch)
+  if swap_ans:
+    ans_batch, ans_y, _ = _ranges_like(x_batch, y_kept, x_kept)
+  else:
+    ans_batch, _, ans_y = _ranges_like(x_batch, x_kept, y_kept)
+  dims = ((ans_y, y_kept), (ans_batch, y_batch))
+  x_contract_sorted_by_y = list(np.take(x_contract, np.argsort(y_contract)))
+  out_axes = np.argsort(list(x_batch) + x_kept + x_contract_sorted_by_y)
+  x_bar = lax.transpose(
+    lax.dot_general(
+      g, y, dims, precision=precision,
+      preferred_element_type=preferred_element_type
+    ),
+    tuple(out_axes)
+  )
+  return x_bar
+
+def dot_general_transpose_rhs(g, x, y, *, dimension_numbers, precision,
+                              preferred_element_type: DTypeLike | None):
+  (x_contract, y_contract), (x_batch, y_batch) = dimension_numbers
+  swapped_dimension_numbers = ((y_contract, x_contract), (y_batch, x_batch))
+  y_bar = dot_general_transpose_lhs(
+    g, y, x, dimension_numbers=swapped_dimension_numbers, precision=precision,
+    preferred_element_type=preferred_element_type,
+    swap_ans=True)
+  return y_bar
 
 @partial(custom_vjp, nondiff_argnums=(0, 1))
 def in_qdq(compute_dtype, q_dtype, inp, scale, amax_history):
-  qin, _, _ = qdq_and_return(
+  qin, _, _ = quantize_and_update(
     inp, q_dtype, scale, amax_history, compute_dtype
   )
   return qin
 
 
 def in_qdq_fwd(compute_dtype, q_dtype, inp, scale, amax_history):
-  qin, new_scale, new_history = qdq_and_return(
+  qin, new_scale, new_history = quantize_and_update(
     inp, q_dtype, scale, amax_history, compute_dtype
   )
   return qin, (new_scale, new_history)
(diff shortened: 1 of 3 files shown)


kaixih (commented):
Overall, the placement of quant and dequant is a bit confusing and the q and dq ops seems to be included in our custom dot_general function. I am trying to summarize the rationale here:
Basically, `q` means a pure quantize without amax logics but `xxx_q` includes both quantization and amax math.
```python
# Our original design:
x(in_qdq), k(in_qdq)->y
dy(out_qdq), x(in_qdq)->dk
dy(out_qdq), k(in_qdq)->dx

# New direct design:
x(in_q), k(in_q)->y(dq)
dy(out_q), x(in_q)->dk(dq??)
dy(out_q), k(in_q)->dx(dq??)
```
`??` indicates the problem about where to place these dq ops. In the original design, we don't need to worry about where the dk and dx are defined, because we don't apply any qdq there. However, in the new design, we need to find them and apply the dq ops explicitly and because we are using jvp mode (forward autograd mode), where we express the grad like:
```
dy = dx@k + x@dk
```
So, it seems we have to include the dq ops inside the dot_general function.

So, if that is the case, should we move all the q and dq into the dot_general function, esp in the jvp:
```
in_q(x)
in_q(y)
y = x@y
dq(y)
dq(dx)
dq(dk)
dy = dx@k + x@dk
in_q(dy)
```

flax/linen/fp8_ops.py
```diff
@@ -142,7 +141,7 @@ def qdq_and_return(x, q_dtype, scale, amax_history, compute_dtype):
   amax_from_history = jnp.max(amax_history, axis=0)
   new_scale = compute_scale(amax_from_history, scale, dtype_max)
 
-  qx = quantize_dequantize(x, q_dtype, new_scale, compute_dtype)
```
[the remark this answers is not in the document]
wenscarl: It's used in the test file. Removed.

kaixih:
I think the new design is much clearer of the idea of direct quantization. By the way, do you think we should create a new Fp8DotGeneral op for it and keep the existing fake quant Op untouched? And then we gradually change downstream uses to migrate to the new op?

wenscarl:
>I think the new design is much clearer of the idea of direct quantization. By the way, do you think we should create a new Fp8DotGeneral op for it and keep the existing fake quant Op untouched? And then we gradually change downstream uses to migrate to the new op?

Praxis doesn't use `Fp8DotGeneralOp` directly but `dot_general_with_precision`. Given that most PAXML models access fp8 from praxis. So instead, I think it makes more sense to keep both fake and direct quant Op [there](https://github.com/google/praxis/pull/69).

flax/linen/fp8_ops.py
```diff
@@ -130,141 +126,251 @@ def compute_amax_history(x, amax_history):
...
     amax_history = lax.convert_element_type(amax_history, jnp.float32)
     scale = lax.convert_element_type(scale, jnp.float32)
 
-  dtype_max = get_fp8_max(q_dtype, jnp.float32)
-  amax_from_history = jnp.max(amax_history, axis=0)
-  new_scale = compute_scale(amax_from_history, scale, dtype_max)
-
-  qx = quantize_dequantize(x, q_dtype, new_scale, compute_dtype)
-
+  new_scale = compute_scale(q_dtype, scale, amax_history)
```
kaixih, line 143: Overall, the changes look good to me. One more thing about the backward compatibility when migrating to the direct quantization. Do you think we should propose a new op `Fp8DirectDotGeneralOp`, like:
```python
class Fp8DirectDotGeneralOp(Fp8DotGeneralOp):
  def __call__(self, *args, **kwargs):
    ...
```

flax/linen/fp8_ops.py
```diff
@@ -277,19 +383,18 @@ def __call__(self, *args, **kwargs):
...
-    k_qdq = in_qdq(
-      comp_dtype, k, self.kernel_scale.value, self.kernel_amax_history.value
-    )
-    y_qdq = dot_general_with_precision(x_qdq, k_qdq, dimension_numbers)  # type: ignore
-    y = out_qdq(
-      comp_dtype,
-      y_qdq,
-      self.output_grad_scale.value,
-      self.output_grad_amax_history.value,
+    y = _q_dot_dq(
```
kaixih, line 386: Let's call it `q_dot_dq` because I guess we would like to use it in other modules like in praxis fp8 einsum.

tests/linen/linen_test.py
```diff
@@ -1248,8 +1248,12 @@ class Fp8Test(absltest.TestCase):
   def test_fp8_dot_general_injection(self):
     # Used to cast the inputs to be representable in FP8, so that the difference
     # of the results from the original gemm and fp8 gemm is small.
+    def quantize_dequantize(x, q_dtype, scale, compute_dtype):
```
kaixih, line 1251: How parameterizing the test_fp8_dot_general_injection to be like:
```python
def test_fp8_dot_general_injection(self, use_direct_quant):
  if use_direct_quant:
    # compare fp32 vs fp8 (from direct quant)
  else:
    # compare fp32 vs fp8 (from fake quant)
```

tests/linen/linen_test.py
```diff
@@ -1265,7 +1267,10 @@ def test_fp8_dot_general_injection(self):
     def run(fp8_injection, expected_shapes):
       p = nn.DenseGeneral(features=64, name='dense')
       if fp8_injection:
-        p.dot_general_cls = nn.Fp8DotGeneralOp
+        p.dot_general_cls = (
+          nn.Fp8DirectDotGeneralOp
+          if use_direct_quant else nn.Fp8DotGeneralOp
```
kaixih, line 1272: Can we do: 
```python
quant_cls = nn.Fp8DirectDotGeneralOp if use_direct_quant else nn.Fp8DotGeneralOp
...
p.dot_general_cls =quant_cls
...
   f'{quant_cls.__name__}_0': {
...
```

wenscarl:
A gentle reminder to @lukaszlew

lukaszlew:
Sorry, I don't have cycles to review this PR. I'm focusing on AQT.

zhangqiaorjc:
Could we get @levskaya to help review since Lukasz is busy with something else these days?

wenscarl:
@levskaya could you take a look?

flax/linen/fp8_ops.py
```diff
@@ -230,6 +240,208 @@ def out_qdq_bwd(compute_dtype, q_dtype, res, g):
...
+    jnp.float8_e5m2,
+    out_grad_scale,
+    out_grad_amax_history,
+    compute_dtype,
+    use_direct_quant=True
+  )
+
+  q_g = quantize(g, jnp.float8_e5m2, new_out_grad_scale, preferred_element_type)
+
+  grad_lhs = _dot_general_transpose_lhs(
```
levskaya, line 406: The JAX team really doesn't like us depending on their internal implementations.  Could we inline this function logic here to make this free-standing?

flax/linen/fp8_ops.py
```diff
@@ -230,6 +240,208 @@ def out_qdq_bwd(compute_dtype, q_dtype, res, g):
...
+    q_rhs,
+    dimension_numbers=dimension_numbers,
+    precision=lax.Precision.HIGHEST,
+    preferred_element_type=preferred_element_type,
+)
+  grad_lhs = dequantize(
+    grad_lhs, preferred_element_type, new_rhs_scale * new_out_grad_scale
+  )
+
+  grad_rhs = _dot_general_transpose_rhs(
```
levskaya, line 418: same comment as above

flax/linen/fp8_ops.py
```diff
@@ -25,6 +25,7 @@
 from jax import numpy as jnp
 from jax._src import core
 from jax._src import dtypes
+from jax._src.lax.lax import _dot_general_transpose_lhs, _dot_general_transpose_rhs
```
levskaya, line 28: better to inline - see below.

levskaya (commented):
Apologies for the long delay - I was visiting rural family who don't have an internet connection when I was pinged here.

flax/linen/fp8_ops.py
kaixih: Right, that 's also our main concern back then. Do you mean we should reimplement the logic of the two _xxx functions here?

levskaya: yeah, they're fairly small functions and you don't need all the generality of them - it's just that JAX may need to change things in the future and we don't want to add external dependencies on their internals.

levskaya:
Sorry we block if trailing spaces are left in the file, there's some after the line `class Fp8Test(parameterized.TestCase):` in the tests (you can see the failed precommit) - could you fix and we can run the tests.   I think things look ok.

flax/linen/fp8_ops.py
```diff
@@ -230,6 +284,205 @@ def out_qdq_bwd(compute_dtype, q_dtype, res, g):
...
+    lhs,
+    q_rhs,
+    dimension_numbers=dimension_numbers,
+    precision=lax.Precision.HIGHEST,
+    preferred_element_type=preferred_element_type,
+  )
+  grad_lhs = dequantize(grad_lhs, preferred_element_type, new_lhs_scale * new_out_grad_scale)
+
+  (x_contract, y_contract), (x_batch, y_batch) = dimension_numbers
+  grad_rhs = dot_general_transpose(
```
kaixih, line 461: I feel maybe it is clearer to follow the existing design of dot_general_transpose_[lhs|rhs] and then the logic of swapping the dims and setting swap_ans can be hidden in them.

flax/linen/fp8_ops.py
```diff
@@ -163,39 +165,91 @@ def compute_amax_history(x, amax_history):
...
+    ans_batch, ans_y, _ = _ranges_like(x_batch, y_kept, x_kept)
+  else:
+    ans_batch, _, ans_y = _ranges_like(x_batch, x_kept, y_kept)
+
+  dims = ((ans_y, y_kept), (ans_batch, y_batch))
+  x_contract_sorted_by_y = list(np.take(x_contract, np.argsort(y_contract)))
+  out_axes = np.argsort(list(x_batch) + x_kept + x_contract_sorted_by_y)
+
+  # Perform dot_general and transpose operations
+  result = jax._src.lax.lax.transpose(
```
kaixih, line 231: Can we `from jax._src import lax`?

flax/linen/fp8_ops.py
```diff
@@ -163,39 +165,91 @@ def compute_amax_history(x, amax_history):
...
+    jax._src.lax.lax.dot_general(g, y, dims, precision=precision,
+                                 preferred_element_type=preferred_element_type),
+    tuple(out_axes)
+  )
+
+  # Convert element type if necessary
+  if result.dtype != x.aval.dtype:
```
kaixih, line 238: I don't think this branch is necessary since our use case is guaranteed to be fp8 in and fp8 out.

kaixih:
@levskaya I think all the issues have been resolved by @wenscarl . Can you help review and merge?

kaixih:
@levskaya Just resolved some formatting issues. I think all the tests should pass now. Can you help review and merge?

kaixih (approved):

kaixih (approved):

levskaya (approved):

flax/linen/fp8_ops.py
```diff
@@ -25,6 +26,8 @@
 from jax import numpy as jnp
 from jax._src import core
 from jax._src import dtypes
+from jax._src.lax import lax
```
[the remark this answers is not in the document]
levskaya: Yeah we actually already import the `lax` symbol just a few lines above, so this is a redundant import line.  I think we can just delete it.

levskaya:
@wenscarl @kaixih - hey also, I invited you as collaborators to this repo, apologies that I hadn't done that earlier, it should immediately trigger presubmit tests once you join.

levskaya:
@wenscarl - I'm seeing a failed test? 
```
E       UserWarning: The function dot_general_with_precision will set the precision/preferred_element_type and disregard any provided values.
```

wenscarl:
@levskaya Thanks for reviewing. All checks are passed.

PR #3925, opened 2024-05-14
[nnx] fix bugs

cgarciae:
# What does this PR do?

* Fixes a bug in `nnx.state` that returned an additional argument when using 2 or more filters.
* Correctly estimates the `length` when using `scan`'s in_axis` argument.


chiamp (approved):
thanks!

PR #3928, opened 2024-05-19
Make FlatState a Mapping instead of a dict

### LINKED ISSUE #3879: No way to call nnx.State.from_flat_path

NeilGirdhar:
* Add nnx.traversals.{flatten_mapping, unflatten_dict}
    * These are modified from traverse_util.{flatten_dict, unflatten_dict}
    * They are annotated so that any future changes to the function (e.g., changing it back to work with dicts only) triggers a type error.

* Minor tweaks to imports:
    * import from collections.abc instead of typing since the latter imports are deprecated.
    * Import from flax.typing instead of flax.core.scope since the latter has been moved.
    * Add from `__future__` import annotations so that the annotations work on Python 3.9.

* Minor tweaks to code:
    * Annotate some private functions to make them easier to understand.
    * When printing a type, print its __qualname__ since that's a bit easier to read (str instead of <class 'str'>).

Fixes #3879

### DIFF
--- a/flax/nnx/__init__.py
+++ b/flax/nnx/__init__.py
@@ -23,6 +23,7 @@ from .nnx import graph as graph
 from .nnx import errors as errors
 from .nnx import helpers as helpers
 from .nnx import compat as compat
+from .nnx import traversals as traversals
 from .nnx.filterlib import All as All
 from .nnx.filterlib import Not as Not
 from .nnx.graph import GraphDef as GraphDef
--- a/flax/nnx/examples/lm1b/models_test.py
+++ b/flax/nnx/examples/lm1b/models_test.py
@@ -26,7 +26,6 @@ import jax.numpy as jnp
 from absl.testing import absltest
 from jax import random
 
-from flax import traverse_util
 from flax import nnx
 from configs import default
 from models import TransformerConfig, TransformerLM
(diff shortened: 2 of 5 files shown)


NeilGirdhar:
@cgarciae (Done, as you requested.)

cgarciae:
@NeilGirdhar looks great! Thanks for doing this.
Approved but left a comment.

cgarciae:
Seems there are some git conflicts.

NeilGirdhar:
Rebased to main.

cgarciae (approved):

PR #3929, opened 2024-05-21
[NNX] Add `LoRA` and `LoRALinear` to NNX

IvyZX:
Provided here are two ways to add LoRA on any layer:

* `nnx.LoRA` with `base_module` arg that can take any layer instance. Easy for model surgery on modules, but will render mismatching `state`s.
* `nnx.LoRALinear` which subclasses `nnx.Linear` and attach a simple `nnx.LoRA` module along the way. Param structure matches with `nnx.Linear`, but need a bit more surgery on runtime. This technique can be applied for any other NNX modules.

Not yet sure which approach is the best, so providing both at this moment.

### DIFF
--- a/flax/experimental/nnx/nnx/nn/lora.py
+++ b/flax/experimental/nnx/nnx/nn/lora.py
@@ -1,0 +1,190 @@
+# Copyright 2024 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Copyright 2023 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+import typing as tp
+
+import jax
+import jax.numpy as jnp
+
+from flax.experimental import nnx
+from flax.experimental.nnx.nnx import rnglib, variables
+from flax.experimental.nnx.nnx.module import Module
+from flax.experimental.nnx.nnx.nn import initializers
+from flax.typing import (
+  Dtype,
+  Initializer,
+)
+
+Array = jax.Array
+Axis = int
+Size = int
+A = tp.TypeVar('A')
+
+default_kernel_init = initializers.lecun_normal()
+
+
+class LoRAParam(variables.Variable[A]): pass
+
+
+class LoRA(Module):
+  """A standalone LoRA layer.
+
+  Example usage::
+
+    >>> from flax.experimental import nnx
+    >>> import jax, jax.numpy as jnp
+    >>> layer = nnx.LoRA(3, 2, 4, rngs=nnx.Rngs(0))
+    >>> layer.lora_a.value.shape
+    (3, 2)
+    >>> layer.lora_b.value.shape
+    (2, 4)
+    >>> # Wrap around existing layer
+    >>> linear = nnx.Linear(3, 4, rngs=nnx.Rngs(0))
+    >>> wrapper = nnx.LoRA(3, 2, 4, base_module=linear, rngs=nnx.Rngs(1))
+    >>> assert wrapper.base_module == linear
+    >>> wrapper.lora_a.value.shape
+    (3, 2)
+    >>> layer.lora_b.value.shape
+    (2, 4)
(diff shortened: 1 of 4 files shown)


flax/experimental/nnx/nnx/nn/lora.py
```diff
@@ -0,0 +1,192 @@
...
+      kernel_init(rngs.params(), (in_features, lora_rank), param_dtype)
+    )
+    self.lora_b = param_type(
+      kernel_init(rngs.params(), (lora_rank, out_features), param_dtype)
+    )
+
+  def __call__(self, x: jax.Array):
+    out = x @ self.lora_a @ self.lora_b
+    if self.base_module is not None:
+      assert callable(self.base_module), "`base_module` must be callable."
```
cgarciae, line 123: Lets raise an error here instead

flax/experimental/nnx/nnx/nn/lora.py
```diff
@@ -0,0 +1,192 @@
...
+      out += self.base_module(x)
+    return out
+
+
+class LoRALinear(nnx.Linear):
```
cgarciae, line 128: It might cleaner to inherit from `LoRA` and create a `Linear` in `__init__` which is passed as the `base_module` to `super().__init__()`. This way you can remove `__call__`.

flax/experimental/nnx/nnx/nn/lora.py
```diff
@@ -0,0 +1,192 @@
...
+      for details.
+    kernel_init: initializer function for the weight matrices.
+    use_lora_param_type: if yes, LoRA params will be of different Param type.
+      This makes it easier to split model state for trainable/nontrainables.
+  """
+  def __init__(
+    self,
+    in_features: int,
+    out_features: int,
+    lora_rank: int,
```
cgarciae, line 96: Maybe this ordering is a bit more intuitive?
```suggestion
    in_features: int,
    lora_rank: int,
    out_features: int,
```

cgarciae (changes requested):

flax/experimental/nnx/nnx/nn/lora.py
```diff
@@ -0,0 +1,192 @@
...
+    self.dtype = dtype
+    self.param_dtype = param_dtype
+    self.lora_param_type = lora_param_type
+    self.base_module = base_module
+
+    param_type = LoRAParam if self.lora_param_type else nnx.Param
+    self.lora_a = param_type(
+      kernel_init(rngs.params(), (in_features, lora_rank), param_dtype)
+    )
+    self.lora_b = param_type(
```
chiamp, line 116: I'm wondering if there are more informative variable names we can use like `self.lora_in` and `self.lora_out`, or is `a` and `b` standard naming conventions for LoRA?

flax/experimental/nnx/nnx/nn/lora.py
```diff
@@ -0,0 +1,192 @@
...
+    in_features: the number of input features.
+    out_features: the number of output features.
+    lora_rank: the rank of the LoRA dimension.
+    base_module: a base module to call and substitute, if possible.
+    dtype: the dtype of the computation (default: infer from input and params).
+    param_dtype: the dtype passed to parameter initializers (default: float32).
+    precision: numerical precision of the computation see `jax.lax.Precision`
+      for details.
+    kernel_init: initializer function for the weight matrices.
+    use_lora_param_type: if yes, LoRA params will be of different Param type.
```
chiamp, line 89:
```suggestion
    use_lora_param_type: if True, LoRA params will be of different Param type.
```

flax/experimental/nnx/nnx/nn/lora.py
```diff
@@ -0,0 +1,192 @@
...
```
chiamp, line 161:
```suggestion
    use_lora_param_type: if ``True``, LoRA params will be of different Param type.
```

flax/experimental/nnx/nnx/nn/lora.py
IvyZX: Done.

flax/experimental/nnx/nnx/nn/lora.py
IvyZX: The reason I provided `LoRALinear` is that this assumes the exact same param structure and argument as `Linear`. This is also how [PyTorch](https://pytorch.org/torchtune/stable/_modules/torchtune/modules/peft/lora.html) provides their `LoRALinear`.

If it's just a `LoRA` instance, the original linear weights will be one level below, inside `base_module`. I think users can easily create something like this on their own, no need to have a `LoRALinear` shortcut for it.

flax/experimental/nnx/nnx/nn/lora.py
IvyZX: Sure! LoRA is a generic layer and this is indeed more intuitive.

Note that the [PyTorch LoRALinear](https://pytorch.org/torchtune/stable/_modules/torchtune/modules/peft/lora.html) implementation is ordered `(in, rank, out)`. To avoid this being confusing, I will move `lora_rank` to a kwarg argument.

flax/experimental/nnx/nnx/nn/lora.py
IvyZX: Yeah unfortunately AFAIK `a` and `b` are the conventions here...

flax/experimental/nnx/nnx/nn/lora.py
IvyZX: I am thinking maybe it's better to make this argument an explicit type instead of a boolean... more customizability!

flax/experimental/nnx/nnx/nn/lora.py
IvyZX: Same here

flax/experimental/nnx/nnx/nn/lora.py
cgarciae: Makes sense.

cgarciae (approved):

PR #3982, opened 2024-06-11
Fixes for batch norm docs

jkarwowski:
Changes:
 - Fixed reference to `params` in `eval_step`
 - Fixed the docstring in `eval_step`
 - Added `.mean()` to losses

The documentation for batch norm is written in such a way, that directly copying from it results in an unpleasant bug where the `params`  used in `eval_step` refer to the initial parameters of the network. This causes metrics to show that the model is not generalising to the test set at all. This change hopefully saves someone else learning flax a similar scratch on the head.

## Checklist
      checks if that's the case).


cgarciae (approved):
Thanks @jkarwowski for doing this!

PR #3991, opened 2024-06-12
Added `nnx.graph.iter_children`

chiamp:
Added `nnx.graph.iter_children`, which returns an iterator of only immediate child `Module`'s (contrasted with `nnx.graph.iter_modules` that returns an iterator of ALL nested `Module`'s). Also updated docstring and tests for both methods.
Maybe we can use more informative names like `iter_all_modules` and `iter_child_modules`?

### DIFF
--- a/flax/nnx/nnx/module.py
+++ b/flax/nnx/nnx/module.py
@@ -28,7 +28,7 @@ from flax.nnx.nnx.graph import GraphDef
 from flax.nnx.nnx.object import Object, ObjectMeta
 from flax.nnx.nnx.graph import GraphState, StateLeaf
 from flax.nnx.nnx.state import State
-from flax.typing import Path, PathParts
+from flax.typing import Key, Path, PathParts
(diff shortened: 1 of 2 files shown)


flax/nnx/nnx/module.py
```diff
@@ -194,26 +195,73 @@ def iter_modules(self) -> tp.Iterator[tuple[PathParts, Module]]:
...
+      >>> model = Block(2, 5, rngs=nnx.Rngs(0))
+      >>> for path, module in model.iter_children():
+      ...  print(path, type(module).__name__)
+      ...
+      batch_norm BatchNorm
+      dropout Dropout
+      linear Linear
+      submodule SubModule
+    """
+    node_dict = graph.get_node_impl(self).node_dict(self)
```
cgarciae, line 260: Nice! This is more general.

cgarciae (approved):

PR #3992, opened 2024-06-12
added self.param to nnx.compat

chiamp:
added self.param to nnx.compat


flax/nnx/nnx/compat/module.py
```diff
@@ -204,6 +206,81 @@ def __init_subclass__(cls, experimental_pytree: bool = False) -> None:
...
+    if self.is_initializing():
+      assert hasattr(
+        self, 'rngs'
+      ), 'The `param` method implicitly calls `self.rngs.params()`, and so the Module must have an `nnx.Rng` in `self.rngs`.'
+      assert not hasattr(
+        self, name
+      ), f'Tried to create a parameter in the `{name}` attribute, but `{name}` is already used.'
+      value = init_fn(self.rngs.params(), *init_args, **init_kwargs)
+      setattr(self, name, Param(value))
+    return getattr(self, name)
```
cgarciae, line 282: We should return `getattr(self, name).value` here

flax/nnx/nnx/compat/module.py
```diff
@@ -204,6 +206,81 @@ def __init_subclass__(cls, experimental_pytree: bool = False) -> None:
...
+        value=Array([[ 0.4154572 , -0.28943744,  0.1879725 ]], dtype=float32)
+      )
+    })
+
+    >>> y = model(x) # model can now be called normally after parameters are initialized
+    """
+    if self.is_initializing():
+      assert hasattr(
+        self, 'rngs'
+      ), 'The `param` method implicitly calls `self.rngs.params()`, and so the Module must have an `nnx.Rng` in `self.rngs`.'
```
cgarciae, line 276: Here we want to use `self.scope.rngs` so we have to check that `self.scope is not None`.

cgarciae: We should also raise a proper error instead of an assert

PR #4005, opened 2024-06-18
Add NNX surgery guide

IvyZX:
Add a guide for common model surgery scenarios and how to use NNX state and modules to fluently surgery around.

### DIFF
--- a/docs/nnx/surgery.md
+++ b/docs/nnx/surgery.md
@@ -1,0 +1,284 @@
+# Model surgery
+
+This guide will demostrate how to do model surgery in NNX with a few real-scenario use cases.
+
+* __Module manipulation__: Pythonic ways to manipulate submodules given a model.
+
+* __Abstact model__: A key trick to play with NNX modules and states without memory allocation.
+
+* __From raw state to model__: How to manipulate parameter state when they are incompatible with existing model code.
+
+* __Partial initialization__: Initializing only part of the model from scratch.
+
+
+```python
+from typing import *
+from pprint import pprint
+import functools
+
+import jax
+from jax import lax, numpy as jnp, tree_util as jtu
+
+from jax.sharding import PartitionSpec, Mesh, NamedSharding
+from jax.experimental import mesh_utils
+import flax
+from flax import nnx
+import flax.traverse_util
+import numpy as np
+import orbax.checkpoint as orbax
+
+key = jax.random.key(0)
+```
+
+
+```python
+class TwoLayerMLP(nnx.Module):
+  def __init__(self, dim, rngs: nnx.Rngs):
+    self.linear1 = nnx.Linear(dim, dim, rngs=rngs)
+    self.linear2 = nnx.Linear(dim, dim, rngs=rngs)
+
+  def __call__(self, x):
+    x = self.linear1(x)
+    return self.linear2(x)
+```
+
+## Pythonic module manipulations
+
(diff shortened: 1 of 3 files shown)


docs/nnx/surgery.md
```diff
@@ -0,0 +1,261 @@
...
+  new_state = nnx.state(get_model())
+  return nnx.State.merge(new_state, old_state)
+
+good_model = nnx.eval_shape(get_lora_model)
+# Note that `old_state` will be deleted after this `partial_init` call.
+new_state = partial_init(old_state, get_lora_model)
+nnx.update(good_model, new_state)
+
+# They are equivalent. The way RNGs are propagated in the new model is also preserved.
+np.testing.assert_allclose(simple_model(x), good_model(x))
```
cgarciae, line 260: I'm a bit confused by this section, it seems to be doing the same operation in two different ways. I was thinking we could show case the use of `nnx.jit` here and use something like:

```python
old_state = nnx.state(TwoLayerMLP(4, rngs=nnx.Rngs(0)))

@functools.partial(nnx.jit, donate_argnums=0, static_argnums=1)
def partial_init(old_state, rngs):
  model = TwoLayerMLP(4, rngs=rngs)
  # add existing create
  nnx.update(model, old_state)
  # create new state
  model.linear1 = nnx.LoRALinear(4, 4, lora_rank=3, rngs=rngs))
  return model
  
  model = partial_init(old_state, nnx.Rngs(0))
```

IvyZX: Hmm, for some reason the code you proposed didn't really save memory.

I added some lines to explain this section and also showed how to print the memory cost at each step. You can see that the `jax.jit` cases skipped the creation of extra unused params in `LoRALinear`. I don't know why your `nnx.jit` version doesn't repro this...

You can run this piece of code in a standalone python file to verify what I've seen. If `nnx.jit` can do this I'm happy to switch to `nnx.jit` instead.

cgarciae: My bad, I put `nnx.update` in the middle but it should be at the end else the `model.linear1 =` assignment overrides the updated state. Here is the fixed code:

```python
  class TwoLayerMLP(nnx.Module):
      def __init__(self, dim, rngs: nnx.Rngs):
        self.linear1 = nnx.Linear(dim, dim, rngs=rngs)
        self.linear2 = nnx.Linear(dim, dim, rngs=rngs)

      def __call__(self, x):
        x = self.linear1(x)
        return self.linear2(x)

    old_state = nnx.state(TwoLayerMLP(4, rngs=nnx.Rngs(0)))

    print(f'Arrays start: {len(jax.live_arrays())}')
    print(len(jax.tree.leaves(old_state)))

    @partial(nnx.jit, donate_argnums=0, static_argnums=1)
    def partial_init(old_state, rngs):
      model = TwoLayerMLP(4, rngs=rngs)
      # create new state
      model.linear1 = nnx.LoRALinear(4, 4, lora_rank=3, rngs=rngs)
      # add existing create
      nnx.update(model, old_state)
      return model

    model = partial_init(old_state, nnx.Rngs(0))

    print(f'Arrays end: {len(jax.live_arrays())}')
    print(len(jax.tree.leaves(nnx.state(model))))
```

IvyZX: Thanks! Added your code and replaced the `jax.jit` code. Thank you!

cgarciae (approved):

PR #4006, opened 2024-06-18
Fix head pytest breakage

IvyZX:
* Removed trivial dependency of `tensorflow` on `jax_utils_test.py`
* Ignored JAX-side deprecation warning introduced by Tensorflow
* No longer skipping `jax.tree_.*` deprecation warnings because we are free of them!
* Pinned `orbax-checkpoint` to `0.5.15` because 16 and 17 introduces new bugs


chiamp (approved):
Thanks!

PR #4021, opened 2024-06-24
[nnx] add Filters guide

cgarciae:
# What does this PR do?

* Adds `filters_guide` to the docs.
* Improves current Filters by making them hashable and equatable.

### DIFF
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -56,8 +56,12 @@ class FlaxError(Exception):
     )
     module_name = self.__class__.__module__
     class_name = self.__class__.__name__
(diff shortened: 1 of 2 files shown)


IvyZX:
This is a nice show of the APIs, but I feel it's a bit too minimal and unclear for users to apply it to their use cases.

Some things might worth adding:
* How is filter usually used? Maybe have a small model and training loop and show how you can customize trainable params with the API easily.
*  Various ways to use `nnx.split()` args. At least showcase when there's no filter predicates, and when there's a function as filter predicate, etc.
* Add API reference links to the functions / types mentioned in the doc.

cgarciae:
Continued in #4028

PR #4022, opened 2024-06-24
[nnx] add jit donation test

cgarciae:
# What does this PR do?

Adds a test using `nnx.jit` + `donate_argnums` to showcase partial initialization.

### DIFF
--- a/flax/nnx/tests/transforms_test.py
+++ b/flax/nnx/tests/transforms_test.py
@@ -345,6 +345,33 @@ class TestJIT(absltest.TestCase):
 
     m.kernel.value.sharding
 
+  def test_donation(self):
+    class TwoLayerMLP(nnx.Module):
+      def __init__(self, dim, rngs: nnx.Rngs):
+        self.linear1 = nnx.Linear(dim, dim, rngs=rngs)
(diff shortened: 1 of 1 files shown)


flax/nnx/tests/transforms_test.py
```diff
@@ -345,6 +345,33 @@ def constrain_object(m):
...
+    def partial_init(old_state, rngs):
+      model = TwoLayerMLP(4, rngs=rngs)
+      # create new state
+      model.linear1 = nnx.LoRALinear(4, 4, lora_rank=3, rngs=rngs)
+      # add existing state
+      nnx.update(model, old_state)
+      return model
+
+    model = partial_init(old_state, nnx.Rngs(0))
+
```
IvyZX, line 372: Can you also add check to verify that `old_state`'s arrays are deleted now?

IvyZX (changes requested):
Thanks for adding this test!

flax/nnx/tests/transforms_test.py
cgarciae: What is the correct way to check they are deleted? One way would be to try to use them and assert there is an error, is this what you are thinking?

PR #4030, opened 2024-06-26
[nnx] stabilize unsafe_pytree

cgarciae:
# What does this PR do?

* Renames `experimental_pytree` to `unsafe_pytree`, making it a stable feature.
* Adds the "Modules as Pytrees" section to the NNX Basics guide.


ASEM000:
Maybe I am missing something, is there is a discussion why is it _unsafe_? My guess would be because of lost refs,  but I am not sure in the context of nnx.

docs/nnx/nnx_basics.md
```diff
@@ -376,3 +376,52 @@ model = nnx.merge(graphdef, params, counts)
...
+model = Block(3, 5, rngs=nnx.Rngs(0))
+
+@jax.jit  # regular jax.jit!
+def forward(model: Block, x: jax.Array):
+  y = model(x)
+  return y, model  # manually propagate state updates
+
+y, model = forward(model, jnp.ones((1, 3)))
+```
+
```
cgarciae, line 406: @ASEM000 here is the explanation, indeed its because of sharing. The `unsafe` naming hopefully gives users a visual queue that they shouldn't use this feature lightly.

docs/nnx/nnx_basics.md
```diff
@@ -376,3 +376,52 @@ model = nnx.merge(graphdef, params, counts)
...
+shared = nnx.Linear(3, 5, rngs=nnx.Rngs(0))
+ma, mb = Foo(shared), Foo(shared)
+
+print(f'Before: {ma.shared is mb.shared = }')
+
+# flatten + unflatten
+ma, mb = jax.tree.map(lambda x: x, (ma, mb))
+
+print(f'After:  {ma.shared is mb.shared = }')
+```
```
cgarciae, line 427: @ASEM000 here is the explanation, indeed its because of sharing. The `unsafe` naming hopefully gives users a visual queue that they shouldn't use this feature lightly.

ASEM000: It is more clear now, thanks.

flax/nnx/nnx/compat/module.py
```diff
@@ -199,8 +199,8 @@ def is_initializing(self) -> bool:
 
     return self._object__state._initializing
 
-  def __init_subclass__(cls, experimental_pytree: bool = False) -> None:
-    super().__init_subclass__(experimental_pytree)
+  def __init_subclass__(cls, unsafe_pytree: bool = False) -> None:
```
IvyZX, line 202: Can we add a link or docstring around here to explain this feature so that it's available somewhere in the API reference? I think this feature warrants explicit explanation (since its caveat is a bit hidden).

docs/nnx/nnx_basics.md
```diff
@@ -376,3 +376,52 @@ model = nnx.merge(graphdef, params, counts)
...
+  return y, model  # manually propagate state updates
+
+y, model = forward(model, jnp.ones((1, 3)))
+```
+
+**WARNING**: The reason the features is called `unsafe` is because NNX's 
+reference semantics are broken by JAX's referential transparency, this 
+is specially problematic when there is shared state between NNX graph nodes 
+as reference identity is lost. Use `unsafe_pytree` only when there' i's only 
+a single top-level Module or there is no shared state between top-level Modules.
```
IvyZX, line 411: > when there' i's only a single top-level Module *or* there is no shared state between top-level Modules.

Should this be an "or" or an "and"? 
Also nit: "when there is only a single..."

cgarciae: Thanks! It's an "or". Changed the wording a bit.

IvyZX (approved):

PR #4047, opened 2024-07-02
Fix typo in `nnx_basics` doc

### LINKED ISSUE #4046: typo in nnx_basics.md
Replace 
Since Modules hold their own state methods can be called directly without the no need for a separate apply method
by
Since Modules hold their own state methods can be called directly without the need for a separate apply method

rajasekharporeddy:
Fixed typo from `without the no need for` to `without the need for`.

Fixes #4046


gnecula (approved):
Thanks!

cgarciae (approved):

cgarciae:
Thanks @rajasekharporeddy for doing this!

PR #4066, opened 2024-07-09
Rename `nnx.compat` to `nnx.bridge`

IvyZX:
To avoid any future confusion of `compat` vs `compact`.


flax/nnx/__init__.py
```diff
@@ -18,11 +18,12 @@
...
 from flax.typing import Initializer as Initializer
 
-from .nnx.compat import wrappers as wrappers
+from .nnx.bridge import wrappers as wrappers
 from .nnx import graph as graph
 from .nnx import errors as errors
 from .nnx import helpers as helpers
-from .nnx import compat as compat
+from .nnx import bridge as bridge
+compat = bridge  # Make the old import name still work
```
cgarciae, line 26: we should consider just removing this to take advantage of the fact that we don't have users yet

cgarciae (approved):

flax/nnx/__init__.py
IvyZX: Good idea - done in the final merged version.

PR #4076, opened 2024-07-11
[NVIDIA] Add a user guide for fp8

kaixih:
This PR adds a user guide on how to utilize FP8 GEMM calls for GPUs. It primarily focuses on the functions provided in [this file](https://github.com/google/flax/blob/main/flax/linen/fp8_ops.py) and explains how to enable the FP8 feature in Flax layers.

cc. @nouiz @nluehr


IvyZX (approved):
Thank you for making this guide!

PR #4078, opened 2024-07-12
[nnx] add extract APIs

cgarciae:
# What does this PR do?

The `extract` APIs helps with manipulating pytrees that contain graph nodes, it provides useful operations like extracting graph nodes from a pytree and replacing them with some pytree structures called Extractables. This API is mainly used in transforms to process the Modules in the input arguments.


IvyZX (approved):
Thanks! Please fix the CI errors.

PR #4081, opened 2024-07-13
Add shape-based lazy init to `LinenToNNX` (prev `LinenWrapper`)

IvyZX:
* Renamed wrappers to `LinenToNNX` and `NNXToLinen` to minimize confusion.
* Moved state initialization of `LinenToNNX` to `__call__` to realize lazy init. This allows it to be a submodule of an NNX module, which doesn't have input args during initialization.
  * User can use `nnx.shaped_init` to do a dry run of `__call__` and initialize the whole state & full graphdef.
* Made state initialization of `LinenToNNX` nested & closer to NNX, aka. each `VariableState` is created for every jax Array, not every collection.

### DIFF
--- a/flax/nnx/nnx/bridge/wrappers.py
+++ b/flax/nnx/nnx/bridge/wrappers.py
@@ -18,10 +18,14 @@ from typing import Any
 
 from flax import nnx
 from flax import linen
+from flax.nnx.nnx import graph
 from flax.nnx.nnx import variables as variableslib
 from flax.nnx.nnx.module import GraphDef, Module
 from flax.nnx.nnx.rnglib import Rngs
 from flax.nnx.nnx.state import State
+from flax.nnx.nnx.object import Object
+import jax
+from jax import tree_util as jtu
 
 M = tp.TypeVar('M', bound=Module)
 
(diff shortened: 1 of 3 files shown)


flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -55,43 +56,52 @@ def _functional_constructor(*args: tp.Any, **kwargs: tp.Any) -> Functional[M]:
...
+    if not hasattr(self, 'states'):
+      _rngs = (
+        {name: stream.key.raw_value for name, stream in rngs.items()}
+        if rngs
+        else {}
+      )
+      # rename default to params
+      if 'params' not in _rngs and 'default' in _rngs:
+        _rngs['params'] = _rngs['default']
+        del _rngs['default']
```
cgarciae, line 91:
```suggestion
        _rngs['params'] = _rngs.pop('default')
```

cgarciae: Now that I think about it, we have to make `Rngs` implement `MutableMapping` for either of these to work.

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -55,43 +56,52 @@ def _functional_constructor(*args: tp.Any, **kwargs: tp.Any) -> Functional[M]:
   return _functional_constructor
 
 
-class LinenWrapper(Module):
+def shaped_init(module: Module, *args, **kwargs):
+  """To trigger init of all `LinenToNNX` module variables and return a wholesome state."""
+  assert callable(module)
+  _ = module(*args, **kwargs)
+  return nnx.split(module)
```
cgarciae, line 63: In this function we should leverage the fact that `Object._object__state._initializing` still exists and set if via something like

```python
def _set_initializing(initializing: bool):
  for _, value in graph.iter_graph(module):
    if isinstance(value, Object):
      value._object__state._initializing = initializing
```
and use the value of `_initializing` to choose between `init` and `apply` when calling the Linen Modules.

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -55,43 +56,52 @@ def _functional_constructor(*args: tp.Any, **kwargs: tp.Any) -> Functional[M]:
...
-      {name: stream.key.value for name, stream in rngs.items()} if rngs else {}
-    )
 
-    variables = {
-      collection: value.value for collection, value in self.states.items()
-    }
+    # Shape-based lazy init of the flax variables
+    if not rngs:
+      rngs = self.rngs
+    if not hasattr(self, 'states'):
```
cgarciae, line 82: Use `self._object__state.initializing` instead, see above.
```suggestion
    if self._object__state.initializing:
```

flax/nnx/nnx/bridge/wrappers.py
IvyZX: The `_rngs` here is a dict instead of `Rngs` class, so this already works.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Added `_set_initializing` to the LinenToNNX wrapper. 
Note that we can't do check on top level modules' `._object__state._initializing` because the top level module might be a pure NNX module with `._object__state._initializing` always False.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Done.

flax/nnx/nnx/bridge/wrappers.py
cgarciae: I'd do something like this:

```python
def _set_initializing(module, initializing: bool): 
  for _, value in graph.iter_graph(module): 
    if isinstance(value, Object): 
      value._object__state._initializing = initializing
      
def shaped_init(module: Module, *args, **kwargs):
  """To trigger init of all `LinenToNNX` module variables and return a wholesome state."""
  module = graph.clone(module) # create a copy
  _set_initializing(module, True)
  assert callable(module)
  try:
    _ = module(*args, **kwargs)
  finally:
    _set_initializing(module, False)
  return nnx.split(module)
```

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -55,43 +58,58 @@ def _functional_constructor(*args: tp.Any, **kwargs: tp.Any) -> Functional[M]:
...
 
-    variables = {
-      collection: value.value for collection, value in self.states.items()
-    }
+    # Shape-based lazy init of the flax variables
+    if not rngs:
+      rngs = self.rngs
+    if self._object__state.initializing:
+      _rngs = (
+        {name: stream.key.raw_value for name, stream in rngs.items()}
```
cgarciae, line 92: We need to generate new keys so Linen Modules get new RNG state every time.
```suggestion
        {name: stream() for name, stream in rngs.items()}
```

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -55,43 +58,58 @@ def _functional_constructor(*args: tp.Any, **kwargs: tp.Any) -> Functional[M]:
...
+        if rngs
+        else {}
+      )
+      # rename default to params
+      if 'params' not in _rngs and 'default' in _rngs:
+        _rngs['params'] = _rngs.pop('default')
+
+      variables = self.module.init(_rngs, *args, **kwargs)
```
cgarciae, line 100: we could use `init_with_output` to avoid calling forward twice

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Done, also renamed to `lazy_init` as discussed offline.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Done.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Done.

cgarciae (approved):

PR #4102, opened 2024-07-23
[linen] add share_scope

cgarciae:
# What does this PR do?

Adds`nn.share_scope` can be used to share a scope between two Modules. This means that any parameters created by any of the Modules will be added to their common scope. This is useful when you want to wrap a Module and extend its functionality without changing the parameter structure.

```python
import flax.linen as nn
import jax
from jax import numpy as jnp, random

class DenseLoRA(nn.Module):
  base: nn.Dense
  rank: int

  def setup(self):
    nn.share_scope(self, self.base)

  @nn.compact
    din, dout = x.shape[-1], self.base.features
    A = self.param('A', nn.zeros_init(), (din, self.rank))
    B = self.param('B', nn.zeros_init(), (self.rank, dout))
    return self.base(x) + x @ A @ B

class Model(nn.Module):
  @nn.compact
    base = nn.Dense(10)  # base scope
    return DenseLoRA(base, rank=2)(x)  # reuse base scope

model = Model()

params = model.init(random.key(0), jnp.ones((1, 5)))['params']
list(params['Dense_0'].keys())
# ['A', 'B', 'kernel', 'bias']
```

### DIFF
--- a/docs/api_reference/flax.linen/module.rst
+++ b/docs/api_reference/flax.linen/module.rst
@@ -11,3 +11,4 @@ Module
 .. autofunction:: init
 .. autofunction:: init_with_output
 .. autofunction:: intercept_methods
+.. autofunction:: share_scope
--- a/flax/errors.py
+++ b/flax/errors.py
@@ -691,6 +691,24 @@ class CallUnbindOnUnboundModuleError(FlaxError):
   def __init__(self):
     super().__init__("Can't call `unbind()` on unbound modules")
 
+class CallShareScopeOnUnboundModuleError(FlaxError):
+  """This error occurs when you are trying to call ``nn.share_scope`` on an unbound
+  Module. For instance, when you try to use ``nn.share_scope`` at the top-level::
+
+    from flax import linen as nn
+
+    class CustomDense(nn.Dense):
+      def __call__(self, x):
+        return super().__call__(x) + 1
+
+    custom_dense = CustomDense(5)
+    dense = nn.Dense(5)  # has the parameters
+
+    nn.share_scope(custom_dense, dense) # <-- ERROR!
+  """
+
+  def __init__(self):
+    super().__init__("Can't call `share_scope` on unbound modules")
 
 class InvalidInstanceModuleError(FlaxError):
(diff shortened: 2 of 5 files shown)


flax/linen/combinators.py
```diff
@@ -112,3 +115,90 @@ def __call__(self, *args, **kwargs):
...
+  else:
+    # Child has its own independent scope, overwrite
+    # parent scope, so that we preserve the sharing
+    object.__setattr__(parent, 'scope', child.scope)
+
+
+class Transparent(Module, Generic[M]):
+  """A Module that shares its scope with an inner Module. This combinator is useful
+  when you want to wrap a Module and extend itsfunctionality without changing the
+  parameter structure.
```
marksandler2, line 132: I would also highlight the difference between 

class DenseLora(Dense): 
   ...

vs

class DenseLora(Transparent[Dense]):
   ...

(The important distinction is that Transparent replaces inheritance with composition, that allows to swap out implementation of inner module with Dense implementation (as long as it has compatible interface), without having to create separate DenseLora for each implementation.

cgarciae:
@marksandler2 wondering if we should just expose the `_assign_shared_scope` as e.g. `nn.share_scope` an have users manually do the call instead of introducing this abstraction? The LoRA example would look like this:

```python
class DenseLoRA(nn.Module):
  inner: nn.Dense
  rank: int

  def setup(self):
    nn.share_scope(self, self.inner)

  @nn.compact
    din, dout = x.shape[-1], self.inner.features
    A = self.param('A', nn.zeros_init(), (din, self.rank))
    B = self.param('B', nn.zeros_init(), (self.rank, dout))
    return self.inner(x) + x @ A @ B

class Model(nn.Module):
  @nn.compact
    return DenseLoRA(nn.Dense(10), rank=2)(x)
```

marksandler2:
Hmm, i like this new proposed function a lot actually. It certainly feels a lot less heavy weight (e.g. we don't need to muck around with Transparent[...] inheritance, and the logic could be made a lot more flexible.

marksandler2:
Can we merge this?

superbobry (approved):

PR #4136, opened 2024-08-21
use uv package manager

cgarciae:
# What does this PR do?

* Adds `uv.lock`.
* Moves `docs/requirements.txt` requirements into `pyproject.toml` under `project.optional-dependencies.docs`.
* Updates `contributing.md` to use `uv`.
* Removed the `--no-*` flags from `run_all_tests.sh` in favor of `--only-*` flags which makes is much easier to use.


docs/contributing.md
```diff
@@ -39,36 +39,28 @@ To contribute code to Flax on GitHub, follow these steps:
...
    with [PyPi](https://pip.pypa.io/en/stable/cli/pip_install/). This enables you to immediately
    test the code after modifying it:
 
    ```bash
    git clone https://github.com/YOUR_USERNAME/flax
    cd flax
-   pip install -e ".[all]"
-   pip install -e ".[testing]"
-   pip install -r docs/requirements.txt
+   uv sync --all-extras
```
levskaya, line 56: how does uv interact with virtual envs?

levskaya: I think maybe we should keep a few lines here about installing with pip for those who don't want to `curl | sh` a new binary.

levskaya (approved):

docs/contributing.md
levskaya: so there's a `uv venv` - we should _definitely_ keep encouraging people to use virtual-envs in our docs.

levskaya (approved):

PR #4145, opened 2024-08-27
Support linen <-> nnx metadata box converging in `nnx.bridge`

IvyZX:
Supports conversion between Linen and NNX metadata boxes.
 * NNX always has a metadata box (aka `nnx.Variable`), so make sure `ToNNX` always make them.
 * `ToLinen` will always convert `nnx.Variable` to a boxed Linen variable. By default, it would be `nnx.bridge.NNXMeta`, a class that captures everything that an `nnx.Variable` should.
 * Already existing boxes of `nn.Partitioned` should be preserved, and its axes names will be translated to `nnx.Variable` sharding fields inside `ToNNX`.

Also refactored variable-related functions that only `bridge` API uses outside of `nnx/variables.py`, into a separate `nnx/bridge/variables.py`.

### DIFF
--- a/flax/nnx/nnx/bridge/variables.py
+++ b/flax/nnx/nnx/bridge/variables.py
@@ -1,0 +1,138 @@
+# Copyright 2024 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Any, TypeVar
+
+import jax
+from flax import struct
+from flax.core import meta
+from flax.nnx.nnx import variables as variableslib
+import typing as tp
+
+
+A = TypeVar('A')
+B = TypeVar('B')
+
+
+#######################################################
+### Variable type <-> Linen collection name mapping ###
+#######################################################
+# Assumption: the mapping is 1-1 and unique.
+
+VariableTypeCache: dict[str, tp.Type[variableslib.Variable[tp.Any]]] = {}
+
+
+def variable_type(name: str) -> tp.Type[variableslib.Variable[tp.Any]]:
+  """Given a Linen-style collection name, get or create its corresponding NNX Variable type."""
+  if name not in VariableTypeCache:
+    VariableTypeCache[name] = type(name, (variableslib.Variable,), {})
+  return VariableTypeCache[name]
+
+
+def variable_type_name(typ: tp.Type[variableslib.Variable[tp.Any]]) -> str:
+  """Given an NNX Variable type, get or create its Linen-style collection name.
+
+  Should output the exact inversed result of `variable_type()`."""
+  for name, t in VariableTypeCache.items():
+    if typ == t:
+      return name
+  name = typ.__name__
+  if name in VariableTypeCache:
+    raise ValueError(
+      'Name {name} is already registered in the registry as {VariableTypeCache[name]}. '
+      'It cannot be linked with this type {typ}.'
+    )
+  register_variable_name_type_pair(name, typ)
+  return name
+
+
+def register_variable_name_type_pair(name, typ, overwrite = False):
+  """Register a pair of variable type name (like Linen collections) and its NNX type."""
+  if not overwrite and name in VariableTypeCache:
+    raise ValueError(f'Name {name} already mapped to type {VariableTypeCache[name]}. '
+                     'To overwrite, call with `overwrite=True`.')
+  VariableTypeCache[name] = typ
+
+
+# add known variable type names
+register_variable_name_type_pair('params', variableslib.Param)
+register_variable_name_type_pair('batch_stats', variableslib.BatchStat)
+register_variable_name_type_pair('cache', variableslib.Cache)
+register_variable_name_type_pair('intermediates', variableslib.Intermediate)
+
+
+def sort_variable_types(types: tp.Iterable[type]):
+  def _variable_parents_count(t: type):
+    return sum(1 for p in t.mro() if issubclass(p, variableslib.Variable))
+  parent_count = {t: _variable_parents_count(t) for t in types}
+  return sorted(types, key=lambda t: -parent_count[t])
+
+
+#############################################
+### NNX Variable <-> Linen metadata boxes ###
+#############################################
+
+
+class NNXMeta(struct.PyTreeNode, meta.AxisMetadata[A]):
+  """Default Flax metadata class for `nnx.VariableState`.
+  """
+
+  var_type: type[variableslib.Variable[tp.Any]] = struct.field(pytree_node=False)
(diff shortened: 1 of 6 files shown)


flax/nnx/nnx/bridge/variables.py
```diff
@@ -0,0 +1,131 @@
...
+      'Name {name} is already registered in the registry as {VariableTypeCache[name]}. '
+      'It cannot be linked with this type {typ}.'
+    )
+  register_variable_name_type_pair(name, typ)
+  return name
+
+
+def register_variable_name_type_pair(name, typ):
+  """Register a pair of variable type name (like Linen collections) and its NNX type."""
+  VariableTypeCache[name] = typ
```
cgarciae, line 62: we should check that the `name` does exist already

flax/nnx/nnx/bridge/variables.py
```diff
@@ -0,0 +1,131 @@
...
+  """
+
+  var_type: type[variableslib.Variable[tp.Any]] = struct.field(pytree_node=False)
+  value: Any = struct.field(pytree_node=True)
+  metadata: dict[str, tp.Any] = struct.field(pytree_node=False)
+
+  def unbox(self) -> A:
+    return self.value
+
+  def replace_boxed(self, val: B) -> 'NNXMeta[B]':
```
cgarciae, line 95: String types not needed in python >= 3.10
```suggestion
  def replace_boxed(self, val: B) -> NNXMeta[B]:
```

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -143,16 +144,17 @@ def __call__(
...
-      def nn_var_to_nnx_state(kp, v):
-        assert isinstance(kp[0], jtu.DictKey)
-        vtype = variableslib.variable_type(kp[0].key)
-        return vtype(v)
-      for col, tree in jtu.tree_map_with_path(nn_var_to_nnx_state, variables).items():
+
+      nnx_vars = jtu.tree_map_with_path(bv.to_nnx_var, variables,
+                                        is_leaf=lambda x: isinstance(x, meta.AxisMetadata))
+      for col, tree in nnx_vars.items():
         self._setattr(col, tree)
```
cgarciae, line 151: A bit more idiomatic:
```suggestion
        setattr(self, col, tree)
```

flax/nnx/nnx/bridge/variables.py
```diff
@@ -0,0 +1,131 @@
...
+def to_linen_var(vs: variableslib.VariableState) -> meta.AxisMetadata:
+  metadata = vs.get_metadata()
+  if 'linen_meta_type' in metadata:
+    if metadata['linen_meta_type'] is not meta.Partitioned:
+      raise ValueError('Not supporting Linen metadata types other than nn.Partitioned')
+    return meta.Partitioned(vs.value, names=metadata['sharding'], mesh=metadata['mesh'])
+  return NNXMeta(vs.type, vs.value, vs.get_metadata())
+
+
+def to_nnx_var(kp: tp.Sequence[Any], x: meta.AxisMetadata | Any) -> variableslib.Variable:
```
cgarciae, line 116: NIT: We could avoid using key paths here by accepting the collection and iterating over all collections when used.
```suggestion
def to_nnx_var(col: str, x: meta.AxisMetadata | Any) -> variableslib.Variable:
```
Its a bit more verbose but easier to understand.

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -162,8 +164,10 @@ def nn_var_to_nnx_state(kp, v):
     # Split out the updates if `mutable` is passed into the Flax module
     if kwargs.get('mutable', False) != False:
       out, updates = out
+      updates = jtu.tree_map_with_path(bv.to_nnx_var, updates,
+                                       is_leaf=lambda x: isinstance(x, meta.AxisMetadata))
       for collection, value in updates.items():
-        self._setattr(collection, jax.tree.map(variableslib.variable_type(collection), value))
+        self._setattr(collection, value)
```
cgarciae, line 170:
```suggestion
        setattr(self, collection, value)
```

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -143,16 +144,17 @@ def __call__(
...
-        assert isinstance(kp[0], jtu.DictKey)
-        vtype = variableslib.variable_type(kp[0].key)
-        return vtype(v)
-      for col, tree in jtu.tree_map_with_path(nn_var_to_nnx_state, variables).items():
+
+      nnx_vars = jtu.tree_map_with_path(bv.to_nnx_var, variables,
+                                        is_leaf=lambda x: isinstance(x, meta.AxisMetadata))
+      for col, tree in nnx_vars.items():
         self._setattr(col, tree)
         self.linen_collections.add(col)
```
cgarciae, line 152: For `linen_collections` we should either implement this with a `list` or create a `PytreeSet`. The issue is that sets are not pytrees so they are treated as static but they are not hashable.

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -225,14 +230,16 @@ def update_variables(self, module):
     types = set(jax.tree.leaves(
       jax.tree.map(lambda x: x.type, state,
                     is_leaf=lambda x: isinstance(x, nnx.VariableState))))
-    types = variableslib.sort_variable_types(types)
+    types = bv.sort_variable_types(types)
```
cgarciae, line 233: `sort_variable_types` expects type `list` but we can change the type to `Iterable`

cgarciae (approved):

flax/nnx/nnx/bridge/variables.py
IvyZX: I think we should allow overwrite of names if user wants to. I'll add an overwrite option there.

flax/nnx/nnx/bridge/variables.py
IvyZX: I got error trying to do this, saying `NNXMeta is not found`. I'll leave it as is for now.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Done.

flax/nnx/nnx/bridge/variables.py
IvyZX: Done. We need the key path to get the collection name, so I just moved that part to wrapper.py, to make the function in variables.py cleaner.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Done.

flax/nnx/nnx/bridge/wrappers.py
IvyZX: `list` is mutable so not hashable either. Curious why that matters though... but I'll make it a tuple.

flax/nnx/nnx/bridge/wrappers.py
```diff
@@ -225,14 +230,16 @@ def update_variables(self, module):
...
     _, *state_by_types = nnx.split(module, *types)
```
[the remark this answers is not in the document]
IvyZX: Thanks but `nnx.split` is more convenient because it always return a tuple...

flax/nnx/nnx/bridge/wrappers.py
IvyZX: Thanks for the catch! Done.

PR #4154, opened 2024-08-29
[nnx] add Pytree

cgarciae:
# What does this PR do?

`nnx.Pytree` is a proxy object that wraps Module and implements the pytree protocol. It makes it easier to use models with APIs that accept pytrees comprate to `split` / `merge` as `Pytree` can call any method and access any attribute from the underlaying object it wraps. 

```python
module = nnx.Pytree(nnx.Linear(2, 3, rngs=nnx.Rngs(0)))
module = jax.tree.map(lambda x: x, module)
assert module.in_features == 2
assert module.out_features == 3
assert module.kernel.shape == (2, 3)

y = module(jnp.ones((1, 2)))
assert y.shape == (1, 3)
```
`Pytree` respect referential transparency, meaning that it wont share state with Module(s) it wraps, internally the input is `split` and the `GraphDef` and `State` are stored, only when accessing attributes the underlying object is materialized using `merge` and cached for later use. When methods are called the `GraphDef` and `State` are updated in place. `Pytree` can be used as a context manager to get access to the underlying object when needed for manual modifications, upon termination the pytree is updated with the new state.

```python
@dataclasses.dataclass
class Counter(nnx.Module):
  count: nnx.BatchStat[int]

pt_counter = nnx.Pytree(Counter(nnx.BatchStat(0)))
assert pt_counter.count.value == 0

with pt_counter as counter:
  counter.count += 1

assert pt_counter.count.value == 1
```


docs/nnx/nnx_basics.md
```diff
@@ -252,16 +254,86 @@ print(f'{y.shape = }')
 nnx.display(model)
 ```
 
-How do NNX transforms achieve this? To understand how NNX objects interact with
-JAX transforms lets take a look at the Functional API.
+## Using Modules as Pytrees
```
IvyZX, line 257: I just realized pytree is a JAX concept, not a wider-spread python concept. 
Let's add this link when we talk about pytree: https://jax.readthedocs.io/en/latest/pytrees.html

docs/nnx/nnx_basics.md
```diff
@@ -252,16 +254,86 @@ print(f'{y.shape = }')
...
+
+y = jnp.ones((3,))
+ts = jnp.linspace(0, 1, 5)
+ts = jnp.linspace(0, 1, 5)
+
+ys = odeint(step_fn, y, ts, model)
+
+print(ys)
+```
+
```
IvyZX, line 313: Add a line to introduce this piece of code, something like: Let's see another example mixing NNX and Equinox modules, as Equinox modules are pure pytrees.

flax/nnx/nnx/pytree.py
```diff
@@ -0,0 +1,213 @@
...
+    return pytree
+
+  @property
+  def graphdef(self) -> graph.GraphDef[A]:
+    return self._nnx__graphdef
+
+  @property
+  def state(self):
+    return self._nnx_state
+
```
IvyZX, line 147: Can we have a util to convert it back to `nnx.Module`, like:
```python
def to_module() -> nnx.Module:
  return nnx.merge(self.graphdef, self.state)
```

flax/nnx/tests/graph_utils_test.py
```diff
@@ -425,61 +425,6 @@ def f_pure(graphdef: nnx.graph.GraphDef[Foo], state):
     assert m2 is m
     assert m2.ref is m2
 
-  def test_call_jit_update(self):
```
IvyZX, line 428: Why do we delete these tests?

cgarciae: We would delete `nnx.call` if we add `Pytree`, these test were moved to `pytree_test.py`.

PR #4155, opened 2024-08-29
[nnx] use jax-style transforms API in nnx_basics

cgarciae:
# What does this PR do?

Updates `nnx_basics` to use the new JAX-like transforms syntax in the Transforms section. Also adds some additional notes about `nnx.scan`.


docs/nnx/nnx_basics.md
```diff
@@ -225,22 +225,24 @@ input (scan over layers).
```
IvyZX, line 225: I think we'd want to explain a bit about scan here. What about changing this line to:

> Next lets take a look at a different example, which uses `nnx.vmap` to create a stack of multiple `MLP` layers and `nnx.scan` to iteratively apply each layer of the stack to the input.

IvyZX (approved):

docs/nnx/nnx_basics.md
cgarciae: Thanks!

docs/nnx/nnx_basics.md
```diff
@@ -225,22 +225,24 @@ input (scan over layers).
 
 Notice the following:
 1. The `create_model` function creates a (single) `MLP` object that is lifted by
-  `nnx.vmap` to have an additional dimension of size `axis_size`.
+  `nnx.vmap` to have an additional dimension.
 2. The `forward` function indexes the `MLP` object's state to get a different set of
   parameters at each step.
-3. `nnx.scan` automatically propagates the state updates for the `BatchNorm` and 
-`Dropout` layers from within `forward` to the `model` reference outside.
+3. The `nnx.scan` transform consciously deviates from its JAX equivalent in order to mimick
```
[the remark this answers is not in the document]
cgarciae: Good idea. I will add add the reference on their first appearance above.

PR #4194, opened 2024-09-14
Align bridge variable tree structures

IvyZX:
Fixing bridge API in a bunch of ways:

1. `ToNNX` will convert the whole variables structure to NNX style. If your underlying Linen module has variable `foo` at collection `bar`, its `ToNNX` version will have an attribute `foo` with type `bar`, instead of an attribute `bar` with a dict `{'foo': ...}`.
This means you can freely put `ToNNX` in the top or middle or back of the whole model layer stack, and the weight pytree structure shouldn't change.
Same goes for `ToLinen` - if your top-level type is Linen, the whole variable tree shall be Linen-style.

2. If you have a vanilla `nnx.Variable` with no sharding metadata, hooks, etc, `ToLinen` will not convert it into an `NNXMeta`, but instead just keep the vanilla JAX array inside. This makes it more intuitive and pytree-structure-proof for any Linen users not using partitioning metadata.

3. `nn.get_partition_spec` now works on `NNXMeta` wrappers, and any other wrapper that has `get_partition_spec` method.

4. Updated the `nnx.bridge` guide accordingly.


flax/nnx/bridge/wrappers.py
```diff
@@ -12,12 +12,14 @@
...
 # limitations under the License.
 
+from collections import defaultdict
 import dataclasses
 import typing as tp
 from typing import Any
 
 from flax import nnx
 from flax import linen
+from flax import traverse_util
```
cgarciae, line 22: NIT: `flax.nnx.traversals` has similar same APIs but with more accurate typing.

flax/nnx/bridge/wrappers.py
```diff
@@ -85,6 +87,50 @@ def lazy_init(fn: Module | tp.Callable[..., tp.Any], *args, **kwargs):
   return fn
 
 
+def _recursive_merge(dict1, dict2):
```
cgarciae, line 90: I think this could be expressed using `traversals` / `traverse_util` as:
```python
flat_map = traversal.flatten_mapping(dict1)
flat_map |= traversal.flatten_mapping(dict2)
return traversal.unflatten_mapping(flat_map)
```
Not sure if there are edge cases though but it seems easy.

flax/nnx/bridge/wrappers.py
```diff
@@ -85,6 +87,50 @@ def lazy_init(fn: Module | tp.Callable[..., tp.Any], *args, **kwargs):
...
+  """Recursively merge two dicts."""
+  for key, value in dict2.items():
+    if key in dict1 and isinstance(dict1[key], dict) and isinstance(value, dict):
+      # Recursively merge nested dictionaries
+      dict1[key] = _recursive_merge(dict1[key], value)
+    else:
+      # Merge non-dictionary values
+      dict1[key] = value
```
cgarciae, line 98: Is mutating `dict1` always safe?

flax/nnx/bridge/wrappers.py
```diff
@@ -85,6 +87,50 @@ def lazy_init(fn: Module | tp.Callable[..., tp.Any], *args, **kwargs):
...
+        nnx_attrs[attr_name] = _recursive_merge(nnx_attrs[attr_name], value)
+      else:
+        nnx_attrs[attr_name] = value     # it's a variable on this layer
+  return nnx_attrs
+
+
+def nnx_attrs_to_linen_vars(nnx_attrs: dict) -> dict:
+  linen_structured = {}
+  for kp, v in traverse_util.flatten_dict(
+      nnx_attrs, is_leaf=lambda _, x: isinstance(x, nnx.Variable | nnx.GraphDef)).items():
```
cgarciae, line 121: I didn't know Unions were now valid in `isinstance`, nice!

cgarciae (approved):

flax/nnx/bridge/wrappers.py
IvyZX: Nice! Adopted.

flax/nnx/bridge/wrappers.py
IvyZX: Looks good! Adopted.

flax/nnx/bridge/wrappers.py
IvyZX: I think using `nnx.traversals` is indeed easier!

PR #4197, opened 2024-09-16
[nnx] add transforms guide

cgarciae:
# What does this PR do?

Adds the `Transforms` guide.

[Preview](https://flax-nnx--4197.org.readthedocs.build/en/4197/guides/transforms.html)

### DIFF
--- a/docs_nnx/guides/transforms.md
+++ b/docs_nnx/guides/transforms.md
@@ -1,0 +1,367 @@
+---
+jupytext:
+  cell_metadata_filter: -all
+  formats: ipynb,md:myst
+  main_language: python
+  text_representation:
+    extension: .md
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.13.8
+---
+
+# Transforms
+JAX transformations in general operate on [Pytrees](https://jax.readthedocs.io/en/latest/pytrees.html) of arrays
+and abide by value semantics, this presents a challenge for Flax NNX which represents Modules as regular Python objects
+that follow reference semantics. To address this, Flax NNX introduces its own set of transformations that extend JAX
+transformations to allow Modules and other Flax NNX objects to be passed in and out of transformations while preserving
+reference semantics.
+
+Flax NNX transformations should feel quite familar to those who have used JAX transformations before as they use the
+same APIs and behave like the JAX transformations when only working with Pytrees of arrays. However, when working with
+Flax NNX objects, they allow Python's reference semantics to be preserved for these objects, this includes:
+* Preserving shared references across multiple objects in the inputs and outputs of the transformation.
+* Propagating any state changes made to the objects inside the transformation to the objects outside the transformation.
+* Enforcing consistency of how objects are transformed when aliases are present across multiple inputs and outputs.
+
+```{code-cell} ipython3
+import jax
+from jax import numpy as jnp, random
+from flax import nnx
+```
+
+Throughout this guide we will use `nnx.vmap` as a case study to demonstrate how Flax NNX transformations work but the principles
+outlined here extend to all transformations.
+
+## Basic Example
+To begin, let's look at a simple example of using `nnx.vmap` to extend an elementwise `vector_dot` function to work on
+batched inputs. We will define a `Weights` Module with no methods to hold some parameters, these weights will be passed
+as an input to the `vector_dot` function along with some data. Both the weights and data will be batched on axis `0` and we will use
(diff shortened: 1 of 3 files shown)


docs_nnx/guides/transforms.md
```diff
@@ -0,0 +1,340 @@
...
+
+y = nnx.vmap(crazy_vector_dot, in_axes=0, out_axes=1)(weights, x)
+
+nnx.display(weights)
+```
+
+> With great power comes great responsibility.
+> <br> \- Uncle Ben
+
+While this feature is very powerful, it must be used with care as it can clash with JAX's underlying assumptions for certain transformations. For example, `jit` expects the structure of the inputs to be stable in order to cache the compiled function, changing the graph structure inside a `nnx.jit`-ed function cause continuous recompilations and performance degradation, `scan` on the other hand only allows a fixed `carry` structure, so adding/removing substates declared as carry will cause an error.
```
IvyZX, line 155: nit:
> For example, `jit` expects the structure of the inputs to be stable in order to cache the compiled function, changing the graph structure inside a `nnx.jit`-ed function cause continuous recompilations and performance degradation, `scan` on the other hand only allows a fixed `carry` structure, so adding/removing substates declared as carry will cause an error.

For example, `jit` expects the structure of the inputs to be stable in order to cache the compiled function, so changing the graph structure inside a `nnx.jit`-ed function cause continuous recompilations and performance degradation. On the other hand, `scan` only allows a fixed `carry` structure, so adding/removing substates declared as carry will cause an error.

docs_nnx/guides/transforms.md
```diff
@@ -0,0 +1,340 @@
...
+    self.count = Count(count)
+
+weights = Weights(
+  kernel=random.uniform(random.key(0), (10, 2, 3)),
+  bias=jnp.zeros((10, 3)),
+  count=jnp.array(0),
+)
+x = jax.random.normal(random.key(1), (10, 2))
+
+
```
IvyZX, line 188: It's probably better to only call vmap once when only one call is needed, to avoid confusion. Same for the example below.

```
state_axes = nnx.StateAxes({nnx.Param: 0, Count: None}) # broadcast Count
@nnx.vmap(in_axes=(state_axes, 0), out_axes=1)
def stateful_vector_dot(weights: Weights, x: jax.Array):
  assert weights.kernel.ndim == 2, 'Batch dimensions not allowed'
  assert x.ndim == 1, 'Batch dimensions not allowed'
  weights.count += 1
  return x @ weights.kernel + weights.bias

y = stateful_vector_dot(weights, x)
y = stateful_vector_dot(weights, x)
```

docs_nnx/guides/transforms.md
```diff
@@ -0,0 +1,340 @@
...
+weights.count
+```
+
+Here count is now a scalar since its not being vectorized. Also, note that `StateAxes` can only be used directly on Flax NNX objects, it cannot be used as a prefix for a pytree of objects.
+
++++
+
+### Random State
+In Flax NNX random state is just regular state. This means that its stored inside Modules that need it and its treated as any other type of state. This is a simplification over Flax Linen where random state was handled by a separate mechanism. In practice Modules usually keep that need random state simply need a references to a `Rngs` object that is passed to them during initialization, and
```
IvyZX, line 207: > In practice Modules usually keep that need random state simply need a references to a Rngs object that is passed to them during initialization, and use it to generate a unique key for each random operation.

What about:
In practice Modules simply need to keep a reference to a Rngs object that is passed to them during initialization, and use it to generate a unique key for each random operation.

IvyZX (approved):
Thanks for making this guide! Super helpful and cool. Just a few nits on wordings.

docs_nnx/guides/transforms.md
```diff
@@ -0,0 +1,340 @@
...
+    self.count = Count(count)
+
+weights = Weights(
+  kernel=random.uniform(random.key(0), (10, 2, 3)),
+  bias=jnp.zeros((10, 3)),
+  count=jnp.arange(10),
+)
+x = jax.random.normal(random.key(1), (10, 2))
+
+def crazy_vector_dot(weights: Weights, x: jax.Array):
```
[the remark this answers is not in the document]
cgarciae: Its a good point. I'll add a variation of the first example using `vmap` over `__call__` so users know that its possible.

cgarciae:
Thanks @IvyZX for the detailed feedback. I've integrated all the suggestions.

PR #4216, opened 2024-09-22
[nnx] add Randomness guide

cgarciae:
# What does this PR do?

Adds Randoness guide.

[Preview](https://flax-nnx--4216.org.readthedocs.build/en/4216)


docs_nnx/guides/randomness.md
```diff
@@ -0,0 +1,245 @@
...
+Random state in Flax NNX is radically simplified compared to systems like Haiku/Flax Linen in that it defines "random state as object state". In essence, this means that random state is just another type of state, and is stored in Variables and help by the models themselves. The main characteristic of the RNG system in Flax NNX is that its **explicit**, its **order-based**, and uses **dynamic counters**. This is a bit different from [Flax Linen's RNG system](https://flax.readthedocs.io/en/latest/guides/flax_fundamentals/rng_guide.html) which is (path + order)-based and uses a static counters.
+
+```{code-cell} ipython3
+from flax import nnx
+import jax
+from jax import random, numpy as jnp
+```
+
+## Rngs, RngStream, and RngState
+Flax NNX provides the `nnx.Rngs` type a the main convenice API for managing random state. Following Flax Linen's footsteps, `Rngs` has the ability to create multiple named RNG streams, each with its own state, for the purpose of allowing for tight control over randomness in the context of JAX transforms. Here's a breakdown of the main RNG-related types in Flax NNX:
```
IvyZX, line 22: nit: 
convenice -> convenience
for the purpose of allowing for tight control -> for the purpose of tight control

docs_nnx/guides/randomness.md
```diff
@@ -0,0 +1,245 @@
...
+
+* **Rngs**: The main user interface. It defines a set of named `RngStream` objects.
```
IvyZX, line 24: Can we add `nnx.` to all these titles? like **nnx.Rngs**,  **nnx.RngStream**,  **nnx.RngState**,  **nnx.RngKey**,  **nnx.RngCount**.

docs_nnx/guides/randomness.md
```diff
@@ -0,0 +1,245 @@
...
+params_key = rngs.params()
+dropout_key = rngs.dropout()
+
+nnx.display(rngs)
+```
+
+Note that the `key` attribute does not change when a new keys are generated.
+
+### Standard stream names
+There are only two standard stream names used by Flax NNX, shown in the table below:
```
IvyZX, line 51: used by Flax NNX -> used by Flax NNX built-in layers

Just to avoid user confusion that other names are not allowed.

docs_nnx/guides/randomness.md
```diff
@@ -0,0 +1,245 @@
...
+
+| Name     | Description                               |
+|----------|-------------------------------------------|
+| `params` | Used for parameter initialization         |
+| `dropout`| Used by `Dropout` to create dropout masks |
+
+`params` is used my most of the standard layers (`Linear`, `Conv`, `MultiHeadAttention`, etc.) during construction to initialize their parameters. `dropout` is used by the `Dropout` and `MultiHeadAttention` to generate dropout masks. Here's a simple example of a model using `params` and `dropout` streams:
```
IvyZX, line 58: nit: `etc.` -> `etc`

docs_nnx/guides/randomness.md
```diff
@@ -0,0 +1,245 @@
...
+Notice that both the `key` and the `count` for the `dropout` stream have a dimension of `5`. This strategy the benefit of being very explicit, but can be a bit limiting as it wouldn't allow the stream to be used outside of the transform:
+
+```{code-cell} ipython3
+try:
+  rngs.dropout()
+except ValueError as e:
+  print(e)
+```
+
+The other option is to use the `nnx.split_rngs` decorator which will automatically split the random state of any `RngStream`s found in the inputs of the function, and will automatically "lower" them once the function call ends so the `Rngs` can be used outside again. `split_rngs` allows passing Filters to the `only` keyword argument to select the `RngStream`s that should be split. Using `split_rngs` is useful in combination with a transform but here we will show a simple example without any transforms to illustrate the concept, we'll use `split_rngs` with a transform on the next section.
```
IvyZX, line 158: I have a feeling that we should introduce and suggest `nnx.split_rngs` first, since it works better with transform cases and doesn't have the drawback that splitting in stream has (aka. can't work outside transform).

Actually, if there isn't any case in which splitting in stream is better, maybe it's fine to simply avoid introducing that... or introducing that as a lower-level view of what `nnx.split_rngs` actually does.

docs_nnx/guides/randomness.md
```diff
@@ -0,0 +1,245 @@
...
+
+f(rngs)
+
+print('Outside:')
+rngs.dropout() # works!
+nnx.display(rngs)
+```
+
+## Transforms
+As stated before, in Flax NNX random state is just another type of state, this means that there is nothing special about it regarding transforms. This means that you should be able to use the state handling APIs of each transform to get the results you want. In this section we will two examples of using random state in transforms, one with `pmap` and another one with `scan`.
```
IvyZX, line 177: > one with `pmap` and another one with `scan`.

What about this to highlight the intention of both examples:
one with `pmap` to split multiple RNG keys and one with `scan` to broadcast a single RNG key.

IvyZX (approved):
Great work! Very informative!

docs_nnx/guides/randomness.md
cgarciae: Good point. I'll think about it. We might as well just delete the other example as stick with `split_rngs`.
The main motivation of showing how to manually lift was to make it more familiar as the operation is explicit but it might be less used.

PR #4242, opened 2024-10-01
Add redirects for Linen guide links in the NNX site scope.

IvyZX:
This prepares for the site flip of `flax-nnx` scope becoming `flax` scope.


levskaya (approved):
we've checked that the redirect is snappy?

IvyZX:
> we've checked that the redirect is snappy?

Yeah it jumps automatically for me on Chrome

PR #4253, opened 2024-10-04
[nnx] add cloudpickle support

cgarciae:
# What does this PR do?

Adds support for `cloudpickle`.


levskaya (approved):
it _looks_ ok but cloudpickle is likely to make our lives exciting at some point down the road :)

PR #4291, opened 2024-10-12
[nnx] improve init_cache docs

cgarciae:
# What does this PR do?

Fixes #4290.


8bitmp3 (approved):
@cgarciae LGTM.

Checked the new code, it throws `ValueError: Autoregressive cache not initialized, call ``init_cache`` first.` as expected ("throws an error because cache isn't initialized").

@IvyZX @levskaya PTAL if a second LGTM is needed.

IvyZX (approved):

Issue #4265, opened 2024-10-08
CustomVJP example with NNX


rodrigodzf:
I wanted to know how to properly use `custom_vjp` with NNX. I could only find the docstring for the `linen` version, but no docstring for the NNX one.

cgarciae:
Here's an example from the docs:

```python
import jax
import jax.numpy as jnp
from flax import nnx

class Foo(nnx.Module):
  def __init__(self, x, y):
    self.x = nnx.Param(x)
    self.y = nnx.Param(y)

@nnx.custom_vjp
def f(m: Foo):
  return jnp.sin(m.x) * m.y

def f_fwd(m: Foo):
  return f(m), (jnp.cos(m.x), jnp.sin(m.x), m)

def f_bwd(res, g):
  inputs_g, out_g = g
  cos_x, sin_x, m = res
  tangent_m = nnx.State(dict(x=cos_x * out_g * m.y, y=sin_x * out_g))
  return (tangent_m,)

f.defvjp(f_fwd, f_bwd)

m = Foo(x=jnp.array(1.), y=jnp.array(2.))
grads = nnx.grad(f)(m)

jax.tree.map(jnp.shape, grads)
```
```
State({
  'x': VariableState(
    type=Param,
    value=()
  ),
  'y': VariableState(
    type=Param,
    value=()
  )
})
```

rodrigodzf:
Excellent! Thank you!

PR #4306, opened 2024-10-17
[nnx] fix custom_vjp

cgarciae:
# What does this PR do?

* Solves the issue where `FwdFn` is called outside of an update context. When this happens `FwdFn` behaves as a pure function.
* Correctly passes the non differentiable states as captures instead of using global state.

Solves #4265.

### DIFF
--- a/flax/nnx/transforms/autodiff.py
+++ b/flax/nnx/transforms/autodiff.py
@@ -19,6 +19,7 @@ import typing as tp
 
 
 from flax import struct
+from flax.core.frozen_dict import FrozenDict
 from flax.nnx import (
   extract,
   filterlib,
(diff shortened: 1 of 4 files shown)


hrbigelow:
Hi @cgarciae - I'm excited to try this fix!

Just wanted to mention a slight discrepancy in jax vs. nnx custom_vjp signatures that are expected:

nnx bwd:  res, (ins_g, outs_g) -> tangent
jax bwd:  res, outs_g -> tangent

The docs here seem t obe showing an example of jax.custom_vjp.

cgarciae:
@hrbigelow this is correct! I expanded a bit more on the docs to reflect this.

BTW: thanks for the wait. I've been trying to get this transform right but its taken some rewrites to solve some edge cases.

IvyZX (approved):

PR #4333, opened 2024-10-25
[nnx] improve transforms guide

cgarciae:
# What does this PR do?

* Adds section about passing Modules by closure.

Note: tracer level checks have been deactivated by JAX so the expected error is not showing up in the guide.


docs_nnx/guides/transforms.md
```diff
@@ -51,14 +51,14 @@ weights = Weights(
   kernel=random.uniform(random.key(0), (10, 2, 3)),
   bias=jnp.zeros((10, 3)),
 )
-x = jax.random.normal(random.key(1), (10, 2))
+count = jax.random.normal(random.key(1), (10, 2))
```
IvyZX, line 54: Why renamed it to `count`? `x` makes more sense as a model input.

docs_nnx/guides/transforms.md
```diff
@@ -302,14 +302,44 @@ def noisy_vector_dot(weights: Weights, x: jax.Array):
...
 
-## Consistent aliasing
+## Rules and limitations
+In this section we will cover some rules and limitations apply when using Modules inside transformations.
+
+### Mutable Module cannot be passed by closure
+
+While Python allows for passing objects as closures to functions, this is generally not supported by Flax NNX transforms. The reason is that because Modules are mutable, it is very easy to insert a tracer into a captured Module and get tracer leakage. To avoid this, Flax NNX Module and Variables check that the current context is valid every time they are performing an update operation.
+
+For example, if we a have stateful Module such as `Counter` that increments a counter every time it is called, and we try to pass it as a closure to a function decorated with `nnx.jit`, we would be leaking the tracer, however Flax NNX will raise an error instead to prevent this:
```
IvyZX, line 319: nit: ", however" -> ". However"

docs_nnx/guides/transforms.md
```diff
@@ -302,14 +302,44 @@ def noisy_vector_dot(weights: Weights, x: jax.Array):
...
+
```
IvyZX, line 320: I think it's worth to add an additional line saying this can be solved by explicitly pass the module `counter` as an argument of `f`.

IvyZX (approved):

docs_nnx/guides/transforms.md
cgarciae: Thanks for catching this! There's a vscode bug when renaming variables in jupyter notebooks. Reverted back to `x`.

docs_nnx/guides/transforms.md
cgarciae: Good idea. I simplified the language and added a comment add the end with the solution.

PR #4362, opened 2024-11-08
Update nnx_basics.ipynb

mohsenh17:
updating count value

# What does this PR do?

This PR refines the Counter class by updating the __call__ method to correctly increment the count attribute's value. Previously, the code attempted to increment the Count instance directly using self.count += 1, which led to an error due to syntax and attribute reference. The new version updates the counter by incrementing self.count.value.

## Checklist
- [*] This PR fixes a minor issue (e.g.: typo or small bug) or improves the docs (you can dismiss the other
      checks if that's the case).


mohsenh17:
The value is not necessary in the new version. My bad

PR #4368, opened 2024-11-11
Lint flax.nn.grad docstring

8bitmp3:
@IvyZX @cgarciae

### DIFF
--- a/flax/nnx/transforms/autodiff.py
+++ b/flax/nnx/transforms/autodiff.py
@@ -233,14 +233,16 @@ def grad(
   tp.Callable[..., tp.Any]
   | tp.Callable[[tp.Callable[..., tp.Any]], tp.Callable[..., tp.Any]]
 ):
-  """Lifted version of ``jax.grad`` that can handle Modules / graph nodes as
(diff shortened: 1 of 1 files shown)


flax/nnx/transforms/autodiff.py
```diff
@@ -233,14 +233,19 @@ def grad(
   tp.Callable[..., tp.Any]
   | tp.Callable[[tp.Callable[..., tp.Any]], tp.Callable[..., tp.Any]]
 ):
-  """Lifted version of ``jax.grad`` that can handle Modules / graph nodes as
-  arguments.
+  """A "lifted" version of `jax.grad <https://jax.readthedocs.io/en/latest/_autosummary/jax.grad.html#jax.grad>`_
```
cgarciae, line 236: I know this was there but we can improve this while we are at it... can you remove this "lifted" terminology. I think "reference-aware" or something like that is better.

8bitmp3: Yes! SGTM. We have discussed not using "lifted" with you and team (cc @levskaya) before.

8bitmp3:
@cgarciae @IvyZX @levskaya PTAL

8bitmp3:
@levskaya @cgarciae PTAL

PR #4453, opened 2024-12-17
Update NNX `State.state.merge` method docs in `statelib.py`

8bitmp3:


8bitmp3:
May need to review/close because of "[nnx] remove State #4418" https://github.com/google/flax/pull/4418

PR #4351, opened 2024-10-31
Linesearch (and lbfgs) support

jlperla:
# What does this PR do?
This PR modifies the `.update()` to support additional arguments required for the Optax `lbfgs` and related algorithms which rely on linesearch methods.

Fixes #4144 

- Additional tests were added to check the `lbfgs()` (where I believe the additional arguments are required for the `linesearch` more generally)
- The key feature needed is a callback to evaluate the objective function at different "state" values.  This needs to use a `split` and `merge` in the optimization.
  -  In cases where the `model_static, model_state = nnx.split(state.model, self.wrt)` has already been done, the `model_static` can be passed into the `.update` to avoid doing the `nnx.split` step with each iteration.  See the `test_jit_linesearch` for more there.

A feature missing here is support for `value_and_grad_from_state ` as described in https://optax.readthedocs.io/en/stable/_collections/examples/lbfgs.html#linesearches-in-practice.  Implementing this would provide some performance advantages as it would allow the optimizer to reuse the gradients/value.

Issue #4144, opened 2024-08-26
Support for optax lbfgs and related optimizers with NNX


jlperla:
I am trying to use L-BFGS and related optimizers with nnx + optax, but running into trouble.  It might be that `optax` has a slightly different optimization interface in those cases: https://optax.readthedocs.io/en/latest/api/optimizers.html#lbfgs with a function called `optax.value_and_grad_from_state` and also a change to the `optimizer.update` interface?

In particular, note that the sample code for these optimizers in https://optax.readthedocs.io/en/latest/api/optimizers.html#lbfgs looks like
```python
opt_state = solver.init(params)
value_and_grad = optax.value_and_grad_from_state(f)
for _ in range(5):
  value, grad = value_and_grad(params, state=opt_state)
  updates, opt_state = solver.update(
     grad, opt_state, params, value=value, grad=grad, value_fn=f
  )
  params = optax.apply_updates(params, updates)
```

So maybe we need the ability to pass a `value_fn` argument on?  Any easy fixes or things I am missing?

### Problem you have encountered:
Take the following implementation LLS with NNX + optax
```python
import jax
import jax.numpy as jnp
from jax import random
import optax
import equinox as eqx
import jax_dataloader as jdl
from jax_dataloader.loaders import DataLoaderJAX
from flax import nnx

N = 64  # samples
M = 2
sigma = 0.001
rngs = nnx.Rngs(42)
theta = random.normal(rngs(), (M,))
X = random.normal(rngs(), (N, M))
Y = X @ theta + sigma * random.normal(rngs(), (N,))  # Adding noise

def residual(model, x, y):
    y_hat = model(x)
    return (y_hat - y) ** 2

def residuals_loss(model, X, Y):
    return jnp.mean(jax.vmap(residual, in_axes=(None, 0, 0))(model, X, Y))

model = nnx.Linear(M, 1, use_bias=False, rngs=rngs)

lr = 0.001
optimizer = nnx.Optimizer(model,
                          optax.sgd(lr) #optax.lbfgs()
                          )

@nnx.jit
def train_step(model, optimizer, X, Y):
    grad_fn = nnx.value_and_grad(residuals_loss, has_aux=False)
    loss, grads = grad_fn(model, X, Y)
    optimizer.update(grads)
    return loss

num_epochs = 500
batch_size = 64
dataset = jdl.ArrayDataset(X, Y)
train_loader = DataLoaderJAX(dataset, batch_size=batch_size, shuffle=True)
[...]

PR #4471, opened 2025-01-06
Update train_state.py

emiresenov:
# What does this PR do?

This PR modifies the `.update()` to support the optional additional arguments `value` and `value_fn` for the `GradientTransformationExtraArgs` class.

These changes are in the same vein as the merged PR #4351, addressing #4144 for the TrainState class.

## Checklist
      checks if that's the case).
      #4351 , #4144 
      #4351 includes a test with these modifications. Additionally, the modifications follow the spec for [`GradientTransformationExtraArgs`](https://optax.readthedocs.io/en/latest/api/transformations.html#optax.GradientTransformationExtraArgs). If you want me to write a file with test cases for the `TrainState` class, let me know. Still, I cannot find any tests for this class in the repository.

### DIFF
--- a/flax/training/train_state.py
+++ b/flax/training/train_state.py
@@ -78,14 +78,19 @@ class TrainState(struct.PyTreeNode):
   tx: optax.GradientTransformation = struct.field(pytree_node=False)
   opt_state: optax.OptState = struct.field(pytree_node=True)
 
-  def apply_gradients(self, *, grads, **kwargs):
+  def apply_gradients(self, *, grads, value=None, value_fn=None, **kwargs):
     """Updates ``step``, ``params``, ``opt_state`` and ``**kwargs`` in return value.
 
     Note that internally this function calls ``.tx.update()`` followed by a call
     to ``optax.apply_updates()`` to update ``params`` and ``opt_state``.
 
+    For  ``optax.GradientTransformationExtraArgs``, the optional ``value`` and 
+    ``value_fn`` are passed to ``.tx.update()``.
+
     Args:
       grads: Gradients that have the same pytree structure as ``.params``.
+      value (optional): value of the objective associated with the current grads update.
+      value_fn (optional): function to evaluate the objective given the model.
       **kwargs: Additional dataclass attributes that should be ``.replace()``-ed.
 
     Returns:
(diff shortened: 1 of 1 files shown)


emiresenov:
I have signed the agreement.

Den mån 6 jan. 2025 kl 12:36 skrev google-cla[bot] ***@***.***

flax/training/train_state.py
```diff
@@ -100,9 +105,20 @@ def apply_gradients(self, *, grads, **kwargs):
       grads_with_opt = grads
       params_with_opt = self.params
 
-    updates, new_opt_state = self.tx.update(
-      grads_with_opt, self.opt_state, params_with_opt
-    )
+    if value is None or value_fn is None:
```
cgarciae, line 108: I think if either of them is not `None` we should pass them.

emiresenov: Good idea. I'll update the PR

emiresenov: I figure the same logic should hold for grad since e.g. [scale_by_polyak](https://optax.readthedocs.io/en/latest/api/transformations.html#optax.scale_by_polyak) also uses`GradientTransformationExtraArgs` and only takes `value` as an additional keyword argument in `.update()`. 

I moved all three optional args to an `update_kwargs` dictionary, adding each to `.update()` only if they are not `None`. Additionally, my former commit should not have sent the `**kwargs` to `.update()` since`TrainState` uses these for something else – I mixed up the usage from the merged PR I was referring to.

See the changes in my latest commit and let me know if there are any issues.

cgarciae:
This version looks better. Can you fix pre-commit issue? Run

```
pip install pre-commit
pre-commit run --all-files
```

cgarciae:
Tangent: curious if `nnx.Optimizer` works with `scale_by_polyak`? I've never used it but we recently added support for Linesearch there as well.

emiresenov:
> This version looks better. Can you fix pre-commit issue? Run
> 
> ```
> pip install pre-commit
> pre-commit run --all-files
> ```

Fixed, see my latest commit!

emiresenov:
> Tangent: curious if `nnx.Optimizer` works with `scale_by_polyak`? I've never used it but we recently added some support for Linesearch in `Optimizer` as well.

I haven't used it either, but I see no reason why it wouldn't work since the last commit of #4351 uses the same logic. I think the first commit of that PR (the one I mistakenly replicated in my first commit) suggested a solution that wouldn't work for `scale_by_polyak` used by `optax.polyak_sgd` since it didn't incorporate `grad` as an optional if only `value` was passed, presumably because the author only had Linesearch in mind.

However, this potential issue was fixed in the last commit of the PR when the author moved everything to their added `kwargs` in `.update()`. With this change, each optional value is passed only if not None – this makes the solution more generalized and also enables `scale_by_polyak` which only takes `value` as an additional keyword argument.

cgarciae (approved):

StefanoPizzamiglio:
@cgarciae, why hasn't this PR been merged yet? I think it could be useful to fix the corresponding bug

PR #4503, opened 2025-01-25
Rename variable string mapping utils and move them to variableslib

IvyZX:
* Renamed `variable_type` and `variable_type_name` to more concise `named_variable` and `variable_name`

* Move the registry that maps `nnx.Variable` types to strings from `nnx.bridge.variables` to `nnx.variablelib`, so that it can be used in core NNX methods like `nnx.Module.sow` and perturb.

### DIFF
--- a/flax/nnx/variablelib.py
+++ b/flax/nnx/variablelib.py
@@ -44,8 +44,6 @@ AxisIndex = int
 AddAxisHook = tp.Callable[[V, AxisIndex, AxisName | None], None]
 RemoveAxisHook = tp.Callable[[V, AxisIndex, AxisName | None], None]
 
-VariableTypeCache: dict[str, tp.Type[Variable[tp.Any]]] = {}
-
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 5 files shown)


flax/nnx/variablelib.py
```diff
@@ -966,3 +964,51 @@ def split_flat_state(
...
+
+
+def named_variable(name: str) -> tp.Type[Variable[tp.Any]]:
+  """Given a Linen-style collection name, get or create its NNX Variable class."""
+  if name not in VariableTypeCache:
+    VariableTypeCache[name] = type(name, (Variable,), {})
+  return VariableTypeCache[name]
+
+
+def variable_name(typ: tp.Type[Variable[tp.Any]]) -> str:
```
cgarciae, line 985: Name suggestions: `variable_type_from_name` and `variable_name_from_type`. A bit more verbose but current names are a bit ambiguous.

cgarciae (approved):

PR #4515, opened 2025-01-30
Add `nnx.Module.perturb`

IvyZX:
Add perturb to the `nnx.Module`, same as Linen. Also added a `nnx.Perturbations` variable class.
Covered with tests.

Note: 
* **Diff from Linen behavior**: Since we no longer have mutable collection check, this will create new perturbation variables even if none is passed into the Module. 
* Perturbation variables requires a sample input to be created, as its shape depends on the input shape.

### DIFF
--- a/flax/nnx/variablelib.py
+++ b/flax/nnx/variablelib.py
@@ -747,6 +747,38 @@ class Intermediate(Variable[A]):
   pass
 
 
+class Perturbation(Intermediate[A]):
+  """:class:`Variable` type that is typically used for
+  :func:`Module.perturb`::
+
+    >>> from flax import nnx
+    >>> import jax, jax.numpy as jnp
+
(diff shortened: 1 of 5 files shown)


flax/nnx/variablelib.py
```diff
@@ -747,6 +747,38 @@ class Intermediate(Variable[A]):
   pass
 
 
+class Perturbations(Variable[A]):
```
cgarciae, line 750: Can we make it singular like the other types? (e.g. `Perturbation`)

cgarciae, line 750: Also, we should inherit from `Intermediate` for all temp Variables.
```suggestion
class Perturbations(Intermediate[A]):
```

cgarciae (approved):

PR #4536, opened 2025-02-10
added support to the accuracy metric for binary classification

### LINKED ISSUE #4456: Accuracy doesn't work for simple binary classifier
Hello,

This is quite a simple issue but the function `nnx.metric.Accuracy`  only works for a network that returns more than one output, i.e. 
`super().update(values=(logits.argmax(axis=-1) == labels))`. Not sure if this is by design. 

I was playing around with a simple binary classifier than only returned a single logit per observation. The Accuracy function gives nonsense for this case.

Anyway, might be worth noting in the doc string what the function expects.

mattbahr:
Fixes #4456 

Adds support for binary classification to the accuracy metric.

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/nnx/training/metrics.py
+++ b/flax/nnx/training/metrics.py
@@ -240,26 +240,64 @@ class Accuracy(Average):
     >>> metrics.reset()
     >>> metrics.compute()
     Array(nan, dtype=float32)
+
+    >>> logits3 = jax.random.normal(jax.random.key(2), (5,))
+    >>> labels3 = jnp.array([0, 1, 0, 1, 1])
+    >>> accuracy = nnx.metrics.Accuracy(threshold=0.5)
+    >>> accuracy.update(logits=logits3, labels=labels3)
+    >>> accuracy.compute()
+    Array(0.8, dtype=float32)
   """
+
+  def __init__(self, threshold: float | None = None, *args, **kwargs):
+    """For binary classification, pass in a float denoting a threshold to determine if a
+    prediction is positive. For example, constructing the metric as
+    ``acc = Accuracy(threshold=0.5)`` would cause any logit greater than or equal to 0.5
+    to be interpreted as a positive classification. For multi-class classification, do
+    not pass in a threshold.
+
+    Args:
+      threshold: for binary classification, determines if a prediction is
+        positive. Defaults to None.
+    """
+    if (threshold is not None) and (not isinstance(threshold, float)):
+      raise TypeError(f'Expected threshold to be a float, got {type(threshold)}')
+
+    self.threshold = threshold
+    super().__init__(*args, **kwargs)
 
   def update(self, *, logits: jax.Array, labels: jax.Array, **_) -> None:  # type: ignore[override]
     """In-place update this ``Metric``.
 
     Args:
-      logits: the outputted predicted activations. These values are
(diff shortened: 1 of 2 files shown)


flax/nnx/training/metrics.py
```diff
@@ -240,26 +240,40 @@ class Accuracy(Average):
...
+
+    >>> logits3 = jax.random.normal(jax.random.key(2), (5,))
+    >>> labels3 = jnp.array([0, 1, 0, 1, 1])
+    >>> metrics.update(logits=logits3, labels=labels3)
+    >>> metrics.compute()
+    Array(0.8, dtype=float32)
   """
 
-  def update(self, *, logits: jax.Array, labels: jax.Array, **_) -> None:  # type: ignore[override]
+  def update(self, *, logits: jax.Array, labels: jax.Array, threshold=0.5, **_) -> None:  # type: ignore[override]
```
cgarciae, line 251: This parameter should be a constructor argument instead.

flax/nnx/training/metrics.py
```diff
@@ -240,26 +240,40 @@ class Accuracy(Average):
...
-        f'Expected logits.ndim==labels.ndim+1, got {logits.ndim} and {labels.ndim}'
+        'Expected logits.ndim==labels.ndim or logits.ndim==labels.ndim+1, '
+        f'got {logits.ndim} and {labels.ndim}'
       )
     elif labels.dtype in (jnp.int64, np.int32, np.int64):
       labels = jnp.astype(labels, jnp.int32)
     elif labels.dtype != jnp.int32:
       raise ValueError(f'Expected labels.dtype==jnp.int32, got {labels.dtype}')
 
+    if logits.ndim==labels.ndim:
```
cgarciae, line 273: please add a comment here that this case is for binary classification

cgarciae (commented):
Hey @mattbahr, thanks for doing this! Apart from the two other comments, I think `threshold` should be `None` by default and the user should explicitly opt-in to binary classification. This way we can have stronger shape checks for each case.

mattbahr:
> Hey @mattbahr, thanks for doing this! Apart from the two other comments, I think `threshold` should be `None` by default and the user should explicitly opt-in to binary classification. This way we can have stronger shape checks for each case.

Good idea! Just made the requested changes. Let me know if there's anything else I can do to improve the implementation.

flax/nnx/training/metrics.py
```diff
@@ -240,26 +240,64 @@ class Accuracy(Average):
...
+
+    >>> logits3 = jax.random.normal(jax.random.key(2), (5,))
+    >>> labels3 = jnp.array([0, 1, 0, 1, 1])
+    >>> accuracy = nnx.metrics.Accuracy(threshold=0.5)
+    >>> accuracy.update(logits=logits3, labels=labels3)
+    >>> accuracy.compute()
+    Array(0.8, dtype=float32)
   """
 
+  def __init__(self, threshold=None, *args, **kwargs):
```
cgarciae, line 252: Add type tyhint:
```suggestion
  def __init__(self, threshold: float | None = None, *args, **kwargs):
```

cgarciae (approved):

flax/nnx/training/metrics.py
mattbahr: Type hint added

PR #4552, opened 2025-02-19
Add linen metadata conversion to linx

IvyZX:
Added Linen variable metadata box conversion to `nnx.bridge.Module`'s `init()` and `apply()`. They will now take and output Linen metadata boxes instead of `VariableState`s.

Added `bridge.with_partitioning` which uses `nnx.with_partitioning` within but will automatically converts to `nn.Partitioned`.

* For custom metadata class, directly use `nnx.with_metadata(..., linen_meta_type=<custom_type>)`.


cgarciae (approved):
This is great! Thanks @IvyZX !

PR #4555, opened 2025-02-19
Raise error if user uses 'name' in bridge module setup

IvyZX:


flax/nnx/bridge/module.py
```diff
@@ -123,7 +123,7 @@ def _module_meta_call(cls: type[M], *args, **kwargs) -> M:
   parent = None
   module: M
 
-  if parent_ctx is not None and parent_ctx.in_compact:
+  if parent_ctx is not None:
```
cgarciae, line 126: you might need to rebase, its removing stuff from the previous PR?

flax/nnx/bridge/module.py
```diff
@@ -253,7 +256,6 @@ def variable(  # type: ignore[invalid-annotation]
     name: str,
     init_fn: tp.Callable[..., A] | None = None,
     *init_args,
-    unbox: bool = True,
```
cgarciae, line 256: we should leave this for compatibility?

flax/nnx/bridge/module.py
cgarciae, line 126: Not sure we should go into this path when we're not inside `compact`. Maybe we could do the check bellow outside?

flax/nnx/bridge/module.py
IvyZX: Done.

flax/nnx/bridge/module.py
IvyZX: Done.

PR #4560, opened 2025-02-20
Avoid returning rng states in bridge module output

IvyZX:
Linen modules don't return them, and frameworks like metadata and variable conversion doesn't really know how to deal with them.


IvyZX:
Closing this - fixed in another PR

PR #4561, opened 2025-02-21
Add deprecation warning to all `nnx.State` methods

IvyZX:
We'd like to clean up `nnx.State` and use built-in Python dict, as proposed by https://github.com/google/flax/pull/4418. 

In concern of active users, set a deprecation warning and let it fire for 3 months before we actually deprecate it.


IvyZX:
Stealing a lot of usage fixes from https://github.com/google/flax/pull/4418 to patch here

cgarciae (approved):

PR #4563, opened 2025-02-21
Fixes #1925

### LINKED ISSUE #1925: Can not assign dict whos key is not string as module attribute

muhrin:
In PR #2007, the issue of having a dict with non-string keys as linen module parameters was fixed but the related issue of having non-string keys as part of the state dict was not.  Specifically, these cases did not work:

    a = {1: 2}

    class Foo(nn.Module):
      def setup(self):
        self.a = FrozenDict(a)  # int as key.

    class Foo(nn.Module):
      a: dict[int, int]  # int as key.

this is because serialization for non-int keys was fixes for ``dict`` but not for ``FrozenDict``.  This uses the code from ``dict`` to add the same behaviour to ``FrozenDict``.

# What does this PR do?

Fixes #1925 

## Checklist
- [ ] This PR fixes a minor issue (e.g.: typo or small bug) or improves the docs.
- [X] This change is discussed in a Github issue/
      [discussion](https://github.com/google/flax/issues/1925) .
- [X] The documentation and docstrings adhere to the
- [X] This change includes necessary high-coverage tests.


muhrin:
Any chance this can get merged?  It's a very small change, and would help my use case a lot.

muhrin:
@cgarciae Any chance you would be willing to review this?

muhrin:
@IvyZX,  do you think you could review this? It's really a very small commit.

IvyZX (approved):

muhrin:
Thank you so much !

Issue #4426, opened 2024-12-10
NNX WeightNorm


rbflx:
Is there an NNX implementation of the [WeightNorm](https://flax-linen.readthedocs.io/en/latest/api_reference/flax.linen/layers.html#flax.linen.WeightNorm) module as used in Linen? I can't seem to find it in the [normalization layers](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/nn/normalization.html) implemented in NNX.

I'm in the process of writing a new model in NNX which would require them.

cgarciae:
We don't have a `WeightNorm` implementation yet in NNX. It should be very easy to implement using `nnx.state` + `nnx.statelib.map` + `nnx.update`. Contributions welcomed.

prakashkagitha:
Is it okay if I take a shot at this implementation? I see that NNX doesn't have InstanceNorm and SpectralNorm from Linen as well.

mattbahr:
@cgarciae I see the old implementation of `WeightNorm` from Linen here: https://flax-linen.readthedocs.io/en/latest/_modules/flax/linen/normalization.html#WeightNorm

Are you suggesting that there's a better way using `nnx.state`, `nnx.statelib.map`, and `nnx.update`?

mattbahr:
I have a branch where I've added the `InstanceNorm` and `SpectralNorm` implementations to NNX. @cgarciae do you think it would be worth opening a PR for these?

AshwinSankar17:
Any updates on this? @mattbahr have you tried to use the weight norm implementation here?

PR #4568, opened 2025-02-24
[nnx] Add NNX WeightNorm.

nilq:
# What does this PR do?

This PR addresses #4426 by implementing an NNX WeightNorm class. This brings the requested normalization class to the NNX library – implemented to be consistent with the corresponding module in Linen.

## Checklist
- [x] This change is discussed in #4426.


flax/nnx/nn/normalization.py
```diff
@@ -835,4 +853,144 @@ def __call__(self, x, *, mask: tp.Optional[jax.Array] = None):
...
+    Returns:
+      Output of the layer using l2-normalized weights.
+    """
+    state = nnx.state(self.layer_instance)
+
+    def apply_weightnorm(path, var_state):
+      path_str = '/'.join(str(k) for k in path)
+
+      if self.variable_filter:
+        for variable_name in self.variable_filter:
```
cgarciae, line 950: We should use the same Filter supported by `nnx.filterlib` which are a bit more general and used thought the library.

cgarciae:
To resolve the pytest issues you will have to add some deprecation warning to the list here:
https://github.com/google/flax/blob/6af8fcb0bfac1eb078bdf5ba41a3899a71dcaf81/pyproject.toml#L188

If this is not from your PR we can fix it in `main` first.

nilq:
@cgarciae Just a friendly ping – in case this might have drowned. 🙂

HyperBlaze456:
Hello, I think the variable filter is not correctly working?
The code does not apply them to extract only the required variables to be normalized, and there is only case where no variable filter is passed.

samanklesaria:
Hi @nilq - we'd like to finally land this PR on top of all the recent changes to nnx! We just merged support for SpectralNorm (#4623), and your WeightNorm code is quite similar. Do you think it might be possible to factor out some of the shared functionality?

samanklesaria:
Closed by #5043

PR #4584, opened 2025-02-27
Util to let bridge module work with NNX submodules

IvyZX:
* Add `current_module` to track the current parent
* Add `wrap_nnx_module` to automate RNG distribution into pure-NNX territory, and to avoid buffer allocation

### DIFF
--- a/flax/nnx/bridge/module.py
+++ b/flax/nnx/bridge/module.py
@@ -106,6 +106,25 @@ def _bind_module(parent: Module, module: Module) -> Module:
   return module
 
 
+def current_context() -> ModuleStackEntry | None:
+  return MODULE_CONTEXT.module_stack[-1]
+
+
+def current_module() -> Module | None:
+  """A quick util to get the current bridge module."""
+  ctx = current_context()
+  if ctx is None:
+    return None
+  return ctx.module
+
+
+def _auto_submodule_name(parent_ctx, cls):
+  """Increment type count and generate a new submodule name."""
+  type_index = parent_ctx.type_counter[cls]
+  parent_ctx.type_counter[cls] += 1
+  return f'{cls.__name__}_{type_index}'
+
+
 class ModuleMeta(nnx_module.ModuleMeta):
 
   def _object_meta_construct(cls, self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


tests/nnx/bridge/module_test.py
```diff
@@ -302,6 +302,48 @@ def __call__(self, x):
...
+        x = self.dropout(x)
+        return x
+
+    class BridgeMLP(bridge.Module):
+      num_layers: int
+      @bridge.compact
+      def __call__(self, x):
+        for i in range(self.num_layers):
+          layer = nnx.bridge.wrap_nnx_module(lambda r: NNXLayer(8, 0.3, rngs=r))
+          setattr(self, f'layer{i}', layer)
```
cgarciae, line 325: I think we can do a bit better and try to mimic compact with auto-naming or optional `name` parameter like this:

```suggestion
          layer = nnx.bridge.compact_init(NNXLayer, 8, 0.3, rngs=self.scope.rngs, name=f'layer{i}')
```

flax/nnx/bridge/module.py
```diff
@@ -106,6 +107,25 @@ def _bind_module(parent: Module, module: Module) -> Module:
   return module
 
 
+def current_context() -> tp.Optional[ModuleStackEntry]:
```
cgarciae, line 110: nit: use `ModuleStackEntry | None`

flax/nnx/bridge/module.py
```diff
@@ -501,3 +518,32 @@ def _get_unbound_fn(method_or_fn: tp.Callable) -> tp.Callable:
...
+  if parent.is_initializing():
+    module = factory(parent.scope.rngs)
+  else:
+    rngs = parent.scope.rngs if parent.scope.rngs else rnglib.Rngs(7)  # dummy
+    module = transforms.eval_shape(factory, rngs)
+    # Make sure the internal rng state is not abstract - other vars shall be
+    if parent.scope.rngs:
+      for _, stream in graph.iter_graph(module):
+        if isinstance(stream, rnglib.RngStream):
+          stream.key.value = rngs[stream.key.tag].key.value
```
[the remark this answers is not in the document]
cgarciae: Maybe the policy could be that we only rng state from `apply` and try to replace it here if the keys and counts are scalars.

cgarciae (approved):

flax/nnx/bridge/module.py
IvyZX: Done.

flax/nnx/bridge/module.py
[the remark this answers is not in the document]
IvyZX: This is indeed tricky. We supply non-abstract RNG state here essentially for any code in the nnx.Module that assumes a complete `nnx.Rngs` instance and call its methods in call time.
I think the cleanest solution is to use `nnx.jit` to partially init the NNX module and only keep all `RngState`. Updated the PR and added a test for it.

PR #4587, opened 2025-02-28
Adding nnx Gemma2-2b (including overall fixes) to examples/gemma

### LINKED ISSUE #4567: Some surprises when adding Gemma2-2b to flax/examples/gemma
Looking forward to getting to grips with `nnx`, I went about implementing the apparently missing `gemma2_2b` model.

After coming up with a suitable `TransformerConfig`, I was surprised that I couldn't get a forward pass to run : So I added 'Grouped-Query Attention' (mirroring what can be seen in the GDM `gemma` library).  

Still no dice, since (although I checked all the params were getting set correctly), garbage logits were being returned.  So I labouriously went through each layer, checking for numerical differences, only to find that in [`modules.Block`](https://github.com/google/flax/blob/main/examples/gemma/modules.py#L338-L358), the ordering of the sub-modules is clearly in the wrong order (look at the `pre_ffw_norm` and the `post_attn_norm`, and the residual paths, for instance):

```python
class Block(nnx.Module):
    #...
  def __call__(self, x, ...):  
    inputs_normalized = self.pre_attention_norm(x)
    cache, attn_output = self.attn(
        inputs_normalized,
        segment_pos,
        cache,
        attn_mask,
    )
    attn_output += x
    residual = attn_output
    attn_output = self.pre_ffw_norm(attn_output)

    if self.use_post_attn_norm:
      attn_output = self.post_attn_norm(attn_output)
    self.sow_config.maybe_sow_rs_after_attention(attn_output, self)

    outputs = self.mlp(attn_output)
    if self.use_post_ffw_norm:
      outputs = self.post_ffw_norm(outputs)
    outputs = residual + outputs
    self.sow_config.maybe_sow_rs_after_ffw(outputs, self)
    return cache, outputs
```

After fixing this up, the model's final `logits` come out pretty close to the GDM `gemma` ones (small errors seem to creep in along the way due to bfloat16/float16).  This is good news - though it is slightly worrying that this example has been *unusable* over the whole period, since its actual functionality as an implementation of gemma models does not appear to have been tested...

Please let me know if you'd like a PR to cover (at minimum) adding a working `gemma2_2b`

mdda:
## Adds Gemma2-2b (including GQA to Attention and fixes to Block)

Fixes #4567 

It includes:
* New `TransformerConfig` for gemma2-2b model
* Renames existing configs to make them uniform
   + NB: This cannot affect existing code, since this module was previously unusable
* Added additional params reading key adjustment (key in gemma2-2b needs remapping from kaggle download)
* Adds GQA to Attention module
* Reorders the operations in Block module so that logits output from overall Transformer are not gibberish
  + logits confirmed to (approximately) match those from GDE `gemma` (flax linen) model
* No new documentation provided
  + This change would make the example in the `nnx` documentation work (did not work before)
* No additional tests provided


cgarciae:
Hey Martin! Thanks for doing this.
Some folks are internally also improving the model. Let me merge their changes first and make sure there are no conflicts with those changes.

mdda:
Ahhh - Brings back memories of PRs for TensorFlow.  Good times!

examples/gemma/transformer.py
```diff
@@ -128,30 +128,31 @@ def gemma_7b(cls):
     )
 
   @classmethod
-  def gemma_27b(cls):
-    num_layers = 46
+  def gemma2_2b(cls):
```
cgarciae, line 131: Why don't we add a new method for `gemma_27b` configuration instead of removing the `gemma2_2b` one?

cgarciae:
Thanks @mdda for doing this, it took a while because there where other changes to the model in the background that were pending. I think this is great. Can you please add a test to check that the GQA configuration works and matches the base version?

Also now need to solve the conflicts, sorry about this, this code is being used by a couple of users internally.

examples/gemma/transformer.py
mdda: They are both in the file - it's just the git that didn't pick up the diff properly.

But also, it makes sense to rename the different 'generations' of gemma, gemma2, gemma3, etc  as separate classes, rather than relying on implicit knowledge about which size came from where.

Moreover, the gemma2_27 didn't have the right normalisation in the attention - will need a separate fix.

I also see that the Google-generated PR adopted the same fix as mine to the Block module.  Good for you!

mdda:
Surprised that the code was being used internally prior to my PR, since the Block module was entirely borked.

cgarciae:
@mdda please take a look at CI, you probably need to run:

```bash
pip install pre-commit
pre-commit run --all-files
```

vfdev-5:
Hey @mdda , I wonder whether you would be able to finalize this PR? If you are busy, we can take over it keeping your commits (such that you will be credited for the work you have done). Please, let me know. Thanks!

PR #4606, opened 2025-03-07
[nnx] add support for standalone Variables

cgarciae:
# What does this PR do?

Adds support for using Variables outside of graph nodes, meaning they can now be passed directly to transforms, including inside pytrees without a parent graph node, and the ability to call `split`, `merge`, and `update` on them directly.

```python
rngs = nnx.Rngs(0)
w = nnx.Param(jax.random.normal(rngs(), (2, 3)))
b = nnx.Param(jnp.zeros((3,)))
count = nnx.Variable(jnp.array(0))

@nnx.jit
def linear(w, b, count, x):
  count += 1
  return x @ w + b[None]

x = jax.random.normal(rngs(), (1, 2))
y = linear(w, b, count, x)

assert count.value == 1
assert y.shape == (1, 3)
```

### DIFF
--- a/flax/nnx/extract.py
+++ b/flax/nnx/extract.py
@@ -127,8 +127,8 @@ class PrefixMapping(abc.ABC):
   ) -> tp.Any: ...
 
 def check_consistent_aliasing(
-  node: tuple[tp.Any, ...],
-  prefix: tuple[tp.Any, ...],
+  node: tp.Any,
+  prefix: tp.Any,
   /,
   *,
   node_prefixes: dict[tp.Any, list[tuple[PathParts, tp.Any]]] | None = None,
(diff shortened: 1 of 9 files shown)


flax/nnx/extract.py
```diff
@@ -343,7 +345,9 @@ def from_tree(
     with graph.merge_context(is_inner, ctxtag) as merge_ctx:
       return jax.tree.map(
         lambda x: merge_fn(merge_ctx, (), prefix, x)
-        if map_non_graph_nodes or is_node_leaf(x)
+        if map_non_graph_nodes
```
IvyZX, line 348: This lambda becomes kinda long & unreadable. Can you refactor it out to a function? Then you can also use this for the `merge_fn` call in line 374.

IvyZX (approved):

PR #4613, opened 2025-03-11
add promote_dtype as a config option for multiple layers

cgarciae:
# What does this PR do?

Adds a `promote_dtype` config option to various layers, defaults to `dtypes.promote_dtype`.

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -16,6 +16,7 @@
 
 from typing import (
   Any,
+  Protocol,
 )
 from collections.abc import Iterable, Sequence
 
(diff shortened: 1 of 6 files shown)


flax/linen/linear.py
```diff
@@ -45,6 +46,10 @@
   LaxPadding,
 )
 
+class PromoteDtypeFn(Protocol):
```
levskaya, line 49: why are there two copies of the protocol?  I see another in typing.py

levskaya (approved):
I'm guessing the user needs something too complex to support in standard fns.  Given the explosion of quantized dtypes I guess this makes sense as an escape hatch.

flax/linen/linear.py
cgarciae: Yeah they are slightly different because the version that Linen users is a bit less `typing` friendly.

PR #4634, opened 2025-03-18
[bridge module] Allow name arg to represent actual submodule path

IvyZX:
Allows `name=` arg to actually represent the submodule path most of the time, while keeping other module attributes.

This arg will automatically set an attribute of this name to the parent module, and it is of slightly higher priority than common attributes. This works in both compact and setup.

This is to accommodate such common patterns of Linen:

```
    class MLP(bridge.Module):
      nlayers: int
      def setup(self):
        self.layers = [Dense(4, name=f'layer_{i}') for i in range(self.nlayers)]
      def __call__(self, x):
        for layer in self.layers:
          x = layer(x)
        return x
```

### DIFF
--- a/flax/nnx/bridge/module.py
+++ b/flax/nnx/bridge/module.py
@@ -143,37 +143,34 @@ def _module_meta_call(cls: type[M], *args, **kwargs) -> M:
 
   name = None
   if parent_ctx is not None:
-    want_name = 'name' in kwargs and 'name' in inspect.get_annotations(cls)
-    if not want_name and not parent_ctx.in_compact and 'name' in kwargs:
-      raise ValueError(
-        f"'name' can only be set in @compact functions. If in setup(), "
-          "use parent's `self.<attr_name> to set the submodule name.")
+    if 'parent' in kwargs:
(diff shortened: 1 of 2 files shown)


flax/nnx/bridge/module.py
```diff
@@ -183,8 +180,9 @@ def _module_meta_call(cls: type[M], *args, **kwargs) -> M:
 
 class AttrPriority(enum.IntEnum):
   HIGH = 1
-  DEFAULT = 2
-  LOW = 3
+  INIT_PARENT = 2
+  DEFAULT = 3
+  LOW = 4
```
cgarciae, line 185: What if we leave some space between them as a escape hatch in case more control is needed?

```python
HIGH = 0
INIT_PARENT = 20
DEFAULT = 50
LOW = 100
```

cgarciae (approved):

flax/nnx/bridge/module.py
IvyZX: Done.

cgarciae:
We should consider adding a check so we don't override existing priorities since we are doing this automatically.

PR #4639, opened 2025-03-19
[nnx] flaxlib types

cgarciae:
# What does this PR do?

* Implement more types with `nanobind` and patch them if `flax_use_flaxlib` is set.
* Slight refactor to `RefMap` and the new `IndexMap`.

Note that `flaxlib` is currently not being use internally and has not been published to pypi, can only be used by building it for testing purposes.

### DIFF
--- a/benchmarks/nnx_simple_training.py
+++ b/benchmarks/nnx_simple_training.py
@@ -116,14 +116,11 @@ def main(argv):
       loss = jnp.mean((y - y_pred) ** 2)
       return {'loss': loss}
 
-    cached_train_step_nnx = nnx.cached_partial(train_step_nnx, model, optimizer)
-    cached_test_step_nnx = nnx.cached_partial(test_step_nnx, model)
-
     for step, batch in enumerate(dataset(X, Y, batch_size)):
-      cached_train_step_nnx(batch)
+      train_step_nnx(model, optimizer, batch)
 
       if step % 1000 == 0:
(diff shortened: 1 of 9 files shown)


benchmarks/nnx_simple_training.py
```diff
@@ -116,14 +116,14 @@ def test_step_nnx(model: MLP, batch):
       loss = jnp.mean((y - y_pred) ** 2)
       return {'loss': loss}
 
-    cached_train_step_nnx = nnx.cached_partial(train_step_nnx, model, optimizer)
-    cached_test_step_nnx = nnx.cached_partial(test_step_nnx, model)
+    # cached_train_step_nnx = nnx.cached_partial(train_step_nnx, model, optimizer)
```
IvyZX, line 119: Remove commented-out lines

flax/nnx/graph.py
```diff
@@ -258,6 +285,11 @@ def __treescope_repr__(self, path, subtree_renderer):
       subtree_renderer=subtree_renderer,
     )
 
+if config.flax_use_flaxlib:
```
IvyZX, line 288: Why does it need to import repeatedly? Can we do one import at top?

IvyZX (approved):

flax/nnx/graph.py
cgarciae: could be done once at the end I guess, else they are rewritten by the actual definition

PR #4640, opened 2025-03-20
Apply vmap jnp.pad over the batch in REFLECT convolution mode

sarlinpe:
# What does this PR do?

This PR fixes an error occurring when exporting flax models with `jax2tf`, polymorphic batch size, and REFLECT padding mode in convolutions. This issue can be minimally reproduced as
```python
import jax.numpy as jnp
import tensorflow as tf
from jax.experimental import jax2tf

def fn_jax(x):
  return jnp.pad(x, pad_width=[(0, 0), (1, 1), (1, 1)], mode='reflect')

fn_tf = jax2tf.convert(fn_jax, polymorphic_shapes=["(b, 14, 14)"])
fn_tf = tf.function(fn_tf, autograph=False, jit_compile=True)
fn_tf.get_concrete_function(tf.TensorSpec([None, 14, 14], tf.float32))
```
which results in 
```
ValueError: Shape polymorphism is supported for jnp.pad with 'reflect' or 'symmetric' padding mode only when it is possible to determine at lowering time that the axis size (= b) is larger than 1 and larger or equal than the padding length (= 0). Error while handling left padding on axis 0.
```
Applying `jax.vmap(functools.partial(jnp.pad, pad_width=[(1, 1), (1, 1)], mode='reflect'))(x)` fixes the issue.

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -16,6 +16,7 @@
 
 from collections.abc import Iterable, Sequence
 from typing import Any, Protocol
+import functools
 
 from flax.core import meta
 from flax.linen import initializers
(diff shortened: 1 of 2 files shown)


sarlinpe:
@cgarciae It would be great to receive your review, thank you!

PR #4643, opened 2025-03-23
Add optional error handlers to frozen dict

### LINKED ISSUE #513: Linen: Consider raising an error when reading variables from submodule before initialization?

mattbahr:
# What does this PR do?

Fixes #513 - implementation based on @cgarciae's [comment](https://github.com/google/flax/issues/513#issuecomment-1237336408)

## Checklist
      checks if that's the case).
- [x] This change is discussed in Github issue #513 

### DIFF
--- a/flax/core/frozen_dict.py
+++ b/flax/core/frozen_dict.py
@@ -16,12 +16,14 @@
 
 import collections
 from types import MappingProxyType
+import typing as tp
 from typing import Any, TypeVar
 from collections.abc import Hashable, Mapping
 
 import jax
 
 from flax import serialization
+from flax.error_handlers import ErrorHandlers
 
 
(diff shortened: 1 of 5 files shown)


cgarciae:
I don't agree with my past self. The main issue seems fundamental to Linen and making FrozenDict more complex seems like a bad tradeoff.

mattbahr:
> I don't agree with my past self on #513. The main issue seems fundamental to Linen and making FrozenDict more complex seems like a bad tradeoff.

Ok, no problem! I'll close this PR. Should #513 be closed as well if it's not worth making the change for Linen?

Issue #4585, opened 2025-02-28
Several tests failing with `AttributeError: module 'jax.api_util' has no attribute 'debug_info'`


GaetanLepage:
Context: bumping `flax` to 0.10.4 on nixpkgs: https://github.com/NixOS/nixpkgs/pull/385676

### System information
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): NixOS unstable
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`: flax 0.10.4, jax 0.5.0, jaxlib 0.5.0
- Python version: 3.12.9
- GPU/TPU model and memory: None
- CUDA version (if applicable):

### Problem you have encountered:
Several tests now fail with `AttributeError: module 'jax.api_util' has no attribute 'debug_info'`:
```
=========================== short test summary info ============================
FAILED tests/core/core_meta_test.py::MetaTest::test_scan_over_layers - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/core/core_scope_test.py::ScopeTest::test_lazy_init - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/core/core_scope_test.py::ScopeTest::test_lazy_init_fails_on_data_dependence - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/core/design/core_big_resnets_test.py::BigResnetTest::test_big_resnet - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/linen/linen_module_test.py::ModuleTest::test_lazy_init - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/linen/linen_module_test.py::ModuleTest::test_lazy_init_fails_on_data_dependence - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/linen/linen_meta_test.py::LinenMetaTest::test_pjit_scan_over_layers - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/linen/linen_recurrent_test.py::RNNTest::test_basic_seq_lengths - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
FAILED tests/linen/linen_recurrent_test.py::RNNTest::test_numerical_equivalence - AttributeError: module 'jax.api_util' has no attribute 'debug_info'
[...]

PR #4648, opened 2025-03-24
Roll up package jax version and uv.lock

IvyZX:
Rolls up the required JAX version because we made `debug_info` changes that only works after `jax==0.5.1`.

Fixes https://github.com/google/flax/issues/4585

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ dependencies = [
     "numpy>=1.23.2; python_version>='3.11'",
     "numpy>=1.26.0; python_version>='3.12'",
     # keep in sync with jax-version in .github/workflows/build.yml
-    "jax>=0.4.27",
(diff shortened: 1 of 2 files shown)


cgarciae (approved):

bkaplowitz:
@IvyZX This issue also appears in v0.10.4 (although I've confirmed it's not present in any earlier version). Might it also be possible to update the uv.lock JAX requirement for that as well?

IvyZX:
> @IvyZX This issue also appears in v0.10.4 (although I've confirmed it's not present in any earlier version). Might it also be possible to update the uv.lock JAX requirement for that as well?

Can you try to upgrade to the latest Flax (0.10.6) and see if the error goes away? We always recommend users to use the latest version because it may fix bug introduced in older version

Issue #4635, opened 2025-03-18
Add a nnx.identity "activation"?


jlperla:
There are many cases where I want to setup a flexible neural network parameterization where I try out different "activation" functions and final layers from the CLI.  A convenient design for this is to use jsonargparse or alternatives and do things like
```python
# Workaround: https://github.com/omni-us/jsonargparse/issues/619#issuecomment-2466451720
nnx.softplus.__deepcopy__ = MethodType(lambda self, memo: self, nnx.softplus)

def my_model(
    width: int = 64,
    depth: int = 4,
    activation: Callable = nnx.relu,
    final_activation: Callable = nnx.softplus,
    ...): 
    # use the settings to create a neural network
if __name__ == "__main__":
    jsonargparse.CLI(my_model)
```
Then I can do things like `python my_model.py --depth=3 --activation=nnx.tanh`

However, a common thing I need to do is swap out different settings for an identity function.  But this isn't available in jax or nnx.  What I would love to do is try things like `python my_model.py --depth=3 --final_activation=nnx.identity`

So, could we add this to the nnx activation functions?  Something trivial like
```python
@jax.jit
def identity(x: ArrayLike) -> Array:
  r"""Identity activation function.

  Returns the argument unmodified.

  Args:
    x : input array

  Returns:
    The argument `x` unmodified.

  Examples:
    >>> nnx.identity(jax.numpy.array([-2., -1., -0.5, 0, 0.5, 1., 2.]))
    Array([-2. , -1. , -0.5, 0. , 0.5, 1. , 2. ], dtype=float32)

  """  
  return x
```

If so, I can create a PR if you tell me which files to add it to.  It seemed like https://github.com/google/flax/blob/main/flax/nnx/nn/activations.py is a natural place.  If a test is required, tell me which file it makes the most sense in?

Ruhaan838:
Okay, you have a nice suggestion that I should allow you to do this. I found that it uses `jax` for activation, so you need to add the file here.
-> Main src for the functions: 
https://github.com/jax-ml/jax/blob/main/jax/_src/nn/functions.py
and also add that here too 
-> actual imports: 
https://github.com/jax-ml/jax/blob/main/jax/nn/__init__.py

jlperla:
I see, so you think the best place for this is in the jax activations which end up exported directly by NNX.  Makes sense, let me post there.

jlperla:
Done.  See https://github.com/jax-ml/jax/pull/27222

[...]

PR #4652, opened 2025-03-25
Added identity export from JAX

jlperla:
Fixes https://github.com/google/flax/issues/4635 

This re-exports the JAX identity "activation" which was merged in https://github.com/jax-ml/jax/pull/27355

This was the only impact I could find in the Flax code itself.  However, given that this has just been merged to JAX (and is not yet part of a tagged release) I am not sure about the merging process and its interaction with version compatibility, etc. @IvyZX


IvyZX:
I'm happy to approve this, but FYI we won't be able to merge it until JAX make another release that includes the `identity` function.

jlperla:
@IvyZX   I bumped the `pyproject.toml` (which in turn told me to sync up with the CI GitHub action file) using the newly released v0.6.0 of JAX.  

I wasn't sure if there were other things that were required here prior to merging?

IvyZX (approved):

jlperla:
Thanks @IvyZX 

I wasn't sure how to check the docs here since I think they autogenerate in some way?

PR #4718, opened 2025-04-19
FEAT: add LoRALinearGeneral

zhengzl18:
1. add `LoRALinearGeneral` since `MultiHeadAttention` uses `LinearGeneral`.
2. modify default value of `axis` in `LinearGeneral.__init__()` so that it is compatible to in_features that is a suquence of ints
3. add `lora_base_module` param to `LoRALinear` so that it's consistent with `LoRA`
4. update docs in lora.py

Since `LoRA` currently doesn't support dot_general, I reshape the input before and after calling lora. I also add a check on the value of `axis` param in `LinearGeneral`, so that it's only valid to apply linear transformation on the last axes of the input. 

I got trouble running the tests, so I couldn't add one.

### DIFF
--- a/flax/nnx/__init__.py
+++ b/flax/nnx/__init__.py
@@ -99,6 +99,7 @@ from .nn.linear import LinearGeneral as LinearGeneral
 from .nn.linear import Einsum as Einsum
 from .nn.lora import LoRA as LoRA
 from .nn.lora import LoRALinear as LoRALinear
+from .nn.lora import LoRALinearGeneral as LoRALinearGeneral
 from .nn.lora import LoRAParam as LoRAParam
 from .nn.normalization import BatchNorm as BatchNorm
 from .nn.normalization import LayerNorm as LayerNorm
--- a/flax/nnx/nn/linear.py
+++ b/flax/nnx/nn/linear.py
@@ -146,7 +146,7 @@ class LinearGeneral(Module):
     in_features: Size | tp.Sequence[Size],
     out_features: Size | tp.Sequence[Size],
     *,
-    axis: Axis | tp.Sequence[Axis] = -1,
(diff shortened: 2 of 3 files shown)


vfdev-5:
@zhengzl18 thanks for the PR, but please first open an issue about that. I'm not totally sure why we need to implement this layer as we can create it using `nnx.LoRA`:
```python
import flax.nnx as nnx

linear = nnx.LinearGeneral(3, 4, rngs=nnx.Rngs(0))
wrapper = nnx.LoRA(3, 2, 4, base_module=linear, rngs=nnx.Rngs(1))
assert wrapper.base_module == linear
print(wrapper.lora_a.shape)  # (3, 2)
print(wrapper.lora_b.shape)  # (2, 4)
y = wrapper(jnp.ones((16, 3)))
print(y.shape)  # (16, 4)
```

zhengzl18:
@vfdev-5 `nnx.LoRA` only takes `in_features` and `out_features` of type `int`, while `nnx.LinearGeneral` takes `in_features` and `out_features` of type `Union[int|Tuple[int]]`. For a "real" `nnx.LinearGeneral` with multiple in or out features, I'm not seeing how to create a lora version of it:
```
linear_general = nnx.LinearGeneral((2, 3), (4, 5), axis=(1, -1), rngs=nnx.Rngs(0))
# wrapper = nnx.LoRA((2, 3), 2, (4, 5), base_module=linear_general, rngs=nnx.Rngs(1)) # This raises an TypeError: Shapes must be 1D sequences of concrete values of integer type, got ((2, 3), 2). 
```

vfdev-5:
@zhengzl18 makes sense, thanks for pointing out these details. I'll ask others about this feature.

PR #4749, opened 2025-05-14
Added keep_rngs arg to MHA to optionally store rngs

vfdev-5:
# What does this PR do?

- Added `keep_rngs` arg to MHA to optionally store rngs as `self.rngs`

Today's MHA behaviour is that provided rngs are stored in the module if `dropout_rate>0`. In this case, module split has parameters and also RNG object:
```python
m = nnx.MultiHeadAttention(num_heads=8,
      in_features=512,
      qkv_features=512,
      dtype=jnp.float32,
      use_bias=False,
      broadcast_dropout=False,
      dropout_rate=0.1,
      rngs=nnx.Rngs(params=123),
)
gd, params, rest = nnx.split(m, nnx.Param, ...)

State({
  'rngs': {
    'params': {
      'count': VariableState( # 1 (4 B)
        type=RngCount,
        value=Array(4, dtype=uint32),
        tag='params'
      ),
      'key': VariableState( # 1 (8 B)
        type=RngKey,
        value=Array((), dtype=key<fry>) overlaying:
        [  0 123],
        tag='params'
      )
    }
  }
})
```

With `keep_rngs` arg we can optionally disable saving rngs in the object and still pass rngs in the call method. In this case we can split the module without RNG object:
```python
rngs = nnx.Rngs(42)
module = nnx.MultiHeadAttention(
   in_features=4,
   num_heads=2,
   qkv_features=4,
   decode=True,
   rngs=rngs,
   dropout_rate=0.5,
   keep_rngs=False
)
assert module.rngs is None
gd, params = nnx.split(module, nnx.Param)
```

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -301,6 +301,10 @@ class MultiHeadAttention(Module):
     decode: whether to prepare and use an autoregressive cache.
     normalize_qk: should QK normalization be applied (arxiv.org/abs/2302.05442).
     rngs: rng key.
+    keep_rngs: whether to store the input rngs as attribute (i.e. `self.rngs = rngs`)
+      (default: True). If rngs is stored, we should split the module as
+      `graphdef, params, nondiff = nnx.split(module, nnx.Param, ...)` where `nondiff`
+      contains RNG object associated with stored `self.rngs`.
   """
 
   def __init__(
@@ -330,6 +334,7 @@ class MultiHeadAttention(Module):
     qkv_dot_general_cls: Any = None,
     out_dot_general_cls: Any = None,
     rngs: rnglib.Rngs,
+    keep_rngs: bool = True,
   ):
     self.num_heads = num_heads
     self.in_features = in_features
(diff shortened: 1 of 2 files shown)


cgarciae (approved):

cgarciae:
This is great!

flax/nnx/nn/attention.py
```diff
@@ -330,6 +334,7 @@ def __init__(
     qkv_dot_general_cls: Any = None,
     out_dot_general_cls: Any = None,
     rngs: rnglib.Rngs,
+    keep_rngs: bool = True,
```
cgarciae, line 337: maybe we make this `False` by default?

vfdev-5: It's true that in both cases it is BC-breaking. It would be good to make users aware of the change with a warning?
Or we can have a three-state flag: None, True, False, with None doing the previous non-breaking behavior and setting default value to None + warning saying about a future change to `keep_rngs=False` as default. What do you think?

Issue #4779, opened 2025-06-13
Better type Module.apply signature


Daraan:
Following up on #2086 and this recent SO question [How to type hint `flax.linen.Module.apply`'s output correctly?](https://stackoverflow.com/q/79658104/12439683).

I would like to add a PR based on my [answer](https://stackoverflow.com/a/79658693/12439683) and add two `typing.overload` signatures that make it more clear when to expect a `tuple` return value - following the contribution guidelines I am opening this issue first.

Based on #2086, the docstring and 

https://github.com/google/flax/blob/893a66024aad7c9ab0d0c5b6cb22578b8b95ae4c/flax/core/scope.py#L1080-L1083

I am certain that only `mutable=False` never returns a tuple, otherwise it will always be a tuple.

PR #4783, opened 2025-06-18
feat: Add overload signatures to module.apply

Daraan:
# What does this PR do?

The `apply` method returns either the output or a `tuple[output, dict]` depending on `mutable` (in `core.apply`).

https://github.com/google/flax/blob/7d4a288491acec7f24e0ae469b41590fb05dd400/flax/core/scope.py#L1080-L1083

The current type hints are were general in the return type `-> Any | tuple[Any, FrozenVariableDict | dict[str, Any]]:` which can be confusing lead to type-checking errors. A related SO [question](https://stackoverflow.com/q/79658104/12439683) how to type `Module.apply` correctly.

**This PR adds two typing.overload signatures to `Module.apply` to differentiate these cases - when to expect a tuple and when not**

See also #2086 as a related discussion.

Resolves:
#4779

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/linen/module.py
+++ b/flax/linen/module.py
@@ -2085,6 +2085,30 @@ class Module(ModuleBase):
     module = self.clone(_deep_clone=True, _reset_names=True, name=None)
     return module, variables
 
+  @overload
+  def apply(
+    self,
+    variables: VariableDict,
+    *args,
+    rngs: PRNGKey | RNGSequences | None = None,
+    method: Callable[..., Any] | str | None = None,
+    mutable: Literal[False] = False,
+    capture_intermediates: bool | Callable[['Module', str], bool] = False,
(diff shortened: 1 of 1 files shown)


Daraan:
Hi, @IvyZX or @cgarciae or ..., this PR is laying around for a while it would be nice if we could make flax more type safe. A review would be great

vfdev-5:
@Daraan thanks for the PR, but `flax.linen` is mostly in a "bugfix" mode. Even if your PR is small, the OSS CI is green, but typing change can lead to errors in the internal CI that flax maintainers wont have time to fix.

If you would like to help with the project please consider contributing to `flax.nnx` part. A good practice is to open an issue first before coding anything.
Thanks!

PR #4795, opened 2025-06-27
Add a JAX/Flax key concepts doc

IvyZX:


docs_nnx/sharp_bits.ipynb
```diff
@@ -0,0 +1,533 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "929920d0",
+   "metadata": {},
+   "source": [
+    "# 🪄 Flax - Traps and Tricks 🪤\n",
```
8bitmp3: To improve SEO, maybe also call it "The Sharp Bits" for JAX folks? WDYT

8bitmp3:
TODO: @8bitmp3 review

8bitmp3:
@IvyZX @cgarciae If the contents are good, let's merge. Then we can improve the style and formatting.

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+# For simulating multi-device environment
+import os
+os.environ["XLA_FLAGS"] = '--xla_force_host_platform_device_count=8'
+
+import numpy as np
+import jax
+import jax.numpy as jnp
+import flax
+from flax import nnx
+from functools import partial
```
cgarciae:
```suggestion
```python
import numpy as np
import jax
import jax.numpy as jnp
import flax
from flax import nnx
from functools import partial
# For simulating multi-device environment
jax.config.update('jax_num_cpu_devices', 8)
```

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+flax_y = linear(x)
+jax_y = jax_linear(x, linear.kernel.value, linear.bias.value)
+assert jnp.array_equal(flax_y, jax_y)
+```
+
+    <class 'flax.nnx.variablelib.Param'>
+    <class 'jaxlib._jax.ArrayImpl'>
+
+
+## Pytrees
```
cgarciae, line 78: A minor correction in this section is that pytrees don't really know about `Array`s but in the context of Flax it might make sense to focus on them? Else, the generic term would be "leaves".

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+
+Your code likely needs more than one `jax.Array`. A **pytree** is a container structure of multiple pytrees, possibly nested. It is a key and handly concept in the JAX world.
+
+Many things are pytrees: Python dicts, lists, tuples, dataclasses, and more. The key is that a pytree can be "flattened" into multiple children, which are either pytrees or individual `jax.Array`s. JAX allows you to  register your own classes as pytree nodes.
+
+Pytree is the primary data holder in JAX. When JAX transforms see a pytree argument, they automatically trace its internal `jax.Array`s when compiling. Therefore, it's crucial to organize your data as pytrees. [JAX pytree documentation](https://docs.jax.dev/en/latest/working-with-pytrees.html) has a thorough overview on pytrees and JAX APIs to manipulate them.
+
+In Flax, a `Module` is not a pytree, but its state is a pytree. This how we implement Flax transforms (e.g. `flax.nnx.vmap` instead of `jax.vmap`) - the model is first converted to a JAX-compatible pytree dict, then the underlying JAX transforms are called.
```
cgarciae, line 88: As of #4811 this is no longer true 😅.

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+
+In the future, Flax Modules will also become pytrees, allowing us to directly using JAX transforms.
```
cgarciae: The future is now! 😅 (#4811) Should we update this section?
Maybe we can merge this an do an update.

cgarciae (approved):

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+assert arrays[0] == linear
+```
+
+## Traced vs. static data
+
+A pytree *contains* JAX arrays, but a pytree is *more than* its JAX arrays. For example, a dictionary keeps information like the key of every array, and it might contain entries that are not JAX arrays. From JAX's standpoint, all data are one of the two types:
+
+* **Traced** ("dynamic") data: JAX will trace them during compilation and optimize the operations upon them. If they stay inside a pytree argument, `jax.tree.flatten` must return them as leaves. They must be data values (`jax.Array`, Numpy array, scalar, etc).
+
+* **"Static"** data: They stay as simple Python objects that don't get traced by JAX.
```
cgarciae, line 120: Would be good to mention that they must implement `__eq__` and `__hash__`.

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+     [ 0.16276604  0.16813846]
+     [ 0.310975   -0.43336964]]
+
+
+
+```python
+# `nnx.Module` is not a pytree (yet). Therefore, flattening it will only return itself.
+arrays, _ = jax.tree.flatten(linear)
+assert len(arrays) == 1
+assert arrays[0] == linear
```
cgarciae: Could be good to keep the `pytreedef` here and show the use of `unflatten`. In the next section this would be useful to specify that "dynamic" data is accumulated in arrays (maybe we should use "leave" here?) and static data is accumulated in pytreedef.

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+
+    model.dim = 4
+    model.traced_dim.value = Traced<~int32[]>with<DynamicJaxprTrace>
+    Code path based on static data value works fine.
+    Code path based on JAX data value throws error: Attempted boolean conversion of traced array with shape bool[].
+    The error occurred while tracing the function jitted at /var/folders/4c/ylxxyg_n67957jf6616c7z5000gbn1/T/ipykernel_25111/4039468265.py:3 for jit. This concrete value was not available in Python because it depends on the value of the argument state['traced_dim'].value.
+    See https://docs.jax.dev/en/latest/errors.html#jax.errors.TracerBoolConversionError
+
+
+## Abstract arrays and pytrees
```
cgarciae:
```suggestion
## Abstract arrays
```

docs_nnx/key_concepts.md
```diff
@@ -0,0 +1,320 @@
...
+        [38;2;156;220;254mvalue[0m[38;2;212;212;212m=[0mShapeDtypeStruct(shape=(8190, 8190), dtype=float32)
+      [38;2;255;213;3m)[0m
+    [38;2;255;213;3m})[0m
+
+
+Once you have an abstract pytree for your function input or output, it's easier to describe how you want your data to be sharded. You should use such a pytree with sharding information to instruct your checkpoint loading library to load your arrays distributedly. Our checkpointing guide contains [an example of how to do this](https://flax.readthedocs.io/en/latest/guides/flax_gspmd.html#load-a-sharded-model-from-a-checkpoint).
+
+## Distributed computing
+
+Another big use case for abstract pytrees is to tell JAX machinery how you want each array to be sharded during any point of your computation.
```
cgarciae, line 245: Would be nice to introduce concept by itself shortly and only mention that `jax.lax.with_sharding_constrain` accepts pytrees as a comment (hopefully the reader gets the idea by now).

docs_nnx/key_concepts.md
8bitmp3: +1

docs_nnx/key_concepts.md
IvyZX: Done.

docs_nnx/key_concepts.md
IvyZX: Yeah, I kinda wanted to express in a more novice-friendly way. Rephrasing it to be technically correct.

docs_nnx/key_concepts.md
IvyZX: Aha! I quickly updated it to briefly show the latest fact. We can have more elaborate and examples later.

docs_nnx/key_concepts.md
IvyZX: Done.

docs_nnx/key_concepts.md
IvyZX: Good idea, added a few lines to show pytreedef as example of static data.

docs_nnx/key_concepts.md
IvyZX: Done.

docs_nnx/key_concepts.md
IvyZX: True, considering that we will have auto-constraint soon, maybe I will make it brief here. The intention is to keep the bulk in the GSPMD guide.

PR #4814, opened 2025-07-11
Modify NNX to use id(variable) instead of nnx.Variables as dictionary

divyashreepathihalli:
- Updated the check_consistent_aliasing function to use id(variable) as dictionary keys instead of using Variables directly
- Changed the dictionary type annotation from dict[tp.Any, ...] to dict[int, ...] to properly reflect the use of variable IDs as keys
- Added a mapping from variable IDs to variables for generating meaningful error messages
- Enhanced error message generation to show the variable type name (e.g., "Param") instead of just the node ID

### DIFF
--- a/flax/nnx/extract.py
+++ b/flax/nnx/extract.py
@@ -50,7 +50,7 @@ def check_consistent_aliasing(
   """Check for consistent aliasing of nodes when extracting graph."""
   if node_prefixes is None:
     node_prefixes = {}
-  
+
   # Store variable references for error messages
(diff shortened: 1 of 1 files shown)


divyashreepathihalli:
@cgarciae - Can you please take a look?

cgarciae (approved):

cgarciae:
Thanks @divyashreepathihalli! Please run pre-commit:

```bash
pip install pre-commit
pre-commit run --all-files
```

PR #4844, opened 2025-07-23
[WIP] Auto mode spmd sharding utilities and docs

IvyZX:
This PR creates a few helpful utilities for users to do auto mode sharding, and removed a few functions that people are not using actively.

`nnx.sharded_init` allows user to create sharded model with one line, without creating custom `jit` functions and impose sharding constraints. Users can now do:
```python
with jax.sharding.use_mesh(mesh):
  sharded_model = nnx.sharded_init(lambda: nnx.Linear(16, 16, nnx.Rngs(0)))
```

If user want to explicitly create their jitted functions, they can do this instead of explicitly extracting specs and applying contraints:
```python
@jax.jit
def sharded_init_model():
  model = nnx.Linear(16, 16, nnx.Rngs(0))
  model = nnx.shard_model(model)
  return graph.split(model)
```

### DIFF
--- a/flax/nnx/spmd.py
+++ b/flax/nnx/spmd.py
@@ -12,19 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import functools
+from functools import partial
 import typing as tp
 
+from flax.core import meta
 import flax.core.spmd as core_spmd
+from flax.nnx import graph
+from flax.nnx import module
 from flax.nnx import variablelib
 from flax.typing import (
-  Array,
-  ArrayPytree,  # pylint: disable=invalid-name
-  PartitionSpecPytree,  # pylint: disable=invalid-name
   Sharding,
 )
 import jax
-from jax.interpreters import pxla
 from jax.sharding import PartitionSpec
 
 A = tp.TypeVar('A')
@@ -32,12 +31,8 @@ F = tp.TypeVar('F', bound=tp.Callable[..., tp.Any])
 PARTITION_NAME = 'partition_name'
 
 
-class HasSharding(tp.Protocol):
-  sharding: tuple[str | None, ...] | None
-
-
-def _has_sharding(x: tp.Any) -> tp.TypeGuard[HasSharding]:
-  return hasattr(x, 'sharding') and x.sharding is not None
+# Transform axis change helpers
+# ------------------------------------------------------------------------------
 
 
 def add_axis(tree: A, index: int, transform_metadata: tp.Mapping) -> A:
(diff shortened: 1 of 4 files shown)


docs_nnx/guides/flax_gspmd.md
```diff
@@ -10,98 +10,119 @@ jupytext:
...
-import os
-os.environ["XLA_FLAGS"] = '--xla_force_host_platform_device_count=8'
-```
+When you define a Flax variable, you can pass in a metadata field called `sharding_names`, to specify how the underlying JAX array should be sharded. This field should be a tuple of names, each of which refer to how an axis of the array should be sharded.
+
+**You must have an existing device mesh** and create a sharding-annotated `nnx.Variable` within its scope. This allows the result variable to be sharded accordingly on those devices. The device mesh can be your actual accelerator mesh, or a dummy fake CPU mesh like in this notebook.
 
 ```{code-cell} ipython3
-from typing import *
+with jax._src.sharding_impls.use_mesh(auto_mesh):
```
cgarciae:
```suggestion
with jax.sharding.use_mesh(auto_mesh):
```

docs_nnx/guides/flax_gspmd.md
```diff
@@ -110,203 +131,100 @@ class DotReluDot(nnx.Module):
...
-
-+++
-
-## Initialize a sharded model
+class MultiDotReluDot(nnx.Module):
+  def __init__(self, depth: int, num_layers: int, rngs: nnx.Rngs):
+    # Annotate the additional axis with sharding=None, meaning it will be
+    # replicated across all devices.
+    @partial(jax.vmap, spmd_axis_name=None)
+    @nnx.set_transform_axis_name(name=None)
```
cgarciae: why `names=None` here?

cgarciae: `set_transform_axis_name` is a bit less flexible than current `nnx.vmap` which has access to all `in_axes` / `out_axes` so can lift axis other than `0` and can avoid lifting in the `None` case.

flax/nnx/spmd.py
```diff
@@ -152,63 +174,15 @@ def get_named_sharding(tree: A, mesh: jax.sharding.Mesh) -> A:
...
-) -> F:
-  return variablelib.with_metadata(
-    initializer,
-    sharding=sharding,
-    mesh=mesh,
-    **metadata,
-  )
+def get_abstract_sharding(init_fn, mesh):
+  with jax._src.sharding_impls.use_mesh(mesh):
+    abs_model = jax.eval_shape(init_fn)
```
cgarciae: In our internal APIs we should assume the user might be using reference sharing and use `nnx.eval_shape` and `nnx.state` and `nnx.update`.

docs_nnx/guides/flax_gspmd.md
IvyZX: This is a temporary API compatibility diff between internal and external JAX that I'll deal with.

docs_nnx/guides/flax_gspmd.md
IvyZX: Agreed that we should just use `nnx.vmap` for auto mode.

flax/nnx/spmd.py
IvyZX: Fixed.

cgarciae (approved):

PR #4860, opened 2025-07-31
Fix string key preservation in replace_by_pure_dict

### LINKED ISSUE #4858: Unconditional cast to integer breaks models with integer keys
```python
from flax import nnx

class MyModule(nnx.Module):
    def __init__(self, rngs):
        self.d = {"0": nnx.Linear(10, 10, rngs=rngs), "1": nnx.Linear(10, 10, rngs=rngs)}

m = MyModule(nnx.Rngs(0))
state = nnx.state(m)
state.replace_by_pure_dict(state.to_pure_dict())
# ValueError: key in pure_dict not available in state: ('d', 0, 'bias')
```

This is caused by the unconditional cast introduced in #4317.

mohsinm-dev:
# What does this PR do?

The `replace_by_pure_dict` function was unconditionally converting string keys to integers, which broke models that intentionally use string keys like "0", "1".

This fix tries exact key match first to preserve user intent, then falls back to integer conversion for checkpoint restoration compatibility.

## Problem

When users deliberately choose string keys in their models, the function converts them to integers even when not needed. This caused key mismatch errors because the state expected string keys, but the function was looking for integer keys.

## Solution

Use a two-step matching approach:
1. Try to find the exact key first (preserves the user's string keys)
2. If not found, try converting strings to integers (handles checkpoint restoration)

## Testing

- Fixes the exact issue reported in #4858
- Maintains compatibility with checkpoint restoration functionality
- All existing tests continue to pass
- No breaking changes introduced

Fixes #4858

## Checklist
- [ ] This PR fixes a minor issue (e.g., typo or small bug) or improves the docs (you can dismiss the other
      checks if that's the case).
        link): https://github.com/google/flax/issues/4858

### DIFF
--- a/flax/nnx/statelib.py
+++ b/flax/nnx/statelib.py
@@ -510,10 +510,16 @@ def replace_by_pure_dict(
     replace_fn = lambda x, v: x.replace(v) if hasattr(x, 'replace') else v
   current_flat = dict(to_flat_state(state))
   for kp, v in traversals.flatten_mapping(pure_dict).items():
-    kp = tuple(map(try_convert_int, kp))
-    if kp not in current_flat:
-      raise ValueError(f'key in pure_dict not available in state: {kp}')
-    current_flat[kp] = replace_fn(current_flat[kp], v)
+    # Try exact match first, then integer conversion
+    if kp in current_flat:
+      matched_key = kp
+    else:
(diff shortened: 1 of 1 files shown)


IvyZX (approved):
Thanks for making this change and sorry about the wait!

I believe the current CI error is unrelated with this PR. Will try to merge in.

PR #4876, opened 2025-08-06
add Rngs random helpers

cgarciae:
# What does this PR do?

Adds shortcut methods for `jax.random.*` functions on `Rngs` and `RngStream`.

Example:
```python
rngs = nnx.Rngs(0)
x = jax.random.uniform(rngs(), (2, 3))
```
can now be simplied to:
```python
rngs = nnx.Rngs(0)
x = rngs.uniform((2, 3))
```

### DIFF
--- a/flax/nnx/rnglib.py
+++ b/flax/nnx/rnglib.py
@@ -83,6 +83,213 @@ class RngStream(Pytree):
       key = jax.random.split(key, split)
     return type(self)(key, tag=self.tag)
 
+  if tp.TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


IvyZX:
Does it work only with `default`? Would it be helpful to have a shortcut for other tags?

cgarciae:
@IvyZX  the main implementation is for `RngStream`s, so you can also do:

```python
rngs = nnx.Rngs(params=0)
x = rngs.params.uniform((2, 3))
```
but they are also available in `Rngs` via `default`.

PR #4891, opened 2025-08-14
Fix nnx.tabulate crash with empty dict/None values (fixes #4889)

### LINKED ISSUE #4889: Edge case in nnx.tabulate when nnx.Module stores empty dictionary
### System information
```
Name: flax
Version: 0.11.1
---
Name: jax
Version: 0.7.0
---
Name: jaxlib
Version: 0.7.0
```
- Python version: 3.11

### Steps to reproduce:
```python
import jax.numpy as jnp
from flax import nnx


class Model(nnx.Module):
    def __init__(self):
        self.foo = {}  # This fails with tabulate
        # self.foo = None  # This fails with tabulate
        # self.foo = 1  # This works
        # self.foo = {"bar": 1}  # This works

    def subroutine(self, foo, x):
        return x
        
    def __call__(self, x):
        return self.subroutine(self.foo, x)


model = Model()
inputs = jnp.zeros((1, 1024))

# This works
output = model(inputs)  # Shape: (1, 1024)

# This works with depth=0
print(nnx.tabulate(model, inputs, depth=0))  # ✓

# This fails with depth >= 1
print(nnx.tabulate(model, inputs, depth=1))  # ✗ AssertionError in flax.nnx.summary's _unflatten_to_simple_structure
```
Output:
```
                   Model Summary
┏━━━━━━┳━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┓
┃ path ┃ type  ┃ inputs          ┃ outputs         ┃
┡━━━━━━╇━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━┩
│      │ Model │ float32[1,1024] │ float32[1,1024] │
├──────┼───────┼─────────────────┼─────────────────┤
│      │       │                 │           Total │
└──────┴───────┴─────────────────┴─────────────────┘

             Total Parameters: 0 (0 B)

  File "/mnt/c/Users/admin/tmp/nnx_tabulate_issue.py", line 29, in <module>
    print(nnx.tabulate(model, inputs, depth=1))  # ✗ AssertionError in flax.nnx.summary's _unflatten_to_simple_structure
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.12/site-packages/flax/nnx/summary.py", line 386, in tabulate
    inputs_repr += _as_yaml_str(input_args)
                   ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.12/site-packages/flax/nnx/summary.py", line 535, in _as_yaml_str
    value = _maybe_pytree_to_dict(value)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.12/site-packages/flax/nnx/summary.py", line 497, in _maybe_pytree_to_dict
    return _unflatten_to_simple_structure(path_leaves)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/admin/.local/lib/python3.12/site-packages/flax/nnx/summary.py", line 523, in _unflatten_to_simple_structure
    assert path[-1] == len(cursor)
           ^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

mohsinm-dev:
Fixes #4889 by handling gaps in sequence indices when JAX tree flattening omits empty containers.

# What does this PR do?

This PR fixes a crash in `nnx.tabulate` when called with `depth >= 1` on modules that contain empty dictionaries `{}` or `None` values.

## Problem
When `nnx.tabulate` processes function arguments that include empty containers (like `{}`), JAX's tree flattening skips these empty containers completely. This creates gaps in the sequence indices, causing an assertion error in the `_unflatten_to_simple_structure` function.

For example, with input `({}, array)`, JAX produces `[((1,), array)]` instead of `[((0,), {}), ((1,), array)]`. The code expected index 0 but got index 1, causing `assert 1 == 0` to fail.

## Solution
Replace the problematic assertion with logic that handles index gaps by filling missing positions with `None` values:

```python
# Before (crashed):
assert path[-1] == len(cursor)
cursor.append(value)

# After (works):
while len(cursor) <= path[-1]:
    cursor.append(None) 
cursor[path[-1]] = value
```

## Test Case
```python
class Model(nnx.Module):
    def __init__(self):
        self.foo = {}  # This used to crash tabulate
    def __call__(self, x):
        return x

# This now works (was failing before):
nnx.tabulate(Model(), jnp.zeros((1, 10)), depth=1)
```

## Checklist
- [x] This change is discussed in a Github issue/discussion (please add a link).
- [ ] The documentation and docstrings adhere to the documentation guidelines.
- [ ] This change includes necessary high-coverage tests. (No quality testing = no merge\!)

### DIFF
--- a/flax/nnx/summary.py
+++ b/flax/nnx/summary.py
@@ -40,14 +40,14 @@ try:
 except ImportError:
   in_ipython = False
 
-# Ensure None is dumped as the string 'None' (and not YAML 'null') for clarity
-try:
-  yaml.SafeDumper.add_representer(
-    type(None),
-    lambda dumper, data: dumper.represent_scalar('tag:yaml.org,2002:str', 'None'),
-  )
-except Exception:
+# Custom YAML dumper to represent None as 'None' string (not YAML 'null') for clarity
+class NoneDumper(yaml.SafeDumper):
   pass
+
+NoneDumper.add_representer(
+  type(None),
+  lambda dumper, data: dumper.represent_scalar('tag:yaml.org,2002:str', 'None'),
+)
 
 class SizeBytes(typing.SizeBytes):
   def __repr__(self) -> str:
@@ -593,9 +593,10 @@ def _as_yaml_str(value) -> str:
   value = _maybe_pytree_to_dict(value)
 
   file = io.StringIO()
-  yaml.safe_dump(
+  yaml.dump(
     value,
     file,
+    Dumper=NoneDumper,
     default_flow_style=False,
     indent=2,
     sort_keys=False,


samanklesaria:
Thanks for the fix! A couple comments:

1. The test case you're using above doesn't actually test for the error observed in #4889 because `self.foo` is never referred to by the `__call__` method (which is all the `tabulate` function is looking at). The `_unflatten_to_simple_structure` function never gets called. As written, `tabulate` works fine on your test case even without your fix. Instead, I'd recommend a test case like the following:
```python
class Model(nnx.Module):
    def subroutine(self, foo, x):
        return x

    def __call__(self, x):
        return self.subroutine({}, x)
        
model = Model()
nnx.tabulate(model, jnp.zeros((1, 8)), depth=1)
 ```
 
 This breaks using the current version of `tabulate`, but works fine with your fix, giving inputs `null` and `float32[1,8]` for the call to `subroutine`. 
 
 2. Your fix only applies when empty dictionary arguments are given before a non-empty argument. If we switch the test case above to the following, the empty input to `subroutine` is ignored by the call to `tabulate`. 
 ```python
class Model(nnx.Module):
    def subroutine(self, foo, x):
        return x

    def __call__(self, x):
        return self.subroutine(x, {})
        
model = Model()
nnx.tabulate(model, jnp.zeros((1, 8)), depth=1)
 ```
This dropping of the empty dictionary argument also occurs when it is passed as a keyword argument instead of a positional one. 

3. Could you add test cases like these to the `tests/nnx/summary_test.py` file?

4. Both `None` and empty dictionaries get represented as `null` in the output table. Ideally, we'd see `None` for None arguments and `{}` for empty dictionary arguments. Would this kind of behavior be possible?

File: flax/nnx/summary.py [hunk 1]
```diff
@@ -520,13 +561,28 @@ def _unflatten_to_simple_structure(xs: list[tuple[tuple[tp.Any, ...], tp.Any]]):
...
+          cursor.append(None)
       cursor[path[-1]] = value
     else:
       assert isinstance(cursor, dict)
       cursor[path[-1]] = value
+  # If original is a sequence and result is a list, pad trailing items
+  if isinstance(original, (list, tuple)) and isinstance(result, list):
+    for i in range(len(result), len(original)):
+      slot = original[i]
+      result.append(_to_simple(slot) if isinstance(slot, (list, tuple, dict)) else slot)
```
samanklesaria: Doesn't `_to_simple` already take care of passing through non (list, tuple dict) items? Seems like we don't need the isinstance check here.

tests/nnx/summary_test.py
```diff
@@ -97,6 +97,73 @@ def __call__(self, x, x2):
...
+
+    model = Model()
+    out = nnx.tabulate(
+      model, jnp.zeros((1, 8)), depth=2, console_kwargs=CONSOLE_TEST_KWARGS
+    )
+    # Distinguish {} and None in output
+    self.assertIn('{}', out)
+    self.assertIn('None', out)
+
+  def test_tabulate_original_user_repro(self):
```
samanklesaria: I feel like this needs a better name. Maybe something like 'test_tabulate_empty_dict_property'?

samanklesaria (commented):
Thanks for the updates! Everything looks good to me.

flax/nnx/summary.py
```diff
@@ -40,6 +40,15 @@
 except ImportError:
   in_ipython = False
 
+# Ensure None is dumped as the string 'None' (and not YAML 'null') for clarity
```
samanklesaria: Unfortunately, I just realized that this will affect any other use of the yaml library as well. If `flax` users try to dump some yaml with this library loaded, they'd see Nones instead of nulls unexpectedly. We have to subclass the SafeDumper to keep these changes localized:

```python
class NoneDumper(yaml.SafeDumper):
  pass

NoneDumper.add_representer(
    type(None),
    lambda dumper, data: dumper.represent_scalar('tag:yaml.org,2002:str', 'None')
  )
```

Then, instead of calling `yaml.safe_dump`, we'd need to call the more generic `yaml.dump_all` with the `Dumper=NoneDumper` kwarg.

mohsinm-dev: Make sense. Also, we're dumping a single document (the value), so I think `yaml.dump()` is semantically more correct than `yaml.dump_all()`

samanklesaria (approved):

PR #4893, opened 2025-08-14
Sharding API improvements (non breaking)

IvyZX:
A splitted version of #4844 that contains all the sharding API benefits but is not breaking:

* Allow Flax to recognize the new style JAX mesh context (which based off `AbstractMesh`).

* Renamed annotation from `sharding` to `sharding_names`, so that `variable.sharding` can just point to the JAX sharding of the actual value.

* Changed CI to run everything on 4 fake CPU devices, so that we no longer to distinguish between single device and multi device tests.

* Removed a few redundant functions like `nnx.with_sharding_constraint`.

### DIFF
--- a/examples/gemma/utils.py
+++ b/examples/gemma/utils.py
@@ -24,6 +24,7 @@ import numpy as np
 from jax.experimental import mesh_utils
 from transformer import TransformerConfig, Transformer
 
+
 from flax import nnx
 from flax.training import train_state
 
@@ -32,6 +33,12 @@ if TYPE_CHECKING:
 
 Dtype = Any
 Shape = tuple[int, ...]
+
+# JAX version compatibility
+if hasattr(jax.sharding, 'use_mesh'):
+  set_mesh = jax.sharding.use_mesh
+else:
+  set_mesh = jax.set_mesh
 
 
 class TrainState(train_state.TrainState):
@@ -165,7 +172,7 @@ def setup_initial_state(
     return state
 
   # Initialization
-  with jax.set_mesh(mesh):
+  with set_mesh(mesh):
     state = sharded_init()
 
   state_sharding = nnx.get_named_sharding(state, mesh)
--- a/flax/nnx/spmd.py
+++ b/flax/nnx/spmd.py
@@ -12,32 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import functools
 import typing as tp
 
 import flax.core.spmd as core_spmd
-from flax.nnx import variablelib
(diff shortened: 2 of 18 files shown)


examples/gemma/utils.py
```diff
@@ -33,6 +34,12 @@
 Dtype = Any
 Shape = tuple[int, ...]
 
+# JAX version compatibility
+if hasattr(jax.sharding, 'use_mesh'):
+  set_mesh = jax.sharding.use_mesh
+else:
+  set_mesh = jax.set_mesh
```
cgarciae, line 41: shouldn't it be backwards?
```suggestion
if hasattr(jax, 'set_mesh'):
  set_mesh = jax.set_mesh
else:
  set_mesh = jax.sharding.use_mesh
```

IvyZX: Well this is effectively equivalent - a released JAX will either have `jax.set_mesh` or `jax.sharding.use_mesh`. If JAX releases today or tomorrow I will roll forward and remove this.

cgarciae: Isn't `jax.sharding.use_mesh` still going to be there?

IvyZX: No it's deleted in current head

flax/nnx/spmd.py
```diff
@@ -152,63 +166,16 @@ def get_named_sharding(tree: A, mesh: jax.sharding.Mesh) -> A:
...
-  mesh: tp.Optional[jax.sharding.Mesh] = None,
-  **metadata: tp.Any,
-) -> F:
-  return variablelib.with_metadata(
-    initializer,
-    sharding=sharding,
-    mesh=mesh,
-    **metadata,
-  )
+def get_abstract_model(init_fn, mesh):
```
cgarciae, line 173: Wondering if we could extract the `ShapeDtypeStruct` mapping procedure into its own `get_shape_dtypes` function such that `get_abstract_model` is not needed and the user could express the same as:

```python
with jax.set_mesh(mesh):
  gdef, abs_state = nnx.split(jax.eval_shape(init_fn))
  abs_state = nnx.get_shape_dtypes(abs_state)
```
Currently `get_abstract_model` feels a bit magical e.g. hard to know what its doing.

examples/gemma/utils.py
cgarciae: oh I see.

flax/nnx/spmd.py
IvyZX: This is a good idea. I'll try to use this in the gspmd guide.

PR #4904, opened 2025-08-22
Update optimizer.py to support masked variable from optax.

### LINKED ISSUE #4901: `optax.contrib.muon` does not seem to work with nnx
Hi!
I wanted to test out the Muon optimizer, which has been implemented in `optax`: [Muon](https://optax.readthedocs.io/en/latest/api/contrib.html#muon). It does however not seem to work with flax.nnx. Here is a short example:

```
class Test(nnx.Module):
    def __init__(self):
        self.a = nnx.Param(jnp.array([1.0]))

    def __call__(self):
        return self.a.sum()
    
tx = optax.contrib.muon(0.1) # does not work
#tx = optax.adam(0.1) # works
model = Test()

optimizer = nnx.Optimizer(model, tx, wrt=nnx.Param) #TypeError
```

It produces the error:
TypeError: tuple indices must be integers or slices, not ellipsis

ywrt:
Per https://github.com/google/flax/issues/4901 the current code doesn't work with masked variable introduced by optax.partition() because it assumes that the value is an array.

This change just uses x.value to get the value directly, instead of using x[...] to get in implicitly.

# What does this PR do?

Fixes #4901

## Checklist
- [X ] This PR fixes a minor issue (e.g.: typo or small bug) or improves the docs (you can dismiss the other
      checks if that's the case).

### DIFF
--- a/flax/nnx/training/optimizer.py
+++ b/flax/nnx/training/optimizer.py
@@ -53,7 +53,7 @@ class OptVariable(OptState):
 def to_opt_state(tree):
   def _to_opt_state(x):
     if isinstance(x, Variable):
-      opt_state = OptVariable(x[...], **x.get_metadata())  # type: ignore
+      opt_state = OptVariable(x.value, **x.get_metadata())  # type: ignore
     else:
       opt_state = OptArray(x)
     return opt_state


ywrt:
Tests don't pass, so this simple fix can't be right. Withdrawn.

cgarciae:
@ywrt not sure why they don't pass but I think you have the right idea

cgarciae:
oh its just mypy, I think you just have to rebase
I ran them locally and it looks good

cgarciae:
seems like `main` is broken, will send a fix

vfdev-5:
@ywrt can you please rebase your PR to move it forward. Thanks !

ywrt:
Ok, I think this is rebased now?

vfdev-5:
@ywrt can you please execute the following on your machine to make this PR up to date with `main` branch:
```bash
git checkout patch-1
# assuming "origin" is pointing to git@github.com:google/flax.git
git pull origin main -r
# assuming "fork" is pointing to your fork: git@github.com:ywrt/flax.git
git push fork patch-1 -f
```
Thanks!

PR #4912, opened 2025-08-26
merge no copy Variables

cgarciae:
# What does this PR do?

Adds `copy_variables` flag to `merge` which defaults to `False`, meaning now merge doesn't create a copy of the Variables when merging. If you call

```
m1 = SomeModule()
m2 = nnx.merge(*nnx.split(m1))
```
then `m1` and `m2` will be different structures but hold the same Variables. This makes `split` and `merge` consistent with jax's `flatten` and `unflatten` in that they don't modify the leaves themselves.


flax/nnx/graph.py
```diff
@@ -1835,6 +1835,7 @@ def merge(  # type: ignore[invalid-annotation]
       _state,
       index_ref=self.index_ref,
       outer_index_outer_ref=outer_index_outer_ref,
+      copy_variables=True,
```
IvyZX, line 1838: `copy_variables` is kinda verbose. Can we call it `clone` or `copy`?

IvyZX (approved):

flax/nnx/graph.py
cgarciae: using `copy` now

PR #4913, opened 2025-08-27
Updates in gemma example: grain dataflow, simplified sharding rules

vfdev-5:


examples/gemma/configs/benchmarks/gemma3_1b_grain.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2024 The Flax Authors.
```
vfdev-5, line 1: Let's rename this file to gemma3_1b_*.py

vfdev-5:
Let's use nnx.set_view and try to use sow under a config

examples/gemma/input_pipeline_grain.py
```diff
@@ -0,0 +1,275 @@
...
+
+import grain
+import jax
+import numpy as np
+import tensorflow_datasets as tfds
+import tokenizer
+from grain.python import MapTransform, MultiprocessingOptions
+from grain.experimental import pick_performance_config
+
+if tp.TYPE_CHECKING:
```
IvyZX: Can you remove the TYPE_CHECKING brackets for imports? Just always import them. I think this is incompatible with our internal pipeline in some way.

Please remember to remove them in all occurrences, thanks!

vfdev-5: Removed them by moving configs into separate files to avoid circular deps

Issue #4924, opened 2025-09-04
Intermediate value capture API via JAX's `hijax.Box`


IvyZX:
JAX has this new `hijax.Box` mechanism that can insert arbitrary values during forward & backward passes. This can be a good alternative to the current `sow` and `perturb` APIs on Flax level.

The feature is not complete yet (e.g., not yet work with vmap/scan, and in some other corner cases). Gonna do some prototying first.

PR #4925, opened 2025-09-04
Add intermediate value captures based off hijax.Box

IvyZX:
Master tracker: #4924

### DIFF
--- a/flax/nnx/capture.py
+++ b/flax/nnx/capture.py
@@ -1,0 +1,73 @@
+# Copyright 2024 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import annotations
+
+import contextlib
+import dataclasses
+import threading
+
+from jax._src.hijax import Box
+
+from flax.nnx.module import Module, CAPTURE_BOX_NAME
+from flax import traverse_util
+
+
+@dataclasses.dataclass
+class CaptureContext(threading.local):
+  model: Module
+
+  def extract_boxes(self, remove: bool = True):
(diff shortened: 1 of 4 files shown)


flax/nnx/capture.py
```diff
@@ -0,0 +1,73 @@
...
+    if self not in CAPTURE_CONTEXTS:
+      raise ValueError('`extract_boxes` only works inside the capture context.')
+    return extract_boxes(self.model, remove=remove)
+
+
+CAPTURE_CONTEXTS = []
+
+
+@contextlib.contextmanager
+def capture_intermediates(model: Module):
```
cgarciae: can we use a different name? I've been wanting to use `capture_intermediate` to mimic Linen's `apply(..., capture_intermediates=True)`.

flax/nnx/capture.py
```diff
@@ -0,0 +1,73 @@
...
+  context = CaptureContext(model)
+  add_boxes(context.model)
+  CAPTURE_CONTEXTS.append(context)
+  try:
+    yield context
```
cgarciae: I was thinking we yield an empty `boxes = {}` dictionary here and then we automatically fill it on exit? This way the API reduce to:

```python
with nnx.capture_intermediates(model) as intms:
  y = model(x)
intms # ready to use outside
```

flax/nnx/capture.py
IvyZX: Renamed to `capture_intms`. Or maybe just `nnx.capture` is better?

I'm also thinking about `enable_capture`, since this have to work with calling `self.capture_.*` methods within model.

flax/nnx/capture.py
IvyZX: Done.

flax/nnx/module.py
```diff
@@ -77,6 +80,40 @@ class Module(Pytree, metaclass=ModuleMeta):
     >>> y = model(x)
   """
 
+  def capture_fwd(self, name: str, value: jax.Array, prefix = ''):
```
cgarciae, line 83: Can we use `sow` and `perturb` instead of these new APIs? I think both APIs could be condition on wether `CAPTURE_BOX_NAME` exists and append the sowed values there or use the current behavior if not.

IvyZX: I think `sow` and `perturb` are pretty confusing names, especially `perturb`. I actually want this to be a chance for us to rename it, at least to be `sow` and `sow_bwd`. 

Regardless, I think it's a good idea to merge into one API, but this box thing isn't quite ready to use for now. I could merge them when they are more mature (like, supports vmap & scan).

PR #4929, opened 2025-09-06
finish pytree guide

cgarciae:
# What does this PR do?

### DIFF
--- a/docs_nnx/guides/pytree.md
+++ b/docs_nnx/guides/pytree.md
@@ -8,59 +8,158 @@ jupytext:
     jupytext_version: 1.13.8
 ---
 
+# Module & Pytree
+
++++
+
+Flax NNX's Modules are by default registered as JAX Pytrees, this allows using them throughout most of JAX APIs but in particular JAX transforms and the `jax.tree.*` functions. Thanks to the pytree protocol a simple NNX program might look like this:
+
 ```{code-cell} ipython3
 from flax import nnx
 import jax
 import jax.numpy as jnp
-import dataclasses
 
-#----------------------
-# helper functions
-#----------------------
-def pytree_structure(pytree, title='pytree structure'):
-  print(f"{title}:")
-  for path, value in jax.tree.leaves_with_path(pytree):
-    print(f"- pytree{jax.tree_util.keystr(path)} = {value!r}")
+class Linear(nnx.Module):
+  def __init__(self, din, dout, rngs: nnx.Rngs):
+    self.din, self.dout = din, dout
+    self.kernel = nnx.Param(rngs.normal((din, dout)))
+
+rngs = nnx.Rngs(0)
+weights = Linear(2, 3, rngs=rngs)
+
+@jax.jit
+def forward(weights, x):
+  return x @ weights.kernel
+
+y = forward(weights, x=rngs.uniform((5, 2)))
+print(f"{y.shape = }")
 ```
 
-## Pytree
+Here `weights`, of type `Linear`, was able to be passed directly to the `jit`-ed function `forward`. Throughout the rest of this guide we will try to answer the questions:
+1. What are pytrees? 
+2. How does NNX implement pytrees?
+
++++
+
+## Pytrees 101
+Most modern ML models have too many Arrays for users to pass around individually, to deal with this JAX developed a way to track Array data in nested structures that still allowed caching for compilation: Pytrees. JAX pytrees are tree structures made of python objects that can be recursively traversed in order to collect an ordered list of leaves and a definition of the tree structure, this is done via the `jax.tree.flatten` function. Most common pytrees are native python containers like `list`, `dict`, and `tuple`, but interestingly it also include `None`. The example bellow shows how to collect all the integer leaves from a nested structure using `flatten`:
+
+```{code-cell} ipython3
+pytree = [
+  {'a': 1},
+  {
+    'b': 2,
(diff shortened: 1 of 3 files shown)


docs_nnx/guides/pytree.md
```diff
@@ -8,60 +8,159 @@ jupytext:
...
+jax.tree_util.register_pytree_node(Foo, flatten_foo, unflatten_foo)
+
+foo = Foo()
+leaves, treedef = jax.tree.flatten(foo)
+print(f"leaves = {leaves}")
+print(f"treedef = {treedef}")
+```
+
+Notice that `'hi'` does not appear in the leaves because `c` is defined as static, but you can see it as part of the `PyTreeDef` structure.
+
```
IvyZX, line 103: Before diving into details of pytree, I kinda want there to be a line explaining *why* we care about pytrees in JAX ecosystem. 
Even a simple mention like saying "JAX uses pytrees to contain and track array data, to optimize its compilations" would do.

docs_nnx/guides/pytree.md
```diff
@@ -8,60 +8,159 @@ jupytext:
...
+  print(f"{title}:")
+  path_leaves, treedef = jax.tree.flatten_with_path(pytree)
+  for path, value in path_leaves:
+    print(f" - pytree{jax.tree_util.keystr(path)} = {value!r}")
+
 pytree_structure(pytree)
 ```
 
-### Attribute Annotations
+As you can see above, only the `data` paths appear in the leaves. However, its very verbose to have to define `static` and `data` for each attribute so `Pytree` has sensible defaults so you can mostly just remove them from them and it will work.
```
IvyZX: nit: `However, its very verbose to have to define `static` and `data` for each attribute, so `Pytree` has sensible defaults. You can remove most of them and it will just work:`

docs_nnx/guides/pytree.md
```diff
@@ -72,18 +171,36 @@ print(f"""
...
 """)
 ```
 
-* remove mixed
-* error on nnx.data/nnx.static in pytrees
+### When to use explicit annotations?
 
 +++
 
+There are cases were you do want to explicitely annotate the attributes to avoid ambiguety or protect yourself against possible edge cases. These include constraining input arguments which might have unexpected types, forcing attributes as data when their type is not treated as data by default, or using `nnx.static` as a way to assert the attribute should not contain data.
```
IvyZX: nit: ambiguety -> ambiguity

docs_nnx/guides/pytree.md
```diff
@@ -72,18 +171,36 @@ print(f"""
...
+    self.bias = nnx.data(None)  # optional values that can be data
+    if use_bias:
+      self.bias = nnx.Param(jnp.array(0.0))
+
+pytree = Bar(1.0, True)
+pytree_structure(pytree)
+```
+
 ### Class Annotations
+Annotations can also be added at the type level via `nnx.Static` and `nnx.Data`, this is useful for creating `dataclasses` but the mechanism also works for regular classes.
```
IvyZX: nit:  `, this` -> `. This`

docs_nnx/guides/pytree.md
```diff
@@ -132,13 +236,14 @@ class Foo(nnx.Pytree):
...
-pytree.b = nnx.data(42)     # annotation to override status
-pytree.c = nnx.static(0.5)  # annotation to override status
+pytree.a = "🤔"  # data status doesn't change
+pytree.b = nnx.data(42)     # explicit annotation overrides status to data
+pytree.c = nnx.static(0.5)  # explicit annotation overrides status to static
 pytree_structure(pytree, "updated")
 ```
 
 ### Attribute checks
+`Pytree` has a variety of checks to prevent a common class of errors in JAX, this includes checking for Arrays being assigned to new `static` attributes:
```
IvyZX: nit:  `, this` -> `. This`

docs_nnx/guides/pytree.md
```diff
@@ -196,64 +306,119 @@ foo = Foo()
...
+def f(m):
+  print(f"Inside: {m.left is m.right = }")
+  return m
+
+m = f(m)
+
+print(f"After:  {m.left is m.right = }")
+```
+
+### pytree=False
```
IvyZX: Let's call the title "Backward compatibility" or "Turning off pytree registration"

docs_nnx/guides/pytree.md
```diff
@@ -196,64 +306,119 @@ foo = Foo()
...
 
+## Module
+
++++
+
+NNX Modules are `Pytree`s that have two additional methods for traking intermediate values: `sow` and `perturb`.
+
++++
+
 ### sow
```
IvyZX, line 420: Should we move sow and perturb into a separate guide? This will make it easier when we move to boxes and capture context.

IvyZX (approved):
Thanks! A few nits but overall great. Approved to unblock.

docs_nnx/guides/pytree.md
cgarciae: Done!

docs_nnx/guides/pytree.md
cgarciae: Thanks!

docs_nnx/guides/pytree.md
cgarciae: Would be a good fit for a Capture Intermediates guide. Adding it here for now.

PR #4948, opened 2025-09-15
Add compute_flops and compute_vjp_flops options to `nnx.tabulate`. 

### LINKED ISSUE #4883: compute_flops and compute_vjp_flops for nnx.tabulate
In Flax 0.11.1 it seems that `nnx.tabulate` doesn't support `compute_flops` and `compute_vjp_flops` like Linen did. I would like to track this feature or hear about alternatives.

samanklesaria:
Closes #4883


flax/nnx/summary.py
```diff
@@ -174,11 +178,27 @@ def _to_dummy_array(x):
...
+      call_info.flops = _get_flops(method, obj, *args, **kwargs)
+    if compute_vjp_flops:
+        call_info.vjp_flops = _get_vjp_flops(method, obj, *args, **kwargs)
     return out
 
   return method_wrapper
 
 
+def _get_flops(fn, *args, **kwargs):
+  e = nnx.jit(fn).lower(*args, **kwargs).compile()
```
vfdev-5:
```suggestion
  e = nnx.jit(fn).lower(*args, **kwargs)
```

flax/nnx/summary.py
```diff
@@ -174,11 +178,27 @@ def _to_dummy_array(x):
...
+  cost = e.cost_analysis()
+  return 0 if cost is None or 'flops' not in cost else cost['flops']
+
+def _get_vjp_flops(fn, *args, **kwargs):
+  g = nnx.grad(fn)
+  e = nnx.jit(g).lower(*args, **kwargs).compile()
```
vfdev-5:
```suggestion
  e = nnx.jit(g).lower(*args, **kwargs)
```

flax/nnx/summary.py
```diff
@@ -174,11 +178,27 @@ def _to_dummy_array(x):
...
   return method_wrapper
 
 
+def _get_flops(fn, *args, **kwargs):
+  e = nnx.jit(fn).lower(*args, **kwargs).compile()
+  cost = e.cost_analysis()
+  return 0 if cost is None or 'flops' not in cost else cost['flops']
+
+def _get_vjp_flops(fn, *args, **kwargs):
+  g = nnx.grad(fn)
```
vfdev-5: Please confirm how this is done in the linen implementation.

samanklesaria:
The implementation used currently will include duplicate rows in the table if one method calls another. Say `f` calls `g`. The implementation will first call `jit(f)` (to be able to call `cost_analysis` on the result), then call `jit(g)` (once again to form a cost analysis). But when we jit `f`, we also will end up tracing `g` a second time. And each time we trace `g`, we add rows to the tabulation table. So: to avoid duplicating rows in the table, we *don't* want to call `jit` on each of the original functions. Instead, can work around this by replacing all the calls that `f` makes to `g` with calls that call the jitted version of `g`. This is almost what is already done by the tabulate function (the only difference is that we modify the functions to add entries to our tabulation table rather than jitting them). Specifically, the current code replaces `f` with `get_method_wrapper(f)` and `g` with `get_method_wrapper(g)`. I advocate replacing `f` with its jitted version instead, and then using these jits to both get the output shapes like we currently do (which is what `eval_shape` does internally) as well as getting the flops if they are requested.

Let's start with a simple pure-jax implementation of this idea. The following function takes a jitted function and gets both its output shape and its flops. The `output_info` actually contains both shape and dtype.

```python
import jax.numpy as jnp
import jax
from functools import partial, wraps

def get_flops_and_shape(jitted, *args, **kwargs):
  e = jitted.lower(*args, **kwargs)
  cost = e.cost_analysis()
  result = 0 if cost is None or 'flops' not in cost else cost['flops']
  return (result, e.out_info)
```

We can use this function to get the output shapes and flops of all the functions in a dictionary. Now, in nnx, we'll be doing this to function attributes of an object rather than a dictionary, but it's easier to understand when we work with dictionaries for now. Instead of calling `self.g` like we would in a class setting, we will have to use `self['g']`.  Here is a simple "object" with two methods: `f` and `g`. 

```python
def g(self, x):
    print("TRACING G")
    return x + jnp.ones(2)

def f(self, x):
    print("TRACING F")
    return self['g'](x + jnp.ones(2))

env = {'f': f, 'g': g}
```

Now, to get both the output shapes and flops for each of these methods (which is what the `tabulate` function does), we can do the following steps:


- Step 1: Argument saving
Modify the methods to save their Tracer arguments when they get traced. 

```python
ARGS = {}
def argsave(f):
    n = f.__name__
    @wraps(f)
    def wrapper(env, x):
        ARGS[n] = x
        return f(env, x)
    return wrapper

env2 = {k: argsave(v) for (k,v) in env.items()}
```

- Step 2: Jitting
Calculate jitted versions of each of the modified methods. Note that `f` will be using the `jits` dictionary to resolve its call to `g`, so we won't get the duplicate tracing issue here. 

```python
jits = {}
for (k,v) in env2.items():
    jits[k] = jax.jit(partial(v, jits))
```

Step 3: Tracing
We can perform a single trace through `f` to save the Tracer arguments associated with both calls to `f` and `g`. 

```python
jits['f'].trace(jnp.ones(2))
```

Step 4: Get Flops and Output Info

Now, we can apply the `get_flops_and_shape` function we defined originally to each jitted method:

```python
print({k: get_flops_and_shape(jits[k], v) for (k,v) in ARGS.items()})
```

We're done! The same idea can be applied to flax in its `tabulate` function. We first turn the objects into dictionaries as follows:
```python
def create_obj_env(object_types):
  "Turn a set of object types into a dictionary mapping (type, method name) pairs to methods"
  result = {}
  for obj_type in object_types:
    for name, top_method in inspect.getmembers(obj_type, inspect.isfunction):
      if not name.startswith('_') or name == '__call__':
        result[(obj_type, name)] = top_method
  return result
```

As the methods are defined in terms of `self.g` rather than `self['g']`, the argument saving wrapper looks a little different:

```python 
def argsave(tracer_args, f):
  "Wrap a function to save its arguments"
  n = f.__name__
  @wraps(f)
  def wrapper(obj, *args, **kwargs):
      tracer_args[(type(obj), n)] = (obj, args, kwargs)
      return f(obj, *args, **kwargs)
  return wrapper
```

Now, `tabulate` will look as follows. We can replace the calls to `get_method_wrapper` with the following. Note that we have to redefine the `flops_and_shape` part because the output_info from the jit we get from nnx looks different from the one jax gives us: 

```python
def get_flops_and_shape(jitted, *args, **kwargs):
  e = jitted.lower(*args, **kwargs)
  cost = e.cost_analysis()
  result = 0 if cost is None or 'flops' not in cost else cost['flops']
  out_info = e.lowered.out_info
  return (result, out_info[2])

def overwrite_methods(env):
  "Overwrite methods with functions from an environment"
  for (obj_type, name), f in env.items():
    setattr(obj_type, name, f)
 
def tabulate(obj, *input_args, **input_kwargs):
  .... # old beginning here, defining object_types
  # Create environment
  env = create_obj_env(object_types)

  # Add argument saving
  tracer_args = {}
  env2 = {k: argsave(tracer_args, v) for (k,v) in env.items()}
  overwrite_methods(env2)

  # Add JIT calculation
  jits = {}
  for key, value in env2.items():
    jits[key] = nnx.jit(value)
  overwrite_methods(jits)

  # Trace the top function
  jits[(type(obj), method)].trace(obj, *input_args, **input_kwargs)

  # Get flops
  result = {k: get_flops_and_shape(jits[k], o, *a, **kw) for (k,(o, a, kw)) in tracer_args.items()}

  # ... continue with assembling the table as usual
```

samanklesaria:
Now `compute_flops` works without duplicates! Next I need to add back `compute_vjp` support.

flax/nnx/summary.py
samanklesaria: Done.

flax/nnx/summary.py
samanklesaria: Looks like linen supports arbitrary vjps, not just gradients. I'll make mine match.

samanklesaria:
Remaining issues:

- There's a lot of single letter variable names. I should make these a bit more clear. 
- I should add more comments similar to my very long comment in the discussion above.

flax/nnx/summary.py
```diff
@@ -340,15 +353,42 @@ def tabulate(
...
-  )
-  nnx.eval_shape(eval_fn, obj, input_args, input_kwargs)
+  # Create a dictionary-version of the object's class. This makes
+  # iteration over methods easier.
+  env = create_obj_env(object_types)
+
+  # Modify all the object's methods to save their Tracer arguments.
+  # tracer_args contains (object, name, args, kwargs) tuples.
+  tracer_args: Tuple[Any, str, Tuple, Dict[str, Any]] = []
+  env2 = {k: argsave(tracer_args, v) for (k,v) in env.items()}
```
vfdev-5:
```suggestion
  env2 = {k: argsave(tracer_args, v) for k, v in env.items()}
```

flax/nnx/summary.py
```diff
@@ -340,15 +353,42 @@ def tabulate(
...
-    object_types=object_types,
-  )
-  nnx.eval_shape(eval_fn, obj, input_args, input_kwargs)
+  # Create a dictionary-version of the object's class. This makes
+  # iteration over methods easier.
+  env = create_obj_env(object_types)
+
+  # Modify all the object's methods to save their Tracer arguments.
+  # tracer_args contains (object, name, args, kwargs) tuples.
+  tracer_args: Tuple[Any, str, Tuple, Dict[str, Any]] = []
```
vfdev-5:
```suggestion
  tracer_args: tuple[tp.Any, str, tuple, dict[str, tp.Any]] = []
```
and similar for other type hints below

flax/nnx/summary.py
vfdev-5: Put a more clear name instead of env2

flax/nnx/summary.py
```diff
@@ -138,80 +141,79 @@ def __repr__(self):
...
+  return wrapper
+
+def overwrite_methods(env):
+  "Overwrite methods with functions from an environment"
+  for (obj_type, name), f in env.items():
+    setattr(obj_type, name, f)
+
+def get_flops(e):
+  cost = e.cost_analysis()
+  return 0 if cost is None or 'flops' not in cost else cost['flops']
```
vfdev-5: let's return int value. In the table I see float value:
```
┏━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ path ┃ type      ┃ inputs          ┃ outputs        ┃ flops   ┃ vjp_flops ┃ BatchStat         ┃ Param                   ┃
┡━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩
│      │ Foo       │ float32[16,100] │ float32[16,50] │ 99580.0 │ 296630.0  │ 40 (160 B)        │ 3,110 (12.4 KB)         │
```

vfdev-5 (commented):
Thanks @samanklesaria 
I left few comments on styling and typing.
I compared linen vs nnx tabulate on an example running on CPU, vjp flops do not exactly match. Can you check if it is due to our code. Thanks

Please also update the description with a summary on what you did with the old code and such that we could ignore intermediate comments as less relevant.

<details>
<summary>
Testing code
</summary>

Linen
```python
import flax.linen as nn
import jax, jax.numpy as jnp

class Foo(nn.Module):
  @nn.compact
  def __call__(self, x):
    h = nn.Dense(20)(x)
    h = nn.relu(h)
    bn = nn.BatchNorm(use_running_average=False)
    h = bn(h)
    return nn.Dense(50)(h)

x = jnp.ones((16, 100))

model = Foo()
table = model.tabulate(jax.random.key(0), x, compute_flops=True, compute_vjp_flops=True)

print(table)
```

and NNX

```python
import flax.nnx as nnx
import jax, jax.numpy as jnp


class Foo(nnx.Module):
  def __init__(self, rngs):
    self.fc1 = nnx.Linear(100, 20, rngs=rngs)
    self.fc1 = nnx.Linear(100, 20, rngs=rngs)
    self.bn = nnx.BatchNorm(20, use_running_average=False, rngs=rngs)
    self.fc2 = nnx.Linear(20, 50, rngs=rngs)

  def __call__(self, x):
    h = self.fc1(x)
    h = nnx.relu(h)
    h = self.bn(h)
    output = self.fc2(h)
    return output


x = jnp.ones((16, 100))

model = Foo(rngs=nnx.Rngs(0))
table = nnx.tabulate(model, x, compute_flops=True, compute_vjp_flops=True)

print(table)
```


</details>

samanklesaria:
> I compared linen vs nnx tabulate on an example running on CPU, vjp flops do not exactly match. Can you check if it is due to our code. Thanks

The way linen calculates vjp flops is the following:
```python
    def apply_vjp(variables, rngs, dynamic_leaves):
      """VJP of `c.module.apply` closed over static keyword arguments."""
      out, vjp_fn = jax.vjp(apply, variables, rngs, dynamic_leaves)
      return vjp_fn(out)

    vjp_flops = _get_flops(apply_vjp, variables, rngs, dynamic_leaves)
```

The way I do it in nnx uses the functional api. This shouldn't matter: if there's side effects in the forward pass, they wouldn't affect computation for the backward pass anyway. Note that there doesn't seem to be a `nnx.jvp` yet. 
```python
def pure_nnx_vjp(f, model, *args, **kwargs):
  "Wrap nnx functional api around jax.vjp. Only handles pure method calls."
  graphdef, state = nnx.split(model)
  def inner(state, *args, **kwargs):
    model = nnx.merge(graphdef, state)
    return f(model, *args, **kwargs)
  return jax.vjp(inner, state, *args, **kwargs)

def do_vjp(*args, **kwargs):
        primals, f_vjp = pure_nnx_vjp(jitted_fn, *args, **kwargs)
        return f_vjp(primals)
rows[i].vjp_flops = _get_flops(jax.jit(do_vjp).lower(object, *args, **kwargs))
```

These two are pretty close, but aren't doing exactly the same thing. The nnx flop calculation will include the `merge` of the graphdef and state after the `jax.vjp` boundary.  I don't know anything about linen, but I assume it handles state differently, so this merge computation is perhaps not included. 

Considering that the vjp flop estimates are very close, and that this `merge` cost *is* a real cost you encounter using nnx, I think the discrepancy is okay.

flax/nnx/summary.py
```diff
@@ -138,80 +141,79 @@ def __repr__(self):
...
+  return wrapper
+
+def overwrite_methods(env):
+  "Overwrite methods with functions from an environment"
+  for (obj_type, name), f in env.items():
+    setattr(obj_type, name, f)
+
+def get_flops(e) -> int:
+  cost = e.cost_analysis()
+  return 0 if cost is None or 'flops' not in cost else int(cost['flops'])
```
vfdev-5: Let's make all these methods private

flax/nnx/summary.py
```diff
@@ -138,80 +141,79 @@ def __repr__(self):
...
-
-      original_methods[obj_type] = methods
-
-    top_method = getattr(type(obj), call_method)
-    top_method(obj, *input_args, **input_kwargs)
-  finally:
-    for obj_type, methods in original_methods.items():
-      for name, top_method in methods.items():  # type: ignore[assignment]
-        setattr(obj_type, name, top_method)
+def pure_nnx_vjp(f, model, *args, **kwargs):
```
vfdev-5:
```suggestion
def _pure_nnx_vjp(f, model, *args, **kwargs):
```

flax/nnx/summary.py
```diff
@@ -138,80 +141,79 @@ def __repr__(self):
...
+  "Wrap nnx functional api around jax.vjp. Only handles pure method calls."
+  graphdef, state = nnx.split(model)
+  def inner(state, *args, **kwargs):
+    model = nnx.merge(graphdef, state)
+    return f(model, *args, **kwargs)
+  return jax.vjp(inner, state, *args, **kwargs)
+
+def get_call_info(jitted, method_name, node_stats, obj, compute_flops: bool, *args, **kwargs):
```
vfdev-5:
```suggestion
def _get_call_info(jitted, method_name, node_stats, obj, compute_flops: bool, *args, **kwargs):
```

flax/nnx/summary.py
```diff
@@ -138,80 +141,79 @@ def __repr__(self):
...
-    for obj_type, methods in original_methods.items():
-      for name, top_method in methods.items():  # type: ignore[assignment]
-        setattr(obj_type, name, top_method)
+def pure_nnx_vjp(f, model, *args, **kwargs):
+  "Wrap nnx functional api around jax.vjp. Only handles pure method calls."
+  graphdef, state = nnx.split(model)
+  def inner(state, *args, **kwargs):
+    model = nnx.merge(graphdef, state)
+    return f(model, *args, **kwargs)
+  return jax.vjp(inner, state, *args, **kwargs)
```
vfdev-5, line 160:
```suggestion
  def inner(model, *args, **kwargs):
    return f(model, *args, **kwargs)
  return jax.vjp(inner, model, *args, **kwargs)
```

vfdev-5:
> Considering that the vjp flop estimates are very close, and that this merge cost is a real cost you encounter using nnx, I think the discrepancy is okay.

I do not think this statement is true.
Here is what linen shows:
```
CPU

                                             Foo Summary
┏━━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┓
┃ path        ┃ module    ┃ inputs          ┃ outputs        ┃ flops ┃ vjp_flops ┃ params                  ┃ batch_stats       ┃
┡━━━━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━┩
│             │ Foo       │ float32[16,100] │ float32[16,50] │ 99580 │ 299980    │                         │                   │
├─────────────┼───────────┼─────────────────┼────────────────┼───────┼───────────┼─────────────────────────┼───────────────────┤

```
and NNX
```
                                                       Foo Summary                                                       
┏━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ path ┃ type      ┃ inputs          ┃ outputs        ┃ flops ┃ vjp_flops ┃ BatchStat         ┃ Param                   ┃
┡━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩
│      │ Foo       │ float32[16,100] │ float32[16,50] │ 99580 │ 296630    │ 40 (160 B)        │ 3,110 (12.4 KB)         │
├──────┼───────────┼─────────────────┼────────────────┼───────┼───────────┼───────────────────┼─────────────────────────┤
```

linen vjp flops number is larger than nnx vjp flops.

On GPU modifyng both `_get_flops` such that:
```python
def get_flops(e) -> int:
  cost = e.cost_analysis()
  if cost is None:
    e = e.compile()
    cost = e.cost_analysis()
  return 0 if cost is None or 'flops' not in cost else int(cost['flops'])
```
we get the following:
LINEN
```
                                                          Foo Summary
┏━━━━━━━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┓
┃ path        ┃ module    ┃ inputs          ┃ outputs        ┃ flops ┃ vjp_flops ┃ params                  ┃ batch_stats       ┃
┡━━━━━━━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━┩
│             │ Foo       │ float32[16,100] │ float32[16,50] │ 4938  │ 13794     │                         │                   │
├─────────────┼───────────┼─────────────────┼────────────────┼───────┼───────────┼─────────────────────────┼───────────────────┤

```
vs NNX
```
                                                       Foo Summary                                                       
┏━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━┳━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ path ┃ type      ┃ inputs          ┃ outputs        ┃ flops ┃ vjp_flops ┃ BatchStat         ┃ Param                   ┃
┡━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━╇━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━┩
│      │ Foo       │ float32[16,100] │ float32[16,50] │ 4938  │ 10384     │ 40 (160 B)        │ 3,110 (12.4 KB)         │
├──────┼───────────┼─────────────────┼────────────────┼───────┼───────────┼───────────────────┼─────────────────────────┤
```

Finally, expected vjp flops can be computed as:
```python
@jax.jit
def fn(model, x):
  return model(x)

def gn(model, x):
    primals, f_vjp = jax.vjp(fn, model, x)
    return f_vjp(primals)

e = jax.jit(gn).lower(model, x)
cost = e.cost_analysis()
if cost is None:
    e = e.compile()
cost = e.cost_analysis()
vjp_flops = 0 if cost is None or 'flops' not in cost else int(cost['flops'])

print("Expected vjp flops:")
print(vjp_flops)
```
and it gives what nnx shows, so, ok to me.

flax/nnx/summary.py
samanklesaria: I don't understand your suggestion here. `state` is defined in line 156. If you remove it, won't line 158 be undefined?

flax/nnx/summary.py
samanklesaria: Done.

flax/nnx/summary.py
vfdev-5: Updated the snippet. The idea is to remove split and merge

samanklesaria: We can't just pass `model` to `jx.vjp` directly like this, can we? Yes, models are pytrees now, but they in general contain a mix of static (graph definitions and choices of activation functions) and dynamic (tensors) leaves. In equinox, we always separate out the static and dynamic pieces of a model, pass them in separately, and reconstruct the result. The flax functional API is essentially duplicating what equinox does. Or am I missing something about how modules work in flax?

vfdev-5: > We can't just pass model to jx.vjp directly like this, can we?

We can. This assumes we consider everything as dynamic pieces in the model.
And this is equivalent to what you had:
```python
graphdef, state = nnx.split(model)
```
where `state` will contain everything: params, rngs, batch stats etc.

samanklesaria: Huh- this is very different to how equinox works. I guess I need to read through the code more to understand how this works. You're right though- using jax directly seems to work just fine. I changed the implementation.

flax/nnx/summary.py
```diff
@@ -138,80 +141,71 @@ def __repr__(self):
...
+    input_kwargs=input_kwargs_info,
+    outputs=output_repr,
+    flops=flops
+  )
 
 
 def filter_rng_streams(row: CallInfo):
   return not issubclass(row.type, nnx.RngStream)
 
+def create_obj_env(object_types):
```
vfdev-5:
```suggestion
def _create_obj_env(object_types):
```

flax/nnx/summary.py
```diff
@@ -138,80 +141,71 @@ def __repr__(self):
...
+  "Turn a set of object types into a dictionary mapping (type, method name) pairs to methods"
+  result = {}
+  for obj_type in object_types:
+    for name, top_method in inspect.getmembers(obj_type, inspect.isfunction):
+      if not name.startswith('_') or name == '__call__':
+        result[(obj_type, name)] = top_method
+  return result
+
+def argsave(tracer_args, f):
```
vfdev-5:
```suggestion
def _argsave(tracer_args, f):
```

flax/nnx/summary.py
```diff
@@ -138,80 +141,71 @@ def __repr__(self):
...
+  return wrapper
+
+def _overwrite_methods(env):
+  "Overwrite methods with functions from an environment"
+  for (obj_type, name), f in env.items():
+    setattr(obj_type, name, f)
+
+def _get_flops(e) -> int:
+  cost = e.cost_analysis()
+  return 0 if cost is None or 'flops' not in cost else int(cost['flops'])
```
vfdev-5:
```suggestion
  cost = e.cost_analysis()
  if cost is None:
    e = e.compile()
    cost = e.cost_analysis()  
  return 0 if cost is None or 'flops' not in cost else int(cost['flops'])
```

samanklesaria: When would it be the case that a cost analysis would only be available for a compiled version (rather than a lowered one)?

samanklesaria: Ah- for cuda! Got it.

flax/nnx/summary.py
```diff
@@ -340,15 +347,42 @@ def tabulate(
...
+    compute_flops, *args, **kwargs)
+    for (object, name, args, kwargs) in tracer_args]
+
+  # Add VJP flops if required
+  if compute_vjp_flops:
+    for i, (object, name, args, kwargs) in enumerate(tracer_args):
+      def do_vjp(*args, **kwargs):
+        primals, f_vjp = jax.vjp(jits[(type(object), name)], *args, **kwargs)
+        return f_vjp(primals)
+      rows[i].vjp_flops = _get_flops(jax.jit(do_vjp).lower(object, *args, **kwargs))
```
vfdev-5: Can't we have this code in `_get_call_info` together with compute_flops option?

vfdev-5:
@samanklesaria please fix mypy

flax/nnx/summary.py
samanklesaria: We can indeed. Done.

flax/nnx/summary.py
```diff
@@ -138,80 +141,82 @@ def __repr__(self):
...
+def _overwrite_methods(env):
+  "Overwrite methods with functions from an environment"
+  for (obj_type, name), f in env.items():
+    setattr(obj_type, name, f)
+
+def _get_flops(e) -> int:
+  cost = e.cost_analysis() or e.compile().cost_analysis()
+  return 0 if cost is None or 'flops' not in cost else int(cost['flops'])
+
+  return 0 if cost is None or 'flops' not in cost else int(cost['flops'])
```
vfdev-5: This line is not needed.

flax/nnx/summary.py
```diff
@@ -340,15 +356,34 @@ def tabulate(
...
+  # Trace the top function (which indirectly traces all the others)
+  jits[(type(obj), method)].trace(obj, *input_args, **input_kwargs)
+
+  # Get call_info
+  rows : list[CallInfo] = [_get_call_info(
+    jits[(type(object), name)], name, node_stats, object,
+    compute_flops, compute_vjp_flops, *args, **kwargs)
+    for (object, name, args, kwargs) in tracer_args]
+
+  _overwrite_methods(env)
```
vfdev-5, line 396: Actually, why do we need this now?

flax/nnx/summary.py
samanklesaria: Whoops! Good catch. Sorry about that.

flax/nnx/summary.py
samanklesaria: If the user ever wants to use the module again, we want them to use its original methods, not our monkey patched ones.

vfdev-5: Please leave a comment that we restore the object back.

samanklesaria: Done.

vfdev-5:
@samanklesaria please squash commits into 1. There is a check on the number of commits in the PR, should be less than 5.

flax/nnx/summary.py
```diff
@@ -138,80 +141,80 @@ def __repr__(self):
...
+  path = node_info.path
+  if method_name != '__call__':
+    path = (*path, method_name)
+
+  vjp_flops = None
+  if compute_vjp_flops:
+    def do_vjp(*args, **kwargs):
+      primals, f_vjp = jax.vjp(jitted, *args, **kwargs)
+      return f_vjp(primals)
+    vjp_flops = _get_flops(jax.jit(do_vjp).lower(obj, *args, **kwargs))
```
cgarciae: There might be some issues here using `jax.jit` and `jax.vjp` if there are non-tree Modules structures in `args` or `kwargs`. Some options here:
1. use `graph.find_duplicates` to detect non-tree structures and raise an error. e.g.
```python
if (duplicates := graph.find_duplicates((args, kwargs)):
  raise ...
```
2. use `split` and `merge` to preserve the non-tree structure.
3. use `nnx.jit` instead of `jax.jit`, however, we don't have `nnx.vjp` so for that one we would need to use option 2.

samanklesaria: The merge approach is in the commit history here (https://github.com/google/flax/commit/c9b10fcb5b3c448aca95edb5196feb1408d844ac). But I like approach (1) better, if that's okay with you. Nnx.jit is already used for compute_flops: it's only compute_vjp_flops that's the tricky one.

samanklesaria:
When I tried to support shared graph structure with split and merge, I got an duplicate row in my tabulation table for the top level call whenever I call it on shared structures. Specifically, say `j = nnx.jit(f)`, and say we've already called `j.trace`. 
Future calls to `jax.vjp(j, ...)` don't result in `f` being re-traced.  However, if we define a version of vjp that supports shared structure...

```python
def nnx_vjp(f, model, *args, **kwargs):
  "Wrap nnx functional api around jax.vjp. Only handles pure method calls."
  graphdef, state = nnx.split(model)
  def inner(state, *args, **kwargs):
    model = nnx.merge(graphdef, state)
    return f(model, *args, **kwargs)
  return jax.vjp(inner, state, *args, **kwargs)
```

... then calling `nnx_vjp(j, ...)` *does* result in `f` being re-traced. This is annoying because it results in duplicate rows being added to the tabulation table. 

Some possible solutions:

1. Don't support shared structures in arguments to tabulate with `compute_vjp`. We plan to deprecate them eventually anyway, and `compute_vjp_flops` is a niche use case. Throw an error like in the previous commit. 
2. Compute the vjp_flops in a separate pass, as I was doing before. This separate pass would only mutate existing entries in the `rows` list rather than adding new ones, so we wouldn't see duplicate rows. 

@vfdev-5 I assume option (2) is the right approach, even though it's a little more ugly?

vfdev-5:
@samanklesaria mypy failure is related to your change.

PR #4953, opened 2025-09-16
add flax_hijax_variable flag

cgarciae:
# What does this PR do?

### DIFF
--- a/flax/configurations.py
+++ b/flax/configurations.py
@@ -27,6 +27,7 @@ class Config:
   flax_pytree_module: bool
   flax_max_repr_depth: int | None
   flax_always_shard_variable: bool
+  flax_hijax_variable: bool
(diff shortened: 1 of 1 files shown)


flax/configurations.py
```diff
@@ -290,4 +291,9 @@ def temp_flip_flag(var_name: str, var_value: bool):
   name='flax_always_shard_variable',
   default=True,
   help='Whether a `nnx.Variable` should always automatically be sharded if it contains sharding annotations.',
+)
+flax_hijax_variable = bool_flag(
```
IvyZX, line 295: Are we using this flag at all?

IvyZX (approved):

flax/configurations.py
cgarciae: for future use so it's, sending it separately for now

PR #4969, opened 2025-09-19
Correctly expose `flax.config.temp_flip_flag` 

IvyZX:
Make sure we can access `temp_flip_flag` via `flax.config.temp_flip_flag`.

Also moved a few helpful API docs from Linen site to NNX.


flax/configurations.py
```diff
@@ -70,6 +70,21 @@ def __repr__(self):
     values_repr = ', '.join(f'\n  {k}={v!r}' for k, v in self._values.items())
     return f'Config({values_repr}\n)'
 
+  @contextmanager
+  def temp_flip_flag(self, var_name: str, var_value: bool):
```
cgarciae, line 75: `Config` should inherit from `threading.local` to make this safe.

cgarciae (approved):

flax/configurations.py
IvyZX: Done.

PR #4980, opened 2025-09-24
Move `ArrayRef` creation to the end of `Variable` creation

IvyZX:
This allows us to make sharding changes to the underlying array before passing it around by reference.

### DIFF
--- a/flax/nnx/variablelib.py
+++ b/flax/nnx/variablelib.py
@@ -242,16 +242,10 @@ class Variable(tp.Generic[A], reprlib.Representable):
     if isinstance(value, VariableMetadata):
       metadata.update(value.metadata)
       value = tp.cast(A, value.raw_value)
+    elif is_array_ref(value):
+      raise ValueError('Cannot pass a ArrayRef directly into Variable init.')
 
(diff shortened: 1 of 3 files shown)


flax/nnx/variablelib.py
```diff
@@ -243,15 +243,7 @@ def __init__(
...
-    if use_ref:
-      if is_array_ref(value):
-        _value = tp.cast(A, value)
-      else:
-        _value = array_ref(jnp.asarray(value))  # type: ignore[assignment]  # type: ignore[assignment]
-    else:
-      _value = value
-
-    object.__setattr__(self, 'raw_value', _value)
+    object.__setattr__(self, 'raw_value', value)
```
cgarciae, line 248: With this change we shouldn't allow `Ref`'s to be passed as an input, can we raise if `is_array_ref(value)` here and then maybe remove the same check bellow?

IvyZX: Discussed offline and we concluded each variable should own its own array/ref and don't allow sharing.

Implemented and changed `mutable_array_test` to reflect it.

PR #4981, opened 2025-09-25
[nnx] preserve the function's type information in jit

cgarciae:
# What does this PR do?

Uses `typing.ParamSpec` to properly annotate `JitWrapped.__call__`.


flax/nnx/transforms/compilation.py
```diff
@@ -166,7 +168,7 @@ def jit(
   backend: tp.Optional[str] = None,
   inline: bool = False,
   abstracted_axes: tp.Optional[tp.Any] = None,
-) -> JitWrapped: ...
+) -> JitWrapped[P, R]: ...
 def jit(
   fun: tp.Callable[..., tp.Any] | type[Missing] = Missing,
```
jburnim: Does the signature for `fun` have to be updated here, as well?

cgarciae: yes, nice catch! updated.

PR #4996, opened 2025-10-03
Fix NNX jit static args with in_shardings issue #4989

### LINKED ISSUE #4989: NNX JIT fails with static args and in shardings
### System information
- OS: MacOS Sequoia 15.1
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`: jax==0.7.2, jaxlib==0.7.2, flax==0.12.0
- Python version: Python 3.12.11
- GPU/TPU model and memory: None
- CUDA version (if applicable): None

```nnx.jit``` does not seem to work when static arguments and ```in_shardings``` are used. For illustration, say we have this function with a static boolean argument ```use_relu```:

```
devices = mesh_utils.create_device_mesh((jax.device_count(),))
mesh = Mesh(devices, axis_names=('data',))

def my_func(x, scale: float, use_relu: bool):
    y = x * scale
    if use_relu:
        y = jnp.maximum(y, 0)
    return y.sum()


x_sharding = NamedSharding(mesh, P('data'))
scale_sharding = None

my_func_jit = nnx.jit(
    my_func,
    in_shardings=(x_sharding, scale_sharding),
    static_argnums=(2,)
)

# Example input
x = jnp.arange(16.0).reshape(4, 4)
scale = 0.5

print("With ReLU:", my_func_jit(x, scale, True))
print("Without ReLU:", my_func_jit(x, scale, False))
```

Following standard jax convention, if we do not specify an ```in_sharding``` for the third static argument ```use_relu```, we get the following error:

```
[/tmp/ipython-input-3302790925.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in <cell line: 0>()
     29 scale = 0.5
     30 
---> 31 print("With ReLU:", my_func_jit(x, scale, True))
     32 print("Without ReLU:", my_func_jit(x, scale, False))

5 frames[/usr/local/lib/python3.12/dist-packages/flax/nnx/transforms/compilation.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in __call__(self, *args, **kwargs)
    430     # run dynamic_cache_context before update_context
    431     with graph.update_context(self):
--> 432       pure_args, pure_kwargs = self._get_pure_args_kwargs(args, kwargs)
    433       pure_args_out, pure_kwargs_out, pure_out = self.jitted_fn(
    434         *pure_args, **pure_kwargs

[/usr/local/lib/python3.12/dist-packages/flax/nnx/transforms/compilation.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in _get_pure_args_kwargs(self, args, kwargs)
    406 
    407   def _get_pure_args_kwargs(self, args, kwargs):
--> 408     pure_args, pure_kwargs = extract.to_tree(
    409       (args, kwargs),
    410       prefix=(self.in_shardings, self.kwarg_shardings)

[/usr/local/lib/python3.12/dist-packages/flax/nnx/extract.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in to_tree(tree, prefix, split_fn, map_non_graph_nodes, ctxtag, check_aliasing)
    223         or graph.is_graph_node(x),
    224       )
--> 225   leaf_prefixes = broadcast_prefix(
    226     prefix,
    227     tree,

[/usr/local/lib/python3.12/dist-packages/flax/nnx/extract.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in broadcast_prefix(prefix_tree, full_tree, prefix_is_leaf, tree_is_leaf)
    124   ).num_leaves
    125   add_leaves = lambda x, subtree: result.extend([x] * num_leaves(subtree))
--> 126   jax.tree.map(
    127     add_leaves,
    128     prefix_tree,

[/usr/local/lib/python3.12/dist-packages/jax/_src/tree.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in map(f, tree, is_leaf, *rest)
    153     - :func:`jax.tree.reduce`
    154   """
--> 155   return tree_util.tree_map(f, tree, *rest, is_leaf=is_leaf)
    156 
    157 

[/usr/local/lib/python3.12/dist-packages/jax/_src/tree_util.py](https://colab.research.google.com/drive/1GnpwciOaeoFbKsS3zLDhDh2mtcju_yWE#) in tree_map(f, tree, is_leaf, *rest)
    358   """Alias of :func:`jax.tree.map`."""
    359   leaves, treedef = tree_flatten(tree, is_leaf)
--> 360   all_leaves = [leaves] + [treedef.flatten_up_to(r) for r in rest]
    361   return treedef.unflatten(f(*xs) for xs in zip(*all_leaves))
    362 

ValueError: Tuple arity mismatch: 3 != 2; tuple: (Array([[ 0.,  1.,  2.,  3.],
       [ 4.,  5.,  6.,  7.],
       [ 8.,  9., 10., 11.],
       [12., 13., 14., 15.]], dtype=float32), 0.5, `True).`
```

This comes from the ```_get_pure_args_kwargs``` function used in nnx.jit which doesn't consider the absence of  ```in_shardings``` for static arguments, therefore there's a mismatch error when trying to prefix the arguments.

Providing an ```in_sharding``` for the static argument will also result in an error, as the jax jitted function used by nnx will fail:

```
import jax
import jax.numpy as jnp
from jax.sharding import Mesh, NamedSharding
from jax.experimental import mesh_utils
from jax.sharding import PartitionSpec as P

devices = mesh_utils.create_device_mesh((jax.device_count(),))
mesh = Mesh(devices, axis_names=('data',))

def my_func(x, scale: float, use_relu: bool):
    y = x * scale
    if use_relu:
        y = jnp.maximum(y, 0)
    return y.sum()


x_sharding = NamedSharding(mesh, P('data'))
scale_sharding = None
static_sharding = None

# use_relu is static, so we do not list it in in_shardings
my_func_jit = nnx.jit(
    my_func,
    in_shardings=(x_sharding, scale_sharding, static_sharding),
    static_argnums=(2,)
)

# Example input
x = jnp.arange(16.0).reshape(4, 4)
scale = 0.5

print("With ReLU:", my_func_jit(x, scale, True))
print("Without ReLU:", my_func_jit(x, scale, False))
```
```
[/tmp/ipython-input-3771445769.py](https://localhost:8080/#) in <cell line: 0>()
     30 scale = 0.5
     31 
---> 32 print("With ReLU:", my_func_jit(x, scale, True))
     33 print("Without ReLU:", my_func_jit(x, scale, False))

1 frames
[/usr/local/lib/python3.12/dist-packages/flax/nnx/transforms/compilation.py](https://localhost:8080/#) in __call__(self, *args, **kwargs)
    431     with graph.update_context(self):
    432       pure_args, pure_kwargs = self._get_pure_args_kwargs(args, kwargs)
--> 433       pure_args_out, pure_kwargs_out, pure_out = self.jitted_fn(
    434         *pure_args, **pure_kwargs
    435       )

    [... skipping hidden 7 frame]

[/usr/local/lib/python3.12/dist-packages/jax/_src/pjit.py](https://localhost:8080/#) in flatten_axis_resources(what, tree, shardings, tupled_args)
   1010                 f"a singleton tuple.")
   1011 
-> 1012     raise ValueError(msg)
   1013 
   1014   axis_tree = shardings

ValueError: pjit in_shardings specification must be a tree prefix of the positional arguments tuple. In particular, pjit in_shardings must either be a Sharding, a PartitionSpec, or a tuple of length equal to the number of positional arguments. But pjit in_shardings is the wrong length: got a tuple or list of length 3 for an args tuple of length 2.
```

Not providing a in_sharding for static arguments should not raise an error.

Already provided above.

### Steps to reproduce:
[Colab here](https://drive.google.com/file/d/1FxiUzijNke6_FRIQxjuEj2ybA1PJDgEY/view?usp=sharing)

### PR DESCRIPTION
**Pull request by mohsinm-dev**
## Problem

When using `nnx.jit` with static arguments and `in_shardings`, users encountered errors:

1. Not providing shardings for static args → "Tuple arity mismatch" error
2. Providing shardings for static args → "pjit in_shardings specification must be a tree prefix" error

## Solution

Fixed the sharding handling logic to properly filter out shardings for static argument positions before passing them to JAX's jit function.

## Changes

- Modified `JitWrapped.__init__` in `compilation.py` to drop shardings for static positions
- Added comprehensive tests to verify both scenarios work correctly

## Testing

- Added `tests/nnx/jit_static_sharding_test.py` with tests for both failing scenarios
- Verified no regressions in existing functionality

Fixes #4989

### DIFF
--- a/flax/nnx/transforms/compilation.py
+++ b/flax/nnx/transforms/compilation.py
@@ -17,6 +17,7 @@ from __future__ import annotations
 import dataclasses
 import functools
 import typing as tp
+import inspect
 
 import jax
 import jax.experimental
@@ -128,14 +129,59 @@ class JitFn:
     out = self.f(*args, **kwargs)
 
     args_out, kwargs_out = extract.clear_non_graph_nodes((args, kwargs))
+    # For packaging outputs, do not attempt to broadcast shardings for args_out;
+    # only the function output `out` needs an output sharding prefix.
     pure_args_out, pure_kwargs_out, pure_out = extract.to_tree(
       (args_out, kwargs_out, out),
-      prefix=(self.in_shardings, self.kwarg_shardings, self.out_shardings),
+      prefix=(None, self.kwarg_shardings, self.out_shardings),
       ctxtag=self.ctxtag,
       split_fn=_jit_split_fn,
     )
 
     return pure_args_out, pure_kwargs_out, pure_out
+
+
+def _normalize_static_argnums(
+  fun: tp.Callable[..., tp.Any],
+  static_argnums: int | tp.Sequence[int] | None,
+  static_argnames: str | tp.Iterable[str] | None,
+) -> tuple[int, ...]:
+  """Normalize static args specification to positional indices.
+
+  This mirrors jax.jit's behavior sufficiently for positional static arguments
+  and named arguments resolvable from the function signature.
+  """
+  idxs: set[int] = set()
+  if static_argnums is not None:
+    if isinstance(static_argnums, int):
+      idxs.add(static_argnums)
+    else:
+      idxs.update(static_argnums)
+
+  if static_argnames is not None:
+    if isinstance(static_argnames, str):
+      names = (static_argnames,)
+    else:
+      names = tuple(static_argnames)
+    try:
+      sig = inspect.signature(fun)
+      # Map parameter names to their positional index among positional params.
+      pos_params: list[str] = []
+      for p in sig.parameters.values():
+        if p.kind in (
+          inspect.Parameter.POSITIONAL_ONLY,
+          inspect.Parameter.POSITIONAL_OR_KEYWORD,
+        ):
+          pos_params.append(p.name)
+        # skip VAR_POSITIONAL, KEYWORD_ONLY, VAR_KEYWORD
+      name_to_idx = {n: i for i, n in enumerate(pos_params)}
+      for n in names:
+        if n in name_to_idx:
+          idxs.add(name_to_idx[n])
+    except (ValueError, TypeError):
+      # Fallback: if signature inspection fails, ignore static_argnames here.
+      pass
+  return tuple(sorted(idxs))
 
 
 @tp.overload
@@ -362,12 +408,34 @@ class JitWrapped(tp.Generic[P, R]):
     functools.update_wrapper(self, fun)
     self.fun: tp.Callable[P, R] = fun
     kwarg_shardings = None
-    self.jax_in_shardings = jax.tree.map(
-      lambda x: extract.NodeStates.from_prefixes(x.shardings, metadata=x)
-      if isinstance(x, StateSharding)
-      else x,
-      in_shardings,
+    # Normalize static argnums/argnames to a tuple of positional indices.
+    self._static_positions: tuple[int, ...] = _normalize_static_argnums(
+      fun, static_argnums, static_argnames
     )
+
+    # Prepare jax_in_shardings for jax.jit by dropping entries that correspond
+    # to static positional arguments when a per-argument sequence is provided.
+    def _map_state_sharding(x):
+      return (
+        extract.NodeStates.from_prefixes(x.shardings, metadata=x)
+        if isinstance(x, StateSharding)
+        else x
+      )
+
+    def _drop_static_from_sequence(seq):
+      # Keep only shardings for dynamic (non-static) positions.
+      result = []
+      for i, v in enumerate(seq):
+        if i not in self._static_positions:
+          result.append(v)
+      return type(seq)(result)
+
+    _in_shardings_mapped = jax.tree.map(_map_state_sharding, in_shardings)
+    if isinstance(_in_shardings_mapped, (tuple, list)) and self._static_positions:
+      self.jax_in_shardings = _drop_static_from_sequence(_in_shardings_mapped)
+    else:
+      # Single spec or no static args: pass through as-is so JAX can broadcast.
+      self.jax_in_shardings = _in_shardings_mapped
     self.jax_out_shardings = jax.tree.map(
       lambda x: extract.NodeStates.from_prefixes(x.shardings, metadata=x)
       if isinstance(x, StateSharding)
@@ -378,11 +446,9 @@ class JitWrapped(tp.Generic[P, R]):
     self.jitted_fn = jax.jit(
       JitFn(fun, in_shardings, out_shardings, kwarg_shardings, self),
       in_shardings=self.jax_in_shardings,
-      out_shardings=(
-        self.jax_in_shardings,
-        kwarg_shardings,
-        self.jax_out_shardings,
-      ),
+      # For outputs, rely on propagation for args/kwargs tuples and only
+      # optionally constrain the actual function output via jax_out_shardings.
+      out_shardings=(None, kwarg_shardings, self.jax_out_shardings),
       static_argnums=static_argnums,
       static_argnames=static_argnames,
       donate_argnums=donate_argnums,
@@ -405,14 +471,55 @@ class JitWrapped(tp.Generic[P, R]):
     return functools.partial(self, obj)
 
   def _get_pure_args_kwargs(self, args, kwargs):
+    # Build a prefix for positional arguments that inserts `None` for static
+    # positions so that prefix broadcasting matches the full argument tuple.
+    def _build_positional_prefix():
+      if self.in_shardings is None:
+        return None
+      # Determine number of positional args and number of dynamic args.
+      num_args = len(args)
+      static_set = set(self._static_positions)
+      num_dynamic = num_args - sum(1 for i in range(num_args) if i in static_set)
+
+      # Expand user-provided in_shardings to a per-dynamic-arg sequence.
+      if isinstance(self.in_shardings, (tuple, list)):
+        dyn_specs = list(self.in_shardings)
+        # If user provided a shorter sequence (e.g., as a prefix), do not try
+        # to out-broadcast here; rely on extract/jax to handle deeper trees.
+        # When lengths mismatch with expected dynamic count, we will not index
+        # past available entries.
+      else:
+        # Single spec: broadcast to all dynamic args
+        dyn_specs = [self.in_shardings] * max(num_dynamic, 0)
+
+      # Now interleave dyn_specs into a full positional prefix, inserting
+      # None for static positions.
+      full_prefix: list[tp.Any] = []
+      dyn_idx = 0
+      for i in range(num_args):
+        if i in static_set:
+          full_prefix.append(None)
+        else:
+          # Guard against too-short dyn_specs: fall back to last available or None
+          if dyn_idx < len(dyn_specs):
+            full_prefix.append(dyn_specs[dyn_idx])
+          else:
+            full_prefix.append(None)
+          dyn_idx += 1
+      return type(args)(full_prefix)
+
+    pos_prefix = _build_positional_prefix()
+    prefix = (
+      (pos_prefix, self.kwarg_shardings)
+      if (pos_prefix is not None or self.kwarg_shardings is not None)
+      else None
+    )
+
     pure_args, pure_kwargs = extract.to_tree(
       (args, kwargs),
-      prefix=(self.in_shardings, self.kwarg_shardings)
-      if self.in_shardings is not None or self.kwarg_shardings is not None
-      else None,
+      prefix=prefix,
       split_fn=_jit_split_fn,
-      check_aliasing=self.in_shardings is not None
-      or self.kwarg_shardings is not None,
+      check_aliasing=(pos_prefix is not None) or (self.kwarg_shardings is not None),
       ctxtag=self,
     )
     return pure_args, pure_kwargs
--- a/tests/nnx/jit_static_sharding_test.py
+++ b/tests/nnx/jit_static_sharding_test.py
@@ -1,0 +1,71 @@
+# Copyright 2024 The Flax Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import jax
+import jax.numpy as jnp
+from jax.experimental import mesh_utils
+from jax.sharding import Mesh, NamedSharding
+from jax.sharding import PartitionSpec as P
+
+from flax import nnx
+
+
+def _mesh_data():
+  devices = mesh_utils.create_device_mesh((jax.device_count(),))
+  return Mesh(devices, axis_names=("data",))
+
+
+def _inputs_with_negatives():
+  # Use inputs with negative values so ReLU has an effect
+  x = jnp.linspace(-1.0, 1.0, 16, dtype=jnp.float32).reshape(4, 4)
+  scale = 0.5
+  return x, scale
+
+
+def _fn(x, scale: float, use_relu: bool):
+  y = x * scale
+  if use_relu:
+    y = jnp.maximum(y, 0)
+  return y.sum()
+
+
+def test_jit_static_no_static_in_shardings():
+  mesh = _mesh_data()
+  x_sharding = NamedSharding(mesh, P("data"))
+
+  f = nnx.jit(_fn, in_shardings=(x_sharding, None), static_argnums=(2,))
+
+  x, scale = _inputs_with_negatives()
+  y_relu = f(x, scale, True)
+  y_no_relu = f(x, scale, False)
+
+  # verify static arg toggles result
+  assert y_relu != y_no_relu
+
+
+def test_jit_static_with_extra_static_entry():
+  mesh = _mesh_data()
+  x_sharding = NamedSharding(mesh, P("data"))
+
+  g = nnx.jit(
+    _fn, in_shardings=(x_sharding, None, None), static_argnums=(2,)
+  )
+
+  x, scale = _inputs_with_negatives()
+  y_relu = g(x, scale, True)
+  y_no_relu = g(x, scale, False)
+
+  # verify static arg toggles result even if provided an extra sharding entry
+  assert y_relu != y_no_relu
+

### CODE REVIEW

File: tests/nnx/jit_static_sharding_test.py [hunk 1]
```diff
@@ -0,0 +1,71 @@
+# Copyright 2024 The Flax Authors.
```
vfdev-5: Why creating a new file for this test? Let's put it in an existing test file

tests/nnx/transforms_test.py
```diff
@@ -420,6 +420,58 @@ def f(m: nnx.Linear, x):
...
+    x_sharding = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec('data'))
+    f = nnx.jit(fn, in_shardings=(x_sharding, None), static_argnums=(2,))
+
+    y_relu = f(x, scale, True)
+    y_no_relu = f(x, scale, False)
+
+    # Verify static arg toggles result
+    self.assertNotEqual(y_relu, y_no_relu)
+
+  def test_jit_static_with_extra_static_entry(self):
```
samanklesaria: I don't think that passing a sharding spec for a static argument is something that  `jax.jit` supports. When I try compilation here with `jax.jit` instead of `nnx.jit`,  I get `got a tuple or list of length 3 for an args tuple of length 2`. So it seems like something `nnx.jit` shouldn't support either.

flax/nnx/transforms/compilation.py
```diff
@@ -362,12 +408,34 @@ def __init__(
...
+    self._static_positions: tuple[int, ...] = _normalize_static_argnums(
+      fun, static_argnums, static_argnames
     )
+
+    # Prepare jax_in_shardings for jax.jit by dropping entries that correspond
+    # to static positional arguments when a per-argument sequence is provided.
+    def _map_state_sharding(x):
+      return (
+        extract.NodeStates.from_prefixes(x.shardings, metadata=x)
+        if isinstance(x, StateSharding)
```
samanklesaria: It would be good to add a test for this case as well.

flax/nnx/transforms/compilation.py
```diff
@@ -362,12 +408,34 @@ def __init__(
...
+        else x
+      )
+
+    def _drop_static_from_sequence(seq):
+      # Keep only shardings for dynamic (non-static) positions.
```
samanklesaria: Seems like this would be easier with a list comprehension one liner: 
`type(seq)([v for i,v in in enumerate(seq) if i not in self._static_positions])`.

flax/nnx/transforms/compilation.py
```diff
@@ -405,14 +471,55 @@ def __get__(self, obj, objtype=None):
...
   def _get_pure_args_kwargs(self, args, kwargs):
+    # Build a prefix for positional arguments that inserts `None` for static
+    # positions so that prefix broadcasting matches the full argument tuple.
+    def _build_positional_prefix():
+      if self.in_shardings is None:
+        return None
+      # Determine number of positional args and number of dynamic args.
+      num_args = len(args)
+      static_set = set(self._static_positions)
+      num_dynamic = num_args - sum(1 for i in range(num_args) if i in static_set)
```
samanklesaria: Is this not just `num_args - len(static_set)`?

flax/nnx/transforms/compilation.py
```diff
@@ -405,14 +471,55 @@ def __get__(self, obj, objtype=None):
...
+        # past available entries.
+      else:
+        # Single spec: broadcast to all dynamic args
+        dyn_specs = [self.in_shardings] * max(num_dynamic, 0)
+
+      # Now interleave dyn_specs into a full positional prefix, inserting
+      # None for static positions.
+      full_prefix: list[tp.Any] = []
+      dyn_idx = 0
+      for i in range(num_args):
```
samanklesaria: It might be slightly cleaner to use an iterator here. 
```python
shard_iter = iter(dyn_specs)
tuple(None if i in static_set else next(dyn_specs) for i in range(num_args))
```

flax/nnx/transforms/compilation.py
```diff
@@ -405,14 +471,55 @@ def __get__(self, obj, objtype=None):
...
+      # Expand user-provided in_shardings to a per-dynamic-arg sequence.
+      if isinstance(self.in_shardings, (tuple, list)):
+        dyn_specs = list(self.in_shardings)
+        # If user provided a shorter sequence (e.g., as a prefix), do not try
+        # to out-broadcast here; rely on extract/jax to handle deeper trees.
+        # When lengths mismatch with expected dynamic count, we will not index
+        # past available entries.
+      else:
+        # Single spec: broadcast to all dynamic args
+        dyn_specs = [self.in_shardings] * max(num_dynamic, 0)
```
samanklesaria: Do we need to explicitly broadcast here? Or will `extract.to_tree` do the broadcasting for us anyway?

tests/nnx/transforms_test.py
```diff
@@ -420,6 +420,58 @@ def f(m: nnx.Linear, x):
...
+      if use_relu:
+        y = jnp.maximum(y, 0)
+      return y.sum()
+
+    # Use inputs with negative values so ReLU has an effect
+    x = jnp.linspace(-1.0, 1.0, 16, dtype=jnp.float32).reshape(4, 4)
+    scale = 0.5
+
+    x_sharding = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec('data'))
+    f = nnx.jit(fn, in_shardings=(x_sharding, None), static_argnums=(2,))
```
samanklesaria: We should test when `in_shardings` needs to be broadcast as well. Add another test case for this.

flax/nnx/transforms/compilation.py
```diff
@@ -128,16 +129,61 @@ def __call__(self, *pure_args, **pure_kwargs):
...
+  and named arguments resolvable from the function signature.
+  """
+  idxs: set[int] = set()
+  if static_argnums is not None:
+    if isinstance(static_argnums, int):
+      idxs.add(static_argnums)
+    else:
+      idxs.update(static_argnums)
+
+  if static_argnames is not None:
```
samanklesaria: Can we add a test for using `static_argnames` in addition to the one you have for `static_argnums`?

flax/nnx/transforms/compilation.py
mohsinm-dev: The explicit broadcasting is not needed here.

pyproject.toml
```diff
@@ -23,6 +23,8 @@ dependencies = [
     "typing_extensions>=4.2",
     "PyYAML>=5.4.1",
     "treescope>=0.1.7",
+    "pytest>=8.4.2",
+    "jaxlib>=0.7.2",
```
samanklesaria: I don't think we need to add these dependencies

flax/nnx/transforms/compilation.py
samanklesaria: So could we just have `_build_positional_prefix` return `self.in_shardings` if it's not a tuple or list?

mohsinm-dev: I thought about it for a while, maybe I am wrong but I think we do need to build the prefix explicitly. `extract.to_tree` can't handle this for us. When we have a single sharding value, `broadcast_prefix` will apply it to ALL argument positions including the static ones. We need to mark static positions with None to prevent this. We're not doing "manual broadcasting" but we're marking which positions should NOT have sharding constraints.

I am thinking of something like this :

```
def _build_positional_prefix():
      if self.in_shardings is None:
          return None

      # Always need to handle static positions by inserting None
      num_args = len(args)
      static_set = set(self._static_positions)

      if isinstance(self.in_shardings, (tuple, list)):
          # Sequence case: align provided shardings with dynamic positions
          dyn_iter = iter(self.in_shardings)
          return type(args)(
              None if i in static_set else next(dyn_iter, None) for i in range(num_args)
          )
      Else:
          # Single sharding case: broadcast to all dynamic positions
          # We still need to build the tuple to mark static positions with None
          return type(args)(
              None if i in static_set else self.in_shardings for i in range(num_args)
          )
```


 Example:
#### If we have: fn(x, scale, use_relu) where use_relu is static
And user provides: in_shardings = some_sharding

#### If we just return some_sharding:
broadcast_prefix will create: [some_sharding, some_sharding, some_sharding]
This wrongly applies sharding to the static arg!

#### What we need:
We build: (some_sharding, some_sharding, None)
This correctly marks position 2 as "no sharding"

vfdev-5: This makes sense. I was thinking how I would tackle this problem and thought about a similar idea as you implemented.
However, I also wonder why we have sharding involved when we transform args, kwargs into pure args, kwargs...

mohsinm-dev: @vfdev-5 Maybe I am incorrect, but the coupling exists because sharding information drives the tree transformation decision, not just the final JAX call.

`extract.to_tree` uses the prefix to decide which arguments to transform:
- Non-None prefix values → transform argument to NodeStates
- None prefix values → leave argument unchanged (static)

1. `broadcast_prefix(single_sharding, args)` → ['SHARD', 'SHARD', 'SHARD'] (applies to ALL positions, including static - incorrect)
2. `broadcast_prefix((shard1, shard2, None), args)` → ['SHARD1', 'SHARD2', None] (correctly marks static position)

And I don't think so without sharding information in the prefix, `extract.to_tree` has no way to distinguish which arguments should be transformed from those that should be preserved as static. The transformation step is where the sharding information is consumed.

vfdev-5:
@mohsinm-dev how about doing the following:
```python
from jax._src import api_util  # We use fun_signature and resolve_argnums

class JitWrapped(...):
  def __init__(self, ...):
    ...
    if isinstance(in_shardings, (list, tuple)):
      # We should reintroduce None values into in_shardings corresponding to static arguments
      fun_signature = api_util.fun_signature(fun)
      _, _, static_argnums, _ = api_util.resolve_argnums(
          fun,
          fun_signature,
          None,
          None,
          static_argnums,
          static_argnames,
      )
      in_shardings = list(in_shardings)
      for static_arg_index in sorted(static_argnums):
        in_shardings.insert(static_arg_index, None)
      in_shardings = tuple(in_shardings)

    self.jitted_fn = jax.jit(
      JitFn(fun, in_shardings, out_shardings, kwarg_shardings, self),
      in_shardings=self.jax_in_shardings,
      out_shardings=(None, None, self.jax_out_shardings),
      ...
```
(full code: https://github.com/vfdev-5/flax/pull/new/nnx-jit-static-args)

The idea is that we would like to use jax functions work arguments manipulations.

For the case, when `in_shardings` is not a tuple but a single sharding value, we can just pass it as is without transforming to a tuple and inserting Nones for static args.

We certainly should write various test cases.

What do you think ?

flax/nnx/transforms/compilation.py
vfdev-5: I was told that `extract.to_tree` with shardings is used to check aliasing for nnx modules. Here is a test doing a similar check: https://github.com/google/flax/blob/bb3c31ac84fe0b2638974a8b9430e77910b70854/tests/nnx/graph_utils_test.py#L765-L773

All decisions about static/non-static args are done in jax jit, not in flax.

mohsinm-dev:
@vfdev-5 Thanks for the suggestion. The current approach works as well, but, as you suggested, depending on the JAX, it is better, and we stay consistent with JAX's behavior for edge cases, such as mixed static_argnums and static_argnames.
I've implemented your approach and added comprehensive tests, all of which passed successfully. 

I've added 5 detailed test cases covering static args, edge cases, and StateSharding. Let me know if you want more scenarios or if any seem redundant.

flax/nnx/transforms/compilation.py
mohsinm-dev: @vfdev-5  You're right, `extract.to_tree` needs the prefix list to have the same length as the total number of arguments. That’s how it can tell when the same module is used more than once.

I tested it: if we give 3 arguments but only have 2 prefix entries, it can’t tell whether argument 0 and argument 2 point to the same object.

That’s why we need to include None entries in the prefix for static arguments. Even though JAX later removes static arguments, NNX still needs all of them first to check for aliasing correctly.

flax/nnx/transforms/compilation.py
```diff
@@ -128,16 +130,61 @@ def __call__(self, *pure_args, **pure_kwargs):
...
-      prefix=(self.in_shardings, self.kwarg_shardings, self.out_shardings),
+      prefix=(None, self.kwarg_shardings, self.out_shardings),
       ctxtag=self.ctxtag,
       split_fn=_jit_split_fn,
     )
 
     return pure_args_out, pure_kwargs_out, pure_out
 
 
+def _normalize_static_argnums(
```
vfdev-5: Can we remove this in favor of `api_util.resolve_argnums`?

flax/nnx/transforms/compilation.py
```diff
@@ -128,16 +130,61 @@ def __call__(self, *pure_args, **pure_kwargs):
     out = self.f(*args, **kwargs)
 
     args_out, kwargs_out = extract.clear_non_graph_nodes((args, kwargs))
+    # For packaging outputs, do not attempt to broadcast shardings for args_out;
+    # only the function output `out` needs an output sharding prefix.
     pure_args_out, pure_kwargs_out, pure_out = extract.to_tree(
       (args_out, kwargs_out, out),
-      prefix=(self.in_shardings, self.kwarg_shardings, self.out_shardings),
+      prefix=(None, self.kwarg_shardings, self.out_shardings),
```
vfdev-5: Let's revert this as this can be a BC-breaking.

flax/nnx/transforms/compilation.py
```diff
@@ -405,14 +479,31 @@ def __get__(self, obj, objtype=None):
     return functools.partial(self, obj)
 
   def _get_pure_args_kwargs(self, args, kwargs):
+    # Build prefix for extract.to_tree
```
vfdev-5: I think we can also revert these changes

flax/nnx/transforms/compilation.py
```diff
@@ -362,27 +409,54 @@ def __init__(
...
       JitFn(fun, in_shardings, out_shardings, kwarg_shardings, self),
       in_shardings=self.jax_in_shardings,
-      out_shardings=(
-        self.jax_in_shardings,
-        kwarg_shardings,
-        self.jax_out_shardings,
-      ),
+      # For outputs, rely on propagation for args/kwargs tuples and only
+      # optionally constrain the actual function output via jax_out_shardings.
+      out_shardings=(None, kwarg_shardings, self.jax_out_shardings),
```
vfdev-5: Finally, when discussed with the team, this part may be tricky as can be BC-breaking:
for example, user jits a function with args such that inside the function they use `jax.lax.with_sharding_constraint` to change temporarily the sharding and expect returned sharding to be as the input one. With `None` in out_shardings, the returned PartitionSpec will change. Here is an example:
```python
import os
os.environ["XLA_FLAGS"] = '--xla_force_host_platform_device_count=8'
os.environ["CUDA_VISIBLE_DEVICES"] = ''

import jax
import jax.numpy as jnp
from jax.sharding import NamedSharding, PartitionSpec as P
import flax.nnx as nnx

mesh = jax.make_mesh((4, 2), ('a', 'b'))

def sharding(*args):
  return NamedSharding(mesh, P(*args))

state_sharding = nnx.StateSharding(
  {
    nnx.PathContains('kernel'): sharding('a', 'b'),
    nnx.PathContains('bias'): sharding('b'),
  }
)

m = nnx.Linear(16, 32, rngs=nnx.Rngs(0))

assert not isinstance(m.kernel.sharding, NamedSharding), m.kernel.sharding

@nnx.jit(in_shardings=(state_sharding,))
def constrain_object(m):
  # m.kernel pspec is P('a', 'b') here
  new_sharding = NamedSharding(mesh, P('b', 'a'))
  m.kernel = jax.lax.with_sharding_constraint(m.kernel, new_sharding)
  # here we changed it to P('b', 'a')
  return None

constrain_object(m)
# Here user expects m.kernel to have pspec as P("a", "b")
# with out_shardings=(None, kwarg_shardings, self.jax_out_shardings) in nnx.jit
# this check will fail:
assert m.kernel.sharding.spec == P("a", "b"), m.kernel.sharding.spec
m.kernel.sharding.spec
``` 

What can we do here? This may be tricky. I'll try to see if we can have both (original bug fixed and out_shardings non-modified). A last resort could be to have a flag that either: 1) fixes the original bug and alters `out_shardings` or 2) keeps the bug and have `out_shardings` untouched.

flax/nnx/transforms/compilation.py
mohsinm-dev: But we need to build the prefix with None at static positions. I think it is essential for the fix.

vfdev-5: Unless I misunderstand your point which you may need to make more explicit, but this code fixes the problem: https://github.com/google/flax/pull/4996/files#diff-e8255d5a9ea23fbc5844e57701cee48a822d924bc05fa9e8330bd14f51688545R438-R452

flax/nnx/transforms/compilation.py
mohsinm-dev: I tested the BC concerns, and it's quite complicated :

`extract.to_tree` prefix (in JitFn.__call__)
Switching `prefix=(self.in_shardings, …)` to `prefix=(None, …)` breaks BC with with_sharding_constraint—sharding doesn’t revert to the input spec.

out_shardings
Using None instead of self.jax_in_shardings breaks BC. But when we keep self.jax_in_shardings and have static args, we hit a length mismatch:

`self.jax_in_shardings` matches only dynamic args (e.g., 2),

actual outputs include static args (e.g., 3 total),

JAX expects `out_shardings[0]` to match the full output length.

So we need BC (use `self.jax_in_shardings`) but also a way to handle static args without mismatches.

Can we keep prefix=(self.in_shardings, …) in JitFn.__call__ for BC. But for the out_shardings issue, we may need a different approach that handles both BC and static args correctly

vfdev-5: I updated my branch: https://github.com/google/flax/compare/main...vfdev-5:flax:nnx-jit-static-args
I think this keeps the BC and fixes the bug.
Can you check it from your side if your tests are failing with that code?
If all good, let's base this PR on my code + your tests. Concerning your tests if you can refactor them into a shorter version, it would be better

flax/nnx/transforms/compilation.py
```diff
@@ -768,31 +799,37 @@ def __call__(self, *pure_args, **pure_kwargs):
...
 def shard_map(
-  f: F,
-  *,
-  mesh: Mesh | AbstractMesh,
-  in_specs: Specs,
-  out_specs: Specs,
-  check_rep: bool = True,
-  auto: frozenset[AxisName] = frozenset(),
-) -> F: ...
+    f: F,
```
vfdev-5, line 798: Let's revert this and below changes as unrelated to the PR.

tests/nnx/transforms_test.py
```diff
@@ -420,6 +420,361 @@ def f(m: nnx.Linear, x):
     y = compiled(m, x)
     self.assertEqual(m.count.value, 2)
 
+  def test_jit_static_no_static_in_shardings(self):
```
vfdev-5: Add this test: https://github.com/google/flax/compare/main...vfdev-5:flax:nnx-jit-static-args#diff-bd2688f54ff71a854c098479cfca7c63914dbaa9fb725369cb733fb321b06447R421

Instead of your new 13 test cases, let's refactor them into 1-2 parametrized tests and try to avoid duplication as much as possible.

tests/nnx/transforms_test.py
```diff
@@ -420,6 +420,196 @@ def f(m: nnx.Linear, x):
     y = compiled(m, x)
     self.assertEqual(m.count.value, 2)
 
+  @parameterized.parameters(
+    {'static_argnums': (2,), 'static_argnames': None},
+    {'static_argnums': None, 'static_argnames': ('use_relu',)},
+  )
+  def test_jit_static_args_with_shardings(self, static_argnums, static_argnames):
```
vfdev-5, line 443: Let's keep just two most relevant tests:
- test_jit_static_args_with_shardings
- test_with_sharding_and_static_args

and remove others

vfdev-5:
@mohsinm-dev please fix the failing test: https://github.com/google/flax/actions/runs/18873989510/job/53861798191?pr=4996

PR #4998, opened 2025-10-04
Removed kernel_init/bias_init atttributes from popular layers

vfdev-5:
Description:
- Removed kernel_init/bias_init atttributes from popular layers
- Added tests

Note about recurrent layers like LSTM, we are storing carry_init in self to be used in initialize_carry method: https://github.com/google/flax/blob/771eadb1986e59a5a2dd82a14188fd1cd7b3849f/flax/nnx/nn/recurrent.py#L224-L225
I'm not sure how we could remove self.carry_init and keep BC for initialize_carry method.
Alternatively, we could add carry_init arg to initialize_carry method and also remove carry_init arg from ctor, but this would be BC-breaking...

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -355,10 +355,6 @@ class MultiHeadAttention(Module):
     self.dropout_rate = dropout_rate
     self.deterministic = deterministic
     self.precision = precision
-    self.kernel_init = kernel_init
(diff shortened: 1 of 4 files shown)


cgarciae (approved):

cgarciae:
Thanks @vfdev-5 !

PR #4999, opened 2025-10-05
Convert linen pytorch guide to nnx

### LINKED ISSUE #4997: Port pytorch migration guide from linen to nnx
In `docs/guides/converting_and_upgrading`, there's a guide to convert PyTorch models to flax ones. But this is only for linen; we should port the guide to `nnx` as `docs_nnx/migrating/convert_pytorch_to_flax.rst`.

samanklesaria:
Closes #4997


docs_nnx/migrating/convert_pytorch_to_flax.rst
```diff
@@ -0,0 +1,297 @@
...
+
+
+Convolutions
+--------------------------------
+
+Let's now look at 2D convolutions. PyTorch uses the NCHW format and Flax uses NHWC.
+Consequently, the kernels will have different shapes. The kernel in PyTorch has shape [outC, inC, kH, kW]
+and the Flax kernel has shape [kH, kW, inC, outC]. Transposing the kernel will do the trick.
+
+.. testcode::
```
cgarciae, line 54: We could use our custom `codediff` extension to nicely display this side by side.

samanklesaria: This defines the flax layer by inserting the same parameters that the pytorch one had. You'd never define a flax layer like this normally (manually inserting into `j_conv.kernel.value`). So outside of the context of the pytorch code, the flax code here doesn't make sense. It wouldn't be useful to isolate it in its own tab, especially as it depends on the pytorch code in the other tab.

PR #5001, opened 2025-10-06
Avoid install, import, or tests with tensorflow-text under Python 3.13+.

jburnim:
Some Flax examples use the `tensorflow-text` package, but `tensorflow-text` does not yet support Python 3.13 -- see https://github.com/tensorflow/text/issues/1421 .

This PR updates these Flax examples to skip running any tests that depend on `tensorflow-text` when the Python version is 3.13 or later.  It also updates the examples to only import `tensorflow-text` when the Python version is earlier than 3.13.


IvyZX (approved):
Lgtm! Please fix tests before adding pull_ready.

Issue #4995, opened 2025-10-03
How to construct freezable Parameters?


pfackeldey:
Hi,
I'm currently playing around with `nnx` (awesome library, I love the way model surgery works!), and JAX's new hijax feature, so, I'm very happy to see this development in https://github.com/google/flax/pull/4960.

My problem is as follows:
I'm trying to construct a `Parameter` class that is freezable (i.e. not differentiable) given a metadata attribute that I called 'frozen':
```python
from typing import TypeVar
from flax import nnx
import jax

T = TypeVar("T")
sg = jax.lax.stop_gradient

class Parameter(nnx.Variable[T]):
  """Freezable parameter class."""

  def __init__(self, value: T, frozen: bool, **kwargs):
    value = jnp.asarray(value)
    if frozen:
      value = sg(value)
    super().__init__(value=value, **kwargs)
    self.frozen = frozen
    
  @property
  def value(self):
    value = super().value
    return sg(value) if self.frozen else value
```
This doesn't seem to work, and also I'd like to have this dynamic (maybe I have to set [this hook](https://github.com/google/flax/blob/main/flax/nnx/variablelib.py#L451-L452) instead?), i.e. when I change this `frozen` from False to True the Parameter (value) should become non-differentiable, something like this:
```python
def fun(param: Parameter):
  ...
  param.frozen = True  # essentially `jax.lax.stop_gradient` for `param.value`
  ...
```
In my understanding this should be able to work (somehow) because `.frozen` is just another metadata attribute that's known at trace/compile time (like `.shape`) and independent of the data (numeric values of the array).

Would this somehow be possible to achieve? Also, how (or if) does this change in the context of the new `hijax` approach?

Thanks, Peter

vfdev-5:
Thanks for the question @pfackeldey !
I'm not sure to be able to fully answer this question, maybe @cgarciae can chime in with ideas.
However, the way you are using `lax.stop_gradient` does not seem to be correct to me, as it should be inside the function under `jax.grad` and not outside. For example:
```python
jax.tree.map(jax.grad(jnp.sin), {"o": jnp.asarray(2.0), "p1": sg(jnp.asarray(0.0))})
# {
#   'o': Array(-0.4161468, dtype=float32, weak_type=True),
#   'p1': Array(1., dtype=float32, weak_type=True)  <---- it is still 1.0 and not 0
# }
```

pfackeldey:
Thanks for your reply @vfdev-5! 
[...]

PR #5007, opened 2025-10-08
feat: add `nnx.set_metadata` to in-place change metadata of the state variables of `nnx.Module`s

pfackeldey:
@cgarciae I've put up the PR as discussed in #4995.
~Additionally, I've added a new filter (`IsDifferentiable`) that lets you split a state into a differentiable and a non-differentiable part. I could imagine that may be in some cases easier to express than through types of variables. If you don't like this, I'm happy to remove it.~

Please let me know what you think :) 

## Checklist
      checks if that's the case).
      link).

### DIFF
--- a/flax/nnx/graph.py
+++ b/flax/nnx/graph.py
@@ -30,8 +30,8 @@ from flax.nnx.proxy_caller import (
   CallableProxy,
   DelayedAccessor,
 )
-from flax.nnx.statelib import FlatState, State
-from flax.nnx.variablelib import Variable, is_array_ref
+from flax.nnx.statelib import FlatState, State, map_state
+from flax.nnx.variablelib import Variable, is_array_ref, V
 from flax.typing import Key, PathParts, is_key_like
 import jax
 import numpy as np
@@ -2883,6 +2883,41 @@ def call(
     return out, split(node)
 
   return CallableProxy(pure_caller)  # type: ignore
+
+
+def set_metadata(node: tp.Any, /, *, only: filterlib.Filter = Variable, **metadata: tp.Mapping[str, tp.Any]) -> None:
+  """Sets the metadata of all :class:`Variable` objects in the given graph node in-place.
+
+  Example::
+
+    >>> from flax import nnx
+    >>> import jax, jax.numpy as jnp
+    ...
+    >>> class Foo(nnx.Module):
+    ...   def __init__(self):
+    ...     self.param = nnx.Param(0.0)
+    ...     self.variable = nnx.Variable(0.0)
+    ...
+    >>> node = Foo()
+    ...
+    >>> # set differentiable to False for all nnx.Param objects
+    >>> nnx.set_metadata(node, differentiable=False, only=nnx.Param)
+    ...
+    >>> # check that only the nnx.Param was updated
+    >>> assert node.param.get_metadata('differentiable') is False
+
+  Args:
+    node: A graph node object.
+    only: A Filter to specify which :class:`Variable` objects to set metadata for.
+    metadata: Key-value pairs to set as metadata on the :class:`Variable` objects.
+  """
+  def _set_metadata(path: PathParts, variable: V) -> None:
+    del path  # unused
+    if isinstance(variable, Variable):
+      variable.set_metadata(**metadata)
+
+  # inplace update of variable_state metadata
+  map_state(_set_metadata, state(node, only))
 
 
 def iter_graph(node: tp.Any, /) -> tp.Iterator[tuple[PathParts, tp.Any]]:
--- a/tests/nnx/variable_test.py
+++ b/tests/nnx/variable_test.py
@@ -133,6 +133,24 @@ class TestVariable(absltest.TestCase):
     x = v.get_metadata('x', default=10)
     self.assertEqual(x, 10)
 
+  def test_set_module_metadata(self):
+    class Module(nnx.Module):
+      def __init__(self):
+        self.v = nnx.Variable(jnp.array(0.0))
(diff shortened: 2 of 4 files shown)


cgarciae:
Hey @pfackeldey ! Thanks for doing this. Two things:
1. Can we split this PR into smaller chunks of standalone changes? This seems like an important topic so we want to get every part right. Maybe lets start with just the `set_metadata` helper, I'll leave some comments.
2. Let me create a small proposal doc so we get the details right first before going into the actual implementation.

pfackeldey:
Thanks @cgarciae for looking at this! Sure, I fully agree :+1:

Looking forward to your comments and the proposal doc. I'm happy to adjust this PR accordingly and work on follow-up PRs :)

flax/nnx/graph.py
```diff
@@ -2885,6 +2885,45 @@ def pure_caller(accessor: DelayedAccessor, *args, **kwargs):
...
+    >>> # check that only the nnx.Param was updated
+    >>> assert node.param.get_metadata('differentiable') is False
+
+  Args:
+    node: A graph node object.
+    only: A Filter to specify which :class:`Variable` objects to set metadata for.
+    metadata: Key-value pairs to set as metadata on the :class:`Variable` objects.
+  Returns:
+    The graph node with updated :class:`Variable` metadata.
+  """
```
cgarciae, line 2913: Because the updates are being done in-place via `set_metadata` there is no need to return a new node. You can change `split(node, only, ...)` with `variable_state = state(node, only)`.

pfackeldey:
Hi @cgarciae, I've addressed your comments. This PR is now implementing the `nnx.set_metadata`-only as an in-place update as you suggested.

flax/nnx/graph.py
```diff
@@ -2885,6 +2885,40 @@ def pure_caller(accessor: DelayedAccessor, *args, **kwargs):
...
+    ...
+    >>> class Foo(nnx.Module):
+    ...   def __init__(self):
+    ...     self.param = nnx.Param(0.0)
+    ...     self.variable = nnx.Variable(0.0)
+    ...
+    >>> node = Foo()
+    ...
+    >>> # set differentiable to False for all nnx.Param objects
+    >>> nnx.graph.set_metadata(node, differentiable=False, only=nnx.Param)
```
cgarciae: we should export this to `nnx`

flax/nnx/graph.py
```diff
@@ -2885,6 +2885,40 @@ def pure_caller(accessor: DelayedAccessor, *args, **kwargs):
...
+    >>> assert node.param.get_metadata('differentiable') is False
+
+  Args:
+    node: A graph node object.
+    only: A Filter to specify which :class:`Variable` objects to set metadata for.
+    metadata: Key-value pairs to set as metadata on the :class:`Variable` objects.
+  """
+  def _set_metadata(path: PathParts, variable: V) -> V:
+    del path  # unused
+    variable.set_metadata(metadata)
```
cgarciae: in general there are other leaf types, check for Variable instances here

tests/nnx/variable_test.py
```diff
@@ -133,6 +133,15 @@ def test_get_set_metadata(self):
...
 
+  def test_set_module_metadata(self):
+    class Module(nnx.Module):
+      def __init__(self):
+        self.v = nnx.Variable(jnp.array(0.0))
+
+    m = Module()
+    self.assertTrue('foo' not in m.v.get_metadata())
+    nnx.set_metadata(m, foo='bar')
+    self.assertTrue(m.v.get_metadata() == {'foo': 'bar'})
```
vfdev-5, line 146: Let's add also a test from the docstring code snippet:
```python
nnx.set_metadata(node, differentiable=False, only=nnx.Param)
```

flax/nnx/graph.py
pfackeldey: It is already, I'll update the doc string :+1:

flax/nnx/graph.py
pfackeldey: Good point, would you make it a no-op for other types? i.e.:
```python
if isinstance(variable, Variable):
  variable.set_metadata(metadata)
```
Or would you prefer an appropriate error instead?

edit: I decided to make it a no-op for other types similar to how `.pure` works.

tests/nnx/variable_test.py
pfackeldey: I'll add that :+1:

vfdev-5:
@pfackeldey reported mypy failure is real (https://github.com/google/flax/actions/runs/18400512694/job/52585035298?pr=5007):
```
flax/nnx/graph.py:2916
```
Please fix it. Thanks!

vfdev-5:
@pfackeldey can you please also expose this new method (`nnx.set_metadata`) in https://github.com/google/flax/blob/main/docs_nnx/api_reference/flax.nnx/graph.rst to make it visible in https://flax--5007.org.readthedocs.build/en/5007/api_reference/flax.nnx/graph.html

The failing test https://github.com/google/flax/actions/runs/18460618237/job/52597425181?pr=5007 is also related to your change, please take a look.

Thanks! Otherwise, it looks good to me

vfdev-5:
@pfackeldey can you please squash commits history of this PR into 1 commit to have a green CI. Thanks!

cgarciae (approved):

cgarciae:
@pfackeldey can you please squash the PR into a single commit, CI checks for this

pfackeldey:
Sorry, @vfdev-5 and @cgarciae I didn't find the time to squash the commits yesterday. I squashed them now, thanks for reviewing and approving!

PR #5008, opened 2025-10-08
Fix nnx tabulate variable hooks

### LINKED ISSUE #5005: nnx.tabulate breaks if `nnx.Variable` implements one of the value hooks
Hi,
(another visualization issue... 😅)
The following gives an error, which it should not I suppose:
```python
from flax import nnx

class P(nnx.Variable):
  def on_set_value(self, value):
    return self.value + 1.0

class M(nnx.Module):
  def __init__(self):
    self.param = P(value=0.0)

  def __call__(self, x):
    return self.param * x

nnx.tabulate(M(), 2.0)
# ---------------------------------------------------------------------------
# RepresenterError                          Traceback (most recent call last)
# [/tmp/ipython-input-2062670779.py](https://qy6gqpckbx-496ff2e9c6d22116-0-colab.googleusercontent.com/outputframe.html?vrz=colab_20251003-060040_RC00_814643377#) in <cell line: 0>()
#      12     return self.param * x
#      13 
# ---> 14 nnx.tabulate(M(), 2.0)
# 
# 12 frames
# [/usr/local/lib/python3.12/dist-packages/yaml/representer.py](https://qy6gqpckbx-496ff2e9c6d22116-0-colab.googleusercontent.com/outputframe.html?vrz=colab_20251003-060040_RC00_814643377#) in represent_undefined(self, data)
#     229 
#     230     def represent_undefined(self, data):
# --> 231         raise RepresenterError("cannot represent an object", data)
#     232 
#     233 SafeRepresenter.add_representer(type(None),
#
# RepresenterError: ('cannot represent an object', <function P.on_set_value at 0x789fe827d3a0>)
```
The same issue exists if `on_get_value` or `on_create_value` is implemented.

Best, Peter

---
System information:
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): macOS Sequoia & Google Colab
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`: 0.12.0
- Python version: 3.12
- GPU/TPU model and memory: -
- CUDA version (if applicable): -

mohsinm-dev:
# What does this PR do?

Fixes #5005

Summary

- Prevents nnx.tabulate from crashing with yaml.representer.RepresenterError when a nnx.Variable stores value hooks (e.g.,
on_set_value) or other non-serializable objects in metadata.
- Filters out non-serializable callables and modules during YAML serialization in flax/nnx/summary.py::_as_yaml_str.
- Adds a deterministic fallback to repr(obj) for unknown non-serializable metadata, ensuring tabulate never fails on arbitrary
custom objects.
- Adds a focused test that validates the repr fallback appears in the tabulate output.

## Checklist

- [x] This PR fixes a minor issue (display-time crash) and includes targeted changes.
- [x] This change is discussed in an issue: #5005
- [x] The documentation and docstrings remain consistent with code behavior (no public API changes).

### DIFF
--- a/flax/nnx/summary.py
+++ b/flax/nnx/summary.py
@@ -543,10 +543,8 @@ def _normalize_values(x):
     return f'type[{x.__name__}]'
   elif isinstance(x, ArrayRepr | SimpleObjectRepr):
     return str(x)
-  elif callable(x):
-    return repr(x)
   else:
-    return x
+    return repr(x)
 
 def _maybe_pytree_to_dict(pytree: tp.Any):
   path_leaves = jax.tree_util.tree_flatten_with_path(pytree)[0]
@@ -649,20 +647,16 @@ def _as_yaml_str(value) -> str:
   value = _maybe_pytree_to_dict(value)
 
   file = io.StringIO()
-  try:
-    yaml.dump(
-      value,
-      file,
-      Dumper=NoneDumper,
-      default_flow_style=False,
-      indent=2,
-      sort_keys=False,
-      explicit_end=False,
-    )
-    return file.getvalue().replace('\n...', '').replace("'", '').strip()
-  except yaml.representer.RepresenterError:
-    # Fallback for non-serializable objects not caught by _normalize_values
-    return repr(value)
+  yaml.dump(
+    value,
+    file,
+    Dumper=NoneDumper,
+    default_flow_style=False,
+    indent=2,
+    sort_keys=False,
+    explicit_end=False,
+  )
+  return file.getvalue().replace('\n...', '').replace("'", '').strip()
 
 
 def _render_array(x):
--- a/tests/nnx/summary_test.py
+++ b/tests/nnx/summary_test.py
@@ -253,78 +253,47 @@ class SummaryTest(absltest.TestCase):
     self.assertEqual(sum([s.startswith("├─") for s in table.splitlines()]), 6)
 
   def test_tabulate_with_variable_hooks(self):
-    """Test that tabulate works with Variables implementing value hooks."""
-    
-    class VarWithSetHook(nnx.Variable):
-        def on_set_value(self, value):
-            # This hook should not cause YAML serialization errors
-            return self.value + 1.0
-
-    class Model(nnx.Module):
-        def __init__(self):
-            self.param = VarWithSetHook(value=0.0)
-
-        def __call__(self, x):
-            return self.param * x
-
-    module = Model()
-    # Should not raise yaml.representer.RepresenterError
-    table_repr = nnx.tabulate(module, 2.0, console_kwargs=CONSOLE_TEST_KWARGS)
-    self.assertIsNotNone(table_repr)
-    # Ensure the table contains expected content
-    self.assertIn('Model Summary', table_repr)
-    self.assertIn('param', table_repr)
-
-  def test_tabulate_with_multiple_hooks_and_metadata(self):
-    """Test tabulate with Variables that have hooks and custom metadata."""
-    
-    class VarWithHooks(nnx.Variable):
-        def on_get_value(self, value):
-            return value
-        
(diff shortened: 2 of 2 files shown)


flax/nnx/summary.py
```diff
@@ -645,6 +646,47 @@ def _as_yaml_str(value) -> str:
 
   value = jax.tree.map(_normalize_values, value)
   value = _maybe_pytree_to_dict(value)
+  
+  # Filter out non-serializable objects before YAML serialization
+  def _filter_non_serializable(obj):
```
vfdev-5: Thanks for the PR @mohsinm-dev !
I think we can fix the issue differently with less code modifications.
There is already `_normalize_values` that could be used to represent callables, for example:
```python
def _normalize_values(x):
  if isinstance(x, type):
    return f'type[{x.__name__}]'
  elif isinstance(x, ArrayRepr | SimpleObjectRepr):
    return str(x)
  elif callable(x):
    return repr(x)
  else:
    return x
```

What do you think?

mohsinm-dev: Thanks for the suggestion, @vfdev-5. I was initially thinking about selectively excluding specific hook methods to keep the output clean. But this is definitely cleaner and fits better with the existing design. I have updated the code.

flax/nnx/summary.py
```diff
@@ -647,16 +649,20 @@ def _as_yaml_str(value) -> str:
...
+      file,
+      Dumper=NoneDumper,
+      default_flow_style=False,
+      indent=2,
+      sort_keys=False,
+      explicit_end=False,
+    )
+    return file.getvalue().replace('\n...', '').replace("'", '').strip()
+  except yaml.representer.RepresenterError:
+    # Fallback for non-serializable objects not caught by _normalize_values
```
vfdev-5: Actually, I wonder why we can't just do this fallback in `_normalize_values`:
```python
def _normalize_values(x):
  if isinstance(x, type):
    return f'type[{x.__name__}]'
  elif isinstance(x, ArrayRepr | SimpleObjectRepr):
    return str(x)
  else:
    return repr(x)  #  <--- here 
```

Less code to maintain.

tests/nnx/summary_test.py
```diff
@@ -252,5 +252,79 @@ def __call__(self, x):
...
+        def __call__(self, x):
+            return self.param * x
+
+    module = Model()
+    # Should not raise yaml.representer.RepresenterError
+    table_repr = nnx.tabulate(module, 2.0, console_kwargs=CONSOLE_TEST_KWARGS)
+    self.assertIsNotNone(table_repr)
+    # Ensure the table contains expected content
+    self.assertIn('Model Summary', table_repr)
+    self.assertIn('param', table_repr)
```
vfdev-5:
```suggestion
    self.assertIn('param', table_repr)
    self.assertIn('on_set_value', table_repr)
```

tests/nnx/summary_test.py
```diff
@@ -252,5 +252,79 @@ def __call__(self, x):
...
+
+  def test_tabulate_with_multiple_hooks_and_metadata(self):
```
vfdev-5: Can we merge these test cases: test_tabulate_with_multiple_hooks_and_metadata and test_tabulate_with_custom_nonserializable_metadata with the above one: test_tabulate_with_variable_hooks.

mohsinm-dev: Yes, I think we can just merge it into the one that makes more sense.

flax/nnx/summary.py
mohsinm-dev: Yes, `_normalize_values` is already called via jax.tree.map() on all values before YAML serialization, so we can use repr(x) as the default fallback.

mohsinm-dev:
> LGTM, thanks @mohsinm-dev !

Thanks for the guidance and feedback!

PR #5009, opened 2025-10-08
Use sphinx 6.2+ for docs, which works with Python 3.13.

### LINKED ISSUE #4972: Upgrade sphinx for python 3.13
If I attempt to build the documentation with python 3.13, I get an error from sphinx about imghdr not existing (because version 3.13 removed it). Sphinx removed this dependency in version 6.2, so we could upgrade it. However, we'd need to adapt the documentation slightly; for example, sphinx's `get_documenter` function, which is used a lot in the docs, no longer exists in 6.2.

jburnim:
Fixes #4972


IvyZX:
Thanks for the fix!

PR #5015, opened 2025-10-09
optimize scan transpositions

cgarciae:
# What does this PR do?

Avoids trivial use of `moveaxis` in `nnx.scan` to allow JAX to perform some optimization for outputs that are simply forwarded from an input.

### DIFF
--- a/flax/nnx/transforms/iteration.py
+++ b/flax/nnx/transforms/iteration.py
@@ -733,7 +733,8 @@ def _scan_split_in(
         if axis is None:
           broadcast_states.append(state)
         elif isinstance(axis, int):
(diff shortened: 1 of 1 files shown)


jburnim:
Doesn't JAX already optimize `jnp.moveaxis(x, 0, 0)` to just `x` during tracing?

E.g.:
```python
def f(x):
    return jnp.moveaxis(x, 0, 0)

jax.make_jaxpr(f)(jnp.zeros((3, 4)))
# { lambda ; a:f32[3,4]. let  in (a,) }
```

cgarciae:
@jburnim JAX uses tracer identity to optimize some of transforms so the test would be more like this:

```python
x = jnp.zeros((3, 4))
assert x is jnp.moveaxis(x, 0, 0) # fails
```

PR #5038, opened 2025-10-21
Fix shard_map documentation link in compilation.py

vfdev-5:
Updated documentation to reflect the correct shard_map reference.


cgarciae:
@vfdev-5 can we update the Args section in the docstring that recently become outdated 🙏

cgarciae (approved):

PR #5043, opened 2025-10-22
Add NNX WeightNorm (update of #4568)

### LINKED ISSUE #4426: NNX WeightNorm

samanklesaria:
# What does this PR do?

This PR rebases the changes in #4568 to avoid merge conflicts with the main branch. That PR resolves #4426 by implementing an NNX WeightNorm class. 

The logic implemented here is similar to that of `nnx.SpectralNorm`, but makes a number of difference design choices:

- The code here uses a lazy initialization strategy rather than `SpectralNorm`'s eager one, which makes the code more concise and doesn't require runtime assertions that initialized parameters exist. 
- The code here uses `map_state` rather than doing modifications in place.
- The code here has a `variable_filter` argument (by default `nnx.PathContains('kernel')`) which specifies which part of the wrapped layer to modify. This is present in the linen version as well.

### DIFF
--- a/flax/nnx/nn/normalization.py
+++ b/flax/nnx/nn/normalization.py
@@ -904,14 +904,59 @@ class WeightNorm(nnx.Module):
     rngs: rnglib.Rngs,
   ):
     self.layer_instance = layer_instance
-    self.feature_axes = feature_axes
+    self.feature_axes = () if feature_axes is None else feature_axes
     self.use_scale = use_scale
     self.scale_init = scale_init
     self.epsilon = epsilon
     self.dtype = dtype
     self.param_dtype = param_dtype
-    self.variable_filter = variable_filter
-    self.rngs = rngs
+    self.variable_filter = nnx.filterlib.to_predicate(variable_filter)
+    self.scales : tp.Optional[dict] = None
+
+    if use_scale:
+      state = nnx.state(self.layer_instance, nnx.Param)
+      def init_scales(param):
+        feature_axes = _canonicalize_axes(param.ndim, self.feature_axes)
+        scale_shape = tuple(param.shape[ax] for ax in feature_axes)
+        return scale_init(rngs['params'], scale_shape)
+      self.scales = nnx.data({
+        path: init_scales(param) for path, param in nnx.to_flat_state(state)
+        if self.variable_filter(path, param)})
+
+  def _weightnorm_inplace(self, path, param):
+      if not self.variable_filter(path, param):
+        return
+
+      if self.feature_axes is None:
+        feature_axes = ()
+        reduction_axes = tuple(range(param.ndim))
+      else:
+        feature_axes = _canonicalize_axes(param.ndim, self.feature_axes)
+        reduction_axes = tuple(i for i in range(param.ndim) if i not in feature_axes)
+
+      value_bar = _l2_normalize(param, axis=reduction_axes, eps=self.epsilon)
+
+      if self.use_scale:
+        if path not in self.scales:
+          raise RuntimeError(
+            f"Could not find the scale corresponding to the param {path} "
+            "in scales dict. Parameters of the layer_instance should not change!"
+          )
+        scale_value = self.scales[path]
+
+        if len(feature_axes) < param.ndim:
+          broadcast_shape = [1] * param.ndim
+          for ax in feature_axes:
+            broadcast_shape[ax] = param.shape[ax]
+          scale_value = scale_value.reshape(broadcast_shape)
+        value_bar = value_bar * scale_value
+
+      cast_args = [param]
+      if self.use_scale:
+        cast_args.append(scale_value)
+
+      final_dtype = dtypes.canonicalize_dtype(*cast_args, dtype=self.dtype)
+      param.value = jnp.asarray(value_bar, final_dtype)
 
   def __call__(self, x: Array, *args, **kwargs) -> Array:
     """Compute the l2-norm of the weights in ``self.layer_instance``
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -34,16 +34,6 @@ repos:
         --extra-keys,
         "cell.metadata.executionInfo cell.metadata.id metadata.kernelspec metadata.vscode metadata.colab cell.metadata.executionInfo.user cell.metadata.executionInfo.user_tz cell.metadata.colab",
       ]
-- repo: https://github.com/astral-sh/ruff-pre-commit
```
vfdev-5, line 37: This change is unrelated to the PR topic, let's revert this change or make it in a separate PR.

flax/nnx/__init__.py
```diff
@@ -31,7 +31,6 @@
 from .graph import GraphDef as GraphDef
 from .graph import GraphState as GraphState
 from .graph import PureState as PureState
-from . import pytreelib as object
```
vfdev-5, line 34: Are the changes in this file except `from .nn.normalization import WeightNorm as WeightNorm` really needed?

flax/nnx/nn/normalization.py
```diff
@@ -849,6 +849,138 @@ def __call__(self, x, *, mask: tp.Optional[jax.Array] = None):
...
+      if self.use_scale:
+        scale_shape = tuple(param_val.shape[ax] for ax in feature_axes)
+        scale_path = path + ("scale",)
+        try:
+          scale_state = state[scale_path]
+          scale_value = scale_state.value
+        except KeyError:
+          key = self.rngs.params()
+          scale_value = self.scale_init(key, scale_shape, self.param_dtype)
+          state[scale_path] = nnx.Param(scale_value)
```
vfdev-5: Can we code this similarly to SpectralNorm?

flax/nnx/nn/normalization.py
```diff
@@ -849,6 +849,138 @@ def __call__(self, x, *, mask: tp.Optional[jax.Array] = None):
...
+      if self.use_scale:
+        cast_args.append(scale_value)
+
+      final_dtype = dtypes.canonicalize_dtype(*cast_args, dtype=self.dtype)
+      new_val = jnp.asarray(value_bar, final_dtype)
+
+      return nnx.Param(new_val)
+
+    state = nnx.map_state(apply_weightnorm, state)
+    nnx.update(self.layer_instance, state)
```
vfdev-5: Can we do an inplace changes like in SpectralNorm?
```
    state = nnx.state(self.layer_instance, nnx.Param)
    for path, param in nnx.to_flat_state(state):
      self._spectral_normalize_inplace(path, param, update_stats=update_stats)
```

.pre-commit-config.yaml
samanklesaria: Whoops- sorry, forgot to remove this.

flax/nnx/nn/normalization.py
samanklesaria: It's actually kind of nice to do lazy initialization here instead of the approach in `SpectralNorm`. We don't need to assert the existence of any specific paths, and we only need to modify the `__call__` method rather than both `__call__` and `__init__`. What if I converted SpectralNorm to this lazy approach instead?

vfdev-5: I think try/except in `__call__` can have a cost when called in the loop

flax/nnx/__init__.py
samanklesaria: Ruff automatically makes these changes using the standard pre-commit hook. The code in main must not have had the pre-commit hook run. Disabling pre-commit hooks runs directly counter to the contribution guide (https://flax.readthedocs.io/en/stable/contributing.html), so I'm disinclined to do it. Perhaps I should start with a PR that just runs pre-commit hooks on main, and then base this PR on it?

vfdev-5: There is test job running pre-commit hooks: https://github.com/google/flax/blob/5a2da6d2e35f1aeebfe50c508602c8f936d9e87f/.github/workflows/flax_test.yml#L34

samanklesaria: When I try to commit without these changes, I get the following:

Screenshot 2025-10-22 at 10 12 37 AM

Why might this not have occurred in the `flax_test` workflow? Do you see something similar?

flax/nnx/nn/normalization.py
```diff
@@ -842,6 +842,139 @@ def __call__(self, x, *, mask: tp.Optional[jax.Array] = None):
       self.epsilon,
     )
 
+class WeightNorm(nnx.Module):
```
vfdev-5, line 845: We need to expose this class into the docs: https://flax--5043.org.readthedocs.build/en/5043/search.html?q=WeightNorm

flax/nnx/nn/normalization.py
```diff
@@ -842,6 +842,139 @@ def __call__(self, x, *, mask: tp.Optional[jax.Array] = None):
...
+    self.param_dtype = param_dtype
+    self.variable_filter = nnx.filterlib.to_predicate(variable_filter)
+
+    if use_scale:
+      state = nnx.state(self.layer_instance, nnx.Param)
+      def init_scales(param):
+        feature_axes = _canonicalize_axes(param.ndim, self.feature_axes)
+        scale_shape = tuple(param.shape[ax] for ax in feature_axes)
+        return scale_init(rngs['params'], scale_shape)
+      self.scales = nnx.data({path: init_scales(param) for path, param in nnx.to_flat_state(state)
```
vfdev-5: Can we define `self.scales` attribute unconditionally, e.g. if use_scale is False as None?

tests/nnx/nn/normalization_test.py
```diff
@@ -323,6 +323,70 @@ def __call__(self, x, *, mask=None):
...
+    variables = linen_model.init(jax.random.key(1), x)
+
+    nnx_model = NNXModel(dtype=dtype, param_dtype=param_dtype, rngs=rngs)
+    nnx_model.dense.kernel.value = variables['params']['dense']['kernel']
+    nnx_model.dense.bias.value = variables['params']['dense']['bias']
+
+    linen_out = linen_model.apply(variables, x)
+
+    nnx_out = nnx_model(x)
+    np.testing.assert_array_equal(linen_out, nnx_out)
```
vfdev-5, line 391: Can we add `self.scales` comparsion when `use_scale=True`?

vfdev-5:
@samanklesaria looks good to me, just few minor things to improve.

vfdev-5:
@samanklesaria mypy is failing now due to the branch: `self.scales = None`.

PR #5045, opened 2025-10-22
Fix ValueError when `nnx.jit` is used with `nnx.custom_vjp`

### LINKED ISSUE #5044: ValueError when `nnx.jit` wraps `nnx.custom_vjp`.
The following code works with `jax.jit` but fails with `nnx.jit`:

```python
    @nnx.custom_vjp
    def f(x, y):
      return jnp.sin(x) * y

    def f_fwd(x, y):
      return f(x, y), (jnp.cos(x), jnp.sin(x), y)

    def f_bwd(res, g):
      cos_x, sin_x, y = res
      return (cos_x * g * y, sin_x * g)

    f.defvjp(f_fwd, f_bwd)
    nnx_out = nnx.jit(f)(jnp.array([1.0, 2.0]), jnp.array([3.0, 4.0]))
```

Here's the error:
```
flax/nnx/transforms/compilation.py:444: in __call__
    pure_args_out, pure_kwargs_out, pure_out = self.jitted_fn(
flax/nnx/transforms/compilation.py:119: in __call__
    args, kwargs = extract.from_tree(
flax/nnx/extract.py:290: in from_tree
    with graph.merge_context(ctxtag, is_inner) as merge_ctx:
../../.local/share/uv/python/cpython-3.11.6-macos-aarch64-none/lib/python3.11/contextlib.py:144: in __exit__
    next(self.gen)
flax/nnx/graph.py:1969: in merge_context
    ctx = current_update_context(ctxtag)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tag = 'custom_vjp_f_4439554496'

    def current_update_context(tag: tp.Hashable) -> UpdateContext:
      """Returns the current active :class:`UpdateContext` for the given tag."""
      if tag not in GRAPH_CONTEXT.update_context_stacks:
>       raise ValueError(f'No update context found for tag {tag!r}.')
E       ValueError: No update context found for tag 'custom_vjp_f_4439554496'.
E       --------------------
E       For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.
```

samanklesaria:
Resolves #5044.

### DIFF
--- a/flax/nnx/transforms/compilation.py
+++ b/flax/nnx/transforms/compilation.py
@@ -113,7 +113,10 @@ class JitFn:
   ctxtag: tp.Hashable
 
   def __post_init__(self):
-    functools.update_wrapper(self, self.f)
+    # Prevent overwriting our ctxtag info with the child function's
+    orig_ctxtag = self.ctxtag
+    functools.update_wrapper(self, self.f, updated=())
+    self.ctxtag = orig_ctxtag
 
   def __call__(self, *pure_args, **pure_kwargs):
     args, kwargs = extract.from_tree(
--- a/tests/nnx/transforms_test.py
+++ b/tests/nnx/transforms_test.py
@@ -191,6 +191,24 @@ class TestJIT(absltest.TestCase):
     assert m.a is a
     assert m.b is b
 
+  def test_jit_custom_vjp(self):
+    @nnx.custom_vjp
+    def f(x, y):
+      return jnp.sin(x) * y
+
+    def f_fwd(x, y):
+      return f(x, y), (jnp.cos(x), jnp.sin(x), y)
+
+    def f_bwd(res, g):
+      cos_x, sin_x, y = res
+      return (cos_x * g * y, sin_x * g)
+
+    f.defvjp(f_fwd, f_bwd)
+
+    nnx_out = nnx.jit(f)(jnp.array([1.0, 2.0]), jnp.array([3.0, 4.0]))
+    jax_out = jax.jit(f)(jnp.array([1.0, 2.0]), jnp.array([3.0, 4.0]))
+    assert (nnx_out == jax_out).all()
+
   def test_cached_unflatten_same_type(self):
     n = 0
 


flax/nnx/transforms/compilation.py
```diff
@@ -113,7 +113,8 @@ class JitFn:
   ctxtag: tp.Hashable
 
   def __post_init__(self):
-    functools.update_wrapper(self, self.f)
+    # Prevent overwriting our ctxtag info with the child function's by passing `updated=()`.
+    functools.update_wrapper(self, self.f, updated=())
```
vfdev-5: As you certainly understand, `self.f` has `ctxtag` attribute and just overwrites already set `self.ctxtag`.
I'm not sure about other side effects of setting `updated=()` here. I would tend to do that in a different way:
```python
    orig_ctxtag = self.ctxtag
    functools.update_wrapper(self, self.f)
    self.ctxtag = orig_ctxtag
```
or a variant similar to that.


For example, if we were creating a custom vjp of a function with ctxtag we could have a similar issue, hopefully CustomVjp class is coded a bit differently compared to `JitFn`:
https://github.com/google/flax/blob/5a2da6d2e35f1aeebfe50c508602c8f936d9e87f/flax/nnx/transforms/autodiff.py#L582-L594

However, I'm not totally sure about this code:
https://github.com/google/flax/blob/5a2da6d2e35f1aeebfe50c508602c8f936d9e87f/flax/nnx/transforms/autodiff.py#L450-L458

flax/nnx/transforms/compilation.py
```diff
@@ -113,7 +113,10 @@ class JitFn:
   ctxtag: tp.Hashable
 
   def __post_init__(self):
-    functools.update_wrapper(self, self.f)
+    # Prevent overwriting our ctxtag info with the child function's
+    orig_ctxtag = self.ctxtag
+    functools.update_wrapper(self, self.f, updated=())
```
vfdev-5:
```suggestion
    functools.update_wrapper(self, self.f)
```

PR #5047, opened 2025-10-23
Add split and key methods to RngStream and Rngs

### LINKED ISSUE #5046: Give RngStream the same interface as Jax's stateful RNGs
Jax will likely support a stateful PRNG object similar to Flax's RngStreams in the future: https://github.com/jax-ml/jax/pull/28845.

Eventually, we'll want to have flax use these native flax objects. To ease that eventual transition, we should give RngStreams the same interface (specifically, the `key` and `split` methods) and express the new explicit-rng docs in terms of this interface.

samanklesaria:
Closes #5046


samanklesaria:
Unfortunately, the `key` attribute on streams already exists. I have renamed the existing attribute `key_`, and replaced its use with the hand sed command `sed -I '' "s/\.key\([^_(]\)/.key_\1/"`.

cgarciae:
Thanks @samanklesaria ! 
I'll have to run all internal tests before merging.

As a follow up PR, we could consider:
* replace all usage of `jax.random.<function>(rngs(), ...)` with `rngs.<function>(...)`. I've done some of this.
* when we do need a key, favor `rngs.key()` or `rngs.some_stream.key()`.

flax/jax_utils.py
```diff
@@ -316,3 +317,41 @@ def unpad(x):
     return out if static_return else jax.tree_util.tree_map(unpad, out)
 
   return pad_shard_unpad_wrapper
+
+
+class _DictOrList(dict):
```
cgarciae, line 322: What is this for?

samanklesaria: It's a utility datatype used by `build_tree_from_paths`, which is an inverse of `jax.tree.leaves_with_path`. It's useful for defining the bridge code porting old rng state to the new names. See the changes to the tutorial.

PR #5052, opened 2025-10-24
deprecate Variable.value

cgarciae:
# What does this PR do?


flax/configurations.py
```diff
@@ -201,6 +201,38 @@ def static_bool_env(varname: str, default: bool) -> bool:
     )
 
 
+def str_flag(name: str, *, default: str, help: str) -> FlagHolder[str]:
```
IvyZX: Is this part of the PR? I didn't see any flags using it

cgarciae: Yeah it was outdated. Updated the branch.

Issue #5053, opened 2025-10-25
nnx.remat (and other transformations) is easily misused.


liudangyi:
nnx.remat needs to be applied to (unbound) functions, but it's easy to apply it to a bound method, e.g.

```
class Model(nnx.Module):

  def __init__(self, *, rngs: nnx.Rngs):
    self.linear = nnx.Linear(16, 32, rngs=rngs)
    self.x_max = jnp.array(0.0)

  def __call__(self, x):
    return nnx.remat(self.block)(x)

  def block(self, x):
    self.x_max = jnp.maximum(self.x_max, x.max())
    return self.linear(x)

model = Model(rngs=nnx.Rngs(0))
model(jnp.ones((3, 16)))
# >>> TraceContextError: Cannot mutate 'Model' from different trace level
```

Instead, we need to change `nnx.remat(self.block)(x)` to `nnx.remat(self.block.__func__)(self, x)`.

Is it possible to automate this? e.g., when the function passed to `nnx.remat` is bound, we automatically unbound it and pass the module as the first argument.

mohsinm-dev:
I tried to reproduce it, transforms must see Modules as explicit args so split/merge and update_context can track mutations. Passing a bound method (e.g., self.block) captures the Module by closure, so it doesn’t flow through the transform, causing trace-level mismatches (TraceContextError). Partials over bound methods and index-based static_argnums further obscure argument positions.

I think we should normalize the callable first: unwrap `functools.partial,` unbind Module methods, reapply partial args/kwargs, and resolve kwargs to positional args. Build the transform with merge_inputs inside and split_inputs outside under update_context. If originally bound, inject the Module as arg0; shift index-based static_argnums by +1. And then it needs to apply the same normalization to the other callable-based transforms.

samanklesaria:
@mohsinm-dev That makes sense to me. Do you want to write up a PR, or shall I?

mohsinm-dev:
@samanklesaria, I will open the PR I am just thinking a bit more of edge cases and see if there can be a better solution. Feel free to suggest if you get an idea.

mohsinm-dev:
@samanklesaria, can you please review the PR. Let me know if you have any feedback. I will work on it.

samanklesaria:
@vfdev-5 noted an easy way around this: use `nnx.remat` as a decorator instead. The methods passed to decorators will be unbound, so the problem goes away:

```python
class Model(nnx.Module):

[...]

PR #5055, opened 2025-10-28
Fix bound method auto-unbinding for NNX transforms

mohsinm-dev:
Resolves issue #5053 by automatically detecting and unbinding Module methods in JAX transforms.

- Adds helper functions to detect bound methods and convert them to unbound callables
- Updates all NNX transforms to handle bound methods seamlessly  
- Includes comprehensive test coverage for the new functionality
- Maintains full backwards compatibility


flax/nnx/transforms/autodiff.py
```diff
@@ -312,15 +316,42 @@ def grad(
       holomorphic=holomorphic,
       allow_int=allow_int,
     )
-  return _grad_general(
-    f,
+  # Detect bound nnx.Module methods and normalize to unbound callable.
+  f_unbound, bound_self, was_bound = _resolve_bound_callable(f)
+
+  # For bound methods, argnums=0 should refer to the module, so we don't shift.
```
samanklesaria: Here, you treat an argnums of zero as a special case. In general, you have the argnums indicating arguments to the *bound* method. But when an argnum is zero, it indicates an argument to the *unbound* method. That inconsistency is a bit confusing- ideally we'd always interpret from the bound perspective or always interpret from the unbound one. One option would be to always interpret the argnums as applying to bound method: that is, shift everything up by one, and then add zero (so that we *always* take gradients against self). But that seems overly restrictive. Another would be to always interpret argnums as applying to the unbound method. But then the default value of zero would unintuitively not differentiate against the first method argument. We could fix this by making the defaults for bound and unbound functions different- use (0) as usual for unbound  functions, but (0,1) for bound ones. This still introduces an inconsistency, but it feels less likely to trip people up than treating zero as a special case. I'm not sure what the right behavior should be- I'll ask around.

flax/nnx/transforms/autodiff.py
```diff
@@ -366,15 +397,39 @@ def value_and_grad(
       holomorphic=holomorphic,
       allow_int=allow_int,
     )
-  return _grad_general(
-    f,
+  # Detect bound nnx.Module methods and normalize to unbound callable.
```
samanklesaria: Could we use the `_shift_indices` function in this case and the previous one, assuming we don't treat zero as a special case? I guess you'd need to make `_shift_indices` handle `DiffState` too.

flax/nnx/transforms/autodiff.py
mohsinm-dev: Thanks, I just realized this, treating 0 as a special case mixes "bound" and "unbound" perspectives and led to real indexing bugs for user args.

Looking at the three options you mentioned, I think the best approach would be: unbound perspective and remove all shifting. This gives a single rule that matches JAX's argnums (index positional args) and our injection of the module at index 0. It fixes the current bug and is easiest to explain: 0 → module, 1 → first user arg, 2 → second, etc. What do you think?

If we want bound-friendly defaults later (e.g., (0,1) for bound), we can layer that on top once the team agrees.

flax/nnx/transforms/autodiff.py
mohsinm-dev: I've created the `_normalize_grad_argnums()` helper function that's now used by both `grad` and `value_and_grad` to eliminate the code duplication.

mohsinm-dev: We can use `_shift_indices`, but right now I am thinking about the unbound approach, where we don't need any shifting. If we later decide to implement shifting-based approaches, using/extending `_shift_indices` would definitely be the right way to go

flax/nnx/transforms/autodiff.py
samanklesaria: Say you do the following:

```
class Model(nnx.Module):
    self.linear = nnx.Linear(16, 32, rngs=rngs)
      
  def block(self, x):
    return jnp.sum(self.linear(x))

model = Model(nnx.Rngs(0))
x = jnp.ones((3, 16))
jax.grad(model.block, argnums=0)(x)
```

Then jax will differentiate wrt x, not model. To follow our language in the previous comments, jax uses the "bound perspective" rather than the "unbound" one. 

Nnx tries to stay as close as possible to the jax transformation API. Now, we could support this by transforming `model.block` into `Model.block`, pass in `model` and `x` as args, and increment all the argnums given to `nnx.grad`. But this means derivation wrt the model itself would be impossible given a bound method argument. This naturally limits the utility of being able to apply `nnx.grad` (or any other transformation) to bound methods. Instead, users should be writing `jjx.grad(Model.bock)(model, x)`.

samanklesaria:
@mohsinm-dev I discussed the issue of nnx transforms on bound methods with the flax maintainers. They think there's just too many subtle bugs that might arise if we try to support all the different nnx transformations. All the different flavors of positional arguments (in_specs, static_argnums, in_axes, nondiff_argnums, etc) would need to be modified, resulting in a pretty substantial chunk of code. Instead, the consensus was that applying nnx transforms to bound methods should result in an error (similar to what you currently have for `nnx.scan`) encouraging users to use decorators or pass self arguments explicitly (as in `nnx.transform(Model.method)(model, x)` rather than `nnx.transform(model.method)(x)`). Apologies for not figuring this out before you put in all the work for the PR! Would you be willing to modify this PR to produce the relevant errors instead?

flax/nnx/transforms/autodiff.py
mohsinm-dev: I think we should align fully with JAX’s semantics for bound methods: when a bound method is passed to nnx.grad/value_and_grad, interpret argnums from the “bound” perspective (i.e., exclude self) and uniformly add +1 to all indices (including DiffState.argnum); when an unbound function is passed, apply no shift. This will also resolve the current bug caused by the special-casing rule (x + 1 if x > 0 else 0), which I added. It also keeps us consistent with how jit/remat already shifts static_argnums for bound methods. As with JAX, this means you can’t differentiate w.r.t. the module when calling grad on a bound method; if needed, use the unbound form instead (e.g., `nnx.grad(Model.block) (model, x)`).

mohsinm-dev:
@samanklesaria Thanks for checking with the team that makes sense. I’ll update the PR to stop supporting bound-method callables across nnx transforms and instead raise a clear, consistent error (like nnx.scan already does) whenever a bound Module method is passed to nnx.grad/value_and_grad, nnx.remat, nnx.jit, nnx.vmap, nnx.pmap, nnx.checkify, and nnx.eval_shape. The error will explain the two supported patterns with examples: (1) use the decorator form on the method (@nnx.remat, @nnx.jit, etc.), which naturally operates on the unbound function, or (2) pass the unbound method
explicitly and supply the instance as the first argument (e.g., `nnx.grad(Model.block)(model, x)` rather than `nnx.grad(model.block)(x))`.

No worries about the work pivot, exploring the argnum complexity was valuable for understanding why bound method support is problematic

flax/nnx/transforms/autodiff.py
```diff
@@ -60,6 +62,7 @@ class DiffState:
   filter: filterlib.Filter
 
 
+
```
samanklesaria, line 65: Nit: remove

flax/nnx/transforms/autodiff.py
```diff
@@ -366,15 +377,23 @@ def value_and_grad(
...
     )
-  return _grad_general(
-    f,
+  # Detect bound nnx.Module methods and raise error.
+  f_unbound, _, was_bound = _resolve_bound_callable(f)
+  
+  if was_bound:
+    _raise_bound_method_error('value_and_grad')
+
+  vg_fn = _grad_general(
```
samanklesaria: Just return the call to `_grad_general` like before.

flax/nnx/transforms/compilation.py
```diff
@@ -346,6 +360,7 @@ def jit(
     backend=backend,
     inline=inline,
     abstracted_axes=abstracted_axes,
+    bound_self=None,
```
samanklesaria: Remove

flax/nnx/transforms/compilation.py
```diff
@@ -372,9 +387,11 @@ def __init__(
     backend: tp.Optional[str] = None,
     inline: bool = False,
     abstracted_axes: tp.Optional[tp.Any] = None,
+    bound_self: tp.Any | None = None,
```
samanklesaria: Remove

flax/nnx/transforms/compilation.py
```diff
@@ -372,9 +387,11 @@ def __init__(
...
   ):
     functools.update_wrapper(self, fun)
     self.fun: tp.Callable[P, R] = fun
+    self._bound_self = bound_self
```
samanklesaria: Remove

flax/nnx/transforms/compilation.py
```diff
@@ -419,6 +436,11 @@ def __get__(self, obj, objtype=None):
     return functools.partial(self, obj)
 
   def _get_pure_args_kwargs(self, args, kwargs):
+    # If this JIT was created from a bound method, inject the Module as arg0
```
samanklesaria: Remove

flax/nnx/transforms/iteration.py
```diff
@@ -1293,6 +1307,12 @@ def forward(x, model):
         transform_metadata=transform_metadata,
     )  # type: ignore[return-value]
 
+  # Bound methods are ambiguous for scan (carry vs broadcast vs mapped state).
```
samanklesaria: Remove line

PR #5056, opened 2025-10-28
Set Mode with Tests

chapman20j:
# What does this PR do?
Implements the set_mode function based on graph.recursive_map
Implements set_mode as a class method in MultiheadAttention, Batchnorm, and Dropout
Implements set_train and set_eval based on set_mode
Adds tests for set_mode

Fixes # (issue)

## Checklist
- [ x] The documentation and docstrings adhere to the [documentation guidelines](https://github.com/google/flax/blob/main/docs/README.md#how-to-write-code-documentation).
- [ x] This change includes necessary high-coverage tests. (No quality testing = no merge!)

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -638,6 +638,51 @@ class MultiHeadAttention(Module):
     self.cached_value = nnx.Cache(jnp.zeros(cache_shape, dtype))
     self.cache_index = nnx.Cache(jnp.array(0, dtype=jnp.int32))
 
+  def set_mode(
+      self,
+      deterministic: bool | None = None,
+      decode: bool | None = None,
+      batch_size: int | Shape | None = None,
+      max_length: int | None = None,
+      **kwargs,
+  ) -> dict:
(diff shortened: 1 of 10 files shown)


flax/nnx/nn/attention.py
```diff
@@ -638,6 +638,58 @@ def init_cache(self, input_shape: Shape, dtype: Dtype = jnp.float32):
     self.cached_value = nnx.Cache(jnp.zeros(cache_shape, dtype))
     self.cache_index = nnx.Cache(jnp.array(0, dtype=jnp.int32))
 
+  def init_cache2(
```
cgarciae: We probably need to pick a better name than `init_cache2`, originally only used it for testing.

cgarciae: We could also just inline this in `set_mode`. The original idea for a separate method was to try to reuse it in `__init__`.

chapman20j: Just inlined the function in set_mode. Thanks for the feedback!

PR #5064, opened 2025-10-31
Update mnist example to use NNX

### LINKED ISSUE #5054: Tensorflow test dependency needs upgrade to support orbax
Say you're on an M3 mac with python 3.11 trying export a simple model like this:

```
import tensorflow as tf
import jax
import jax.numpy as jnp
from orbax.export import JaxModule, ExportManager, ServingConfig

yy = jax.random.normal(jax.random.key(9), 256)

def my_pred(yy, x):
    return jnp.dot(xx, yy)

xx = jnp.ones(256)

jax_module = JaxModule(yy, my_pred)

def to_tfspec(a):
    return tf.TensorSpec(shape=a.shape, dtype=a.dtype)

sig = jax.tree.map(lambda a: to_tfspec(jax.typeof(a)), [xx])

export_mgr = ExportManager(jax_module, [
    ServingConfig('serving_default', input_signature=sig)
])

output_dir='/tmp/blahblah'
export_mgr.save(output_dir)
```

We get the following error: "XlaCallModule from your TensorFlow installation supports up to serialization version 9 but the serialized module needs version 10. You should upgrade TensorFlow, e.g., to tf_nightly."

I guess this means we need to depend on `tf_nightly` if we want docs to demonstrate orbax functionality. 

Upon installing `tf_nightly`, however, we get another issue:

"Detected mismatched Protobuf Gencode/Runtime major versions when loading tensorflow/core/framework/attr_value.proto: gencode 6.31.1 runtime 5.29.5. "

I tried upgrading protobuf, but the issue doesn't go away. How do I get the protobuf runtime to use version 6.31.1 too?

samanklesaria:
Update mnist example to use NNX. Closes #5054.  Same as #5002, only not coming from copybara (so potentially easier to merge with changers).

### DIFF
--- a/docs_nnx/mnist_tutorial.ipynb
+++ b/docs_nnx/mnist_tutorial.ipynb
@@ -28,7 +28,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# !pip install flax"
+    "# !pip install -U \"jax[cuda12]\"\n",
+    "# !pip install -U flax"
    ]
   },
   {
--- a/docs_nnx/mnist_tutorial.md
+++ b/docs_nnx/mnist_tutorial.md
@@ -25,7 +25,8 @@ Let’s get started!
 If `flax` is not installed in your Python environment, use `pip` to install the package from PyPI (below, just uncomment the code in the cell if you are working from Google Colab/Jupyter Notebook):
 
 ```{code-cell} ipython3
-# !pip install flax
+# !pip install -U "jax[cuda12]"
+# !pip install -U flax
 ```
 
 ## 2. Load the MNIST dataset


docs_nnx/mnist_tutorial.ipynb
```diff
@@ -14,14 +14,8 @@
     "\n",
```
chapman20j, line 14: This doesn't run on Google colab. Can you please update the flax import and any other necessary imports?

samanklesaria: I didn't change any of the pip install or import statements. What's going wrong on Colab for you? It seems to work for me.

chapman20j: I think updating the line
```python
# !pip install flax
```
to
```python
# !pip install -U "jax[cuda12]"
# !pip install -U flax
```
addresses this. Giving the cuda installation could help with running the example quickly in a colab. When I run without the `-U` flag I get errors since the flax installation is `0.10.7` which is behind the current flax version.

samanklesaria:
Closing in favor of #5188

PR #5066, opened 2025-11-01
add nnx.dataclass

cgarciae:
Adds `nnx.dataclass` and allows `nnx.data` and `nnx.static` to be used as field specifiers. `nnx.Data` and `nnx.Static` are deprecated, the main reason is because python currently turns type annotations into strings when forward references are involved and there is no good solution for detecting the annotations.

### DIFF
--- a/docs_nnx/guides/pytree.ipynb
+++ b/docs_nnx/guides/pytree.ipynb
@@ -395,8 +395,8 @@
    "id": "8055b72c",
    "metadata": {},
    "source": [
-    "### Class Annotations\n",
-    "Annotations can also be added at the type level via `nnx.Static` and `nnx.Data`. This is useful for creating `dataclasses` but the mechanism also works for regular classes."
+    "### Dataclasses\n",
+    "`nnx.Pytree` dataclasses can be created by using the `nnx.dataclass` decorator. To control the status of each field, `nnx.static` and `nnx.data` can be used as `field` specifiers."
    ]
   },
   {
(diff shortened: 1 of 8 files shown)


docs_nnx/guides/pytree.ipynb
```diff
@@ -395,8 +395,8 @@
    "id": "8055b72c",
    "metadata": {},
    "source": [
-    "### Class Annotations\n",
-    "Annotations can also be added at the type level via `nnx.Static` and `nnx.Data`. This is useful for creating `dataclasses` but the mechanism also works for regular classes."
+    "### Dataclasses\n",
+    "To create `nnx.Pytree` dataclasses, `nnx.static` and `nnx.data` can be used as `field` specifiers for `nnx.dataclass`."
```
jburnim: What do you think of starting this paragraph with something like?:
  "The `nnx.dataclass` annotation can be used to create a subclass of `nnx.Pytree` that is also a dataclass."
or:
  "To create a dataclass that is also an `nnx.Pytree`, use the `nnx.dataclass` annotation."
(And then mentioning `nnx.static` and `nnx.data` in the next sentence.)

docs_nnx/guides/pytree.ipynb
```diff
@@ -621,15 +654,15 @@
   },
```
jburnim, line 654: Just above this -- looks like "in inside" should just be "inside".

flax/nnx/pytreelib.py
```diff
@@ -191,13 +198,21 @@ def is_data(value: tp.Any, /) -> bool:
...
-  value: tp.Any
-
-
-def static(value: A, /) -> A:
+@tp.overload
+def static(value: A, /) -> A: ...
+@tp.overload
+def static(
+  *,
+  default: A | None = None,
```
jburnim: Does the default need to be `dataclasses.MISSING`?

Otherwise, if the user doesn't specify a default, do we end up specifying a default value of `None`?

flax/nnx/pytreelib.py
```diff
@@ -191,13 +198,21 @@ def is_data(value: tp.Any, /) -> bool:
...
+  default_factory: tp.Callable[[], A] | None = None,
```
jburnim: `dataclasses.MISSING` here, too?

flax/nnx/pytreelib.py
```diff
@@ -52,31 +52,25 @@
...
-
-
-def data(value: A, /) -> A:
+@tp.overload
+def data(value: A, /) -> A: ...
+@tp.overload
+def data(
+  *,
+  default: A | None = None,
+  default_factory: tp.Callable[[], A] | None = None,
```
jburnim: `MISSING` for `default` and `default_factory` here?

tests/nnx/module_test.py
```diff
@@ -91,8 +91,8 @@ class Foo(nnx.Pytree):
   def test_consistent_attrs_dataclass(self):
     @dataclasses.dataclass
     class Foo(nnx.Pytree):
-      a: nnx.Data[int]
```
jburnim, line 94: Will `nnx.Data` and `nnx.Static` continue to work until we remove them?  If so, should we leave in tests of `nnx.Data` and `nnx.Static` until then?

flax/nnx/pytreelib.py
cgarciae: Added `dataclasses.MISSING` to `default`, but the real implementation doesn't care about this.

cgarciae: Ok, using a bit of "type: ignore" we can use `dataclasses.MISSING` here, initially didn't want to do this because there is no `dataclasses.MissingType`.

tests/nnx/module_test.py
cgarciae: Flax CI raises on warnings so we are forced to fix the test. I'll try to add a warning ignore and duplicate the test.

docs_nnx/guides/pytree.ipynb
cgarciae: Thanks! Fixed.

tests/nnx/module_test.py
cgarciae: Created a new text for nnx.Data/Static and added a rule to ignore to the deprecation warning in `pyproject.toml`.

flax/nnx/pytreelib.py
cgarciae: Added use of `dataclasses.MISSING` for `default` but not `default_factory`. As said elsewhere, these defaults are just for typing, the real implementation doesn't use this.

flax/nnx/pytreelib.py
cgarciae: Same as above.

Issue #5067, opened 2025-11-03
nnx.tabulate expects concrete values in unreleased flax>0.12.0


DBraun:
I have the latest flax installed with `pip install -U git+https://github.com/google/flax.git` (specifically https://github.com/google/flax/commit/74985b29404a8da3ab767fdfbbf8ab5ecb532574). nnx.tabulate is not working because it seems to require shapes to be concrete.

### System information
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): WSL2 on Windows Pro
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`:
```
Name: flax
Version: 0.12.0
Summary: Flax: A neural network library for JAX designed for flexibility
Home-page: https://github.com/google/flax
Author:
Author-email: Flax team <flax-dev@google.com>
License:
Location: /home/admin/.local/lib/python3.12/site-packages
Requires: jax, msgpack, numpy, optax, orbax-checkpoint, PyYAML, rich, tensorstore, treescope, typing_extensions
Required-by: clu, evosax, jax-ai-stack, jraphx
---
Name: jax
Version: 0.8.0
Summary: Differentiate, compile, and transform Numpy code.
Home-page: https://github.com/jax-ml/jax
Author: JAX team
Author-email: jax-dev@google.com
License: Apache-2.0
Location: /home/admin/.local/lib/python3.12/site-packages
Requires: jaxlib, ml_dtypes, numpy, opt_einsum, scipy
Required-by: chex, clu, evosax, flax, jax-ai-stack, jaxloudnorm, jraphx, optax, orbax-checkpoint, orbax-export
---
Name: jaxlib
Version: 0.8.0
Summary: XLA library for JAX
Home-page: https://github.com/jax-ml/jax
Author: JAX team
Author-email: jax-dev@google.com
License: Apache-2.0
Location: /home/admin/.local/lib/python3.12/site-packages
Requires: ml_dtypes, numpy, scipy
Required-by: chex, clu, jax, jraphx, optax, orbax-export
```
- Python version: 3.12
- GPU/TPU model and memory: RTX 4080
- CUDA version (if applicable): 12

### Problem you have encountered:
MRE:
```python
from jax import numpy as jnp
from flax import nnx


class Net(nnx.Module):
    def __init__(self):
        self.rngs = nnx.Rngs(0)

    def __call__(self, x):
        return self.rngs.uniform((x.shape[0], 10))


if __name__ == '__main__':
    net = Net()
[...]

PR #5069, opened 2025-11-03
Use nnx split during `tabulate`

samanklesaria:
This PR addresses #5067. Currently, the `tabulate` function tries to `jit` compile every single method of the input object and its decedents so as to support the 'compute_flops' option. If these methods have dynamic shapes or do other things jit doesn't support, this will break. Specifically, consider the following code:

```python
class Net(nnx.Module):
    def __init__(self):
        self.rngs = nnx.Rngs(0)

    def __call__(self, x):
        return self.rngs.uniform((x.shape[0], 10))

net = Net()
x = jnp.zeros((4, 8))
nnx.tabulate(net, x, console_kwargs={"width": 200})
```

We get this error on main:

```
   def uniform(...):
      key, _ = _check_prng_key("uniform", key)
      dtype = dtypes.check_and_canonicalize_user_dtype(
          float if dtype is None else dtype)
>     shape = core.canonicalize_shape(shape)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E     TypeError: Shapes must be 1D sequences of concrete values of integer type, got (JitTracer<~int32[]>, JitTracer<~int32[]>).
E     If using `jit`, try using `static_argnums` or applying `jit` to smaller subfunctions.
```

With flax 0.12 (and also after this PR), we get the following:

```
┏━━━━━━━━━━━━━━┳━━━━━━┳━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ path         ┃ type ┃ inputs       ┃ outputs       ┃ RngState ┃
┡━━━━━━━━━━━━━━╇━━━━━━╇━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━┩
│              │ Net  │ float32[4,8] │ float32[4,10] │ 2 (12 B) │
├──────────────┼──────┼──────────────┼───────────────┼──────────┤
│ rngs/uniform │ Rngs │ - 4          │ float32[4,10] │ 2 (12 B) │
│              │      │ - 10         │               │          │
├──────────────┼──────┼──────────────┼───────────────┼──────────┤
│ rngs         │ Rngs │              │ key<fry>[]    │ 2 (12 B) │
├──────────────┼──────┼──────────────┼───────────────┼──────────┤
│              │      │              │         Total │ 2 (12 B) │
└──────────────┴──────┴──────────────┴───────────────┴──────────┘
                                                                 
                   Total Parameters: 2 (12 B) 
```

This PR allows users to look at input and output shapes of dynamic methods by avoiding jitting the methods. Specifically, it tries to trace the jitted version of each method, but will fall back to the non-jitted version if it encounters an error, and will return 0 as the flops of a non-jittable function as is the convention inherited from linen tabulate.

### DIFF
--- a/tests/nnx/summary_test.py
+++ b/tests/nnx/summary_test.py
@@ -225,6 +225,32 @@ class SummaryTest(absltest.TestCase):
     # We should see 3 calls per block, plus one overall call
     self.assertEqual(sum([s.startswith("├─") for s in table.splitlines()]), 7)
 
+  def test_time_complexity(self):
+    counter = []
+
+    class Block(nnx.Module):
+      def __init__(self, rngs):
+        self.linear = nnx.Linear(2, 2, rngs=rngs)
+
+      def __call__(self, x):
+        counter.append(1)
+        return self.linear(x)
+
+    class Model(nnx.Module):
+      def __init__(self, rngs):
+        for d in range(10):
+          setattr(self, f"linear{d}", Block(rngs))
+
+      def __call__(self, x):
+        for d in range(10):
+          x = getattr(self, f"linear{d}")(x)
+        return x
+
+    m = Model(nnx.Rngs(0))
+    x = jnp.ones((4, 2))
+    nnx.tabulate(m, x, compute_flops=True, compute_vjp_flops=False)
+    self.assertEqual(len(counter), 10)
+
   def test_shared(self):
     class Block(nnx.Module):
       def __init__(self, linear: nnx.Linear, *, rngs):


samanklesaria:
When there's structure sharing, we wind up with duplicate rows in our table again. We need to implement some kind of check that we don't trace the same method twice with the same arguments.

flax/nnx/rnglib.py
```diff
@@ -58,6 +58,7 @@ def _to_keyless(
 
 
 def _function_to_method(random_f):
```
samanklesaria, line 59: Necessary so that tabulate can identify calls to the rng random methods in its tables.

flax/nnx/summary.py
```diff
@@ -130,14 +165,14 @@ def __str__(self):
 
 @dataclasses.dataclass
 class CallInfo:
+  call_order: int
   object_id: int
   type: type
   path: statelib.PathParts
-  input_args: tuple[tp.Any, ...]
-  input_kwargs: dict[str, tp.Any]
+  inputs_repr: str
```
samanklesaria, line 140: We can reuse the table representation as a hashable value to filter for when we're adding duplicate rows to the table.

tests/nnx/summary_test.py
```diff
@@ -41,9 +41,10 @@ def __call__(self, x):
 
     foo = Foo(nnx.Rngs(0))
     x = jnp.ones((1, 32))
-    table_repr = nnx.tabulate(
+    table_repr_ = nnx.tabulate(
```
samanklesaria, line 44: Added because whenever something goes wrong, we have to print the table to debug. And having the repr on a separate line makes it easier to print the table later.

vfdev-5:
@samanklesaria can you please post the output of nnx.tabulate on the user example in the PR description.

flax/nnx/summary.py
```diff
@@ -367,39 +417,21 @@ def tabulate(
...
-  for key, value in saver_env.items():
-    jits[key] = nnx.jit(value)
+  # Information is recorded in post-order, but should be presented as a pre-order traversal.
+  # This keeps track of the order of calls.
+  counter = itertools.count(0)
+
+  # Modify all the object's methods to save their lowered JIT representations.
+  rows : list[CallInfo] = []
+  jits = {k: _argsave(counter, rows, MaybeJit(v), node_stats, compute_flops, compute_vjp_flops)
+    for k,v in env.items()}
```
vfdev-5:
```suggestion
    for k, v in env.items()}
```

vfdev-5:
> @samanklesaria can you please post the output of nnx.tabulate on the user example in the PR description.

I meant how looks the output of the tabulate with this PR

flax/nnx/summary.py
```diff
@@ -151,14 +187,13 @@ def __repr__(self):
 
 
 def _to_dummy_array(x):
-  if isinstance(x,jax.ShapeDtypeStruct):
```
vfdev-5, line 154: Can you explain this change?

samanklesaria: Actually, that should be unnecessary in the current version of the PR. I'll remove it.

samanklesaria:
> ... the names of the variables and functions a bit unrelated to what they do:
> * `_argsave`
> * `jits`

I could call it `maybejits` instead of `jits` (as it's a list of `MaybeJit`). And as far as `_argsave` goes, I could call it `_save_call_info` instead?

cgarciae (approved):

flax/nnx/summary.py
```diff
@@ -206,13 +183,61 @@ def _create_obj_env(object_types):
...
+    primals, f_vjp = jax.vjp(f, *args, **kwargs)
+    return f_vjp(primals)
+
+  method_name = f.__name__
+
+  @functools.partial(jax.jit)
+  def jit_f(graphdef, state):
+    args, kwargs = nnx.merge(graphdef, state)
+    return f(*args, **kwargs)
+  jit_f._seen = set()
```
cgarciae: Q: is it necessary to attach `_seen` as an attribute? wouldn't it be the same to have a as a separate variable?

flax/nnx/summary.py
```diff
@@ -367,39 +392,21 @@ def tabulate(
...
-  # This would add items to `tracer_args`, resulting in duplicate rows in the table.
-  if compute_vjp_flops:
-    for i, row in enumerate(rows):
-      object, method_name, args, kwargs = tracer_args[i]
-      def do_vjp(*args, **kwargs):
-        primals, f_vjp = _pure_nnx_vjp(jits[(type(object), method_name)], *args, **kwargs)
-        return f_vjp(primals)
-      row.vjp_flops = _get_flops(jax.jit(do_vjp).lower(object, *args, **kwargs))
+  # Sort call info in pre-order traversal order
+  rows.sort(key=lambda x: x.call_order)
```
cgarciae, line 413: nice!

flax/nnx/summary.py
samanklesaria: Good point. It would be better to have it as a separate variable. Let me add that.

tests/nnx/summary_test.py
```diff
@@ -224,6 +225,32 @@ def __call__(self, x):
     # We should see 3 calls per block, plus one overall call
     self.assertEqual(sum([s.startswith("├─") for s in table.splitlines()]), 7)
 
+  def test_complexity(self):
+    counter = []
```
samanklesaria, line 229: I can't use an integer counter and write `counter += 1` in the call, because this violates python's scope rules for some reason. This is a hack to work around the issue.

samanklesaria:
Closing in favor of #5186

PR #5070, opened 2025-11-03
support explicit sharding in eager sharding

cgarciae:
# What does this PR do?

Adds support for explicit sharding in Variable's eager sharding. Instead of using `with_sharding_constraint`, the `shard_value` function now uses `jax.jit` over a simple identity function to shard the output using `out_shardings`.

```python
from flax import nnx
import jax
import jax.numpy as jnp

mesh = jax.make_mesh(
  (2, 2),
  ('row', 'col'),
  axis_types=(jax.sharding.AxisType.Auto, jax.sharding.AxisType.Explicit),
)
v = nnx.Variable(
  jnp.ones((4, 4)),
  sharding_names=('row', 'col'),
  mesh=mesh,
)
```

### DIFF
--- a/flax/core/spmd.py
+++ b/flax/core/spmd.py
@@ -32,6 +32,9 @@ def get_pspec(sharding_names, sharding_rules = None) -> PartitionSpec:
     return PartitionSpec(*from_sharding_rules(sharding_names, rules))
   return PartitionSpec(*sharding_names)
 
+def _apply_sharding(value, sharding):
+  with jax.disable_jit(False):
+    return jax.jit(lambda x: x, out_shardings=sharding)(value)
 
 def shard_value(value, sharding_names, sharding_rules, mesh):
   if not sharding_names:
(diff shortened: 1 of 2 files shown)


jburnim (approved):

flax/core/spmd.py
```diff
@@ -32,6 +32,8 @@ def get_pspec(sharding_names, sharding_rules = None) -> PartitionSpec:
     return PartitionSpec(*from_sharding_rules(sharding_names, rules))
   return PartitionSpec(*sharding_names)
 
+def _apply_sharding(value, sharding):
+  return jax.jit(lambda x: x, out_shardings=sharding)(value)
```
jburnim: Should we wrap this line in?:
```python
with jax.disable_jit(False):
  ...
```
so that `_apply_sharding` works even if inside a `jax.disable_jit(True)`?

cgarciae: great idea! Added `disable_jit(False)` in `_apply_sharding` and a test.

PR #5072, opened 2025-11-04
fix List and improve Pytree

cgarciae:


tests/nnx/partitioning_test.py
```diff
@@ -150,9 +150,6 @@ def test_get_paritition(self):
       d=5.0,
     )
 
-    # test Variables not shared
-    self.assertIsNot(vars(m.a)[0], vars(m)['b'])
```
IvyZX, line 154: Why remove this?

IvyZX (approved):

tests/nnx/partitioning_test.py
cgarciae: doesn't make sense here, probably copied from another test in the past

PR #5079, opened 2025-11-07
Add nnx.use_eager_sharding context manager

samanklesaria:
This PR lets users set whether or not eager sharding is applied locally rather than globally. It follows the same approach as `use_refs` or `use_hijacks` and other flax config contexts.

### DIFF
--- a/docs_nnx/guides/flax_gspmd.md
+++ b/docs_nnx/guides/flax_gspmd.md
@@ -29,6 +29,7 @@ import jax
 from jax import numpy as jnp
 from jax.sharding import PartitionSpec as P, NamedSharding, AxisType
 import optax
+import flax
 from flax import nnx
 
 # Ignore this if you are already running on a TPU or GPU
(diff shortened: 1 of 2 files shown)


docs_nnx/guides/flax_gspmd.md
```diff
@@ -46,11 +46,19 @@ In this guide we use a standard FSDP layout and shard our devices on two axes -
...
 import flax
-assert flax.config.flax_always_shard_variable is True
+nnx.eager_sharding(True)
+assert nnx.using_eager_sharding()
+```
+
+The `nnx.eager_sharding` function can also be used as a context manager to toggle the eager sharding feature within a specific scope.
+
+```{code-cell} ipython3
+with nnx.eager_sharding(False):
```
cgarciae: These should be `use_eager_sharding`.

cgarciae (approved):

PR #5086, opened 2025-11-13
Updated pyproject.toml

vfdev-5:
Replaced default torch cuda installation with cpu version only


vfdev-5:
Let's confirm that installing flax into env with pytorch + nvidia cuda support does not remove pytorch cuda support

vfdev-5:
> Let's confirm that installing flax into env with pytorch + nvidia cuda support does not remove pytorch cuda support

This is fine, existing pytorch installation is not altered by installing flax from source.

PR #5089, opened 2025-11-13
Rename sharding_names to sharding_metadata

samanklesaria:
This PR improves the variable sharding API. Specifically:

- It renames sharding_names to sharding_metadata (still supporting sharding_names as a property)
- It generalizes the spmd APIs to allow sharding_metadata to be a NamedSharding

### DIFF
--- a/flax/nnx/variablelib.py
+++ b/flax/nnx/variablelib.py
@@ -1094,7 +1094,11 @@ class Variable(tp.Generic[A], reprlib.Representable, metaclass=VariableMeta):
       metadata['on_remove_axis'] = var_t.on_remove_axis
 
     if 'sharding' in metadata:
-      metadata['sharding_names'] = metadata.pop('sharding')
+      metadata['sharding_metadata'] = metadata.pop('sharding')
+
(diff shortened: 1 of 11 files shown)


flax/nnx/variablelib.py
```diff
@@ -1133,6 +1138,9 @@ def _check_can_update(self):
       )
 
   def __getattr__(self, name: str) -> tp.Any:
+    if name == 'sharding_names': # for backward compatibility
+      warnings.warn("'sharding_names' is deprecated. Use 'sharding_metadata' instead.", DeprecationWarning)
+      return self.sharding_metadata
```
cgarciae: we should use `has_metadata` and `get_metadata` in the lines bellow and teach these methods how to handle `sharding_names`.

samanklesaria: I note that `get_metadata` makes a copy of the metadata whenever you call it. This isn't the case when we use the getattr method here normally. Does this matter?

IvyZX (approved):

PR #5094, opened 2025-11-17
Allow substring matching in `nnx.PathContains`

### LINKED ISSUE #5075: `nnx.PathContains` should support regex
The current implementation of `nnx.PathContains` only looks for exact matches in the path. This makes it difficult to select, e.g., a set of layers which have been given a similar name. To illustrate, my use case would be to use `nnx.split` to separate convolutional layers and linear layers, which I respectively named "conv1", "conv2", etc. and "fc1", "fc2", etc. To my knowledge, there are also no other filters that support this.

My proposal is a very simple modification method, namely:
```
@dataclasses.dataclass(frozen=True)
class PathContains:
  key: Key

  def __call__(self, path: PathParts, x: tp.Any):
    pattern = re.compile(self.key)
    return any(pattern.match(segment) for segment in path)

  def __repr__(self):
    return f'PathContains({self.key!r})'
```

Let me know if it is worth opening a pull request for this, and whether any efficiency constraints should be taken into account.

thijs-vanweezel:
# What does this PR do?

As per issue #5075, this PR slightly modifies nnx.PathContains to allow for substring matching. This is important in order to match all layers which have been given a similar name (e.g., "conv1", "conv2", etc.). As suggested, the modification implements an `exact` parameter, which, if set to `False`, ensures that `key` is matched against each element of `path`. Note; since a `Key` instance is not guaranteed to implement `.__contains__`, the `PathParts` are first converted to string.

P.S.: Although the tests were eventually passed, it appeared that JAX had to be built from source to satisfy recently added imports (an instruction perhaps worth adding to this [guide](https://flax.readthedocs.io/en/latest/contributing.html)), and that pre-existing typing errors in mypy required setting RUN_MYPY=false.

Fixes #5075 

## Example
```
import jax
from flax import nnx
from jax import numpy as jnp

# Modified LeNet-5 for 36X60 images
class LeNet(nnx.Module):
    def __init__(self, key):
        super().__init__()
        self.conv1 = nnx.Conv(1, 8, (4,4), rngs=key, padding="VALID")
        self.conv2 = nnx.Conv(8, 16, (4,4), rngs=key, padding="VALID")
        self.fc1 = nnx.Linear(6*12*16, 128, rngs=key)
        self.fc2 = nnx.Linear(128, 64, rngs=key)
        self.fc3 = nnx.Linear(64, 16, rngs=key)
    
model = LeNet(nnx.Rngs(0))

# Split model into fc and conv layers
struct, fc, conv, rest = nnx.split(
    model, 
    nnx.All(nnx.PathContains("fc", exact=False), nnx.PathContains("kernel")), 
    nnx.All(nnx.PathContains("conv", exact=False), nnx.PathContains("kernel")),
    ...
)
```

## Checklist

### DIFF
--- a/flax/nnx/filterlib.py
+++ b/flax/nnx/filterlib.py
@@ -89,7 +89,7 @@ class WithTag:
 
 @dataclasses.dataclass(frozen=True)
 class PathContains:
-  key: Key
+  key: Key | str
   exact: bool = True
 
   def __call__(self, path: PathParts, x: tp.Any):


thijs-vanweezel:
@vfdev-5 good suggestion. It has been added now.

vfdev-5:
Please rebase your PR on the latest `main` branch to keep only your updates

flax/nnx/filterlib.py
```diff
@@ -90,12 +90,15 @@ def __repr__(self):
 @dataclasses.dataclass(frozen=True)
 class PathContains:
   key: Key
+  exact: bool = True
 
   def __call__(self, path: PathParts, x: tp.Any):
-    return self.key in path
+    if self.exact:
+      return self.key in path
+    return any(str(self.key) in str(part) for part in path)
```
vfdev-5, line 98: Is it on purpose the cast of key and part to str? I mean, is it necessary to do that?


> Note; since a Key instance is not guaranteed to implement .__contains__, the PathParts are first converted to string.

I'm not sure whether this is necessary. whatever the class of key, we can always do `key in part`, IMO.

vfdev-5:
The failure of the CI is related to the broken `main` which is now fixed (please rebase)

flax/nnx/filterlib.py
thijs-vanweezel: Hi, the reason this is necessary is that `in` requires the operands to implement a `__contains__` method, which instances of `Key` are not guaranteed to implement. Also, this functionality seems to be intended primarily for use with string or integer keys.

Edit: I agree that any class that is passed as key supports `in`, but unless I’m mistaken, the tests did not pass.

vfdev-5: I do not think Key should implement `__contains__` method, it works as expected.
```python
class A: pass    

a = A()

path = [(1, "b"), (2, "v"), (3, a), (4, A())]

key = a

print([str(key) in str(part) for part in path])
# [False, False, True, False]
print([key in part for part in path])
# [False, False, True, False]
```
Can you provide an example?

thijs-vanweezel: Will get back in the (cet) morning!

thijs-vanweezel: It's a typing issue: Recall that `PathParts` is of type `tuple[Key,...]`. Therefore, the `in` operator essentially calls `Key() in Key()`, which raises an error because Key-types are not required to implement this (see flax/typing:47). Extending your example with the dummy class A:

```
class A: pass

key = A()
path: tuple[A,...] = (A(), A(), A())

print([str(key) in str(part) for part in path])
# [False, False, False]
print([key in part for part in path])
# TypeError: argument of type 'A' is not iterable
```

Even with more realistic keys and pathparts, such as integers, I see it as required to convert both the key and the pathparts to string:
```
key1 = 0
key2 = "conv"
path = ("backbone_list", 0, "conv1")

print([str(key1) in str(part) for part in path])
# [False, True, False]
print([str(key2) in str(part) for part in path])
# [False, False, True]
print([key1 in part for part in path])
# TypeError: 'in <string>' requires string as left operand, not int
print([key2 in part for part in path])
# TypeError: argument of type 'int' is not iterable
```

During the PyType tests, this results in the following failure:
```
flax/nnx/filterlib.py:98:15: error: in __call__: unsupported operand type(s) for 'in': 'part: flax.typing.Key' and 'flax.typing.Key' [unsupported-operands]
  No attribute '__contains__' on 'part: flax.typing.Key'

    return any(self.key in part for part in path)
```

vfdev-5: Sounds good, thanks! Maybe we could also detail that key can be `Key | str`, as for exact=False we do a partial match as strings?

thijs-vanweezel: Agreed; done.

vfdev-5 (approved):
LGTM, thanks @thijs-vanweezel !

PR #5097, opened 2025-11-18
Fixed deprecated .value usage failing CI tests

vfdev-5:


IvyZX:
Got errors from some internal user:
```
  File "flax/nnx/variablelib.py", line 1715, in __eq__
    other = other[...]
            ~~~~~^^^^^
  File "flax/nnx/variablelib.py", line 1664, in __getitem__
    return self.get_value(index=key)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "flax/nnx/variablelib.py", line 1419, in get_value
    value = value[index]
            ~~~~~^^^^^^^
KeyError: Ellipsis
```

tests/nnx/variable_test.py
```diff
@@ -95,6 +95,18 @@ def test_binary_ops(self):
 
     self.assertEqual(v1[...], 5)
 
+  @parameterized.product(
+    v1=[jnp.array([1, 2]), jnp.array(2), 3],
```
IvyZX: Google's internal tests does not support using JAX before the test's `__main__` starts. This means you cannot use JAX in `parameterized.product`. You can do `np.array` instead and convert it within the test.

PR #5117, opened 2025-11-28
Add Optimization Cookbook

samanklesaria:
# What does this PR do?

This PR adds a guide that shows some common techniques for working with Flax models during optimization. These include:

- Calculation of Exponential Moving Averages
- Optimizing only a low rank addition to certain weights (LORA)
- Using different learning rates for different parameters to implement the maximal update parameterization
- Using second order optimizers like LBFGS. 
- Specifying sharding for optimization state that differs from that of parameter state
- Gradient accumulation

This document emphasizes a style as close to pure jax as possible: to that end, it shows how the flax version of each technique only requires minor deviation from the often more intuitive pure-jax version. 

# Warnings:

- This PR requires the fix #5171 , which is why the diff looks so big here. This PR shouldn't be merged until after #5171 
- This PR includes both the notebook file used to construct the optimization examples as well as the RST file that contains what gets used in the documentation. When review is complete (and I don't need to actually test any of the python snippets anymore) I will delete the notebook file.

### DIFF
--- a/docs_nnx/guides/optimization_cookbook.md
+++ b/docs_nnx/guides/optimization_cookbook.md
@@ -1,0 +1,278 @@
+---
+jupyter:
+  jupytext:
+    formats: ipynb,md
+    main_language: python
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
+      jupytext_version: 1.13.8
+---
+
+# A Flax Optimization Cookbook
+
+
+This notebook goes through some common problems in nontrivial training loops for Flax models. For clarity, all sections below will be training the following toy model. We allow extra keyword arguments so that the sharding and dtype can be determined on an instance by instance basis. 
+
+```python
+import jax
+from flax import nnx
+jax.config.update('jax_num_cpu_devices', 8)
+import jax.numpy as jnp
+import functools as ft
+import optax
+import matplotlib.pyplot as plt
+
+lecun_normal = jax.nn.initializers.lecun_normal()
+
+rngs = nnx.Rngs(0)
+
+def make_model(rngs, init=lecun_normal):
+    return nnx.Sequential(
+        nnx.Linear(2,8, rngs=rngs, kernel_init=init),
+        nnx.Linear(8,8, rngs=rngs, kernel_init=init))
+
+def loss_fn(model, x, y):
+    return jnp.mean((model(x) - y) ** 2)
+```
+
+We'll operate on the following fake data:
+
+```python
+x = rngs.normal((32, 2))
+y = rngs.normal((32, 8))
+```
+
+# Exponential Moving Average
+
+Neural network see increased robustness when, rather than using only the weights available at the end of training, we use an exponential moving average of the weights produced throughout training. To modify the standard Flax training loop to accomodate calculating exponential moving averages, we can introduce a new `nnx.Variable` subclass for EMA Params, along with a function that converts all variables in a module to this subclass. 
+
+```python
+class EmaParam(nnx.Variable):
+  @classmethod
+  def from_variable(cls, var):
+    return cls.from_metadata(jnp.copy(var.get_value()), var.get_metadata())
+    
+def as_ema_params(node):
+  return jax.tree.map(
+    EmaParam.from_variable,
+    node,
+    is_leaf=lambda x: isinstance(x, nnx.Variable),
+  )
+```
+
+Now, we'll add a method to update the EMA params based on current model values. 
+
+```python
+class EMA(nnx.Pytree):
+  def __init__(self, params, decay: float, *, only: nnx.filterlib.Filter = ...):
+    self.decay = decay
+    self.filter = only
+    self.ema_params = nnx.data(as_ema_params(nnx.state(params, only)))
+
+  def update(self, new_params):
+    def _update(ema_param, new_param):
+      ema_param[...] = (
+        self.decay * ema_param[...] + (1.0 - self.decay) * new_param[...]
+      )
+    jax.tree.map(
+      _update,
+      self.ema_params,
+      nnx.state(new_params, self.filter),
+      is_leaf=lambda x: isinstance(x, nnx.Variable),
+    )
+```
+
+
+The training loop proceeds as normal, but calls `ema.update` after each optimizer step. 
+
+```python
+# initialization
+model = make_model(rngs)
+ema = EMA(model, decay=0.9)
+
+# simulate parameter update
+def double(param):
+  param[...] *= 2.0
+jax.tree.map(double, model, is_leaf=lambda x: isinstance(x, nnx.Variable))
+ema.update(model)
+
+@nnx.jit
+def train_step(model, optimizer, ema, x, y):
+  loss, grads = nnx.value_and_grad(loss_fn)(model, x, y)
+  optimizer.update(model, grads)
+  ema.update(model)
+  return loss
+
+optimizer = nnx.Optimizer(
+  model,
+  tx=optax.adam(1e-3),
+  wrt=nnx.Param,
+)
+losses = []
+for _ in range(50):
+  loss = train_step(model, optimizer, ema, x, y)
+  losses.append(loss)
+plt.plot(losses);
+```
+
+# Low Rank Adaptation
+
+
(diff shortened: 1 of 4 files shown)


cgarciae:
I feel we could simplify the intro by doing the following:

1. Define a single model at the begining, simply reuse it on all examples (its just a guide).
2. Inline the loss function.

```python
model = nnx.Sequential(
  nnx.Linear(2,8, rngs=rngs),
  nnx.relu,
  nnx.Linear(8,8, rngs=rngs),
)

optimizer = nnx.Optimizer(
  model,
  tx=optax.adam(1e-3),
  wrt=nnx.Param)

...

@nnx.jit
def train_step(model, optimizer, ema, x, y):
  loss_fn = lambda m, x, y: jnp.sum((m(x) - y) ** 2)
  loss, grads = nnx.value_and_grad(loss_fn)(model, x, y)
  optimizer.update(model, grads)
  ema.update(model)
  return loss
```

docs_nnx/guides/opt_cookbook.rst
```diff
@@ -0,0 +1,447 @@
...
+      wrt=nnx.Param)
+
+    def ema_update(ema, new_val, decay=0.9):
+        return decay * ema + (1 - decay) * new_val
+
+    class Ema(nnx.Module):
+        def __init__(self, model):
+            self.ema = nnx.clone(model) # Make a copy
+        def update(self, model):
+            self.ema = tree.map(ema_update, self.ema, model)
```
cgarciae:
```suggestion
    class Ema(nnx.Module):
        def __init__(self, model, decay=0.9):
            self.decay = decay
            self.ema = jax.tree.map(jnp.copy, model) # Make a copy
        def update(self, model):
            def ema_update(ema, new_val):
              return decay * ema + (1 - self.decay) * new_val
            self.ema = tree.map(ema_update, self.ema, model)
```

docs_nnx/guides/opt_cookbook.rst
```diff
@@ -0,0 +1,447 @@
...
+    :title: NNX, Jax
+    :sync:
+
+    def add_rank2_lora(path, node):
+      if isinstance(node, nnx.Linear):
+        return nnx.LoRA(node.in_features, 2, node.out_features, base_module=node, rngs=rngs)
+      return node
+
+    base_model = nnx_model(rngs)
+    model = nnx.recursive_map(add_rank2_lora, base_model)
```
cgarciae: would be nice to `nnx.display(model)` or print here. I guess RST is a bit of a limitation.

docs_nnx/guides/opt_cookbook.rst
```diff
@@ -0,0 +1,447 @@
...
+          value=loss,
+          value_fn=loss_fn_state)
+        return loss
+
+    model = nnx_model(rngs)
+
+    nnx_optimizer = nnx.Optimizer(
+      model,
+      tx=optax.lbfgs(1e-3),
+      wrt=nnx.Param)
```
cgarciae:
```suggestion
      wrt=nnx.LoRAParam)
```

cgarciae: We also need to explain that we are calculating the gradients for all the parameters in `grad` but are filtering them here and it might be ok because of DCE.

docs_nnx/guides/opt_cookbook.rst
```diff
@@ -0,0 +1,447 @@
...
+In Flax, we will also initialize a partitioned optax optimizer. But unlike the Jax case where we use the model as the PyTree, we'll need to use the model's *state* PyTree here instead. These state PyTrees aren't compatible with ordinary jax tree mapping; instead of ``jax.tree.map_with_path``, we'll need to use the NNX version, which is called ``nnx.map_state``.
+
+.. codediff::
+    :title: NNX, Jax
+    :sync:
+
+    model = nnx_model(rngs)
+    state = nnx.state(model, nnx.Param)
+    rates = {'kernel': optax.adam(1e-3), 'bias': optax.adam(1e-2)}
+    param_tys = nnx.map_state(lambda p, v: list(p)[-1], state)
```
cgarciae: I thin this is enough:
```suggestion
    param_tys = nnx.map_state(lambda p, v: p[-1], state)
```

cgarciae: We could also use `jax.tree. map_with_path` as in the JAX example.

docs_nnx/guides/opt_cookbook.rst
```diff
@@ -0,0 +1,447 @@
...
+
+.. codediff::
+    :title: NNX, Jax
+    :sync:
+
+    mesh = jax.make_mesh((2, 4), ("x", "y"),
+                         axis_types=(AxisType.Explicit, AxisType.Explicit))
+    jax.set_mesh(mesh)
+
+    ghost_model = jax.eval_shape(lambda: nnx_model(nnx.Rngs(0), out_sharding=P('x', 'y')))
```
cgarciae: instead of creating this fake model it would be a good opportunity to create the `optimizer_sharding` API on Variable before finishing this guide.

flax/nnx/helpers.py
```diff
@@ -185,6 +186,17 @@ def __setitem__(self, index: int | slice, value: A | tp.Iterable[A]) -> None:
     else:
       raise TypeError('Invalid index type')
 
+  def _graph_node_set_key(self, key: str, value: tp.Any):
```
cgarciae, line 198: Is this still needed?

cgarciae:
After fully reading the guide I'm getting the sense that having the JAX versions makes explanations a bit longer and slightly harder to understand (cause you have to mentally filter for the version you are interested in) and having the JAX version doesn't necessarily make understanding the NNX version easier.

samanklesaria:
> After fully reading the guide I'm getting the sense that having the JAX versions makes explanations a bit longer and slightly harder to understand (cause you have to mentally filter for the version you are interested in) and having the JAX version doesn't necessarily make understanding the NNX version easier.

Fair enough! I'll convert it to nnx-only.

flax/nnx/helpers.py
samanklesaria: I don't think so. Removing it.

docs_nnx/guides/opt_cookbook.rst
samanklesaria: Done.

docs_nnx/guides/optimization_cookbook.md
```diff
@@ -0,0 +1,242 @@
...
+
+rngs = nnx.Rngs(0)
+
+def make_model(rngs, **metadata):
+    return nnx.Sequential(
+        nnx.Linear(2,8, rngs=rngs, kernel_metadata=metadata),
+        nnx.Linear(8,8, rngs=rngs, kernel_metadata=metadata))
+
+def loss_fn(model, x, y):
+    return jnp.sum((model(x) - y) ** 2)
```
cgarciae: should we use `mean` here?

docs_nnx/guides/optimization_cookbook.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+# Exponential Moving Average
+
+Neural network see increased robustness when, rather than using only the weights available at the end of training, we use an exponential moving average of the weights produced throughout training. To modify the standard Flax training loop to accomodate calculating exponential moving averages, we can introduce a new `nnx.Variable` subclass for EMA Params, along with a function that converts all variables in a module to this subclass. 
+
+```python
+class EmaParam(nnx.Variable):
+  @classmethod
+  def from_variable(cls, var):
```
cgarciae, line 54: Seems like this classmethod could just be a function in `as_ema_params`?

docs_nnx/guides/optimization_cookbook.md
```diff
@@ -0,0 +1,278 @@
...
+
+```python
+# initialization
+model = make_model(rngs)
+ema = EMA(model, decay=0.9)
+
+# simulate parameter update
+def double(param):
+  param[...] *= 2.0
+jax.tree.map(double, model, is_leaf=lambda x: isinstance(x, nnx.Variable))
```
cgarciae, line 98: we could use the new `nnx.map` here.

PR #5141, opened 2025-12-17
Handle pure bodies in nnx.fori_loop

### LINKED ISSUE #5113: Examples how to use fori_loop for gradient accumulation & more clear exceptions
Hey folks,

It would be nice to have some examples in Flax NNX documentation on how to do gradient accumulation with nnx.fori_loop.

I've tried lots of different combination of "how to pass arguments, what I can add to closure and what I can't" myself and haven't succeed. 

My train step looks something like this:

```python

# TrainSession is NamedTuple with 'model: nnx.Module' and 'optimizer: nnx.Optimizer' fields

def get_training_step(
    forward_pass_cfg: ForwardPassConfigBase,
    batch_sharding: PyTree,
    num_minibatches: int = 1,
) -> Callable[
    [PyTree, TrainSession],
    tuple[DeviceArray, PyTree],
]:
    forward_pass_fn = get_forward_pass(forward_pass_cfg)

    @nnx.jit(donate_argnums=(1,))
    def training_step(
        batch: PyTree,
        train_session: TrainSession,
    ) -> tuple[DeviceArray, PyTree]:

        def _loop_body(
            minibatch_idx,
            carry: tuple[nnx.State, DeviceArray, TrainSession],
        ) -> tuple[nnx.State, DeviceArray, TrainSession]:
            g_accum, loss_accum, train_session = carry
            minibatch = get_sharded_minibatch(
                batch=batch,
                batch_sharding=batch_sharding,
                minibatch_idx=minibatch_idx,
                num_minibatches=num_minibatches,
            )

            def _loss_fn(model) -> tuple[DeviceArray, PyTree]:
                return forward_pass_fn(
                    model=model,
                    batch=minibatch,
                    step=train_session.optimizer.step.value,
                    config=forward_pass_cfg,
                )

            (loss, outputs), grads = nnx.value_and_grad(_loss_fn, has_aux=True)(
                train_session.model
            )
            g_accum = jax.tree.map(
                lambda gm, g: gm + g, g_accum, grads
            )
            loss_accum = loss_accum + loss
            return g_accum, loss_accum, train_session

        g_accum = jax.tree.map(jnp.zeros_like, nnx.state(train_session.model))
        g_accum, loss_accum, train_session = nnx.fori_loop(
            lower=0, 
            upper=num_minibatches,
            body_fun=_loop_body,
            init_val=(g_accum, 0.0, train_session)
        )

        g_accum = jax.tree.map(lambda g: g / num_minibatches, g_accum)
        loss_accum = loss_accum / num_minibatches

        train_session.optimizer.update(train_session.model, g_accum)
        return loss_accum, {}

    return training_step
```

And I'm getting the following error:

```
ValueError: nnx.fori_loop requires body function's input and output to have the same reference and pytree structure, but they differ. If the mismatch comes from `outer_index` field, you might have modified reference structure within the body function, which is not allowed.
```

But I'm not sure I what does it mean and how can I fix it.

It would be nice if flax provided examples on documentation on how to approach it properly, what can be done, what can't be done, how to work with closures, etc.

samanklesaria:
Fixes #5113

Currently, `nnx.fori_loop` only works with loop bodies that perform some kind of mutation. This is because the the call to `_add_fake_index_mapping(pure_init_val)` in line 1625 of `flax/nnx/transforms/iteration.py` adds an `outer_index` attribute that won't be added in the corresponding pure body. To fix this, this PR checks what the output of the loop function looks like first. If it has an outer index, we add a fake index mapping. Otherwise, we leave it be.

### DIFF
--- a/flax/nnx/transforms/iteration.py
+++ b/flax/nnx/transforms/iteration.py
@@ -1617,7 +1617,7 @@ def fori_loop(lower: int, upper: int,
 
   pure_init_val = extract.to_tree(init_val, ctxtag='fori_loop')
   body = ForiLoopBodyFn(body_fun)
-  pure_out = body(lower, pure_init_val)
+  pure_out = jax.eval_shape(body, lower, pure_init_val)
   pure_init_val = _reconsile_index_mapping(pure_init_val, pure_out)
   pure_out = jax.lax.fori_loop(lower, upper,
                                body, pure_init_val,


flax/nnx/transforms/iteration.py
```diff
@@ -1548,27 +1562,10 @@ def __post_init__(self):
...
   @graph.update_context('fori_loop_body')
-  def __call__(self, i, pure_val):
-    # Removing the dummy index mapping being added outside of body function.
-    pure_val_in = _remove_index_mapping(pure_val)
-
+  def __call__(self, i, pure_val_in):
     val = extract.from_tree(pure_val_in, ctxtag='fori_loop_body', is_inner=True)
     out = self.f(i, val)
     pure_out = extract.to_tree(out, ctxtag='fori_loop_body')
-
```
samanklesaria, line 1558: It's unfortunate, but we can't have this error here anymore. The first time we call the body function, we know that the body function's output won't match. We manually fix the structure in `nnx.fori_loop`, using the output as a guide. In practice, I don't think this matters. The error you get without this handler is still clear- it just refers to mismatched structures in `jax.lax.scan` rather than `nnx.fori_loop`.

vfdev-5:
What would happen if the function changes the graphdef by adding a new attribute like when calling `nnx.Module.sow` ?
A similar issue happend with `nnx.scan` on modified graphdef.

samanklesaria:
> What would happen if the function changes the graphdef by adding a new attribute like when calling `nnx.Module.sow` ? A similar issue happend with `nnx.scan` on modified graphdef.

If the function changes the graphdef, you'll get an error about the structures not matching. We also get an error even if the changes are only temporary (using `nnx.sow` with `nnx.pop` afterwards), which matches what you were talking about with `nnx.scan`:

```python
class MLP(nnx.Module):
  def __call__(self):
    self.sow(nnx.Intermediate, "blah", jnp.ones(2))

def sow_fn(i, model):
  model()
  nnx.pop(model, nnx.Intermediate)
  return model

def test_sow():
  model = MLP()
  nnx.fori_loop(0, 2, sow_fn, model)
```
In this case, "blah" appears as a static attribute in the pytree result of `sow_fn` even though `nnx.pop` should have removed it, and we get another mismatched structure error. But this is a bug in `nnx.pop` rather than `nnx.fori_loop`. If we run

```
model = MLP()
sow_fn(2, model)
print(nnx.split(model))
```

we get a NodeDef with `attributes=[('_pytree__nodes', Static(value={'_pytree__state': True, '_pytree__nodes': False, 'blah': True}))`. As this bug has nothing to do with `nnx.fori_loop`, I propose that we address it in a separate issue and PR. The separate issue is #5146 which is fixed by #5147

flax/nnx/transforms/iteration.py
```diff
@@ -1619,13 +1616,11 @@ def fori_loop(lower: int, upper: int,
...
 
   pure_init_val = extract.to_tree(init_val, ctxtag='fori_loop')
-
-  # Adding the expected reference mapping to `pure_init_val` to match
-  # `body_fun`'s output pytree structure, to make JAX happy.
-  pure_init_val = _add_fake_index_mapping(pure_init_val)
-
+  body = ForiLoopBodyFn(body_fun)
+  pure_out = body(lower, pure_init_val)
+  pure_init_val = _reconsile_index_mapping(pure_init_val, pure_out)
```
IvyZX, line 1621: My only concern is that it runs the function two times. Perhaps jit will optimize this out, since values inside the first `pure_out` was never used? 

Can we have a small benchmark on it and put it to `flax/benchmarks`?

IvyZX (changes requested):
Thanks for fixing this!

flax/nnx/transforms/iteration.py
samanklesaria: I just changed it to use `jax.eval_shape` instead of calling `body` directly. This should prevent duplicate runs after tracing.

PR #5133, opened 2025-12-09
Make `nnx.pop` remove sown attributes.

samanklesaria:
Fixes #5130
This does not remove the Fingerprint check (also mentioned in the issue). This just fixes the pop bug.

PR #5147, opened 2025-12-18
Have nnx.pop remove attributes from pytree__nodes

### LINKED ISSUE #5146: Nnx.pop leaves module with new static attribute
Consider the following code:

```python
class MLP(nnx.Module):
  def __call__(self):
    self.sow(nnx.Intermediate, "blah", jnp.ones(2))

def sow_fn(model):
  model()
  nnx.pop(model, nnx.Intermediate)
  return model

model = MLP()
sow_fn(model)
print(nnx.split(model))
```

We get a NodeDef with `attributes=[('_pytree__nodes', Static(value={'_pytree__state': True, '_pytree__nodes': False, 'blah': True}))`. But `nnx.pop` should result in a pytree structure without references to the intermediates. Ideally, we'd like to be able to use pop and sow with loops like `nnx.fori_loop`: unless the structure is preserved exactly between iterations, these combinators aren't going to work.

samanklesaria:
Closes #5146

### DIFF
--- a/flax/nnx/pytreelib.py
+++ b/flax/nnx/pytreelib.py
@@ -936,6 +936,10 @@ class Pytree(reprlib.Representable, metaclass=PytreeMeta):
   def _graph_node_pop_key(self, key: str):
     if not isinstance(key, str):
       raise KeyError(f'Invalid key: {key!r}')
+    if hasattr(self, '_pytree__nodes'):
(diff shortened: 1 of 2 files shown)


samanklesaria:
Whoops- I forgot that I already have a PR for this: #5133
Closing.

Issue #5161, opened 2026-01-03
[docs]:  Improve docs for NNX API Reference


ibbyml:
Hi,

I’d like to improve the NNX API reference docs (not Linen). While working on a smaller docs issue ([#5107](https://github.com/google/flax/issues/5107)), I noticed a few recurring problems across the NNX API reference: incomplete indexing/toctrees, inconsistent Sphinx cross-references vs code formatting, and missing/unclear docstrings for user-facing symbols. Some pages also look copied from older docs and don’t match the current NNX surface.

---

Example:

- `flax.nnx.unflatten` in the graph section has a kwarg `outer_index_outer_ref=None`, but its docstring refers to `index_ref_cache` (which doesn’t exist).

<img width="818" height="484" alt="Image" src="https://github.com/user-attachments/assets/117ddeb6-8aff-4ba4-ac13-4ab06038af38" />

---

To keep review manageable, I’ll do this as a series of small PRs, going section by section in the [API Reference page](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/index.html). To keep the scope manageable I'll focus on docs + docstrings only (no behavioral changes).

Is it ok if I send small PRs starting with `flax.nnx.graph`? 

Thanks!

vfdev-5:
Hi @ibbyml, thanks for pointing that out. Yes, keeping the docs updated during the development can be a bit complex. If you can pointing out all these inconsistencies and send small fixes, it would be awesome.

The team is also working on how to improve the merging process on the approved PRs, but meanwhile having the fixes on GitHub would be very helpful!

ayushkumrdev:
Hi @ibbyml and @vfdev-5, this sounds like a great initiative! To help parallelize the effort, I can take ownership of auditing the flax.nnx.nn module (Linear, Conv, Attention, etc.).

I'll check for the same inconsistencies (missing args, broken cross-refs, and outdated Linen-style docstrings) and submit them as separate, small PRs. Does that sound good?

santihdzs:
I'd like to take the filterlib section. The filter classes (WithTag, PathContains, OfType, Any, All, Not, Everything, Nothing) are exported but currently render with no docstrings in the API reference, I'll add short docstrings for each based on the descriptions in the Filters guide.

greekera1000:
Hi @vfdev-5 , I'd like to take the `flax.nnx.training` section (Metrics, Optimizer, EMA) as part of this docs audit. A few concrete inconsistencies I've already spotted there:

[...]

PR #5175, opened 2026-01-07
[Docs]: Updated NNX graph API Reference

ibbyml:
# Updated NNX graph API Reference

This PR expands and restructures the NNX graph API documentation to make the graph system easier to understand and navigate.

Changes:
- Reorganized the graph API reference into clearly scoped sections (graph state APIs, state utilities, traversal, context management, variable utilities, etc.).
- Added detailed docstrings and usage descriptions for many previously undocumented or minimally documented graph functions, including:
	- flatten / unflatten
	- state, update, pop, clone, pure
	- graph traversal utilities (iter_graph, recursive_map, find_duplicates)
	- functional helpers (call)
	- context management utilities (split_context, merge_context, update_context)
- Clarified the roles of GraphDef, GraphState, GraphFlatState, and PureState and how they relate to split / merge and functional graph usage.
- Introduced small documentation‑only typing wrappers (via typing.Annotated) for graph state aliases so they can be documented as named concepts.

Fixes: (#5161)

## Checklist
- [x] This change is discussed in a Github issue/[discussion](https://github.com/google/flax/discussions) (https://github.com/google/flax/issues/5161).

### DIFF
--- a/flax/nnx/graph.py
+++ b/flax/nnx/graph.py
@@ -1401,7 +1401,7 @@ def _cached_partial(f: tp.Callable[..., tp.Any], *cached_args):
     >>> loss = cached_step(jnp.ones((10, 2)), jnp.ones((10, 3)))
 
   Note:
-    Cached graph nodes are cloned but share the same Variable objects, ensuring
+    Cached graph nodes are cloned but share the same ``Variable`` objects, ensuring
     state propagates correctly. The graph structure must remain unchanged after
     each call (temporary mutations like ``sow`` are allowed if cleaned up via
     ``pop``).
@@ -2001,7 +2001,7 @@ def update_context(tag: tp.Hashable):
     tag: A hashable identifier for this context (used to match split/merge calls).
 
   Returns:
-    A context manager (and decorator) for tracking graph state updates.
+    An ``UpdateContext`` object which can be used as a context manager (and decorator) for tracking graph state updates.
   """
 
   return UpdateContextManager(tag=tag)
@@ -2621,7 +2621,7 @@ def call(
 def set_metadata(
   node: tp.Any, /, *, only: filterlib.Filter = Variable, **metadata: tp.Any
 ) -> None:
-  """Set metadata on Variables in a graph node in-place.
+  """Set metadata on :class:`Variable` objects in a graph node in-place.
 
   Example usage::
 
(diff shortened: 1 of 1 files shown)


ibbyml:
Hi. Happy to adjust or revert any of the structural or example changes if there’s a preference. I plan to work on the NNX's [object](https://flax.readthedocs.io/en/latest/api_reference/flax.nnx/object.html) API reference if that's OK.

Thanks!

vfdev-5:
Thanks for working on the docs improvements @ibbyml !
This PR has large changes, I'll try to review it, but there is however one thing which may block it from landing. Rewriting docs into another format can be very opinionated. So, it would be preferable to maybe split the PR into clear small pieces (adding new docstrings or improving existing ones) such that everyone is OK with and leave opinionated pieces to another PR where we can find a middle-ground. What do you think?

For example, the point:

> Reorganized the graph API reference into clearly scoped sections (graph state APIs, state utilities, traversal, context management, variable utilities, etc.).

Personally, I'm not finding it more clear than the previous docs. For example, here we see `graph` header and then `State` header, it is unclear whether it is a submodule or something else. If you want to emphasize a common property we may need rewrite that in a more clear way.

image

What can done is that we may inspire from Jax ref API: https://docs.jax.dev/en/latest/jax.numpy.html

ibbyml:
Thanks for the review @vfdev-5. Definitely makes sense. I'll split these up then we can individually work on how the docs should be formatted. Afterwards, I'll apply the same formatting choices to the other parts of API Reference docs. I also agree that maybe we should take inspiration from the Jax docs.

Closing this PR and I'll open smaller ones without the opinionated changes.

vfdev-5:
We also talked yesterday in the team about exposing certain graph classes
and finally the idea is the opposite. GraphState, GraphFlatState and
PureState are remnants from earlier prototyping, we should not make them
documented but deprecate/remove somehow. Classes like NodeDef are also
private even if they have a docstring. It may be unclear now what is public
and what is private in the graph module but our high level goal is to
improve that. The rule of thumb: nnx/__init__.py public API exposure. We
can fix / improve docstring of these classes. Let's not add anything new in
this submodule and add docstrings for unexposed classes/aliases...

Sorry, this may appear that we go back and forth with this, but as you have
correctly pointed out previously we should improve our docs for the users.
It remains to clarify from our side public/private exposure.

PR #5180, opened 2026-01-07
feat(nnx): add Grouped Query Attention (GQA) support

### LINKED ISSUE #5177: Support QGA in nnx dot_product_attention
Jax's [dot product attention](https://docs.jax.dev/en/latest/_autosummary/jax.nn.dot_product_attention.html#jax.nn.dot_product_attention) supports grouped query attention by allowing the number of key/value heads to be different from the number of query heads. Flax's dot production attention (which differs from Jax's because it allows sowing attention weights into a Module) doesn't currently support this. We should add it. 

Example:
```python
import numpy as np
import jax.numpy as jnp
from flax import nnx

B = 2
S = 5
T = 4
N = 6
K = 3
H = 7

query = jnp.array(np.random.randn(B,T,N,H))
key = jnp.array(np.random.randn(B,S,K,H))
value = key

jax.nn.dot_product_attention(query, key, value) # works just fine

nnx.dot_product_attention(query, key, value) # fails
```

ayulockedin:
# What does this PR do?

This PR adds support for Grouped Query Attention (GQA) to nnx.dot_product_attention.

Previously, nnx.dot_product_attention required the number of heads in Query, Key, and Value to be identical. This caused a shape mismatch error when trying to use GQA configurations (where multiple Query heads share a single Key/Value head).

Changes Implemented:

- Added broadcasting logic in dot_product_attention_weights to repeat Key heads to match Query heads.
- Added broadcasting logic in dot_product_attention to repeat Value heads to match the Attention Weights.
- Added a validation check to ensure query_heads is divisible by key_heads (raising a clear ValueError if not).
- Added a new test file tests/nnx/nn/gqa_test.py covering valid GQA shapes and invalid configuration handling.

This change brings nnx into parity with jax.nn.dot_product_attention, enabling modern architectures (like Llama 3) to be implemented in NNX.

Fixes #5177

## Checklist

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -109,16 +109,43 @@ def dot_product_attention_weights(
 
   assert query.ndim == key.ndim, 'q, k must have same rank.'
   assert query.shape[:-3] == key.shape[:-3], 'q, k batch dims must match.'
-  assert query.shape[-2] == key.shape[-2], 'q, k num_heads must match.'
   assert query.shape[-1] == key.shape[-1], 'q, k depths must match.'
+
+  # check if we need to broadcast Key heads to match Query heads
+  is_gqa = False
+  if query.shape[-2] != key.shape[-2]:
+    q_heads = query.shape[-2]
+    k_heads = key.shape[-2]
+
+    if q_heads % k_heads != 0:
+      raise ValueError(
+        f"Query heads ({q_heads}) must be multiple of "
+        f"Key heads ({k_heads}) for Grouped Query Attention."
+      )
+
+    n_rep = q_heads // k_heads
+    is_gqa = True
+    # Reshape Query: [..., Q, H_k * n_rep, D] -> [..., Q, H_k, n_rep, D]
+    query = query.reshape(query.shape[:-2] + (k_heads, n_rep, query.shape[-1]))
+    # Expand Key: [..., K, H_k, D] -> [..., K, H_k, 1, D]
+    key = jnp.expand_dims(key, axis=-2)
+
+    # Contract: q(h)gd, k(h)1d -> hgqk (h=H_k, g=n_rep)
+    einsum_str = '...qhgd,...kh1d->...hgqk'
+  else:
+    q_heads = query.shape[-2]
+    einsum_str = '...qhd,...khd->...hqk'
+    assert query.shape[-2] == key.shape[-2], 'q, k num_heads must match.'
 
   # calculate attention matrix
   depth = query.shape[-1]
   query = query / jnp.sqrt(depth).astype(dtype)
+
   # attn weight shape is (batch..., num_heads, q_length, kv_length)
-  attn_weights = jnp.einsum(
-    '...qhd,...khd->...hqk', query, key, precision=precision
-  )
+  attn_weights = jnp.einsum(einsum_str, query, key, precision=precision)
+
+  if is_gqa:
+      attn_weights = attn_weights.reshape(attn_weights.shape[:-4] + (q_heads, attn_weights.shape[-2], attn_weights.shape[-1]))
 
   # apply attention bias: masking, dropout, proximity bias, etc.
   if bias is not None:
@@ -145,9 +172,12 @@ def dot_product_attention_weights(
   # apply attention dropout
   if not deterministic and dropout_rate > 0.0:
     keep_prob = 1.0 - dropout_rate
+    # use original key.ndim because we might have expanded key dim
+    ndim_base = key.ndim - 1 if is_gqa else key.ndim
+
     if broadcast_dropout:
       # dropout is broadcast across the batch + head dimensions
-      dropout_shape = tuple([1] * (key.ndim - 2)) + attn_weights.shape[-2:]
+      dropout_shape = tuple([1] * (ndim_base - 2)) + attn_weights.shape[-2:]
       keep = random.bernoulli(dropout_rng, keep_prob, dropout_shape)  # type: ignore
     else:
       keep = random.bernoulli(dropout_rng, keep_prob, attn_weights.shape)  # type: ignore
@@ -224,17 +254,15 @@ def dot_product_attention(
   """
   query, key, value = promote_dtype((query, key, value), dtype=dtype)  # type: ignore[bad-unpacking]
   dtype = query.dtype
+
   assert key.ndim == query.ndim == value.ndim, 'q, k, v must have same rank.'
   assert (
     query.shape[:-3] == key.shape[:-3] == value.shape[:-3]
   ), 'q, k, v batch dims must match.'
-  assert (
-    query.shape[-2] == key.shape[-2] == value.shape[-2]
-  ), 'q, k, v num_heads must match.'
   assert key.shape[-3] == value.shape[-3], 'k, v lengths must match.'
 
   # Criteria that invoke the more optimized dot product attention
-  if dropout_rate == 0.0 and module == None:
+  if dropout_rate == 0.0 and module is None:
     # make sure qkv batch are compressed to one dim
     query_shape = query.shape
     if len(query_shape) > 4:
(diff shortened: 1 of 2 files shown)


ayushkumrdev:
Hi @cgarciae . Could you please take a look at this PR when you have a moment?  Thanks!

samanklesaria:
Hi @ayulockedin - thanks for the PR! Would you be able to add some tests explicitly comparing the results of ` nnx.dot_product_attention` to ` jax.nn.dot_product_attention` to make sure they produce the same results for some random inputs? You'll need to give the nnx one a `module` argument so it doesn't just call the jax one.

ayushkumrdev:
@samanklesaria  on it thx

ayushkumrdev:
@samanklesaria Thanks for the review! I've added test_gqa_parity_with_jax to tests/nnx/nn/gqa_test.py.

It forces the internal NNX python implementation (by passing a dummy module) and compares the output against jax.nn.dot_product_attention (where I manually broadcast the GQA inputs) to ensure numerical equivalence. All tests are passing.

flax/nnx/nn/attention.py
```diff
@@ -107,6 +107,20 @@ def dot_product_attention_weights(
   query, key = promote_dtype((query, key), dtype=dtype)  # type: ignore[bad-unpacking]
   dtype = query.dtype
 
+  # check if we need to broadcast Key heads to match Query heads
+  if query.ndim == key.ndim and query.shape[-2] != key.shape[-2]:
```
samanklesaria: Maybe move this after the assertion that q and k must have the same rank, so we don't need the conditional?

ayushkumrdev:
@samanklesaria Good catch. I've refactored this so the rank assertions happen before broadcasting, which allowed me to simplify the logic. Looks good now??

flax/nnx/nn/attention.py
```diff
@@ -108,10 +108,24 @@ def dot_product_attention_weights(
...
+    k_heads = key.shape[-2]
+
+    if q_heads % k_heads != 0:
+      raise ValueError(
+        f"Query heads ({q_heads}) must be a multiple of "
+        f"Key heads ({k_heads}) for Grouped Query Attention."
+      )
+
+    n_rep = q_heads // k_heads
+    key = jnp.repeat(key, n_rep, axis=-2)
```
samanklesaria: I'm worried that using `jnp.repeat` here might lead to GQA not actually producing memory savings (as the key/value tensors will get expanded to the same size anyways). The corresponding jax implementation uses vmap with broadcasting rather than `jnp.repeat` (see line 993 of `jax/_src/nn/funtions.py`). I could certainly be wrong about this, but I think it might be better to copy their vmap approach.

ayushkumrdev: @samanklesaria That's a valid point about memory usage,  jnp.repeat definitely materialises the array, which we want to avoid. I was planning to refactor this using broadcasting with jnp.einsum (reshaping the query and adding a singleton dimension to the key/value) to achieve the same memory savings without copying data.

Does that approach sound good to you, or would you strictly prefer I use vmap to match the implementation in _src/nn/functions.py

samanklesaria: Broadcasting with einsum sounds fine to me.

ayushkumrdev: Great! I've pushed the changes using einsum and reshaping. The tests pass and it avoids the memory overhead.

flax/nnx/nn/attention.py
```diff
@@ -224,17 +254,17 @@ def dot_product_attention(
...
-  assert (
-    query.shape[-2] == key.shape[-2] == value.shape[-2]
-  ), 'q, k, v num_heads must match.'
   assert key.shape[-3] == value.shape[-3], 'k, v lengths must match.'
 
   # Criteria that invoke the more optimized dot product attention
-  if dropout_rate == 0.0 and module == None:
+  # We skip this optimization for GQA (mismatched heads) to use manual broadcasting
+  if (dropout_rate == 0.0 and module == None and 
+      query.shape[-2] == key.shape[-2] == value.shape[-2]):
```
samanklesaria: We should follow this path (using jax's dot product attention) as long as module is None. It works fine with GQA, so no need for a check here.

samanklesaria:
@ayulockedin  Looks like pre-commit hooks are failing. Make sure you've set up pre-commit hooks as in https://flax.readthedocs.io/en/stable/contributing.html

ayushkumrdev:
Agreed. I removed the extra shape checks. The code now falls back to jax.nn.dot_product_attention whenever dropout_rate == 0 and module is None. @samanklesaria thx alot for reviewing and helping me :)

samanklesaria:
@ayulockedin once all the tests pass I'll give it another look, but so far I don't see any major issues.

ayushkumrdev:
@samanklesaria there was another small typo error which made the checks fail but i have fixed that with the recent commit should be good to run the checks again thx

flax/nnx/nn/attention.py
```diff
@@ -224,17 +254,16 @@ def dot_product_attention(
...
   assert key.ndim == query.ndim == value.ndim, 'q, k, v must have same rank.'
   assert (
     query.shape[:-3] == key.shape[:-3] == value.shape[:-3]
   ), 'q, k, v batch dims must match.'
-  assert (
-    query.shape[-2] == key.shape[-2] == value.shape[-2]
-  ), 'q, k, v num_heads must match.'
   assert key.shape[-3] == value.shape[-3], 'k, v lengths must match.'
 
+  # We skip this optimization for GQA (mismatched heads) to use manual broadcasting
```
samanklesaria: Remove this comment

tests/nnx/nn/gqa_test.py
```diff
@@ -0,0 +1,66 @@
...
+        B, T, D = 1, 4, 8
+        query = jnp.ones((B, T, 5, D))
+        key   = jnp.ones((B, T, 2, D))
+        value = key
+
+        try:
+            nnx.dot_product_attention(query, key, value)
+            assert False, "Should have raised ValueError"
+        except ValueError as e:
+            # Fixed expectation to match your code's error message
```
samanklesaria: Looks like a comment from Claude. Remove.

tests/nnx/nn/gqa_test.py
```diff
@@ -0,0 +1,66 @@
...
+        query = jax.random.normal(k1, (B, T, num_heads_q, D))
+        key   = jax.random.normal(k2, (B, S, num_heads_kv, D))
+        value = jax.random.normal(k3, (B, S, num_heads_kv, D))
+
+        # Manually repeat heads for JAX reference
+        n_rep = num_heads_q // num_heads_kv
+        key_jax = jnp.repeat(key, n_rep, axis=-2)
+        value_jax = jnp.repeat(value, n_rep, axis=-2)
+
+        jax_out = jax.nn.dot_product_attention(query, key_jax, value_jax)
```
samanklesaria: You don't want to manually repeat here. The whole point is to handle unmatched head sizes.

tests/nnx/nn/gqa_test.py
ayushkumrdev: @samanklesaria ya this was the typo which was failing the CI checks, i will do a final cleanup and commit

flax/nnx/nn/attention.py
ayushkumrdev: on it

tests/nnx/nn/gqa_test.py
ayushkumrdev: yup will remove repeat here forgot to change since did the broadcasting logic in attention.py

samanklesaria (approved):
Looks good to me!

ayushkumrdev:
@samanklesaria Just a heads up: I've opened Issue #5198  to track the follow-up work for the MultiHeadAttention module updates.

Once this functional PR lands, I plan to tackle that issue to bring full GQA parity to the class API (adding num_key_value_heads to __init__). Just wanted to link the two contextually so we have a roadmap!

cgarciae:
@ayulockedin @samanklesaria lets move the tests to `attention_test.py`.

ayushkumrdev:
Done! I've moved the GQA tests to attention_test.py (added as the TestGQADotProductAttention class) and removed the separate test file.

I also ran the pre-commit hooks and squashed everything into a single clean commit. Ready for review! @cgarcia

PR #5181, opened 2026-01-08
Reduce warning message of about GCS usage

wdhongtw:
# What does this PR do?

The warning about TensorFlow not being found is now only logged when 
io utilities are called with an argument starting with "gs://",
instead of always being logged if TensorFlow is not installed.

## Checklist
- [ ] This change includes necessary high-coverage tests. (No quality testing = no merge!)

### DIFF
--- a/flax/io.py
+++ b/flax/io.py
@@ -21,11 +21,41 @@ import glob as glob_module
 import importlib
 import os
 import shutil
+from typing import Callable, ParamSpec, TypeVar
 from enum import Enum
 
 from absl import logging
 
 from . import errors
+
+
+_T = TypeVar('_T')
+_P = ParamSpec('_P')
+
+
+def _check_gs_path(func: Callable[_P, _T]) -> Callable[_P, _T]:
+  """Decorator to check for gcs paths and warn if tf is not available."""
+
+  _gs_warning = (
+    'Tensorflow library not found, tensorflow.io.gfile '
+    'operations will use native shim calls. '
+    "GCS paths (i.e. 'gs://...') cannot be accessed."
+  )
+
+  def wrapper(*args: _P.args, **kwargs: _P.kwargs):
+    assert io_mode is not None
+    if io_mode == BackendMode.DEFAULT:
+      values = list(args) + list(kwargs.values())
+      str_values = [a for a in values if isinstance(a, str)]
+
+      # we still check against "mode" param of GFile function, but it's fine.
+      if any(v.startswith('gs://') for v in str_values):
+        logging.warning(_gs_warning)
+
+    return func(*args, **kwargs)
+
+  return wrapper
+
 
 # Global Modes and selective import of tensorflow.io gfile.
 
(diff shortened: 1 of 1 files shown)


vfdev-5:
Thanks for the PR @wdhongtw 
Can you share the context of the problem you are trying to solve by removing the warning ?

wdhongtw:
Hi @vfdev-5 

I'm working on migration some ML workload from GPU to TPU, and using `torchax`
to speed up the process, as it provides a simple bridge to use any PyTorch model on TPU.

Currently any lib/application import `torchax` will generate a log like

> WARNING:root:Tensorflow library not found, tensorflow.io.gfile operations will use native shim calls. GCS paths (i.e. 'gs://...') cannot be accessed.

After some investigation, is because that `torchax` import `flax.training.checkpoints`, which import `flax.io`

- <https://github.com/google/torchax/blob/28f662c53510aa9de1706da7009cf2134476edbc/torchax/checkpoint.py#L22>
- <https://github.com/google/flax/blob/fefc4c9bb42045a2678511c03c2fd708dde2b994/flax/training/checkpoints.py#L45>

Also for other application depends on `torchax`, e.g. vLLM-TPU, the warning log always showing up at first few line.

I thought this warning log should be irrelevant to most open source lib/application, also the log can be confusing to users that use `flax` but without `tensorflow` lib. Here I try to preserve the log, but postponed its timing to when it is actually needed.

vfdev-5:
Thanks for giving the context! 

It's true that such warning can be annoying in the project that are not supposed to have TF.
In flax, io module is a wrapper around the tf gfile mainly...

Another more important point, flax training checkpoint code was relevant before [orbax checkpoint manager](https://orbax.readthedocs.io/en/latest/guides/checkpoint/orbax_checkpoint_101.html#managing-checkpoints) introduced, today users will see a deprecation warning if use methods like `flax.training.checkpoints.save_checkpoint`: https://github.com/google/flax/blob/a7929b37829bb108ed64fb813cd92694bbd4dda0/flax/training/checkpoints.py#L707-L716

wdhongtw:
Thanks @vfdev-5 for reminding this.

Since that the API this PR addressing is part of depredated interface, I should change the caller side instead.
Closing PR now. 🙂‍↕️

vfdev-5:
We have to update our docs as well to clearly state these deprecations in the modules. Sorry about that!

wdhongtw:
Hi @qihqi, wondering whether you are interested in this.

I tried to implemented same functionality through `orbax` CheckpointManager and/or Checkpointer, but there seems no one-to-one migration, while leaving no warning log. :(

These materials may help

- <https://flax-linen.readthedocs.io/en/latest/guides/converting_and_upgrading/orbax_upgrade_guide.html>
- <https://orbax.readthedocs.io/en/latest/guides/checkpoint/api_refactor.html>

vfdev-5:
I would suggest to use: https://flax.readthedocs.io/en/latest/guides/checkpointing.html#save-checkpoints tutorial instead of flax-linen one.

PR #5188, opened 2026-01-13
Update mnist example to use NNX (clone of #5064)

samanklesaria:
This is clone of #5064 to work around internal CI errors.


IvyZX (changes requested):
I'm surprised that this didn't run on GitHub, but you'll need to convert the `jax2tf_test.py` too. Otherwise everything looks good!

IvyZX:
> I'm surprised that this didn't run on GitHub, but you'll need to convert the `jax2tf_test.py` too. Otherwise everything looks good!

Resolved offline & merged

PR #5192, opened 2026-01-16
fix(nnx): ensure PathContains('') matches all paths

### LINKED ISSUE #4660: Filtering with `nnx.PathContains('')` returns an empty state
Filtering a module with `nnx.PathContains('')` always returns an empty state.
Consider the following example:
```python
from flax import nnx

class CNN(nnx.Module):

  def __init__(self, *, rngs: nnx.Rngs):
    self.conv1 = nnx.Conv(1, 32, kernel_size=(3, 3), rngs=rngs)
    self.conv2 = nnx.Conv(32, 64, kernel_size=(3, 3), rngs=rngs)
    ...

  def __call__(self, x):
    ...

model = CNN(rngs=nnx.Rngs(0))

graph, state, rest = nnx.split(model, nnx.PathContains(''), ...)
print(state == nnx.State({})) # True
```
I would argue that the empty path `''` is contained in every path, much like `'' in any_string` is always true.
Therefore, I'd expect `state` to contain the full state and `rest` to be empty.
What do you think?

### System information
- OS Platform and Distribution (e.g., Linux Ubuntu 16.04): Ubuntu 24.04
- Flax, jax, jaxlib versions (obtain with `pip show flax jax jaxlib`: Flax 0.10.2, jax 0.5.0
- Python version: 3.12

ayulockedin:
# What does this PR do?

Title: feat(nnx): support PathContains('') as wildcard (User Ergonomics)

Description:

Problem Overview
This PR addresses the usability friction highlighted in Issue #4660.

Currently, nnx.PathContains('') returns an empty state. While this is technically correct given the underlying tuple implementation (checking if '' is inside the tuple), it contradicts the strong intuition users bring from Python strings and other systems where "empty string" implies "match all" or "root".

As noted in the issue, users expect '' to behave like a wildcard. When it silently fails, it leads to confusion and debugging time.

The Improvement
I have updated PathContains to explicitly handle '' as a wildcard.

Old Behavior: PathContains('') -> Returns Empty (Matches nothing).

New Behavior: PathContains('') -> Returns Identity (Matches everything).

This effectively makes PathContains('') an alias for PathContains(True) or ..., but using the syntax users naturally try first.

Implementation
I modified PathContains.__call__ in flax/nnx/filterlib.py to return True immediately if self.key is ''. This incurs negligible overhead but aligns the API with user expectations.

Verification
-New Unit Test: Added test_path_contains_empty_string in filters_test.py which confirms '' now captures the full state.

-Existing Tests: All existing filter tests pass; this change is backward compatible as '' previously matched nothing useful.

(Relates to #4660)

Fixes #4660

## Checklist
- [ ] The documentation and docstrings adhere to the [documentation guidelines](https://github.com/google/flax/blob/main/docs/README.md#how-to-write-code-documentation).

### DIFF
--- a/flax/nnx/filterlib.py
+++ b/flax/nnx/filterlib.py
@@ -93,6 +93,8 @@ class PathContains:
   exact: bool = True
 
   def __call__(self, path: PathParts, x: tp.Any):
+    if self.key == '':
+      return True
     if self.exact:
       return self.key in path
     return any(str(self.key) in str(part) for part in path)
--- a/tests/nnx/filters_test.py
+++ b/tests/nnx/filters_test.py
@@ -36,5 +36,18 @@ class TestFilters(absltest.TestCase):
     self.assertIn('backbone2', backbones_state)
     self.assertNotIn('head', backbones_state)
 
+  def test_path_contains_empty_string(self):
+    class Model(nnx.Module):
+      def __init__(self):
+        self.a = nnx.Param(1)
+        self.b = nnx.Param(2)
+
+    model = Model()
+    state = nnx.state(model, nnx.PathContains(''))
+
+    self.assertLen(state, 2)
+    self.assertIn('a', state)
+    self.assertIn('b', state)
+
 if __name__ == '__main__':
   absltest.main()


ayushkumrdev:
@samanklesaria  Hey could u check this PR out when you have a moment? Thanks!

vfdev-5:
@ayulockedin I do not agree that `PathContains('')` should match all paths. To return all paths users should not filter or use other options like `...` / `True`. As Cristian answered in the linked issue, that this behavior is expected: https://github.com/google/flax/issues/4660#issuecomment-2758938330, I would close this PR.

Please, leave a comment about what you would like to propose before submitting a PR especially fixing a closed issue.
Thanks for your other contributions!

ayushkumrdev:
@vfdev-5 Thank you for the review and the clarification. I missed the context in the linked issue regarding the expected behavior of PathContains('')—my apologies for the oversight.
​I also appreciate the advice on the contribution workflow. I will make sure to open a discussion or issue to propose changes before submitting a PR in the future, especially regarding closed issues.
​I’ll go ahead and close this PR now. Thanks again for your time!

PR #5205, opened 2026-01-27
Add out_sharding argument to Embed layer call

samanklesaria:
This PR adds an `out_sharding` argument to `Embed` layers. Because `jnp.take` (which is what `Embed` wraps) does not provide an `out_sharding` argument itself, we handle this with `auto_axes`. The PR includes tests that this works as expected, and that without the `out_sharding` argument, mismatched sharding for arrays and indices is expected.

### DIFF
--- a/tests/nnx/spmd_test.py
+++ b/tests/nnx/spmd_test.py
@@ -231,7 +231,7 @@ class TestSPMD(parameterized.TestCase):
       emb = nnx.Embed(num_embeddings=8, features=4, rngs=nnx.Rngs(0))
       emb = reshard(emb, P("X"))
       sharded_array = reshard(jnp.arange(4), P("Y"))
-      self.assertRaises(ValueError, emb, sharded_array)
+      self.assertRaises(Exception, emb, sharded_array)
       self.assertEqual('float32[4@X,4]',
         str(jax.typeof(emb(sharded_array, out_sharding=P("X")))))
 


tests/nnx/spmd_test.py
```diff
@@ -225,6 +225,16 @@ def test_out_sharding_linear_layers(self):
         assert 'float32[2@X,4]' in str(jax.typeof(layer(sharded_array)))
         assert 'float32[2@X,4@Y]' in str(jax.typeof(layer(sharded_array, out_sharding=P("X", "Y"))))
 
+  def test_out_sharding_embed(self):
+    mesh = jax.make_mesh((2, 2), ("X", "Y"), axis_types=(AxisType.Explicit, AxisType.Explicit))
+    with jax.set_mesh(mesh):
+      emb = nnx.Embed(num_embeddings=8, features=4, rngs=nnx.Rngs(0))
+      emb = reshard(emb, P("X"))
+      sharded_array = reshard(jnp.arange(4), P("Y"))
+      self.assertRaises(Exception, emb, sharded_array)
```
IvyZX, line 234: Why does this raise exception? What is the exception?

samanklesaria: It should raise an exception because the shardings don't match (and we're using explicit sharding). The exception type itself isn't exposed- it's in `jax._src`.

IvyZX (approved):

PR #5218, opened 2026-02-03
Fix Sequential unwrapping for static attributes

### LINKED ISSUE #5217: Sequential Layers are Invalid Pytrees
Say you do the following:

```python
from flax import nnx
from jax import tree
model = nnx.Sequential(nnx.Linear(2,8, rngs=nnx.Rngs(0)))
tree.map(lambda x: x + 1, model)
```

You'll get the following error: 
```
TypeError: attribute name must be string, not 'int'
```

This is because of how flax pytrees do unwrapping. Specifically, they only convert data children's names from ints to strings, not static children. This can easily be fixed by adding the requisite transformation to `pytreelib.py`.

samanklesaria:
Fixes #5217 
Depends on #5222

### DIFF
--- a/tests/nnx/module_test.py
+++ b/tests/nnx/module_test.py
@@ -39,6 +39,14 @@ class PytreeTest(absltest.TestCase):
 
     self.assertEqual(jax.tree.leaves(foo), [1])
 
+  def test_sequential_map(self):
+    model = nnx.Sequential(nnx.Linear(2,8, rngs=nnx.Rngs(0)))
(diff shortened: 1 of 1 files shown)


samanklesaria:
As this change is a prerequisite for #5217 as well, I added a fix here too. Fixes #5217

samanklesaria:
#5126 fixed the underlying issue here. But there's still tests I added for this issue that are valuable, so this PR only contains the tests now.

PR #5222, opened 2026-02-03
Pin doc deps to lower versions

samanklesaria:
This PR attempts to fix the doctests that have been failing lately by pinning ipython.


samanklesaria:
Found something that works! Now time for a binary search: let's remove half of these pins and see what happens.

samanklesaria:
Looks like tqdm was the culprit.

PR #5223, opened 2026-02-04
Remove nnx.List

samanklesaria:
An an effort to simplify Flax Modules, this PR replaces the wacky `nnx.List` that pretends to be a list but is really a dict that converts its indices to strings on the fly with a regular python list. This should simplify the code considerably, and make faster flattening possible down the road.

### DIFF
--- a/flax/nnx/helpers.py
+++ b/flax/nnx/helpers.py
@@ -27,7 +27,7 @@ from flax.nnx.proxy_caller import ApplyCaller
 from flax.nnx.rnglib import Rngs
 from flax.nnx.statelib import State
 from flax.training.train_state import struct
-from flax.nnx.variablelib import Variable
+from flax.nnx.pytreelib import data
 
 A = tp.TypeVar('A')
 M = tp.TypeVar('M', bound=Module)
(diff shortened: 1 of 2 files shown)


cgarciae:
@samanklesaria Thanks! Two comments:
1. We probably want to add `Rngs.split` as well (including the `only` argument similar to `Rngs.fork`).
2. If we add `Rngs.split` we should deprecate the `split` argument in `fork`, meaning, copy over the implementation instead of calling `fork`.

cgarciae:
Wait no before implementing the above, I think the important thing is to figure out how we are going to recommend users to conditionally split RngStream when passing it to a transform like `vmap` when you need to split some streams and others not?

Here are some ideas:

1.
```python
rngs = nnx.Rngs(params=0, dropout=1)
rngs = rngs.split(5, only='dropout')

@nnx.vmap(in_axis=(None, 0), ...)
def f(params, dropout):
  rngs = nnx.Rngs(params=params, dropout=dropout)
  ...
  
f(rngs.params, rngs.dropout)
```
2.
```python
rngs = nnx.Rngs(params=0, dropout=1)
rngs = rngs.split(5, only='dropout')
graphdef, dropout, params = nnx.split(rngs, 'dropout', ...)

@nnx.vmap(in_axis=(None, 0), ...)
def f(params, dropout):
  rngs = nnx.merge(graphdef, params, dropout)
  ...
  
f(params, dropout)
```
3.
```python
rngs = nnx.Rngs(params=0, dropout=1)

@nnx.vmap(in_axis=(None, 0), ...)
def f(params, dropout):
  rngs = nnx.Rngs(params=params, dropout=dropout)
  ...
  
f(rngs.params, rngs.dropout.split(5))
```

PR #5225, opened 2026-02-06
Pin optax version to fix CI issues

samanklesaria:
CI is failing because the most recent optax seems to break things. This PR pins it to the old version.


jakevdp:
I ran into this in trying to update jax-ai-stack. Pinning a specific minor version of a dependency can make it difficult for users to find a combination of packages that works together. Could you pin a version range instead? Has the issue that prompted this been reported to optax?

jakevdp:
(in the meantime, I'll solve this by not updating flax beyond v0.12.3, and I suspect others will do this as well)

PR #5231, opened 2026-02-09
Add 'opt_sharding' option for Variables

samanklesaria:
This PR lets the 'opt_sharding' metadata key shard Variables' state differently from their optimizer state.

### DIFF
--- a/docs_nnx/guides/flax_gspmd.ipynb
+++ b/docs_nnx/guides/flax_gspmd.ipynb
@@ -132,7 +132,7 @@
     "\n",
     "Let's begin by sharding the simplest component possible - a Flax variable.\n",
     "\n",
-    "When you define a Flax variable, you can pass in a metadata field called `out_sharding`, to specify how the underlying JAX array should be sharded. This field should be a tuple of names, each of which refer to how an axis of the array should be sharded. This sharding will also apply to any `nnx.Optimizer` created from the variable. To specify a different sharding for optimization, use the `opt_sharding` metadata field. \n",
+    "When you define a Flax variable, you can pass in a metadata field called `optimizer_sharding`, to specify how the underlying JAX array should be sharded. This field should be a tuple of names, each of which refer to how an axis of the array should be sharded. This sharding will also apply to any `nnx.Optimizer` created from the variable. To specify a different sharding for optimization, use the `opt_sharding` metadata field. \n",
     "\n",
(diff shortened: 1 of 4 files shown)


docs_nnx/guides/flax_gspmd.ipynb
```diff
@@ -128,63 +125,24 @@
...
    "cell_type": "markdown",
+   "id": "f36ff900",
    "metadata": {},
    "source": [
     "## Shard a single-array model\n",
     "\n",
     "Let's begin by sharding the simplest component possible - a Flax variable.\n",
     "\n",
-    "When you define a Flax variable, you can pass in a metadata field called `out_sharding`, to specify how the underlying JAX array should be sharded. This field should be a tuple of names, each of which refer to how an axis of the array should be sharded.\n",
+    "When you define a Flax variable, you can pass in a metadata field called `out_sharding`, to specify how the underlying JAX array should be sharded. This field should be a tuple of names, each of which refer to how an axis of the array should be sharded. This sharding will also apply to any `nnx.Optimizer` created from the variable. To specify a different sharding for optimization, use the `opt_sharding` metadata field. \n",
```
IvyZX: The lint error was interesting. I'm not sure what happened but since this PR is a simple case, perhaps you can just: reset this ipynb to head, open it in pure text (json), add this line, and run `jupyter nbconvert --to markdown docs_nnx/guides/flax_gspmd.ipynb` to generate the markdown file. 

That should guarantee you have valid ipynb and md files, minimally changed.

flax/nnx/training/optimizer.py
```diff
@@ -49,7 +49,10 @@ class OptVariable(OptState):
 def to_opt_state(tree):
   def _to_opt_state(x):
     if isinstance(x, Variable):
-      opt_state = OptVariable(x.get_value(), **x.get_metadata())  # type: ignore
+      opt_metadata = x.get_metadata()
+      if 'opt_sharding' in opt_metadata:
+        opt_metadata['sharding_names'] = opt_metadata.pop('opt_sharding')
```
cgarciae:
```suggestion
        opt_metadata['out_sharding'] = opt_metadata.pop('opt_sharding')
```

cgarciae:
Nit: `opt_sharding` is very similar to `out_sharding` (the new name for `sharding_names`), I don't know if this can be a very silly source of error. What do you think? We could also consider `optimizer_sharding`.

samanklesaria:
> Nit: `opt_sharding` is very similar to `out_sharding` (the new name for `sharding_names`), I don't know if this can be a very silly source of error. What do you think? We could also consider `optimizer_sharding`.

I like `optimizer_sharding`. Agreed, it seems like `opt_sharding` might be too close.

PR #5245, opened 2026-02-12
Track explicit sharding in Variables

samanklesaria:
Consider the following:

```python
import jax
import jax.numpy as jnp
from flax import nnx
from jax.sharding import reshard
mesh1 = jax.make_mesh((2, 2), ("a", "b"),
                      axis_types=(AxisType.Explicit, AxisType.Explicit))
jax.set_mesh(mesh1)

x = reshard(jnp.ones((4,4)), P('a', 'b'))

class Foo(nnx.Module):
    def __init__(self, x):
        self.xv = xv = nnx.Param(x)
foo = Foo(x)
```

Currently, `foo.xv` does not result in the variable's `named_sharding` or `mesh` being set. But these should be inherited from the sharding of the value. This PR ensures that `foo.xv` sets the proper sharding metadata.

### DIFF
--- a/flax/nnx/variablelib.py
+++ b/flax/nnx/variablelib.py
@@ -1703,13 +1703,13 @@ class Variable(tp.Generic[A], reprlib.Representable, metaclass=VariableMeta):
       and any(self._raw_value.sharding.spec)
     ):
       if 'out_sharding' in self._var_metadata and self._var_metadata['out_sharding'] is not None:
-        assert self._var_metadata['out_sharding'] == self._raw_value.sharding.spec, \
-          "Mismatched out_sharding: Flax metadata {} does not match exlicit sharding time {}".format(
-            self._var_metadata['out_sharding'], self._raw_value.sharding.spec)
+        if self._var_metadata['out_sharding'] != self._raw_value.sharding.spec:
+          raise ValueError("Mismatched out_sharding: Flax metadata {} does not match explicit sharding time {}".format(
+            self._var_metadata['out_sharding'], self._raw_value.sharding.spec))
       if 'mesh' in self._var_metadata and self._var_metadata['mesh'] is not None:
-        assert self._var_metadata['mesh'] == self._raw_value.sharding.mesh, \
-          "Mismatched meshes: Flax metadata {} does not match explicit sharding type {}".format(
-            self._var_metadata['mesh'], self._raw_value.sharding.mesh)
+        if self._var_metadata['mesh'] != self._raw_value.sharding.mesh:
+          raise ValueError("Mismatched meshes: Flax metadata {} does not match explicit sharding type {}".format(
+            self._var_metadata['mesh'], self._raw_value.sharding.mesh))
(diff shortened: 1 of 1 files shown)


flax/nnx/variablelib.py
```diff
@@ -1408,16 +1397,19 @@ def __init__(
...
         value,
-        metadata['out_sharding'],
+        metadata['sharding_names'],
         metadata.get('sharding_rules', None),
         metadata.get('mesh', None),
       )
     if ref:
       value = jax.new_ref(value)  # type: ignore
     object.__setattr__(self, '_raw_value', value)
+    if hasattr(value, 'sharding'):
```
cgarciae: we should only do this if `sharding_names` is not present, also its now called `out_sharding`

flax/nnx/variablelib.py
```diff
@@ -1761,6 +1766,14 @@ def set_value(self, value: A, *, index: tp.Any = MISSING):
       else:
         self._raw_value[index] = value  # type: ignore
 
+    if (
+      hasattr(self._raw_value, 'sharding')
+      and isinstance(self._raw_value.sharding, NamedSharding)
+      and any(self._raw_value.sharding.spec)
+    ):
+      self._var_metadata['out_sharding'] = self._raw_value.sharding.spec
```
cgarciae, line 1713: Maybe we only do this if `out_sharding` is not provided? What happens when both are defined?

flax/nnx/spmd.py
```diff
@@ -74,10 +74,10 @@ def remove_field(fields, index, value):
 
   def _remove_axis(x: tp.Any):
     if isinstance(x, variablelib.Variable):
-      if hasattr(x, 'out_sharding') and x.out_sharding is not None:
-        x.set_metadata(
-          out_sharding=remove_field(x.out_sharding, index, axis_name)
-        )
+      if hasattr(x, 'out_sharding') and x.out_sharding:
```
cgarciae: is this needed?

samanklesaria: I think I ran into this when `x.out_sharding` is an empty tuple. We try to remove a field from the empty tuple and get an error. But none of the tests seem to fail when I remove this part, so I don't remember when I encountered this error. I'll try to reproduce it and add a test for it.

flax/nnx/variablelib.py
samanklesaria: As my test `test_set_value_with_different_sharding` below assumes, I had thought that the raw value's sharding always takes precedence. I could see the argument, however, for throwing an error if there's a mismatch instead.

flax/nnx/spmd.py
samanklesaria: Without the hasattr check, you get the error `BatchTracer has no attribute out_sharding` in `tests/nnx/spmd_test.py::TestSPMD::test_add_remove_axis_in_transform`. Without the `x.out_sharding` check, however, nothing bad seems to happen. I removed it.

PR #5246, opened 2026-02-13
Generalize out_sharding to work with NamedSharding and Format

samanklesaria:
Like the reverted PR #5089, this PR generalized flax's spmd APIs to allow sharding_metadata to be NamedSharding or Format objects.

### DIFF
--- a/flax/core/spmd.py
+++ b/flax/core/spmd.py
@@ -56,7 +56,7 @@ def apply_rules(sharding, sharding_rules):
 
 def _apply_sharding(value, sharding, mesh):
   if isinstance(sharding, Format):
-    return jax.device_put(value, sharding)
+    return jax.lax.with_sharding_constraint(value, sharding)
   if mesh.are_all_axes_explicit:
     return jax.sharding.reshard(value, sharding)
   elif mesh.are_all_axes_auto:


flax/core/spmd.py
```diff
@@ -19,20 +19,44 @@
...
 from flax.core import meta
+from jax.experimental.layout import Format
 from flax.typing import (
     LogicalRules,
     Sharding,
 )
 
 def get_pspec(sharding, sharding_rules = None) -> PartitionSpec:
   """Given an `nnx.Variable`, return its `PartitionSpec`."""
+  return apply_rules(sharding, sharding_rules)   # type: ignore
```
samanklesaria, line 30: I'm maintaining this function for backwards compat, but because it now works on NamedShardings and Formats too, we can't have the name be called 'get_pspec'. What this function really does it apply a mapping transformation to a sharding. So I've called the more general version 'apply_rules'.

flax/core/spmd.py
```diff
@@ -19,20 +19,44 @@
...
+
+def map_sharding(f, sharding):
+  if isinstance(sharding, PartitionSpec) or isinstance(sharding, tuple):
+    return PartitionSpec(*map(f, sharding))
+  elif isinstance(sharding, NamedSharding):
+    return NamedSharding(sharding.mesh, map_sharding(f, sharding.spec)) # type: ignore
+  elif isinstance(sharding, Format):
+    return Format(sharding.layout, map_sharding(f, sharding.sharding))  # type: ignore
```
cgarciae, line 38: I guess the `else` case is `return None` so we are implicitly doing this?

flax/core/spmd.py
```diff
@@ -19,20 +19,44 @@
...
-    return PartitionSpec(*from_sharding_rules(sharding, rules))
-  return PartitionSpec(*sharding)
+    rules = {alias: on_mesh for (alias, on_mesh) in composite_rules(context_rules, sharding_rules)}
+  else:
+    rules = {}
+  return map_sharding(lambda a: rules.get(a, a), sharding)
 
 def _apply_sharding(value, sharding, mesh):
+  if isinstance(sharding, Format):
+    return jax.device_put(value, sharding)
```
cgarciae: This might not work inside of `jit`, we should just pass it to `reshard` or `with_sharding_constraint`, if they don't support it yet then maybe we just tell the user to no use `eager_sharding`.

flax/core/spmd.py
samanklesaria: Exactly. Should I make this explicit?

flax/core/spmd.py
samanklesaria: You can't use `reshard` with a Format. You get the error:
```
TypeError: `out_sharding` argument of reshard only supports instances of `NamedSharding` or `PartitionSpec`. 
```

But `with_sharding_constraint` seems to work.

flax/core/spmd.py
cgarciae: yeah maybe we can call it get_sharding or something? Seems like the return type also doesn't match.

PR #5179, opened 2026-01-07
feat(nnx): add preferred_element_type to attention.py and recurrent.py, out_sharding to recurrent.py

aarushisingh04:
## What does this PR do?

This PR adds support for `preferred_element_type` and `out_sharding` to core NNX
Attention and RNN modules, aligning them with existing NNX layers such as
`Linear` and `Conv`.

The change improves API consistency and gives users fine-grained control over
numerical precision and distributed sharding at the module level, without
changing any default behavior.

---

## Motivation

In NNX, low-level layers like `Linear` and `Conv` already expose
`preferred_element_type` and `out_sharding`, but higher-level modules that rely
on these layers such as `MultiHeadAttention` and RNN cells did not forward or
expose these parameters.

Negatives/Difficulties -
- Controlling numerical precision for attention and recurrent modules
- Applying sharding constraints to module outputs in distributed settings
- Writing consistent code when mixing `Linear` / `Conv` with Attention or RNNs

This PR closes that gap by plumbing the existing parameters through these
modules in a fully backward-compatible way.

---

### 1. Attention module enhancements (`attention.py`)

Updated:
- `dot_product_attention`
- `dot_product_attention_weights`
- `MultiHeadAttention`

Changes:
- Added `preferred_element_type` to function signatures and forwarded it to
  internal `jnp.einsum` calls
- Added `out_sharding` to `MultiHeadAttention.__call__` to allow sharding of the
  output projection
- Updated docstrings to document the new parameters

---

### 2. RNN module enhancements (`recurrent.py`)

Updated:
- `LSTMCell`
- `OptimizedLSTMCell`
- `SimpleCell`
- `GRUCell`

Changes:
- Added `preferred_element_type` to `__init__` and forwarded it to internal
  `Linear` layers
- Added `out_sharding` to `__call__` for all cells
- Standardized and completed docstrings (Args / Returns)

---

## Backward Compatibility

- No existing behavior is changed
- All new parameters are optional
- Default numerical precision and sharding remain unchanged unless explicitly
  specified by the user

---

## Tests Added

A new targeted test suite was added:

- `tests/nnx/technical_support_test.py`

This suite verifies:
- `preferred_element_type` propagation for Attention and RNN modules
- `out_sharding` exposure and forwarding in `MultiHeadAttention` and RNN cells

Specific tests include:
- `test_mha_preferred_element_type`
- `test_mha_out_sharding_signature`
- `test_lstm_preferred_element_type`
- `test_gru_preferred_element_type`
[...]

PR #5253, opened 2026-02-15
feat(nnx): add preferred_element_type to attention module

aarushisingh04:
## What does this PR do?

adds `preferred_element_type` parameter to `dot_product_attention_weights`, `dot_product_attention` and `MultiHeadAttention` in `flax/nnx/nn/attention.py`.

split from [#5179](https://github.com/google/flax/pull/5179) for easier review.

**changes:**
- `dot_product_attention_weights`: forwards `preferred_element_type` to `jnp.einsum`
- `dot_product_attention`: forwards to both `jnp.einsum` calls and conditionally applies `.astype()` only when `preferred_element_type is not None`
- `MultiHeadAttention.__init__`: accepts `preferred_element_type` and passes it to all `LinearGeneral` layers (q/k/v/out)
- `MultiHeadAttention.__call__`: forwards `preferred_element_type` to `self.attention_fn`

**tests:**
- adds `test_preferred_element_type` to `attention_test.py`

## Checklist

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -101,9 +101,8 @@ def dot_product_attention_weights(
       the logits to mask out the non-causal parts of the attention matrix,
       but other implementations like cudnn will avoid computing the
       non-causal regions, providing speedups.
-    preferred_element_type: Optional parameter controls the data type output by
-      the dot product. This argument is passed to ``dot_general`` function.
-      See ``jax.lax.dot`` for details.
+    preferred_element_type: numerical precision of the computation, see
+      `jax.lax.dot_general` for details.
 
   Returns:
     Output of shape `[batch..., num_heads, q_length, kv_length]`.
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -63,6 +63,7 @@ def dot_product_attention_weights(
   module: Module | None = None,
   promote_dtype: PromoteDtypeFn = dtypes.promote_dtype,
   is_causal: bool = False,
+  preferred_element_type: Dtype | None = None,
 ):
   """Computes dot-product attention weights given query and key.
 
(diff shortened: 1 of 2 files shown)


samanklesaria (approved):

IvyZX (commented):
We already have `precision` to control the op precision and `dtype` to control the param precision. What's a use case for this additional API?

samanklesaria:
> We already have `precision` to control the op precision and `dtype` to control the param precision. What's a use case for this additional API?

Good point. It seems like these parameters cover the use case for a `preferred_element_type` parameter. @aarushisingh04 does that make sense to you?

aarushisingh04:
@IvyZX @samanklesaria 

thank you for reviewing, i'll address the mentioned points below-

i believe `precision` and `dtype` don't cover this use case. `precision` controls how carefully the hardware computes (e.g. `HIGHEST` uses more precise accumulators) but does not change the output `dtype`, as with bfloat16 inputs and `precision=Precision.HIGHEST`, you still get bfloat16 attention weights back. 

`dtype` promotes the inputs before the einsum but similarly doesn't let you accumulate in a higher dtype. so, `preferred_element_type` is the only way to get float32 attention logits when inputs are bfloat16, without manually inserting `.astype()` calls or casting the entire Q/K/V tensors to float32. this is the same design choice `FlashAttention` makes for numerical stability.

>just as a note, the motivation behind this PR was that `preferred_element_type` was already added to `nnx.Linear*`, `nnx.Conv*`, and `nnx.Einsum` in [PR #4920](https://github.com/google/flax/pull/4920), included in the Flax 0.12.0 release. this PR meant to extend the same pattern to `MultiHeadAttention` which was not covered by #4920.

if the pr is green-lighted after this discussion i will fix the docstrings (which currently incorrectly describe it as "numerical precision")

- preferred_element_type: numerical precision of the computation, see `jax.lax.dot_general` for details.

to a more accurate description

- preferred_element_type: Optional parameter controls the data type output by the dot product. This argument is passed to ``dot_general`` function. See ``jax.lax.dot`` for details.


and strengthen the tests to cover the mixed-precision use case.

PR #5255, opened 2026-02-15
feat(nnx): add out_sharding to recurrent cells

aarushisingh04:
## What does this PR do?

adds `out_sharding` parameter to the `__call__` methods of recurrent cell classes in `flax/nnx/nn/recurrent.py`: `LSTMCell`, `OptimizedLSTMCell`, `SimpleCell`, and `GRUCell`.

split from [#5179](https://github.com/google/flax/pull/5179) for easier review.

**changes:**
- adds `*, out_sharding=None` to `__call__` of all 4 cell types
- forwards `out_sharding` to all internal `Linear.__call__()` invocations
- adds `__call__` docstrings with `out_sharding` documented for `SimpleCell` 

**tests:**
- adds `test_out_sharding_signature` for `SimpleCell` in `recurrent_test.py`

## Checklist

### DIFF
--- a/tests/nnx/nn/recurrent_test.py
+++ b/tests/nnx/nn/recurrent_test.py
@@ -642,5 +642,26 @@ class TestRNN(absltest.TestCase):
     self.assertEqual(model.lstm.cell.recurrent_dropout.rngs.count[...], 1)
 
 
+class TestCellSharding(absltest.TestCase):
+  def test_out_sharding_signature(self):
+    rngs = nnx.Rngs(0)
+    
+    cell_types = [
+      (nnx.SimpleCell, {'in_features': 4, 'hidden_features': 4}),
+      (nnx.LSTMCell, {'in_features': 4, 'hidden_features': 4}),
+      (nnx.OptimizedLSTMCell, {'in_features': 4, 'hidden_features': 4}),
+      (nnx.GRUCell, {'in_features': 4, 'hidden_features': 4}),
+    ]
+
+    for cell_cls, kwargs in cell_types:
+      with self.subTest(cell_cls=cell_cls.__name__):
+        model = cell_cls(**kwargs, rngs=rngs)
+        carry = model.initialize_carry((1, 4), rngs=rngs)
+        x = jnp.ones((1, 4))
+        # Just verify it accepts out_sharding=None without error
+        out = model(carry, x, out_sharding=None)
+        self.assertLen(out, 2)  # All cells return (new_carry, output) or equivalent structure
+
+
 if __name__ == '__main__':
   absltest.main()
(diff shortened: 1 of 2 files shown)


samanklesaria (approved):

IvyZX (approved):

tests/nnx/nn/recurrent_test.py
```diff
@@ -642,5 +642,26 @@ def __call__(self, x):
...
+    ]
+
+    for cell_cls, kwargs in cell_types:
+      with self.subTest(cell_cls=cell_cls.__name__):
+        model = cell_cls(**kwargs, rngs=rngs)
+        carry = model.initialize_carry((1, 4), rngs=rngs)
+        x = jnp.ones((1, 4))
+        # Just verify it accepts out_sharding=None without error
+        out = model(carry, x, out_sharding=None)
+        self.assertLen(out, 2)  # All cells return (new_carry, output) or equivalent structure
```
IvyZX: This test is not valid because it doesn't test when out_sharding has a value, which is the point of this PR.

You can use https://docs.jax.dev/en/latest/jax.sharding.html#jax.sharding.Sharding.is_equivalent_to to check sharding.

IvyZX (changes requested):

tests/nnx/nn/recurrent_test.py
aarushisingh04: thanks for catching that, i have made the requested change

tests/nnx/nn/recurrent_test.py
```diff
@@ -642,5 +642,36 @@ def __call__(self, x):
     self.assertEqual(model.lstm.cell.recurrent_dropout.rngs.count[...], 1)
 
 
+class TestCellSharding(absltest.TestCase):
+  def test_out_sharding(self):
+    """Test that out_sharding correctly applies sharding to cell outputs."""
+    rngs = nnx.Rngs(0)
+    devices = np.array(jax.devices())
+    mesh = jax.sharding.Mesh(devices, axis_names=('x',))
+    sharding = jax.sharding.NamedSharding(mesh, jax.sharding.PartitionSpec())
```
IvyZX: `jax.sharding.PartitionSpec()` is the default sharding, aka. everything is replicated & nothing sharded. Using this as test doesn't make sense.

IvyZX (changes requested):

tests/nnx/nn/recurrent_test.py
aarushisingh04: applied the fix here, changed `PartitionSpec()` → `PartitionSpec('x',)` to actually shard along the mesh axis, set `batch_size = jax.device_count()` so the axis can be split. 

also added `AxisType.Explicit` + `with mesh:` for correct axis resolution.

PR #5270, opened 2026-02-23
Add split method to RngStream

samanklesaria:
This is a small convenience method for `RngStream`s. Specifically, `self.split(k) = self.fork(split=k)`.

### DIFF
--- a/flax/nnx/rnglib.py
+++ b/flax/nnx/rnglib.py
@@ -29,6 +29,7 @@ from flax.nnx.variablelib import Variable
 from flax.nnx import filterlib
 from flax.nnx.pytreelib import Pytree
 from flax.typing import MISSING, Key, Missing
+import warnings
 
 F = tp.TypeVar('F', bound=tp.Callable[..., tp.Any])
 A = tp.TypeVar('A')
@@ -123,7 +124,17 @@ class RngStream(Pytree):
     self.count[...] += 1
     return key
 
+  def split(self, k: int | tuple[int, ...]):
+      key = random.split(self(), k)
+      return type(self)(key, tag=self.tag)
+
   def fork(self, *, split: int | tuple[int, ...] | None = None):
+    if split is not None:
+      warnings.warn(
+        "The 'split' argument of 'fork' is deprecated; use the 'split' method instead.",
+        DeprecationWarning,
+        stacklevel=2,
+      )
     key = self()
     if split is not None:
       key = random.split(key, split)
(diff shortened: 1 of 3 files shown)


cgarciae:
We should add `split` to `Rngs` as well. Also I was wondering how to allow users to easily pass the `Rngs` differently to transforms, at least for `vmap` we can allow creating a prefix tree via a `Rngs.prefix` classmethod so users could do something like this:

```python
rngs = Rngs(params=0, dropout=1)
...
rngs = rngs.split(5, only='dropout')

@vmap(in_axes=Rngs.prefix(params=None, dropout=0), ...)
def f(rngs):
  ...
```
Implementation could look something like:
```python
class Rngs:
  @classmethod
  def prefix(cls, default=None, **kwargs):
    if default is not None:
      kwargs['default'] = default
    rngs = cls() # init empty
    for name, value in kwargs.items():
      setattr(rngs, name, nnx.data(value)) # set the inputs directly as attributes
    return rngs
```

samanklesaria:
> Also I was wondering how to allow users to easily pass the `Rngs` differently to transforms, at least for `vmap` we can allow creating a prefix tree via a `Rngs.prefix` classmethod.

@cgarciae  It seems like the situation where we want to vmap different parts of a pytree with different batch axes extends beyond Rngs. The most generic way of doing this, as you mentioned before, is split and merge:

```python
rngs = nnx.Rngs(params=0, dropout=1)
rngs = rngs.split(5, only='dropout')
graphdef, dropout, params = nnx.split(rngs, 'dropout', ...)

@nnx.vmap(in_axis=(None, 0), ...)
def f(params, dropout):
  rngs = nnx.merge(graphdef, params, dropout)
  ...
  
f(params, dropout)
```

But I agree that the 'prefix' method you propose above is easier to use. What if we add it to Pytree instead? That way, it could be used for sharing some subset of parameters when vmapping modules as well.

samanklesaria:
> > Also I was wondering how to allow users to easily pass the `Rngs` differently to transforms, at least for `vmap` we can allow creating a prefix tree via a `Rngs.prefix` classmethod.

> ... the 'prefix' method you propose above is easier to use. What if we add it to Pytree instead? That way, it could be used for sharing some subset of parameters when vmapping modules as well.

After thinking about it some more, the method you propose works in a subtly different way to how StateAxes based filtering works. Your `prefix` method essentially filters on attribute. But `StateAxes`, when given a string as a filter, looks for a `tag` attribute of the underlying objects and compares it. Hypothetically, someone could store a `RngKey` with a different tag into an `Rngs` object with a different attribute. Is this concerning?

I also assume I should get rid of using `StateAxes` everywhere in the documentation for Rngs.

PR #5292, opened 2026-03-02
Remove optax pin

samanklesaria:
This PR will eventually turn the optax pin in #5225 to a range. For the moment, however, it removes the restrictions entirely to review what exactly the CI errors were.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ dependencies = [
     # keep in sync with jax-version in .github/workflows/build.yml
     "jax>=0.8.1",
     "msgpack",
-    "optax==0.2.6",
+    "optax",
     "orbax-checkpoint",
     "tensorstore",
     "rich>=11.1",


samanklesaria:
Interesting! Whatever was bugging CI before no longer seems to be an issue. The original issue was https://github.com/google/flax/actions/runs/21534529121/job/62379371155, and it used optax 0.2.6. It's very unclear what exactly about optax was causing the kernel to die (as there was no error message)- we know only that when the optax dependency was changed, everything worked. When unrestricted, optax 0.2.7 was used in this PR. We might need to simply disallow optax 0.2.6, allowing either < 0.2.6 or >= 0.2.7.

samanklesaria:
Strange- optax 0.26 works as well on main! I guess we should just revert #5225 now.

matthewfeickert:
@samanklesaria As there has been no release of `optax` since PR https://github.com/google/flax/pull/5225 which was related to [`optax` `v0.2.7`](https://github.com/google-deepmind/optax/releases/tag/v0.2.7) this means that this was actually not an `optax` issue but a transative dependency conflict, correct?  For packaging `flax` `v0.12.4` on conda-forge the `optax` pin will then be patched out. As @jakevdp mentioned in https://github.com/google/flax/pull/5225#issuecomment-3984566238 having pinned dependencies in a public release is disruptive and also [dooms that release to eventually break](https://iscinumpy.dev/post/bound-version-constraints/).

PR #5325, opened 2026-03-12
Add nnx.prefix

samanklesaria:
Followup to #5270. Adds `nnx.prefix`, which can be used to provide 'in_axes' specifications based on a filters.  Note that this depends on a separate PR for temporary configuration changes.

### DIFF
--- a/tests/nnx/rngs_test.py
+++ b/tests/nnx/rngs_test.py
@@ -221,6 +221,8 @@ class TestRngs(parameterized.TestCase):
       def __init__(self, rngs: nnx.Rngs):
         self.linear = nnx.Linear(20, 10, rngs=rngs)
         self.drop = nnx.Dropout(0.1, rngs=rngs)
+      def __call__(self, x):
+        return self.drop(self.linear(x))
 
     with nnx.graphlib.set_graph_updates(False):
       with nnx.graphlib.set_graph_mode(False):
@@ -232,6 +234,10 @@ class TestRngs(parameterized.TestCase):
         assert model.drop.rngs.count[...].shape == (5,)
         bias = model.linear.bias[...]
         assert all(jnp.allclose(x,y) for (x,y) in zip(bias, bias[1:]))
+
+        # This is the same as just using 0 as the model in_axes
+        prefix2 = nnx.prefix(model, {nnx.Variable: 0})
+        nnx.vmap(Model.__call__, in_axes=(prefix2,None))(model, jnp.ones(20))
 
 
   def test_random_helpers(self):


samanklesaria:
Interestingly, the following works:

```python
      def my_fn(rngs):
        model = Model(rngs)
        return (model, rngs)
      prefix = nnx.prefix(rngs, {'dropout': 0})
      model, new_rngs = jax.vmap(my_fn, out_axes=(0, prefix), in_axes=(prefix,))(rngs)
 ```
 So why does `nnx.vmap`, which is basically doing the same thing in tree mode, not work?
 
 As far as I can tell, this is because the `updates` returned by `nnx.vmap` don't have exactly the same pytree shape as the inputs (and therefore don't match the pytree prefixes of the outputs). Specifically, there are `None`s wherever the inputs have been left unchanged. And jax can't handle this discrepancy.

**Comment by samanklesaria** [comment]
Found a fix. The issue is that when we mask variables that haven't changed, we need the prefix for these positions to be None. So when constructing the prefix, we just need to ensure that we treat `nnx.Variable` nodes as leaves rather than recursing into them.

File: flax/nnx/pytreelib.py [hunk 1]
```diff
@@ -112,6 +113,18 @@ def __init__(self):
...
+def prefix(pytree, filter_map):
+  preds = [(filterlib.to_predicate(k), v) for k, v in filter_map.items()]
+
+  def lookup(path, value):
+    for (pred,obj) in preds:
+      if pred(path, value):
+        return obj
+    return None
+
+  return jax.tree.map_with_path(lookup, pytree,
```
cgarciae, line 125: recently added `nnx.map` which can be used here, else you have to convert from the jax path format to the nnx path format

samanklesaria: This would require that the `prefix` filter only select Variable nodes. Which is probably what we usually want, but a little less flexible than using `jax.tree.map_with_path` as I have here.

samanklesaria: I get a strange error if I try to switch to `nnx.map`:
```
E     At that key path, the prefix pytree vmap out_axes has a subtree of type
E         <class 'flax.nnx.rnglib.RngKey'>
E     but at the same key path the full pytree has a subtree of different type
E         <class 'flax.nnx.extract.Mask'>.
```
What's `nnx.extract.Mask`? I guess I have some digging to do.

samanklesaria: For now, I'm just converting the jax path format to the nnx one, which makes the tests pass. But I'll investigate why `nnx.map` produces different behavior.

samanklesaria:
Added an additional test which is failing. Will debug.

samanklesaria:
Merged internally.

PR #5326, opened 2026-03-12
Added support for data masking in Average, Accuracy and MultiMetric

vfdev-5:


flax/nnx/training/metrics.py
```diff
@@ -99,22 +99,32 @@ def reset(self) -> None:
...
+      raise ValueError(f"If mask is provided, {self.argname} should be a jax array")
+    if isinstance(values, (int, float)):
+      self.total[...] += values
+      self.count[...] += 1
+    elif mask is None:
+      self.total[...] += values.sum()
+      self.count[...] += values.size
+    else:
+      self.total[...] += (values * mask.astype(values.dtype)).sum()
+      self.count[...] += mask.sum().astype(self.count.dtype)
```
cgarciae, line 127: `dtype` casting might not be necesarry as `[...] =` preserves dtype.

vfdev-5: Without casting `self.count[...] += mask.sum()` it gives an error:
```
E       FutureWarning: scatter inputs have incompatible types: cannot safely cast value from dtype=float32 to dtype=int32 with jax_numpy_dtype_promotion=standard. In future JAX releases this will result in an error.
```

PR #5344, opened 2026-03-17
Fix bug in rng split

samanklesaria:
This fixes an issue with the recently added `split` function for rngs. If the number of splits was not specified, the previous implementation would not fork the old rng. This could lead to unintuitive behavior. In this PR, split implies fork. The code is also cleaned up.


samanklesaria:
On second thoughts, I'm not sure it's a good idea to automatically fork non-split streams. Closing this PR for now.

PR #5363, opened 2026-03-20
Update the RNG guide with split and without StateAxes

samanklesaria:
Now that the `fork` methods for RNGs have been split off from the `split` methods, we need to update the RNG guide to use the `split` methods (or users will see deprecation warnings when they try to follow the guide). Also: `StateAxes` is not tree-mode compatible, so we have to remove their use in the guide.


flax/nnx/rnglib.py
```diff
@@ -438,7 +438,7 @@ def items(self):
 
   def split(self, k: tp.Mapping[filterlib.Filter, int | tuple[int, ...]] | int | tuple[int, ...]):
     """
-    Splits the keys of the newly created ``Rngs`` object.
+    Splits the keys of a ``Rngs`` object.
```
samanklesaria, line 463: Should probably be a separate PR, but it's such a minor fix

docs_nnx/guides/randomness.md
```diff
@@ -1,24 +1,28 @@
...
+```python
+import jax
 from flax import nnx
+import jax.numpy as jnp
 
 class Model(nnx.Module):
   def __init__(self, *, rngs: nnx.Rngs):
     self.linear = nnx.Linear(20, 10, rngs=rngs)
-    self.drop = nnx.Dropout(0.1, deterministic=False)
+    self.drop = nnx.Dropout(0.1)
```
cgarciae: Lets keep `deterministic=False` by default. Not sure if maybe this PR just need rebasing.

cgarciae (approved):

PR #5384, opened 2026-03-30
Rename vars_as to with_vars

samanklesaria:
This PR is part of a series (extracted from #5376) that standardizes the naming conventions around NNX views, unifying behavior when possible. Ideally all view API names would now have a with_ or as_ prefix. Additionally all view functions will expose an only argument that accepts a filter, be implemented using nnx.map or nnx.recursive_map, and expose a graph argument.


cgarciae:
@samanklesaria this PR needs fixing, currently include change that include `auto_create_variables`

cgarciae (approved):

PR #5405, opened 2026-04-10
Port Jax AI stack minigpt example

samanklesaria:
This PR copies the Jax AI stack example https://docs.jaxstack.ai/en/latest/JAX_for_LLM_pretraining.html to the flax examples directory unchanged. The nnx apis used still seem to be appropriate, so no adjustments are required.

### DIFF
--- a/docs_nnx/examples/minigpt.md
+++ b/docs_nnx/examples/minigpt.md
@@ -1,0 +1,508 @@
+---
+jupyter:
+  jupytext:
+    formats: md,ipynb
+    main_language: python
+    text_representation:
+      extension: .md
+      format_name: markdown
+      format_version: '1.3'
+      jupytext_version: 1.13.8
+---
+
+# Train a miniGPT language model with JAX
+
+
+This tutorial demonstrates how to use JAX, [Flax NNX](http://flax.readthedocs.io) and [Optax](http://optax.readthedocs.io) for language model (pre)training using data and tensor [parallelism](https://jax.readthedocs.io/en/latest/notebooks/Distributed_arrays_and_automatic_parallelization) for [Single-Program Multi-Data](https://en.wikipedia.org/wiki/Single_program,_multiple_data)).
+
+Here, you will learn how to:
+
+- Define the miniGPT model with Flax and JAX automatic parallelism
+- Load and preprocess the dataset
+- Create the loss and training step functions
+- Profile for hyperparameter tuning
+
+
+## Setup
+
+We will use [Tiktoken](https://github.com/openai/tiktoken) for tokenization and [Grain](https://google-grain.readthedocs.io/en/latest/index.html) for data loading.
+
+```python outputId="cb9e04fe-46e1-4363-b48a-9e91623abd82"
+import jax
+```
+
+Get the [TinyStories dataset from Hugging Face](https://huggingface.co/datasets/roneneldan/TinyStories). We only use the training split.
+
+```python outputId="e6eff24e-5578-4277-a0f9-24e27bd91ee0"
+!wget https://huggingface.co/datasets/roneneldan/TinyStories/resolve/main/TinyStories-train.txt?download=true -O TinyStories-train.txt
+```
+
+Import the necessary modules, including JAX NumPy, Flax NNX, Optax, Grain, pandas, and Tiktoken:
+
+```python
+import jax
+import jax.numpy as jnp
+
+import flax.nnx as nnx
+import optax
+
+from dataclasses import dataclass
+from jax.sharding import PartitionSpec as P, reshard
+import grain.python as pygrain
+import pandas as pd
+import tiktoken
+import time
+```
+
+<!-- #region -->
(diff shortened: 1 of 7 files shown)


docs_nnx/examples/minigpt.md
```diff
@@ -0,0 +1,559 @@
...
+- `devices`: This will take the value of [`jax.experimental.mesh_utils((4, 2))`](https://jax.readthedocs.io/en/latest/jax.experimental.mesh_utils.html), enabling us to build a device mesh. It is a NumPy ndarray with JAX devices (a list of devices from the JAX backend as obtained from [`jax.devices()`](https://jax.readthedocs.io/en/latest/_autosummary/jax.devices.html#jax.devices))..
+- `axis_names`, where:
+  - `batch`: 4 devices along the first axis - i.e. sharded into 4 - for data parallelism; and
+  - `model`: 2 devices along the second axis - i.e. sharded into 2 -  for tensor parallism
+
+Let's instantiate `Mesh` as `mesh` and declare the TPU configuration to define how data and model parameters are distributed across the devices:
+<!-- #endregion -->
+
+```python
+mesh = Mesh(mesh_utils.create_device_mesh((4, 2)), ("batch", "model"))
```
vfdev-5: I think we can use jax.make_mesh here instead of experimental jax feature

docs_nnx/examples/minigpt.md
```diff
@@ -0,0 +1,559 @@
...
+                                  kernel_init=nnx.with_partitioning(nnx.initializers.xavier_uniform(), NamedSharding(mesh, P(None, 'model'))),
+                                  bias_init=nnx.with_partitioning(nnx.initializers.zeros_init(), NamedSharding(mesh, P('model'))),
+                                  rngs=rngs)
+        # The second dropout with `flax.nnx.Dropout`.
+        self.dropout2 = nnx.Dropout(rate=rate)
+        # Second layer normalization with `flax.nnx.LayerNorm`.
+        self.layer_norm2 = nnx.LayerNorm(epsilon=1e-6,
+                                         num_features=embed_dim,
+                                         scale_init=nnx.with_partitioning(nnx.initializers.ones_init(), NamedSharding(mesh, P(None, 'model'))),
+                                         bias_init=nnx.with_partitioning(nnx.initializers.zeros_init(), NamedSharding(mesh, P(None, 'model'))),
```
vfdev-5: Let's use scale/bias metadat instead of overriding the initializers

docs_nnx/examples/minigpt.md
```diff
@@ -0,0 +1,559 @@
...
+%load_ext tensorboard
+```
+
+As we're going to be running this model a number of times, we need some scaffolding to more easily compare our work. For a baseline, we'll need to perform some warmup to guarantee that our code is JIT'd and that our TPUs are warm. For improved comparability, we'll only start tracing after we've finished warmup.
+
+```python
+trace_dir = "/tmp/jax-trace/"
+
+def loop_step(batch, step):
+    input_batch = jnp.array(jnp.array(batch).T)
```
vfdev-5: Do we need to wrap it twice?

docs_nnx/examples/minigpt.md
```diff
@@ -0,0 +1,559 @@
...
+
+```python
+# Defines the loss function using `optax.softmax_cross_entropy_with_integer_labels`.
+def loss_fn(model, batch):
+    logits = model(batch[0])
+    loss = optax.softmax_cross_entropy_with_integer_labels(logits=logits, labels=batch[1]).mean()
+    return loss, logits
+
+# Define the training step with the `flax.nnx.jit` transformation decorator.
+@nnx.jit
```
vfdev-5: let's donate model and optimizer args

docs_nnx/examples/minigpt.md
samanklesaria: Good catch. Fixed.

docs_nnx/examples/minigpt.md
samanklesaria: I didn't know about those arguments! Fixed.

docs_nnx/examples/minigpt.md
samanklesaria: Done, I think.

samanklesaria:
Actually, it seems we need to do a little more to make this work with the current NNX. Currently getting some sharding errors. Will debug.

docs_nnx/examples/minigpt.md
samanklesaria: Good catch. Fixed.

samanklesaria:
Sharding issues resolved. Have to wait until I have access to multiple gpus to run the full notebook though.

samanklesaria:
I also exposed the `is_causal` argument from the `dot_product_attention` function to `MultiHeadAttention`. It seems like the user is otherwise expected to generate the causal mask themselves.

cgarciae:
Looks good but there are some merge issues.

cgarciae (approved):

PR #5409, opened 2026-04-13
Fix intermediates for Gemma example

vfdev-5:


examples/gemma/train_cfg.py
```diff
@@ -0,0 +1,113 @@
...
+  # Profiling
+  with_profiler_step_trace: bool = False
+
+  # Dataflow choice: grain or TF tensor
+  input_pipeline_type: str = "grain"  # ["grain", "tf"]
+  # Flax specif configs, mostly for benchmarks
+  use_nnx_tree_mode: bool = False
+  use_nnx_transforms: str = "no"  # ["all", "no", "grad-only", "jit-only"]
+
+  sow_config: tuple[str, ...] | None = None
```
vfdev-5: Let's check if we can have `dict[str, bool|int]` and `mldict` and be able to set it from CLI

examples/gemma/sampler.py
```diff
@@ -223,7 +223,8 @@ def _sample_step(
     last_token = last_token.reshape((batch_size, 1))
 
     transformer = nnx.merge(graphdef, params)
-    logits, cache = transformer(
+    transformer = nnx.capture(transformer, nnx.Intermediate)
+    (logits, cache), intermediates = transformer(
```
cgarciae, line 227:
```suggestion
    (logits, cache), intermediates = nnx.capture(transformer, nnx.Intermediate)(
```

examples/gemma/train.py
```diff
@@ -174,23 +177,32 @@ def jax_train_step(
   def loss_fn(params, rngs):
     """loss function used for training."""
     module = nnx.merge(graphdef, params, nondiff)
+    if with_capture:
+      module = nnx.capture(module, nnx.Intermediate)
 
-    logits, _ = module(
+    output = module(
```
cgarciae, line 183:
```suggestion
    if with_capture:
      forward = nnx.capture(module, nnx.Intermediate)
    else:
      forward = module

    output = forward(
```

PR #5423, opened 2026-04-16
Move diffusion example to other examples directory

samanklesaria:
I previously had added the diffusion example to the `docs_nnx/examples` directory. It seems like, comparing to the other examples, it should really be in the main `examples` directory. This also will fix the broken link in the examples rst file.


samanklesaria:
Rethinking this approach. Closing in favor of #5424

PR #5439, opened 2026-04-24
Fix doctests for cached_partial in tree mode

samanklesaria:
This PR fixes the docstring for cached_partial so that it will not throw a doctest error when graph updates are made false by default.


IvyZX (approved):

samanklesaria:
Fixed by #5443

PR #5448, opened 2026-04-27
Use nnx.prefix rather than StateAxes in split_rngs docstring

samanklesaria:
As `StateAxes` will eventually be deprecated, this PR changes the docstring for `split_rngs` to avoid the use of `StateAxes`.


samanklesaria:
Closed in favor of #5442

PR #5458, opened 2026-05-11
Add pyrefly type checking: pyrefly.toml config, CI integration

vfdev-5:
!!! PR based on https://github.com/google/flax/pull/5460

Agent-Logs-Url: https://github.com/vfdev-5/flax/sessions/d862656c-0c03-46b4-971e-55ff771b19b9

### DIFF
--- a/flax/linen/linear.py
+++ b/flax/linen/linear.py
@@ -189,6 +189,7 @@ class DenseGeneral(Module):
     inputs, kernel, bias = self.promote_dtype(
       inputs, kernel, bias, dtype=self.dtype
     )
+    assert inputs is not None and kernel is not None
 
     if self.dot_general_cls is not None:
       dot_general = self.dot_general_cls()
--- a/pyrefly.toml
+++ b/pyrefly.toml
@@ -1,0 +1,26 @@
+# Pyrefly configuration - migrated from mypy
+# Only type-check flax/linen/linear.py for now; expand as issues are resolved.
+project-includes = ["flax/linen/linear.py"]
+
+preset = "legacy"
+ignore-missing-imports = [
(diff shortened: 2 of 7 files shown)


flax/linen/linear.py
```diff
@@ -189,6 +189,7 @@ def bias_init_wrap(rng, shape, dtype=jnp.float32):
     inputs, kernel, bias = self.promote_dtype(
       inputs, kernel, bias, dtype=self.dtype
     )
+    assert inputs is not None and kernel is not None
```
vfdev-5, line 192: This prevents pyrefly ignore comment on the code below:
```python
out = dot_general(
      inputs,  # pyrefly: ignore [bad-argument-type]
      kernel,  # pyrefly: ignore [bad-argument-type]
      ((axis, contract_ind), (batch_dims, batch_ind)),
      precision=self.precision,
    )
```
and pyrefly is right on as `self.promote_dtype` output type hint is `Array | None`, so they deduce that inputs and kernel can be also None.

pyrefly.toml
```diff
@@ -0,0 +1,26 @@
+# Pyrefly configuration - migrated from mypy
+# Only type-check flax/linen/linear.py for now; expand as issues are resolved.
+project-includes = ["flax/linen/linear.py"]
```
vfdev-5, line 3: Right now the check runs over single file to show that there is no more errors like `# pyrefly: ignore [bad-class-definition]` (as in previous attempt: https://github.com/google/flax/pull/5437/changes#diff-cfcdd6089c08daca13a27f79e07b856585c5f9b6910570f9cc3e7be76fd746b1R103)

PR #5467, opened 2026-05-21
Fix MGUCell hidden state update

### LINKED ISSUE #4608: Bug in MinimalGatedUnit
https://github.com/google/flax/blob/d59132d9b17cdf2333ef72f6d7c96a21d81b71ba/flax/linen/recurrent.py#L725

This should be `1 - f`, according to the paper. Confusion arose around the effect of the "forget" gate (in LSTM and GRU papers, information is passed through when `f` is high, but in MGU paper it is the opposite). Variable `f` from the MGU paper, is effectively `1 - f` in Flax (it is the portion that is contributes to short-term response, or `n` in Flax-speak). From the paper:

> In MGU, the forget gate f_t is first generated, and the element-wise product between 1 - f_t and h_{t−1} becomes part of the new hidden state h_t. The portion of h_{t-1} that is "forgotten" (f_t h_{t−1}) is combined with x_t to produce h_bar_t, the short-term response. A portion of h_bar_t (determined again by f_t) form the second part of h_t.

lanarkite99:
# What does this PR do?

Fixes #4608.

This fixes the `nn.MGUCell` hidden-state update equation to match the Minimal Gated Unit recurrence.

The current implementation computes:

```python
h' = (1 - f) * n + f * h
```
but according to the MGU formulation, the gate should mix the previous hidden state and candidate activation as:

```python
h' = (1 - f) * h + f * n
```
This PR updates the implementation and the displayed equation in the MGUCell docstring.

It also expands the existing MGU regression test so the cell output is checked against a manually computed recurrence for both:

```python
reset_gate=True
reset_gate=False
```

Tested with:
```python
python -m pytest tests/linen/linen_test.py -k mgu -q
```
Result:2 passed, 99 deselected

### DIFF
--- a/flax/linen/recurrent.py
+++ b/flax/linen/recurrent.py
@@ -624,7 +624,7 @@ class MGUCell(RNNCellBase):
       \begin{array}{ll}
       f = \sigma(W_{if} x + b_{if} + W_{hf} h) \\
       n = \tanh(W_{in} x + b_{in} + f * (W_{hn} h + b_{hn})) \\
-      h' = (1 - f) * n + f * h \\
+      h' = (1 - f) * h + f * n \\
       \end{array}
 
   where x is the input and h is the output of the previous time step.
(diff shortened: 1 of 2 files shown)


lanarkite99:
@cgarciae could you please take a look at this PR when you have a moment to comment or approve? 
Thanks!

lanarkite99:
@cgarciae your inputs or some directions would be highly valuable here?
Thanks!

PR #5475, opened 2026-05-29
nnx.clone: add arrays parameter to also copy underlying buffers

### LINKED ISSUE #5461: nnx.clone creates buffer copies with the same IDs, causing errors with donate_argnums
There are a number of cases where I would like to use nnx.clone to create a copy of a network (e.g., target networks in RL) inside an nnx.Module. This works fine in general, but I've found that combining this with jax.jit's donate_argnums option on the module state causes errors related to donating the same buffer twice, 

I've provided a minimal reproducible example below, with assertions showing that the ID of the nnx.clone'd buffer is the same as the original:

JAX version 0.6.2, Flax version 0.10.7

```
import functools

import jax
from flax import nnx

class TestModule(nnx.Module):
        self.linear = nnx.Linear(in_features=4, out_features=2, rngs=rngs)
        self.linear_copy = nnx.clone(self.linear)

rngs = nnx.Rngs(0)
test_module = TestModule(rngs)

linear_state = nnx.state(test_module.linear)
linear_copy_state = nnx.state(test_module.linear_copy)

online_kernel_array = linear_state.kernel.value
target_kernel_array = linear_copy_state.kernel.value

assert id(linear_state.kernel.value) == id(linear_copy_state.kernel.value)
assert online_kernel_array.unsafe_buffer_pointer() == target_kernel_array.unsafe_buffer_pointer()

# Test donate_argnums with nnx.clone'd layers.
@functools.partial(jax.jit, donate_argnums=(0,))
def update(state_pytree):
    return jax.tree.map(lambda x: x + 0.01, state_pytree)

state = nnx.state(test_module)
state = update(state)
```

This yields the error

```
jax.errors.SimplifiedTraceback: For simplicity, JAX has removed its internal frames from the traceback of the following exception. Set JAX_TRACEBACK_FILTERING=off to include these.

The above exception was the direct cause of the following exception:

  File "/Users/johnzhou/Library/Application Support/JetBrains/PyCharm2025.2/scratches/clone_test.py", line 29, in <module>
    state = update(state)
jaxlib._jax.XlaRuntimeError: INVALID_ARGUMENT: Attempt to donate the same buffer twice in Execute() (flattened argument 2, replica 0, partition 0, first use: 0). Toy example for this bug: `f(donate(a), donate(a))`.
```

Either using `copy.deepcopy` or something like

```
def clone(module: nnx.Module):
    graphdef, state = nnx.split(module)
    state_copy = jax.tree_util.tree_map(jnp.copy, state)
    return nnx.merge(graphdef, state_copy)
```

works fine. I'm not sure if this is expected behavior, but noting this behavior in the nnx.clone docs might be useful for users?

dparikh79:
# What does this PR do?

Adds `arrays=True` to `nnx.clone` per #5461. When set, the underlying `jax.Array` buffers of each `Variable` are copied so the clone has independent memory and works with `jit(donate_argnums=...)`. Default `False` so existing callers see no change. Setting `arrays=True` forces `variables=True` because a buffer copy without a fresh wrapper has nowhere to live.

Implementation: `jax.tree.map(jax.numpy.copy, state)` between `split` and `merge`. Used `jnp.copy` rather than `jnp.array` to preserve sharding on multi-device arrays.

Expanded the docstring to call out the default buffer-sharing per @samanklesaria's note that it should be better documented.

Two regression tests in `tests/nnx/module_test.py`. `test_clone_arrays_distinct_buffers` asserts `unsafe_buffer_pointer()` differs after `arrays=True` and matches for the default clone. `test_clone_arrays_donate_argnums` is @johnlyzhou's exact `donate_argnums` repro from the issue. Existing `test_clone` still passes. Full sweep: 77 passed / 1 skipped in `module_test.py`, 152 passed in `graph_utils_test.py`.

Fixes #5461

## Checklist
- [x] This change is discussed in a Github issue/[discussion](https://github.com/google/flax/discussions): #5461

### DIFF
--- a/flax/nnx/graphlib.py
+++ b/flax/nnx/graphlib.py
@@ -2878,8 +2878,20 @@ def pop(
     return states
 
 
-def clone(node: Node, variables: bool = True, *, graph: bool | None = None) -> Node:
+def clone(
+  node: Node,
+  variables: bool = True,
+  *,
+  arrays: bool = False,
+  graph: bool | None = None,
+) -> Node:
   """Create a deep copy of the given graph node.
+
+  By default the cloned :class:`Variable` wrappers are new objects but
+  share their underlying ``jax.Array`` buffers with the original. This
+  is cheap, but ``jit(donate_argnums=...)`` will reject the clone
+  because JAX refuses to donate the same buffer twice. Pass
+  ``arrays=True`` to also copy the underlying buffers.
 
   Example usage::
 
@@ -2894,6 +2906,11 @@ def clone(node: Node, variables: bool = True, *, graph: bool | None = None) -> N
     node: A graph node object.
     variables: If ``True`` (default) copies of the :class:`Variable` objects are created,
       otherwise the Variables are shared between the original and cloned node.
+    arrays: If ``True``, the underlying ``jax.Array`` buffers of each
+      :class:`Variable` are also copied so the clone has independent
+      memory and is compatible with ``jit(donate_argnums=...)``. If True,
+      ``variables`` should be also ``True``, otherwise a ValueError is raised.
+      Defaults to ``False`` (buffers shared with the original).
     graph: If ``True`` (default), uses graph-mode which supports the full
       NNX feature set including shared references. If ``False``, uses
       tree-mode which treats Modules as regular JAX pytrees, avoiding
@@ -2902,6 +2919,13 @@ def clone(node: Node, variables: bool = True, *, graph: bool | None = None) -> N
     A deep copy of the :class:`Module` object.
   """
   graphdef, state = split(node, graph=graph)
+  if arrays:
+    if not variables:
+      raise ValueError("If arrays is True, variables should be also True")
+    state = jax.tree.map(jax.numpy.copy, state)
+    # Optimization: as state is already a copy,
+    # set variables to False to avoid another copy
+    variables = False
   return merge(graphdef, state, copy=variables)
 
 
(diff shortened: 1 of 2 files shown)


samanklesaria (approved):
Looks good to me!

flax/nnx/graphlib.py
```diff
@@ -2902,6 +2919,9 @@ def clone(node: Node, variables: bool = True, *, graph: bool | None = None) -> N
     A deep copy of the :class:`Module` object.
   """
   graphdef, state = split(node, graph=graph)
+  if arrays:
```
cgarciae, line 2922: we should raise a ValueError if `arrays=True` and `variables=False` as this implementation currently does not support it.

flax/nnx/graphlib.py
```diff
@@ -2902,6 +2919,9 @@ def clone(node: Node, variables: bool = True, *, graph: bool | None = None) -> N
...
+    state = jax.tree.map(jax.numpy.copy, state)
+    variables = True
```
cgarciae: since the `jax.tree.map` effectively creates copies of all the Variables then `you could actually set:
```suggestion
    variables = False
```

cgarciae (changes requested):

vfdev-5:
@dparikh79 please update the PR according to the suggestions. Thanks!

cgarciae (approved):

PR #5481, opened 2026-06-02
Add RoPE embeddings

samanklesaria:
This PR adds support for RoPE. Specifically, a new function `dot_product_attention_with_rope` can be used as the `attention_fn` argument for `nnx.MultiHeadAttention`.

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -332,6 +332,144 @@ def dot_product_attention(
   return out
 
 
+class RoPE(Module):
+  """Rotary Position Embedding (RoPE).
+
+  Precomputes and stores cosine and sine frequency tensors for rotary
+  positional encoding as described in "RoFormer: Enhanced Transformer
+  with Rotary Position Embedding" (https://arxiv.org/abs/2104.09864).
+
+  Example usage::
+
+    >>> from flax import nnx
+    >>> import functools
+    >>> rope = nnx.RoPE(embedding_size=64, max_seq_len=128)
+    >>> layer = nnx.MultiHeadAttention(
+    ...     num_heads=8, in_features=512, qkv_features=512,
+    ...     attention_fn=functools.partial(
+    ...         nnx.dot_product_attention_with_rope, rope=rope),
+    ...     decode=False, rngs=nnx.Rngs(0))
+
+  Args:
+    theta: Base frequency for sinusoidal functions. Default is 10000.
+    embedding_size: Optional. Size of each embedding vector (i.e. head_dim). Must be
+      even. If provided together with ``max_seq_len``, sin/cos tables are
+      precomputed and cached as module Variables.
+    max_seq_len: Optional. Maximum sequence length for precomputed frequencies. Must
+      be provided together with ``embedding_size``.
+  """
+
+  def __init__(
+    self,
+    theta: float = 10000.0,
+    embedding_size: int | None = None,
+    max_seq_len: int | None = None,
+  ):
+    self.theta = theta
+    self.cached_sin: nnx.Variable | None
+    self.cached_cos: nnx.Variable | None
+    if (embedding_size is None) != (max_seq_len is None):
+      raise ValueError('Either both `embedding_size` and `max_seq_len` or none of them must be provided.')
+    if embedding_size is not None and max_seq_len is not None:
+      positions = jnp.arange(float(max_seq_len))
+      sin, cos = self._rotation_for(embedding_size, positions)
+      self.cached_sin = nnx.Variable(sin)
+      self.cached_cos = nnx.Variable(cos)
+    else:
+      self.cached_sin = None
+      self.cached_cos = None
+
+  @staticmethod
+  def _rotate_half(x: Array) -> Array:
+    d_2 = x.shape[-1] // 2
+    return jnp.concatenate([-x[..., d_2:], x[..., :d_2]], axis=-1)
+
+  def _rotation_for(self, embedding_size: int, input_positions: Array):
+    seq_len = input_positions.shape[-1]
+    if embedding_size <= 0 or embedding_size % 2 != 0:
+      raise ValueError('`embedding_size` must be positive and even.')
+    freqs = 1.0 / (
+      self.theta ** (jnp.arange(0.0, embedding_size, 2) / embedding_size)
+    )
+    freqs_outer = jnp.outer(input_positions, freqs)
+    return jnp.sin(freqs_outer), jnp.cos(freqs_outer)
+
+  def __call__(self, x: Array, input_positions: Array | None = None) -> Array:
+    """Apply rotary positional encoding.
+
+    Args:
+      x: Input array of shape ``(..., seq_length, embedding_size)``.
+      input_positions: Optional position array of shape ``(..., seq_len,)``.
+        If ``None`` and precomputed tables exist (i.e. ``embedding_size``
+        and ``max_seq_len`` were given at init), the cached tables are
+        used. Otherwise positions default to ``jnp.arange(seq_len)``.
+
+    Returns:
+      Array of same shape with RoPE applied.
+    """
+    seq_len, embedding_size = x.shape[-2:]
+    if input_positions is None and self.cached_sin is not None and self.cached_cos is not None:
+      sin = self.cached_sin.value[:seq_len]
+      cos = self.cached_cos.value[:seq_len]
+    else:
+      if input_positions is None:
+        input_positions = jnp.arange(float(seq_len))
+      elif input_positions.shape[-1] != seq_len:
+        raise ValueError('`input_positions` must have the same seq_len as `x`.')
+      sin, cos = self._rotation_for(embedding_size, input_positions)
+    cos = jnp.concatenate([cos, cos], axis=-1)
+    sin = jnp.concatenate([sin, sin], axis=-1)
+    # Broadcast sin/cos to match any leading batch dims of x.
+    n_prefix = x.ndim - cos.ndim
+    if n_prefix > 0:
+      cos = jnp.reshape(cos, (1,) * n_prefix + cos.shape)
+      sin = jnp.reshape(sin, (1,) * n_prefix + sin.shape)
+    return x * cos + self._rotate_half(x) * sin
+
+
+def dot_product_attention_with_rope(
+  query: Array,
+  key: Array,
+  value: Array,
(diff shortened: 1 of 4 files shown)


flax/nnx/nn/attention.py
```diff
@@ -319,6 +319,106 @@ def reshape_4d(x):
   return out
 
 
+class RoPE(Module):
```
vfdev-5, line 335: Please, check if we should also expose this in the docs, API refs etc.

flax/nnx/nn/attention.py
```diff
@@ -319,6 +319,106 @@ def reshape_4d(x):
...
+      raise ValueError('`embedding_size` must be positive and even.')
+    self.embedding_size = embedding_size
+    self.theta = theta
+    freqs = 1.0 / (
+      theta ** (jnp.arange(0.0, embedding_size, 2) / embedding_size)
+    )
+    t = jnp.arange(float(max_seq_len))
+    freqs_outer = jnp.outer(t, freqs)
+    self.cos_cached = nnx.Variable(jnp.cos(freqs_outer).astype(dtype))
+    self.sin_cached = nnx.Variable(jnp.sin(freqs_outer).astype(dtype))
```
vfdev-5: In multiple JAX repositories I saw that sin and cos are constructed from `segment_pos` (absolute token positions):
- our [gemma example](https://github.com/google/flax/blob/0f128b5141a7ab40df430d280bc86d3929ccb1ce/examples/gemma/positional_embeddings.py#L45)
- bonsai and [jax-llm-examples](https://github.com/tachikaw/jax-llm-examples/blob/a2212758065c0dc7af34d8fb7b97f7fb0e4b4f9c/llama3/llama3_jax/model.py#L984)
- [gemma4](https://github.com/google-deepmind/gemma/blob/5621d2dbc89b306b53f65aee455f9054e97f184d/gemma/gm/nn/gemma4/_transformer.py#L449-L452)

The idea is that if the input sequence is not packed, but just padded, our implementation would mostly work as expected. In case of packed sequence where multiple sentences are inserted in the same sequence:
```
[<bos>, 1, 2, 3, <eos>, <bos>, 4, 5, <eos>, <pad>, <pad>]
```
then the input positions (`segment_pos`) would be:
```
[0, 1, 2, 3, 4, 0, 1, 2, 3, 0, 0]
```
so, RoPE may be computed differently.
On the other hand, current `MHA.__call__` does not accept any `positions` arg, so we can't pass it to RoPE...


In PyTorch, [basic implementation](https://colab.research.google.com/drive/11SKfzvMotuvvXNqY9qBpsD2RQX1PK7rP?usp=sharing) does something similar to your implementation, but cos and sin cached from the input `x`.

samanklesaria: My implementation was based on the one in equinox, which doesn't pass explicit positions. 

- We could give `MHA` a new optional `positions` argument which we thread through to the attention_fn. But this could break users' custom attention_fn implementations if they weren't expecting the argument. 
- We could make a MHA subclass with a call method that accepts a `positions` argument. Say, `PackedMHA`? This is a little more ugly, but wouldn't be a breaking change.

vfdev-5: Can we do the following?
```python
class MHA:

  def __call__(self, ..., input_positions: Array | None = None):
    ...
    attn_kwargs = {}
    if input_positions is not None:
       attn_kwargs["input_positions"] = input_positions
    x = self.attention_fn(
      query,
      key,
      value,
      mask=mask,
      dropout_rng=dropout_rng,
      dropout_rate=self.dropout_rate,
      broadcast_dropout=self.broadcast_dropout,
      deterministic=deterministic,
      dtype=self.dtype,
      precision=self.precision,
      module=self if sow_weights else None,
      is_causal=is_causal,
      **attn_kwargs,
    )


def dot_product_attention_with_rope(..., rope, input_positions: Array | None = None, **kwargs)
  # handle properly input_positions is None
  # input_positions: (B, S)
  apply = jax.vmap(rope, in_axes=(-2, -1), out_axes=(-2, -1))
  query = apply(query, input_positions)
  key = apply(key, input_positions)
  ...
```

samanklesaria: I like that! So users that aren't already calling MHA with input positions won't have their custom attention_fn break!

samanklesaria: Should we still keep the cached `sin` and `cos` vectors for use when `input_positions=None`?  Might be slightly faster for the non-packed case, as we wouldn't need to rebuild them. But the interface could be nicer if we just rebuilt them every time, so that the `RoPE` constructor wouldn't need `max_seq_len` and `embedding_size` arguments (dynamically getting them from the input `x`). What do you think?

vfdev-5: We can cache them after the first call ?

samanklesaria: So we'd check if the `input_positions` is the same as the cached one at each call. If so, we'd use the cache (populated on the first call) and otherwise, we'd generate it dynamically? 

To work in tree mode, we'd need to create the Variables for the cache at initialization. But then we could write to these variables on the first `__call__`.

samanklesaria: If we're doing cross attention, things get trickier. The sequence lengths for the keys might not be the same as the lengths for the values. Caching a single value wouldn't account for this. Caching everything would break if the packing of each batch might be different.

vfdev-5: Good point, actually, not sure if seen RoPE for cross-attention, but it makes sense that we may need to have q_positions and kv_positions or even k_positions, v_positions. But it can become rather cumbersome the API finally

samanklesaria: I ended up at a compromise API. If the user wants caching, they can specify the max size during initialization, like I had before. Otherwise, we don't cache, and compute the rotation matrices on the fly. 

This also means we don't have to use QDD for the cache, which might become deprecated if we end up switching to hijax variables and QDD is removed. One less thing to worry about down the road.

flax/nnx/nn/attention.py
```diff
@@ -319,6 +319,128 @@ def reshape_4d(x):
...
+  """
+
+  def __init__(
+    self,
+    theta: float = 10000.0,
+    embedding_size: int | None = None,
+    max_seq_len: int | None = None,
+  ):
+    self.theta = theta
+    if sum([embedding_size is None, max_seq_len is None]) % 2 != 0:
```
vfdev-5: This works and nicely short but a bit less readable.
Here is another one liner for that (neither easily readable):
```python
if (embedding_size is None) != (max_seq_len is None):
  raise ValueError('Either both `embedding_size` and `max_seq_len` or none of them must be provided.')
```

flax/nnx/nn/attention.py
```diff
@@ -319,6 +319,128 @@ def reshape_4d(x):
...
+      raise ValueError('Either both `embedding_size` and `max_seq_len` or none of them must be provided.')
+    if embedding_size is not None and max_seq_len is not None:
+      positions = jnp.arange(float(max_seq_len))
+      sin, cos = self._rotation_for(embedding_size, positions)
+      self.cached_sin = nnx.Variable(sin)
+      self.cached_cos = nnx.Variable(cos)
```
vfdev-5, line 377: Should we have self.cached_sin and cos defined as None for else case, so do not do `hasattr(self, 'cached_sin')`

samanklesaria: I guess then we'd check for whether `self.cached_sin` is `None` in that case. Seems like it's equally complicated either way. Any arguments in favor of the `None` version?

flax/nnx/nn/attention.py
samanklesaria: Yours is more readable. I'll change it.

tests/nnx/nn/attention_test.py
```diff
@@ -383,7 +383,41 @@ def _run(m):
     np.testing.assert_allclose(nnx_out, jax_out, atol=1e-3, rtol=1e-3)
 
 
-if __name__ == '__main__':
-  absltest.main()
+class TestRoPE(absltest.TestCase):
+
```
IvyZX, line 408: Can you add a test on the RoPE calculation itself? like the output of a `RoPE()` module vs. the handwritten computation of RoPE.

tests/nnx/nn/attention_test.py
```diff
@@ -383,7 +383,41 @@ def _run(m):
...
+    q_rot = rope_at(q_vec, i)
+    k_rot = rope_at(k_vec, j)
+    dot_original = jnp.dot(q_rot, k_rot)
+
+    q_rot_shifted = rope_at(q_vec, i + d)
+    k_rot_shifted = rope_at(k_vec, j + d)
+    dot_shifted = jnp.dot(q_rot_shifted, k_rot_shifted)
+
+    np.testing.assert_allclose(dot_original, dot_shifted, atol=1e-5)
+
```
IvyZX, line 549: Can you also make a test on `dot_product_attention_with_rope`?

tests/nnx/nn/attention_test.py
samanklesaria: What if we tested against equinox's RoPE?

vfdev-5: This would bring another testing dependency, I would avoid. Check if existing deps like keras can provide RoPE....

samanklesaria: Fair enough. If we already have massing dependencies like tensorflow and keras, equinox seems pretty small. But Keras seems to have a RoPE implementation. It won't be quite as natural a comparison, but that seems like the best path forward.

samanklesaria: Hm- actually, it seems like https://keras.io/keras_hub/api/modeling_layers/rotary_embedding/ is part of keras_hub, not raw keras. But torch.onnx.ops.rotary_embedding exists, and pytorch is already a dependency. Let's use that.

flax/nnx/nn/attention.py
vfdev-5: >  Any arguments in favor of the None version?

I prefer checking against None vs if an object has a attribute which should be a bit more expensive to check in a dictionary.

vfdev-5:
@samanklesaria can you fix the merging conflict, thanks!

vfdev-5:
@samanklesaria can you check mypy CI job if it is real? Thanks!

IvyZX (approved):

PR #5490, opened 2026-06-09
Handle capture variables in `nnx.jit`. 

samanklesaria:
This PR is the first in a series that splits #5485 into separate pieces for each transform.

### DIFF
--- a/flax/nnx/extract.py
+++ b/flax/nnx/extract.py
@@ -17,6 +17,7 @@ from collections import namedtuple
 import dataclasses
 import functools
 import inspect
+import types
 import typing as tp
 
 from flax import struct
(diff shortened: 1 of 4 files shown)


flax/nnx/transforms/compilation.py
samanklesaria, line 527: Make the captured pieces a separate kwarg to`labeled`.

flax/nnx/transforms/compilation.py
```diff
@@ -464,7 +465,14 @@ def __call__(self, *args, **kwargs):
     )
     if self.graph:
       args, kwargs = extract.from_tree2((args, kwargs))
-    out = self.f(*args, **kwargs)
+    n_captured = len(self.captured_info)
+    user_args = args[n_captured:]
+    if self.captured_info:
```
samanklesaria: Always pass. Don't condition.

flax/nnx/extract.py
```diff
@@ -1119,6 +1121,158 @@ def to_masked(tree, all_updates: OrderedDict[tp.Any, Updates]):
...
+    return functools.partial(new_func, *f.args, **f.keywords)
+
+  changed = False
+  cells: list | None = None
+
+  # Replace captured Variables in this function's closure.
+  if hasattr(f, '__closure__') and f.__closure__ is not None:
+    cells = list(f.__closure__)
+    for i, cell in enumerate(cells):
+      try:
```
IvyZX, line 1197: Can we reduce the try-catch antipatterns with explicit check?

flax/nnx/extract.py
```diff
@@ -1119,6 +1121,158 @@ def to_masked(tree, all_updates: OrderedDict[tp.Any, Updates]):
...
+  # Recurse into __wrapped__ and update the cell referencing the old inner.
+  if hasattr(f, '__wrapped__'):
+    old_inner = f.__wrapped__
+    new_inner = _replace_closure_cells(old_inner, id_to_replacement)
+    if new_inner is not old_inner:
+      if cells is None and hasattr(f, '__closure__') and f.__closure__ is not None:
+        cells = list(f.__closure__)
+      if cells is not None:
+        for i, cell in enumerate(cells):
+          try:
```
IvyZX, line 1218: Also here

flax/nnx/transforms/compilation.py
```diff
@@ -426,31 +426,62 @@ class SimpleJitFn:
   donate_argnames: frozenset[str]
   graph: bool
   update_shardings: tuple[tp.Any, ...]
+  captured_info: tp.Optional[list[object]] = None
```
cgarciae:
```suggestion
  captured_info: list[tp.Any] = dataclasses.field(init=False)
```

samanklesaria: This doesn't work because, on line 617, we actually want to pass in the argument. Actually, there's no reason to have a default here. I'll just remove the `None`.

Issue #5483, opened 2026-06-04
`metrics` does not work well with `vmap`


narang99:
flax.metrics.Average does not work well with vmap.

consider code like this:
```python
@nnx.vmap(in_axes=(0, None, None, None, None, None), out_axes=0)
def parallel_init_models(
    seeds, input_dim, n_components, scaled_hyperparameters, init_strategy, lr
):
    return init_single_model(
        seeds, input_dim, n_components, scaled_hyperparameters, init_strategy, lr
    )


def init_single_model(
    s, input_dim, n_components, scaled_hyperparameters, init_strategy, lr
):
    rngs = nnx.Rngs(s)
    model = Autoencoder(input_dim, n_components, rngs)
    # Initialize model parameters using shared hyperparameters
    optimizer = nnx.Optimizer(model, optax.adam(lr), wrt=nnx.Param)
    metrics = nnx.MultiMetric(
        loss=nnx.metrics.Average("loss"),
        recon_loss=nnx.metrics.Average("recon_loss"),
        weight_loss=nnx.metrics.Average("weight_loss"),
        codes_loss=nnx.metrics.Average("codes_loss"),
        mse=nnx.metrics.Average("mse"),
    )
    return model, optimizer, metrics
```

`parallel_init_seeds` is used to create multiple models, which helps me train different seeds for the same input.  
The shapes in the beginning are good for metrics, models and optimizers.

Running `metrics.reset` breaks the metrics though.
Before reset
```python
MultiMetric( # MetricState: 30 (120 B)
  codes_loss=Average( # MetricState: 6 (24 B)
    argname='codes_loss',
    count=MetricState( # 3 (12 B)
      value=Array(shape=(3,), dtype=dtype('int32'))
    ),
    total=MetricState( # 3 (12 B)
      value=Array(shape=(3,), dtype=dtype('float32'))
    )
  )
...
```

After calling `metrics.reset()`, it becomes
```python
MultiMetric( # MetricState: 10 (40 B)
  codes_loss=Average( # MetricState: 2 (8 B)
    argname='codes_loss',
    count=MetricState( # 1 (4 B)
      value=Array(0, dtype=int32)
    ),
    total=MetricState( # 1 (4 B)
      value=Array(0., dtype=float32)
    )
  ),
  ...
```

Note the shape difference. The code also seems to assign it to a static 0 array https://github.com/google/flax/blob/main/flax/nnx/training/metrics.py#L97

[...]

PR #5491, opened 2026-06-11
Add vmap reset() regression test for nnx.metrics (#5483)

chenkuanliao:
# What does this PR do?

This adds a regression test for the `nnx.metrics` behavior reported in #5483.

`nnx.metrics.Average.reset()` (and `Welford.reset()`) re-zero their state by assigning a scalar (`jnp.array(0, ...)`). When a metric is built under `nnx.vmap`, its state carries a leading batch axis of shape `(N,)`. On the reporter's versions (flax 0.12.0 / jax 0.7.2) the scalar assignment replaced the whole array, collapsing `(N,)` to `()`, so a later vmapped `update` failed with:

    ValueError: vmap was requested to map its argument along axis 0, which implies that its rank should be at least 1, but is only 0 (its shape is ())

While investigating I found the crash **no longer reproduces on `main`**: the `Variable` state layer has since changed so a full-index assignment broadcasts a scalar into the existing array in place instead of replacing it, so `reset()` preserves the batch axis today. I confirmed it still reproduces on flax 0.12.0 / jax 0.7.2 and is gone on `main` independent of any single commit.

Since there is no longer a live crash to fix on `main`, this PR is scoped as a **regression test only**, to lock in the current vmap `reset()` behavior so it can't silently break again.

## What's in this PR

- `tests/nnx/metrics_test.py` — `test_vmap_reset_preserves_shape`: construct a `MultiMetric(loss=Average('loss'))` under `nnx.vmap` (state shape `(N,)`), call `reset()`, assert the state stays shape `(N,)`, then run a vmapped `update` and assert `compute()['loss']` returns the expected per-batch values.

## Checklist
- [x] This change is discussed in a Github issue/[discussion](https://github.com/google/flax/discussions) (please add a link). — https://github.com/google/flax/issues/5483

### DIFF
--- a/tests/nnx/metrics_test.py
+++ b/tests/nnx/metrics_test.py
@@ -265,6 +265,29 @@ class TestMetrics(parameterized.TestCase):
       )
     )
 
+  def test_vmap_reset_preserves_shape(self):
+    n = 3
+
+    @nnx.vmap(in_axes=0, out_axes=0)
+    def init_metrics(seed):
+      del seed
+      return nnx.MultiMetric(loss=nnx.metrics.Average('loss'))
+
+    metrics = init_metrics(jnp.arange(n))
+    self.assertEqual(metrics.loss.total[...].shape, (n,))
+
+    metrics.reset()
+    self.assertEqual(metrics.loss.total[...].shape, (n,))
+
+    @nnx.vmap(in_axes=(0, 0), out_axes=None)
+    def do_update(m, value):
+      m.update(loss=value)
+
+    do_update(metrics, jnp.arange(n, dtype=jnp.float32))
+    np.testing.assert_array_equal(
+      metrics.compute()['loss'], jnp.arange(n, dtype=jnp.float32)
+    )
+
 
 if __name__ == '__main__':
   absltest.main()
--- a/tests/nnx/metrics_test.py
+++ b/tests/nnx/metrics_test.py
@@ -269,8 +269,7 @@ class TestMetrics(parameterized.TestCase):
     n = 3
 
     @nnx.vmap(in_axes=0, out_axes=0)
-    def init_metrics(seed):
-      del seed
+    def init_metrics(_):
       return nnx.MultiMetric(loss=nnx.metrics.Average('loss'))
 
     metrics = init_metrics(jnp.arange(n))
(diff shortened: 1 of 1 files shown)


tests/nnx/metrics_test.py
```diff
@@ -265,6 +265,29 @@ def test_average(self, with_mask, scalar_values):
       )
     )
 
+  def test_vmap_reset_preserves_shape(self):
+    n = 3
+
+    @nnx.vmap(in_axes=0, out_axes=0)
+    def init_metrics(seed):
+      del seed
```
samanklesaria: Why do we need to `del seed`?

tests/nnx/metrics_test.py
```diff
@@ -265,6 +265,29 @@ def test_average(self, with_mask, scalar_values):
...
+      return nnx.MultiMetric(loss=nnx.metrics.Average('loss'))
+
+    metrics = init_metrics(jnp.arange(n))
+    self.assertEqual(metrics.loss.total[...].shape, (n,))
+
+    metrics.reset()
+    self.assertEqual(metrics.loss.total[...].shape, (n,))
+
+    @nnx.vmap(in_axes=(0, 0), out_axes=None)
```
samanklesaria: This next part doesn't seem necessary for testing the shape reset issue. I'd leave it out.

tests/nnx/metrics_test.py
chenkuanliao: The argument was only there to provide a mapped axis for `nnx.vmap`; it is not used by the metric initialization. I have replace `seed` with `_` and remove `del seed` to make that clearer, addressed in `abc2dbe`. Thanks for pointing this out!

tests/nnx/metrics_test.py
chenkuanliao: got it! Since the issue is specifically that `reset()` collapses the vmapped state shape, so the shape assertion after `reset()` should be enough for this regression test. I have remove the update/compute part to keep the test focused, addressed in `abc2dbe`

samanklesaria (approved):
Looks good to me! I'll get some other eyes on it, and then we should be good!

IvyZX (approved):

PR #5503, opened 2026-06-16
Validate dropout_rate in dot_product_attention

### LINKED ISSUE #5497: flax.nnx.dot_product_attention does not validate dropout_rate (negative/NaN silently disables dropout)
### System information
- Python version: Python 3.11

I found this through fuzzing-based testing. In a CPU-only setup, `flax.nnx.nn.attention.dot_product_attention` never validates `dropout_rate`. A negative or NaN `dropout_rate` is silently accepted and silently disables dropout, because the dropout block is guarded by `if not deterministic and dropout_rate > 0.0` and the comparison is False for negative/NaN, so the requested regularization simply is not applied. Separately, a positive `dropout_rate` without a `dropout_rng` raises `TypeError: unexpected PRNG key type <class 'NoneType'>` from inside `jax.random.bernoulli`, instead of telling the user that `dropout_rng` is required.

`dropout_rate` outside `[0, 1)` (negative, `>= 1`, or NaN) raises a clear `ValueError` naming the value, and a positive `dropout_rate` with `deterministic=False` and no `dropout_rng` raises a clear "dropout_rng is required" error, rather than silently disabling dropout or surfacing a PRNG-internal `TypeError`. (`torch.nn.Dropout` has validated this range since 0.3.1.)

`dropout_rate=-0.5` or `nan` -> no error, dropout silently skipped
`dropout_rate=0.5` without `dropout_rng` -> `TypeError: unexpected PRNG key type <class 'NoneType'>`

### Steps to reproduce:

```python
import jax
import flax.nnx as nnx

q = jax.random.normal(jax.random.PRNGKey(0), (2, 16, 4, 16)); k = q; v = q
drng = jax.random.PRNGKey(1)

attn(q, k, v, dropout_rate=-0.5, deterministic=False, dropout_rng=drng)  # no error, dropout silently off
attn(q, k, v, dropout_rate=0.5)                                          # TypeError: unexpected PRNG key type <class 'NoneType'>
```

mohsinm-dev:
Fixes #5497

Reject `dropout_rate` outside `[0, 1)` (including NaN/inf) with a clear `ValueError`, and raise early when `dropout_rng` is missing for active dropout.


tests/nnx/nn/attention_test.py
```diff
@@ -383,6 +383,44 @@ def _run(m):
     np.testing.assert_allclose(nnx_out, jax_out, atol=1e-3, rtol=1e-3)
 
 
+class TestDropoutRateValidation(parameterized.TestCase):
```
vfdev-5, line 386: Reduce the number of tests to essential only and use parameterize as much as you can to reduce the number of lines

samanklesaria (approved):
Looks good to me!

IvyZX (approved):

PR #5505, opened 2026-06-16
replace bare asserts with ValueError in NNX attention validation

### LINKED ISSUE #5496: flax.nnx.dot_product_attention silently returns wrong-shaped output under python -O (rank asserts dropped)
### System information
- Python version: Python 3.11

I found this through fuzzing-based testing. In a CPU-only setup, `flax.nnx.nn.attention.dot_product_attention` guards its q/k/v rank and shape invariants with bare `assert` statements. Python drops `assert` under `-O` (or `PYTHONOPTIMIZE=1`), which is common in production serving, and on the non-fast path (when `dropout_rate > 0` or a `module` is provided, so the code routes through `dot_product_attention_weights` instead of `jax.nn.dot_product_attention`) the broadcasting einsum then silently returns a wrong-shaped result instead of raising. With a rank mismatch on the dropout path I get `AssertionError` normally and a silent `(2, 16, 4, 8)` output under `-O`. The fast path (`dropout_rate == 0`, `module is None`) is backstopped by `jax.nn.dot_product_attention`'s own checks, so the silent case is specifically the dropout / `module` path used during training.

The rank/shape invariants enforced with `if not (...): raise ValueError(...)` so they hold under `-O`, giving the same clear error with and without the flag, instead of a silently wrong-shaped attention output.

`python    repro.py` -> `AssertionError: q, k, v must have same rank.`
`python -O repro.py` -> no error, returns shape `(2, 16, 4, 8)`

### Steps to reproduce:

```python
import jax
import flax.nnx as nnx

q = jax.random.normal(jax.random.PRNGKey(0), (2, 16, 4, 8))  # rank 4
k = jax.random.normal(jax.random.PRNGKey(0), (16, 4, 8))     # rank 3 (mismatch)
v = jax.random.normal(jax.random.PRNGKey(0), (16, 4, 8))     # rank 3 (mismatch)
out = attn(q, k, v, dropout_rate=0.1, dropout_rng=jax.random.PRNGKey(1))
print("result shape:", out.shape)
# python    -> AssertionError: q, k, v must have same rank.
# python -O -> result shape: (2, 16, 4, 8)   (silently wrong, no error)
```

mohsinm-dev:
Fixes #5496

`dot_product_attention`, `dot_product_attention_weights`, and `combine_masks` used bare `assert` for input shape validation. Under `python -O`, these are stripped and invalid inputs silently produce wrong-shaped output on the dropout/module path.

Replaced with `if/raise ValueError`. Also removed one dead-code assert in the non-GQA else branch where the condition was already guaranteed by the enclosing if/else.

### DIFF
--- a/flax/nnx/nn/attention.py
+++ b/flax/nnx/nn/attention.py
@@ -49,6 +49,17 @@ from flax.typing import (
 Array = jax.Array
 
 
+def _validate_qkv_shapes(query, key, value=None):
+  arrays = (query, key, value) if value is not None else (query, key)
+  names = 'q, k, v' if value is not None else 'q, k'
+  if not all(a.ndim == arrays[0].ndim for a in arrays[1:]):
+    ndims = ', '.join(str(a.ndim) for a in arrays)
+    raise ValueError(f'{names} must have same rank, but got {ndims}.')
+  if not all(a.shape[:-3] == arrays[0].shape[:-3] for a in arrays[1:]):
+    shapes = ', '.join(str(a.shape) for a in arrays)
+    raise ValueError(f'{names} batch dims must match, but got shapes {shapes}.')
+
+
 def dot_product_attention_weights(
   query: Array,
   key: Array,
@@ -107,9 +118,11 @@ def dot_product_attention_weights(
   query, key = promote_dtype((query, key), dtype=dtype)  # type: ignore[bad-unpacking]
   dtype = query.dtype
 
-  assert query.ndim == key.ndim, 'q, k must have same rank.'
-  assert query.shape[:-3] == key.shape[:-3], 'q, k batch dims must match.'
-  assert query.shape[-1] == key.shape[-1], 'q, k depths must match.'
+  _validate_qkv_shapes(query, key)
+  if query.shape[-1] != key.shape[-1]:
+    raise ValueError(
+      f'q, k depths must match, but got {query.shape[-1]} and {key.shape[-1]}.'
+    )
 
   # check if we need to broadcast Key heads to match Query heads
   is_gqa = False
@@ -135,7 +148,6 @@ def dot_product_attention_weights(
   else:
     q_heads = query.shape[-2]
     einsum_str = '...qhd,...khd->...hqk'
-    assert query.shape[-2] == key.shape[-2], 'q, k num_heads must match.'
 
   # calculate attention matrix
   depth = query.shape[-1]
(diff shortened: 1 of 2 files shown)


tests/nnx/nn/attention_test.py
```diff
@@ -383,6 +383,51 @@ def _run(m):
     np.testing.assert_allclose(nnx_out, jax_out, atol=1e-3, rtol=1e-3)
 
 
+class TestDotProductAttentionValidation(parameterized.TestCase):
```
vfdev-5, line 386: Let's make this test parameterized with just two tests methods and all these options as parameters

flax/nnx/nn/attention.py
```diff
@@ -255,11 +263,21 @@ def dot_product_attention(
   query, key, value = promote_dtype((query, key, value), dtype=dtype)  # type: ignore[bad-unpacking]
   dtype = query.dtype
 
-  assert key.ndim == query.ndim == value.ndim, 'q, k, v must have same rank.'
-  assert (
-    query.shape[:-3] == key.shape[:-3] == value.shape[:-3]
-  ), 'q, k, v batch dims must match.'
-  assert key.shape[-3] == value.shape[-3], 'k, v lengths must match.'
+  if not (query.ndim == key.ndim == value.ndim):
```
vfdev-5: If we can refactor these a bit different but looking redundant checks it would be great. Not all, but at least 2 first with value as optional arg for the checks.

vfdev-5 (commented):
Thanks @mohsinm-dev , I left few comments

tests/nnx/nn/attention_test.py
```diff
@@ -383,6 +383,31 @@ def _run(m):
...
+      nnx.dot_product_attention(q, k, v, dropout_rate=0.1,
+                                dropout_rng=jax.random.key(0))
+
+  def test_valid_dropout_path(self):
+    q = jax.random.normal(jax.random.key(0), (2, 16, 4, 8))
+    k = jax.random.normal(jax.random.key(1), (2, 16, 4, 8))
+    v = jax.random.normal(jax.random.key(2), (2, 16, 4, 8))
+    out = nnx.dot_product_attention(q, k, v, dropout_rate=0.1,
+                                    dropout_rng=jax.random.key(3))
+    assert out.shape == q.shape
```
vfdev-5: If you could make all checks in a single test, let's remove this test as it does not bring any value, IMO.

mohsinm-dev: Make sense, Removed it.

vfdev-5 (approved):
LGTM

PR #5518, opened 2026-06-27
Fix nnx.cond tracing cache misses by caching SimpleCondFn wrappers

### LINKED ISSUE #5512: flax.nnx.cond causes tracing cache misses
Using `nnx.cond` causes tracing cache misses (and thus prevents persistent compilation cache loads) due to creation of transient SimpleCondFn passed to `jax.lax.cond`. This is very inconvenient when using nnx.cond with larger modules that take a while to compile.

```
TRACING CACHE MISS at .../flax/nnx/transforms/transforms.py:623:19 (cond):
  never seen function:
    true_fn id=17864732047504 defined at .../repro_cond_miss.py:15
  but seen another function defined on the same line; maybe the function is
  being re-defined repeatedly, preventing caching?
W0622 10:13:43.770388 1182679 partial_eval.py:2164] TRACING CACHE MISS at .../flax/nnx/transforms/transforms.py:623:19 (cond):
  never seen function:
    false_fn id=17864737580928 defined at .../repro_cond_miss.py:19
  but seen another function defined on the same line; maybe the function is
  being re-defined repeatedly, preventing caching?
W0622 10:13:43.771588 1182679 dispatch.py:204] Finished tracing cond for jit in 0.000418901 sec
```

Here's a repro code that prints the cache misses (including the jax-level prints)

```python
import weakref
from absl import app
from flax import nnx
from flax.nnx.transforms import transforms as nnx_transforms
import jax

jax.config.update("jax_log_compiles", True)
jax.config.update("jax_explain_cache_misses", True)


def true_fn(x):
  return x + 1.0


def false_fn(x):
  return x - 1.0


original_SimpleCondFn = nnx_transforms.SimpleCondFn


def run_test(name, patch_fn):
  print(f"\n--- Running Test: {name} ---", flush=True)

  patch_fn()

  def test_fn(pred, x):
    # Pass graph_updates=False to force the path that uses SimpleCondFn
    return nnx.cond(pred, true_fn, false_fn, x, graph_updates=False)

  for i in range(3):
    print(f"\nIteration {i}:", flush=True)
    pred = jax.numpy.array(i % 2 == 0)
    res = test_fn(pred, 1.0)
    jax.block_until_ready(res)


# Buggy implementation (original behavior, but with prints)
def apply_buggy_patch():
  # We want to intercept calls to SimpleCondFn to print when they are created
  def buggy_SimpleCondFn_factory(f, graph):
    instance = original_SimpleCondFn(f, graph)
    print(
        f"  [BUGGY] Created new wrapper for {f.__name__} (id={id(f)}):"
        f" wrapper_id={id(instance)}",
        flush=True,
    )
    return instance

  nnx_transforms.SimpleCondFn = buggy_SimpleCondFn_factory


# Patched implementation (using cache)
_repro_cache = weakref.WeakKeyDictionary()


def apply_patched_patch():
  _repro_cache.clear()  # Clear cache for clean run

  def patched_SimpleCondFn_factory(f, graph):
    if f not in _repro_cache:
      _repro_cache[f] = {}
    if graph not in _repro_cache[f]:
      instance = original_SimpleCondFn(f, graph)
      _repro_cache[f][graph] = instance
      print(
          f"  [PATCHED] Cache MISS for {f.__name__} (id={id(f)}). Created"
          f" wrapper_id={id(instance)}",
          flush=True,
      )
    else:
      instance = _repro_cache[f][graph]
      print(
          f"  [PATCHED] Cache HIT for {f.__name__} (id={id(f)}). Reused"
          f" wrapper_id={id(instance)}",
          flush=True,
      )
    return instance

  nnx_transforms.SimpleCondFn = patched_SimpleCondFn_factory


def main(argv):
  if len(argv) > 1:
    raise app.UsageError("Too many command-line arguments.")

  print(f"Global true_fn id: {id(true_fn)}")
  print(f"Global false_fn id: {id(false_fn)}")

  # Run buggy first
  run_test("BUGGY (Original)", apply_buggy_patch)

  # Run patched
  run_test("PATCHED", apply_patched_patch)
  print("\n--- Test Finished ---", flush=True)


if __name__ == "__main__":
  app.run(main)
```

chenkuanliao:
# What does this PR do?

`nnx.cond` (and `nnx.switch`) wrap each branch function in a transient
`SimpleCondFn` adapter before handing it to `jax.lax.cond`. A new wrapper
object is created on every call, and `SimpleCondFn` is declared with
`@dataclasses.dataclass(eq=False)`, so it falls back to **identity-based**
hashing/equality.

JAX keys its trace cache (and, by extension, the persistent compilation cache)
on the identity/hash of the callable passed to `lax.cond`. Because a fresh
wrapper is produced each call — even when the underlying branch function is the
exact same object — JAX treats every invocation as a never-before-seen function.
This produces repeated tracing cache misses and prevents the persistent
compilation cache from ever being reused:

```
TRACING CACHE MISS at .../flax/nnx/transforms/transforms.py:623 (cond): never seen function
... being re-defined repeatedly, preventing caching
```

For large modules with long compile times, this means recompilation on every
run instead of a fast warm start from the on-disk cache.

This PR memoizes the `SimpleCondFn` wrappers so that the same underlying branch
function reuses the same wrapper instance, giving JAX a stable object identity
and restoring trace/compilation cache hits. The cache uses a
`weakref.WeakKeyDictionary` keyed on the original function so wrappers are
collected automatically when the branch function is garbage-collected (no leak).

This is a performance-only change: the computation and results are unchanged —
only the identity of the wrapper handed to `jax.lax.cond` is stabilized.

Fixes #5512

## Notes / scope

- The wrapper cache is keyed on `(f, graph)`, so it helps the common case where
  branches are stable function references (e.g. bound methods / module-level
  functions). Freshly-created lambdas or closures defined inside the call site
  are genuinely new objects each call and will not (and should not) be cached.
- An alternative approach would be to make `SimpleCondFn` hash-by-value
  (`__eq__`/`__hash__` over `(f, graph)`). The `WeakKeyDictionary` approach was
  chosen because it guarantees true object reuse and does not depend on how JAX
  combines hashing with `is` checks.

## Checklist
- [x] This change is discussed in a Github issue/[discussion](https://github.com/google/flax/discussions) (please add a link). — https://github.com/google/flax/issues/5512

### DIFF
--- a/flax/nnx/transforms/iteration.py
+++ b/flax/nnx/transforms/iteration.py
@@ -17,6 +17,7 @@ from collections import deque
 import dataclasses
 import functools
 import typing as tp
+import weakref
 
 
 from flax import struct
@@ -30,6 +31,7 @@ from flax.nnx.statelib import State
 from flax.nnx.transforms.transforms import (
   resolve_kwargs,
   _resolve_bound_callable,
+  _get_cached_wrapper,
   _raise_bound_method_error,
 )
 from flax.typing import Leaf, Missing, PytreeDeque
@@ -1911,6 +1913,14 @@ class SimpleWhileLoopCondFn:
     return self.f(val)
 
 
+_SIMPLE_WHILE_LOOP_BODY_FN_CACHE: weakref.WeakKeyDictionary = (
+  weakref.WeakKeyDictionary()
+)
+_SIMPLE_WHILE_LOOP_COND_FN_CACHE: weakref.WeakKeyDictionary = (
+  weakref.WeakKeyDictionary()
+)
+
+
 @dataclasses.dataclass(eq=False)
 class WhileLoopCondFn:
   f: tp.Callable[..., tp.Any]
@@ -2050,8 +2060,18 @@ def while_loop(cond_fun: tp.Callable[[T], tp.Any],
   if graph_updates is None:
     graph_updates = graphlib.set_graph_updates.current_value()
   if not graph or not graph_updates:
-    simple_body_fn = SimpleWhileLoopBodyFn(body_fun, graph=graph)
-    simple_cond_fn = SimpleWhileLoopCondFn(cond_fun, graph=graph)
+    simple_body_fn = _get_cached_wrapper(
+      _SIMPLE_WHILE_LOOP_BODY_FN_CACHE,
+      SimpleWhileLoopBodyFn,
+      body_fun,
+      graph,
+    )
+    simple_cond_fn = _get_cached_wrapper(
+      _SIMPLE_WHILE_LOOP_COND_FN_CACHE,
+      SimpleWhileLoopCondFn,
+      cond_fun,
+      graph,
+    )
 
     if graph:
       init_val = extract.to_tree2(init_val)
@@ -2092,6 +2112,11 @@ class SimpleForiLoopBodyFn:
       out = extract.to_tree2(out)
     extract.check_same_variables(val_in, out, 'fori_loop')
     return out
+
+
+_SIMPLE_FORI_LOOP_BODY_FN_CACHE: weakref.WeakKeyDictionary = (
+  weakref.WeakKeyDictionary()
+)
 
 
 @dataclasses.dataclass(eq=False)
@@ -2166,7 +2191,12 @@ def fori_loop(lower: int, upper: int,
   if graph_updates is None:
     graph_updates = graphlib.set_graph_updates.current_value()
   if not graph or not graph_updates:
-    simple_body_fn = SimpleForiLoopBodyFn(body_fun, graph=graph)
+    simple_body_fn = _get_cached_wrapper(
+      _SIMPLE_FORI_LOOP_BODY_FN_CACHE,
+      SimpleForiLoopBodyFn,
+      body_fun,
+      graph,
+    )
 
     if graph:
       init_val = extract.to_tree2(init_val)
--- a/flax/nnx/transforms/transforms.py
+++ b/flax/nnx/transforms/transforms.py
@@ -19,6 +19,7 @@ import dataclasses
 import functools
 import inspect
 import typing as tp
+import weakref
 
 from jax._src import checkify as checkify_lib
 
@@ -47,6 +48,7 @@ G = tp.TypeVar('G', bound=tp.Callable[..., tp.Any])
 M = tp.TypeVar('M', bound=Module)
 MA = tp.TypeVar('MA', bound=Module)
 N = tp.TypeVar('N', bound=Module)
+W = tp.TypeVar('W')
 StrInt = tp.TypeVar('StrInt', str, int)
 AxisName = tp.Hashable
 Leaves = list[Leaf]
@@ -585,6 +587,56 @@ class SimpleCondFn:
     return out, updates
 
 
+_SIMPLE_COND_FN_CACHE: weakref.WeakKeyDictionary = weakref.WeakKeyDictionary()
+
+
+def _get_cached_wrapper(
+  cache: weakref.WeakKeyDictionary,
+  wrapper_type: tp.Callable[..., W],
+  f: tp.Callable[..., tp.Any],
+  graph: bool,
+) -> W:
+  """Returns a cached wrapper for ``(f, graph)``.
+
+  Only weak-referenceable, hashable callables can be cached; other callables
+  fall back to a freshly constructed wrapper.
+  """
+  try:
+    weakref.ref(f)
+    hash(f)
+  except TypeError:
+    return wrapper_type(f, graph=graph)
+
+  by_graph = cache.get(f)
+  if by_graph is None:
+    by_graph = {}
+    cache[f] = by_graph
+  wrapper = by_graph.get(graph)
+  if wrapper is None:
+    wrapper = wrapper_type(f, graph=graph)
+    by_graph[graph] = wrapper
+  return wrapper
+
+
+def _get_simple_cond_fn(f: tp.Callable[..., tp.Any], graph: bool) -> SimpleCondFn:
+  """Returns a cached ``SimpleCondFn`` for ``(f, graph)``.
+
+  ``jax.lax.cond`` / ``jax.lax.switch`` key their tracing cache on the identity
+  of the callables they trace. Constructing a fresh ``SimpleCondFn`` on every
+  ``cond``/``switch`` call gives each branch a new identity, defeating that
+  cache and forcing re-tracing (and blocking the persistent compilation cache);
+  see https://github.com/google/flax/issues/5512. Reusing the same wrapper for
+  a given ``(f, graph)`` keeps a stable identity so repeated calls hit the
+  cache.
+
+  ``f`` is held weakly so locally-defined branches stay collectable. Only
+  weak-referenceable, hashable callables can be cached. Callable instances whose
+  classes define ``__slots__`` must include ``'__weakref__'`` in ``__slots__``;
+  otherwise this function falls back to constructing a fresh wrapper.
+  """
+  return _get_cached_wrapper(_SIMPLE_COND_FN_CACHE, SimpleCondFn, f, graph)
+
+
 def cond(
   pred,
   true_fun: tp.Callable[..., A],
@@ -622,8 +674,8 @@ def cond(
     variables = extract.check_no_aliases('cond', operands=operands)
     out, updates = jax.lax.cond(
       pred,
-      SimpleCondFn(true_fun, graph=graph),
-      SimpleCondFn(false_fun, graph=graph),
+      _get_simple_cond_fn(true_fun, graph),
+      _get_simple_cond_fn(false_fun, graph),
       *operands,
     )
     if graph:
@@ -677,7 +729,7 @@ def switch(
     variables = extract.check_no_aliases('switch', operands=operands)
     out, updates = jax.lax.switch(
       index,
-      [SimpleCondFn(f, graph=graph) for f in branches],
+      [_get_simple_cond_fn(f, graph) for f in branches],
       *operands,
     )
     if graph:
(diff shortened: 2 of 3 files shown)


flax/nnx/transforms/transforms.py
```diff
@@ -585,6 +586,38 @@ def __call__(self, *operands):
...
+  ``f`` is held weakly so locally-defined branches stay collectable. Callables
+  that do not support weak references fall back to a freshly constructed
+  wrapper.
+  """
+  try:
+    by_graph = _simple_cond_fn_cache.get(f)
+    if by_graph is None:
+      by_graph = {}
+      _simple_cond_fn_cache[f] = by_graph
+  except TypeError:
```
samanklesaria: This is the error for when `f` doesn't support weak refs, right? Can we make this more specific? Also- can we document this limitation in the docstring (that the caching won't work for objects with `__slots__`)?

flax/nnx/transforms/transforms.py
```diff
@@ -585,6 +586,38 @@ def __call__(self, *operands):
     return out, updates
 
 
+_simple_cond_fn_cache: weakref.WeakKeyDictionary = weakref.WeakKeyDictionary()
```
samanklesaria: Can we make this all caps, like `GRAPH_REGISTRY` is in `flax/nnx/graphlib.py`? This seems to be the global variable convention.

tests/nnx/transforms_test.py
```diff
@@ -7121,6 +7121,61 @@ def false_fn(m):
...
+      def __call__(self, x):
+        return x
+
+    f = Slotted()
+    w1 = nnx_transforms._get_simple_cond_fn(f, True)
+    w2 = nnx_transforms._get_simple_cond_fn(f, True)
+    self.assertIsNot(w1, w2)
+
+  def test_cond_reuses_branch_wrappers(self):
+    from flax.nnx.transforms import transforms as nnx_transforms
```
samanklesaria: I don't love mocking `SimpleCondFn` with a counting function. Could we just transform a function with `nnx.cond` as normal and check that jax's cache isn't growing when we call it multiple times? I believe you can get this with the `._cache_size` attribute. Same thing for the switching test below.

flax/nnx/transforms/transforms.py
chenkuanliao: Addressed in `cb1e6a6`. I made the `weakref` fallback explicit and documented the caching limitation for non-weak-referenceable callables.

flax/nnx/transforms/transforms.py
chenkuanliao: Addressed in `da4e73`. I renamed the cache global to follow the all-caps global variable convention.

tests/nnx/transforms_test.py
chenkuanliao: Addressed in `4883f40`. I updated the 2 tests to verify cache reuse without mocking internal wrapper construction.

tests/nnx/transforms_test.py
```diff
@@ -7121,6 +7121,53 @@ def false_fn(m):
...
+    def false_fn(x):
+      return x - 1
+
+    @jax.jit
+    def f(pred, x):
+      return nnx.cond(
+        pred, true_fn, false_fn, x, graph=False, graph_updates=False
+      )
+
+    self.assertEqual(f._cache_size(), 0)
```
samanklesaria: This test doesn't actually lead to recompilation on main. My idea of using the `cache_size` actually won't work here- sorry! Instead, how about something like this:

```
outer = 0

def branch(x):
  global outer
  outer += 1
  return x

def f(pred, x):
  return jax.lax.cond(pred, branch, branch, x)

for _ in range(2):
  res = f(1.0, 1.0)
  self.assertEqual(outer, 1)
```

For lax, this passes. For flax, it doesn't.

samanklesaria:
@chenkuanliao  It seems that the same problem occurs in `nnx.while_loop` and `nnx.fori_loop`. Could you add fixes for them too?

tests/nnx/transforms_test.py
chenkuanliao: Addressed in `9d4b8cd`. I replaced the cache-size assertions with trace-count side-effect tests.

chenkuanliao:
regarding the similar problems in `nnx.while_loop` and `nnx.fori_loop`, I have added wrapper caching for `nnx.while_loop` and `nnx.fori_loop` with regression tests, which are addressed in `6e8f81d`

flax/nnx/transforms/iteration.py
```diff
@@ -1911,6 +1912,64 @@ def __call__(self, val):
...
 
+_SIMPLE_WHILE_LOOP_BODY_FN_CACHE: weakref.WeakKeyDictionary = (
+  weakref.WeakKeyDictionary()
+)
+_SIMPLE_WHILE_LOOP_COND_FN_CACHE: weakref.WeakKeyDictionary = (
+  weakref.WeakKeyDictionary()
+)
+
+
+def _get_simple_while_loop_body_fn(
```
samanklesaria: This same pattern is repeated for all your _get_simple... functions. Can you refactor things a bit to get rid of the code duplication?

chenkuanliao: Addressed in `d98fb0b`. I refactored the repeated wrapper-cache logic into a shared helper and updated the control-flow wrappers to use it.

chenkuanliao:
I noticed the commit-count check is failing because this PR now has 7 commits. Before I update the branch, would squashing the current PR commits into a single commit be the preferred way to resolve this for Flax, or is there a better project practice I should follow? Thank you!

samanklesaria:
> I noticed the commit-count check is failing because this PR now has 7 commits. Before I update the branch, would squashing the current PR commits into a single commit be the preferred way to resolve this for Flax, or is there a better project practice I should follow? Thank you!

Yes, you should always squash PRs to flax manually before pushing.

tests/nnx/transforms_test.py
```diff
@@ -7121,6 +7121,49 @@ def false_fn(m):
...
+
+      def __call__(self, x):
+        return x
+
+    f = Slotted()
+    w1 = nnx_transforms._get_simple_cond_fn(f, True)
+    w2 = nnx_transforms._get_simple_cond_fn(f, True)
+    self.assertIsNot(w1, w2)
+
+  def test_cond_reuses_branch_wrappers(self):
```
samanklesaria: Can the switch, cond, while and fori versions of this be a single parameterized test?

chenkuanliao: Addressed in `3f801d1`. I combined the cond, switch, while_loop, and fori_loop wrapper-reuse checks into a single parameterized test. thanks!

tests/nnx/transforms_test.py
```diff
@@ -7121,6 +7121,114 @@ def false_fn(m):
...
+    elif control_flow == 'while_loop':
+      cond_trace_count = 0
+      body_trace_count = 0
+
+      def cond_fn(x):
+        nonlocal cond_trace_count
+        cond_trace_count += 1
+        return x < 1
+
+      def body_fn(x):
```
samanklesaria: Could these body functions (also in the fori branch) be the same `branch` function used above? Maybe we could even make the cond function be the branch! This would simplify the assertion check at the end too.

chenkuanliao: Addressed in `53ebd9c`. I reused the same `branch` function across the cond, switch, while_loop, and fori_loop cases and simplified the trace-count assertion

tests/nnx/transforms_test.py
```diff
@@ -7121,6 +7121,86 @@ def false_fn(m):
...
+          0, 2, branch, x, graph=False, graph_updates=False
+        )
+
+      expected = False
+
+    else:
+      raise AssertionError(f'Unknown control flow: {control_flow}')
+
+    for _ in range(2):
+      self.assertEqual(f(init_value), expected)
```
samanklesaria: We don't care about the return value in this test- just use the assertion below.

samanklesaria: Instead of creating a function in each branch and then calling it at the end, what if you just stored the result in `result` in each branch, and then made an assertion about `result`?

chenkuanliao: got it! I removed the return-value assertion in `b567144`

chenkuanliao: I simplified the test so each branch executes the relevant control-flow op directly and then asserts `trace_count`, instead of defining a local function and calling it afterward.

tests/nnx/transforms_test.py
```diff
@@ -7121,6 +7121,78 @@ def false_fn(m):
...
+
+      def __call__(self, x):
+        return x
+
+    f = Slotted()
+    w1 = nnx_transforms._get_simple_cond_fn(f, True)
+    w2 = nnx_transforms._get_simple_cond_fn(f, True)
+    self.assertIsNot(w1, w2)
+
+  @parameterized.named_parameters(
```
samanklesaria: What about `@parameterized.parameters('cond', 'switch', 'while_loop', 'fori_loop')` instead?

chenkuanliao: Addressed in `fe461ec`. I switched this to `@parameterized.parameters(...)` as suggested. thanks!

samanklesaria (approved):
Looks good to me!

flax/nnx/transforms/transforms.py
cgarciae, line 567: Feels like we could explore changing the dataclass definition to so the hash depends on `f`, this way we could get the caching for free on the jax side?

samanklesaria: Jax traces with

```python
@weakref_lru_cache(maxsize=None, explain=explain)
def trace_to_jaxpr(fun, in_avals, debug_info, ...):
    ....
```

The `weakref_lru_cache` function's docstring says "The cache will take a weakref to the first argument of the wrapped function and strong refs to all other arguments." This means the first argument (Our SimpleCondFn) will be compared by *identity*, not by `__eq__/__hash__`. So we can't rely on existing caching. Rolling our own, as in this PR, is the only choice.

PR #5532, opened 2026-07-22
Exclude haiku doctests from CI

samanklesaria:
This PR fixes the CI's recent doctest failures. Currently, we get "AttributeError: jax.core.DropVar was deprecated in JAX v0.10.0 and removed in JAX v0.11.0." when running doctests. This is actually a problem with haiku: (dm-haiku==0.0.16's jaxpr_info.py:45 still references jax.core.DropVar, which the latest jax removed). The simplest fix is to just not run guides/converting_and_upgrading/haiku_migration_guide during doctest CI.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -73,6 +73,11 @@ templates_path = ['_templates']
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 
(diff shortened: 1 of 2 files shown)


IvyZX (changes requested):
Approach LGTM. But the CI doctest is still broken?

samanklesaria:
> Approach LGTM. But the CI doctest is still broken?

It seems there's an NNX version of the haiku migration guide as well. I'll exclude that from doctests too.

PR #5539, opened 2026-07-29
implement axes_scan's broadcast semantics via dce

mattjj:
By using DCE instead of partial evaluation, we unblock JAX_REMAT3=1 and the deletion of general partial eval in JAX.


mattjj:
Oops, this isn't ready...

PR #5564, opened 2026-08-19
fix(traverse_util): validate keys in flatten_dict when separator is specified (#5554)

### LINKED ISSUE #5554: flatten_dict with a separator cannot round-trip keys containing the separator, and raises an opaque error on non-string keys
### Summary

`flax.traverse_util.flatten_dict`, when called with a separator, produces flattened keys that `unflatten_dict` cannot faithfully restore for two classes of input. In one case the result is a silently corrupted round-trip; in the other it is a low-level `TypeError` from deep inside the call. Both are reachable with ordinary dictionaries, and both are surprising given that the `flatten_dict` docstring directs the reader to `unflatten_dict` "on how to restore the nested dictionary structure."

The proposal is to fail fast with a clear, actionable error when a separator cannot represent the given keys unambiguously, rather than corrupt the data or raise an opaque error.

### Case 1: a key that contains the separator round-trips to a different structure

```python
from flax.traverse_util import flatten_dict, unflatten_dict

xs = {"a/b": 1, "c": {"d": 2}}
flat = flatten_dict(xs, sep="/")
back = unflatten_dict(flat, sep="/")

print(flat)          # {'a/b': 1, 'c/d': 2}
print(back)          # {'a': {'b': 1}, 'c': {'d': 2}}
print(back == xs)    # False
```

The key `"a/b"` and the nested path `("c", "d")` both flatten to a slash-joined string, so `unflatten_dict` splits `"a/b"` into a two-level path and the original structure is lost. No error is raised.

### Case 2: a non-string key under a separator raises an opaque error

```python
from flax.traverse_util import flatten_dict

flatten_dict({1: {2: 3}}, sep="/")
# TypeError: sequence item 0: expected str instance, int found
```

The failure surfaces from `sep.join(path)` inside `_flatten`, so the message describes an internal `join` rather than the actual problem, which is that a separator cannot be applied to non-string keys.

### Why this is worth addressing

`flatten_dict` and `unflatten_dict` are commonly used to move parameter and state trees to and from a flat, string-keyed representation, for example when interfacing with checkpoint formats or logging. Silent structural corruption of such a tree is difficult to detect downstream, and the non-string-key error gives no indication of its cause. Integer keys and keys containing punctuation both occur in real parameter dictionaries.

### Proposed change

When `sep` is not `None`, validate each key before joining and raise a single, descriptive error for the two conditions that make the flattened key ambiguous or ill-formed:

- a key that is not a string, and
- a string key that contains `sep`.

For example:

```
ValueError: flatten_dict with sep='/' requires string keys that do not
contain the separator; got key 'a/b' at path ('a/b',)
```

This turns two silent or opaque failure modes into one clear error, and leaves every input that currently round-trips correctly unchanged. The tuple-key mode (`sep=None`) is unaffected, since it already round-trips these inputs faithfully.

I am happy to prepare the pull request, including regression tests for both cases and a short note in the `flatten_dict` docstring, if this direction is agreeable.

### Environment

- flax 0.12.8
- jax 0.11.0
- numpy 2.5.2
- Python 3.12

Ultron09:
### Summary

Fixes #5554

When `flax.traverse_util.flatten_dict` is called with a separator (`sep is not None`):
1. **Case 1 (Structural corruption)**: A string key containing the separator (e.g. `{"a/b": 1, "c": {"d": 2}}` with `sep="/"`) flattens to `{"a/b": 1, "c/d": 2}`. When `unflatten_dict(..., sep="/")` is called, `"a/b"` is split into a nested path `{"a": {"b": 1}}`, silently corrupting the round-trip structure without error.
2. **Case 2 (Opaque TypeError)**: A non-string key (e.g. `{1: {2: 3}}` with `sep="/"`) surfaces an internal `TypeError: sequence item 0: expected str instance, int found` from deep inside string joining.

### Solution

- Validates each key in `_flatten` when `sep is not None` to ensure it is a string and does not contain `sep`.
- Raises a descriptive `ValueError`:
  ```
  ValueError: flatten_dict with sep='/' requires string keys that do not contain the separator; got key 'a/b' at path ('a/b',)
  ```
- Updates `flatten_dict` docstrings to document the string/separator requirement and `ValueError`.
- Added unit tests in `traverse_util_test.py` covering non-string keys, separator-containing keys, nested dictionary paths, and verifying `sep=None` behavior remains unaffected.

cc @IvyZX @cgarciae @jheek @Amey-Thakur

### DIFF
--- a/flax/traverse_util.py
+++ b/flax/traverse_util.py
@@ -91,6 +91,12 @@ def _flatten(xs, prefix, keep_empty_nodes, is_leaf, sep):
   result = {}
   is_empty = True
   for key, value in xs.items():
+    if sep is not None:
+      if not isinstance(key, str) or sep in key:
+        raise ValueError(
+          f"flatten_dict with sep='{sep}' requires string keys that do not "
+          f"contain the separator; got key {key!r} at path {prefix + (key,)}"
+        )
     is_empty = False
     path = prefix + (key,)
     result.update(_flatten(value, path, keep_empty_nodes, is_leaf, sep))
@@ -130,9 +136,13 @@ def flatten_dict(xs, keep_empty_nodes=False, is_leaf=None, sep=None):
       leaf (i.e., should not be flattened further).
     sep: if specified, then the keys of the returned
       dictionary will be ``sep``-joined strings (if
-      ``None``, then keys will be tuples).
+      ``None``, then keys will be tuples). When specified,
+      all keys must be strings that do not contain ``sep``.
   Returns:
     The flattened dictionary.
+  Raises:
+    ValueError: if ``sep`` is specified and any dictionary key is not a string
+      or contains the separator ``sep``.
   """
   assert isinstance(
     xs, (flax.core.FrozenDict, dict)
--- a/tests/traverse_util_test.py
+++ b/tests/traverse_util_test.py
@@ -175,6 +175,43 @@ class TraversalTest(absltest.TestCase):
       },
     )
 
+  def test_flatten_dict_sep_key_containing_sep(self):
+    xs = {'a/b': 1, 'c': {'d': 2}}
+    with self.assertRaisesRegex(
+      ValueError,
+      r"flatten_dict with sep='/' requires string keys that do not contain the separator; got key 'a/b' at path \('a/b',\)",
+    ):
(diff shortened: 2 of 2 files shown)


samanklesaria (approved):
Looks good to me!

PR #5573, opened 2026-08-28
Fix nnx.capture with List modules in scan

### LINKED ISSUE #5571: nnx.capture fails with nnx.scan when the model contains nnx.List (mixed str/int pytree dict keys)
### System information

- Flax 0.12.7, JAX 0.10.1, Python 3.14, macOS arm64 (CPU)

### Problem

`nnx.capture` fails when the wrapped function runs an `nnx.scan` and the model contains an `nnx.List` (directly or via `nnx.Sequential`):

```
ValueError: Comparator raised exception while sorting pytree dictionary keys.
...
TypeError: '<' not supported between instances of 'str' and 'int'
```

The same model captures fine eagerly and under `nnx.jit` alone; the same scan captures fine when the model stores its layers as plain attributes instead of an `nnx.List`. Only the combination capture + scan + `nnx.List` fails.

### Reproduction

```python
import jax.numpy as jnp
from flax import nnx


class Model(nnx.Module):
        self.layers = nnx.List([nnx.Linear(4, 4, rngs=rngs) for _ in range(2)])

    def __call__(self, x):
        for layer in self.layers:
            x = layer(x)
        self.sow(nnx.Intermediate, "out", x)
        return x


def rollout(model, x):
    state_axes = nnx.StateAxes({nnx.Intermediate: 0, ...: nnx.Carry})
    return nnx.scan(
        lambda m, x: m(x), in_axes=(state_axes, nnx.Carry), out_axes=nnx.Carry, length=3
    )(model, x)


x = jnp.ones(4)

# eager: OK
nnx.capture(Model(nnx.Rngs(0)), nnx.Intermediate)(x)

# nnx.jit: OK
nnx.capture(nnx.jit(lambda m, x: m(x)), nnx.Intermediate)(Model(nnx.Rngs(0)), x)

# nnx.scan: fails
nnx.capture(rollout, nnx.Intermediate)(Model(nnx.Rngs(0)), x)
```

Output:

```
capture(model)               OK
capture(jit(model))          OK
capture(scan rollout)        ValueError: Comparator raised exception while sorting pytree
                             dictionary keys.  (cause: TypeError: '<' not supported between
                             instances of 'str' and 'int')
```

Replacing the `nnx.List` with two plain attributes (`self.l0`, `self.l1`) makes the scan case pass and the sown value comes out stacked as expected, so the scan/StateAxes usage itself appears correct.

### Analysis

`capture()`'s wrapper sets `m.__captures__ = pytreelib.data(...)` on every module yielded by `iter_modules(module)` (flax/nnx/module.py, `wrapper`). `iter_modules` yields `nnx.List` (it is a `Module` subclass), so the `List` node — whose existing children are keyed by integers — gains a string-keyed entry. When `jax.lax.scan` later flattens the carry (first hit in `api_util.debug_info` → `flatten_with_path`), the dict with mixed `int`/`str` keys cannot be key-sorted and the comparator raises.

A possible fix is to skip integer-keyed container modules (`nnx.List`/`nnx.Dict`-style nodes) when planting `__captures__` — a sow on such a container has no stable attribute name anyway — or to store the capture buffers out-of-band rather than as attributes.

### What you expected to happen

The scan case behaves like the eager and jit cases: `(result, intermediates)` with the sown value stacked along the scan axis.

aswanth-07:
## What does this pull request do?

`nnx.capture` now uses a dedicated string key that remains orderable beside integer container keys. This lets `nnx.List` participate in capture without producing mixed-key sorting failures under `nnx.scan`, while preserving capture support for `nnx.List` subclasses that call `sow` in their own methods.

Fixes #5571

## What did you check?

- Added a regression covering `nnx.capture` around `nnx.scan` with a model containing an `nnx.List`.
- Added a subclass regression that calls `sow` on the list itself and verifies its stacked captured values.
- `python -m pytest tests/nnx/module_test.py::TestCapture -q`: 10 passed.
- `ruff check flax/nnx/module.py tests/nnx/module_test.py`: passed.
- `python -m py_compile flax/nnx/module.py tests/nnx/module_test.py`: passed.
- `git diff --check`: passed.
- The original scan regression fails on pristine main with `ValueError: Comparator raised exception while sorting pytree dictionary keys`.

AI assistance disclosure: AI-assisted development tools were used during investigation and implementation. The reported tests were run against the final diff.

## Checklist

- [x] This PR fixes a minor issue.
- [x] This change is discussed in a GitHub issue.
- [x] This change includes a regression test.

### DIFF
--- a/flax/nnx/module.py
+++ b/flax/nnx/module.py
@@ -882,6 +882,8 @@ def capture(fn: tp.Callable[P, R] | type[variableslib.Variable], *var_types: typ
 
       # Initialize __captures__ as a tuple of Variables (one per type)
       for path, m in iter_modules(module):
+        if getattr(m, '_pytree__has_int_keys', False):
+          continue
         # Create initial dicts for each variable type
         initial_dicts = {}
         for var_type in var_types:
--- a/tests/nnx/module_test.py
+++ b/tests/nnx/module_test.py
@@ -287,6 +287,34 @@ class TestCapture(parameterized.TestCase):
       jnp.broadcast_to(model.w.get_value()[None, :], (3, 4)))
     self.assertEqual(intermediates['y'].get_value()[0].shape, (3,))
 
+  def test_scan_with_list(self):
+    class Model(nnx.Module):
+      def __init__(self, rngs: nnx.Rngs):
+        self.layers = nnx.List([
+          nnx.Linear(4, 4, rngs=rngs) for _ in range(2)
+        ])
+
+      def __call__(self, x):
+        for layer in self.layers:
+          x = layer(x)
+        self.sow(nnx.Intermediate, 'out', x)
+        return x
+
+    def rollout(model, x):
+      state_axes = nnx.StateAxes({nnx.Intermediate: 0, ...: nnx.Carry})
+      return nnx.scan(
+        lambda m, x: m(x),
+        in_axes=(state_axes, nnx.Carry),
+        out_axes=nnx.Carry,
+        length=3,
+      )(model, x)
+
+    model = Model(nnx.Rngs(0))
+    _, intermediates = nnx.capture(rollout, nnx.Intermediate)(model, jnp.ones(4))
+
+    self.assertEqual(intermediates['out'][0].shape, (3, 4))
+    self.assertFalse(hasattr(model.layers, '__captures__'))
+
 
   @parameterized.parameters(True, False)
   def test_fwd_bwd(self, graph_mode):


samanklesaria:
This would prevent `nnx.List` subclasses from running `self.sow` in their methods. Perhaps not a common use case, but still one worth considering. I don't think blocking the `__captures__` attribute is the move here. Instead, we need to make the `__captures__` key comparable with ints.

aswanth-07:
Updated this in `ac7e375`. `nnx.capture` now uses a dedicated string key that remains orderable beside integer container keys, so `nnx.List` subclasses still receive capture state and can call `sow`. I added a subclass regression that verifies the list's captured values under `nnx.scan`. The complete `TestCapture` class passes with 10 tests, and Ruff passes on both changed files. Thanks for pointing out the subclass case.

aswanth-07:
Closing this in favor of #5585, which carries the maintainer-owned implementation of the same mixed-key ordering fix. Thanks for the design feedback and for taking it forward.

PR #5582, opened 2026-09-10
fix(nnx): skip Sequence modules in capture to prevent mixed key types (fixes #5571)

### LINKED ISSUE #5571: nnx.capture fails with nnx.scan when the model contains nnx.List (mixed str/int pytree dict keys)

jdymitarai:
### Summary

Fixes #5571.

When `nnx.capture` is used with functions containing `nnx.scan` on models that contain sequence containers (such as `nnx.List` or `nnx.Sequential`), execution failed with:
```
TypeError: '<' not supported between instances of 'str' and 'int'
ValueError: Comparator raised exception while sorting pytree dictionary keys.
```

### Cause
`nnx.capture` initializes `__captures__` on every `Module` yielded by `iter_modules(module)`. `nnx.List` is a `Module` subclass whose child elements in the pytree are keyed by integers (`0, 1, ...`). Setting attribute `__captures__` on the sequence container adds a string key to an otherwise integer-keyed pytree node. When JAX transforms (like `jax.lax.scan` carry flattening) sort dictionary keys via `<` comparator, comparing `int` and `str` raises a `TypeError`.

### Fix
- In `flax/nnx/module.py`: skip `Sequence` modules (e.g. `nnx.List`) when creating `__captures__` and wrapping methods in `capture`. Intermediate states are still captured as expected in child modules and enclosing parent modules.
- In `tests/nnx/module_test.py`: add regression test `test_scan_with_nnx_list` in `TestCapture` verifying `nnx.capture` with `nnx.scan` on a model containing an `nnx.List`.

### Test Plan
- Run module tests:
  ```bash
  pytest tests/nnx/module_test.py
  ```
  All 82 tests passed.


jdymitarai:
Closing this PR to avoid review overhead on the team. The context remains in the thread for future reference if helpful. Thanks for your time.

PR #5588, opened 2026-09-17
Count broadcast mask entries in nnx.metrics.Average

Nicholas022400701:
# What does this PR do?

`nnx.metrics.Average.update` accepts a `mask` whose shape only needs to be broadcastable to the shape of `values`, as its docstring says. The total was computed from `values * mask`, which broadcasts, but the count was computed from `mask.sum()`, which does not. A mask smaller than `values` therefore gave a count that was too small and an average that was too large. `Accuracy` and `MultiMetric` forward their masks to this method, so they had the same problem.

```python
import jax.numpy as jnp
from flax import nnx

average = nnx.metrics.Average()
values = jnp.arange(6, dtype=jnp.float32).reshape(2, 3)
average.update(values=values, mask=jnp.array([True, False, True]))
average.compute()  # 5.0 on main, but the four kept entries 0, 2, 3 and 5 have mean 2.5
```

The fix broadcasts the mask to the shape of `values` before summing it, so the total and the count are always computed over the same entries. Masks that already have the shape of `values` behave exactly as before, the existing `test_average` and `test_multimetric` cases cover that.

## Tests

`test_average_broadcast_mask` covers a mask that broadcasts along a trailing dimension and one that broadcasts along a leading dimension, and `test_accuracy_broadcast_mask` covers the same situation through `Accuracy`. `test_average_broadcast_mask_under_jit` runs the `(3,)` mask over `(2, 3)` values under `jax.jit` and `nnx.jit`, following the review comment below. On `main` the five new test cases fail with counts of 2, 1, 2, 2 and 2 instead of 4, 3, 4, 4 and 4, while the existing 19 tests in the file pass. With the fix all 24 tests pass and `ruff check` is clean. Python 3.12, jax 0.11.1, CPU.

- With the fix, 24 passed: https://github.com/Nicholas022400701/oss-ci/actions/runs/35287363861
- The same tests on main, 5 failed and 19 passed: https://github.com/Nicholas022400701/oss-ci/actions/runs/35287363835

## Checklist
- [ ] The documentation and docstrings adhere to the [documentation guidelines](https://github.com/google/flax/blob/main/docs/README.md#how-to-write-code-documentation).


mohsinm-dev:
Nice fix! Could we also add a small jax.jit test with a lower-rank broadcastable mask, such as (3,) for (2, 3) values? That would keep this behavior covered under transformed execution.

Nicholas022400701:
Thanks for the review, @mohsinm-dev. Added `test_average_broadcast_mask_under_jit` in a82a630. It runs `Average.update` with a `(3,)` mask over `(2, 3)` values under both `jax.jit` and `nnx.jit`, using the same split, merge and static graphdef pattern as `test_multimetric_split_merge_under_jit`, and checks that the count is 4, the total is 10 and the average is 2.5. Both cases fail on `main` with a count of 2 and pass with the fix. The runs are linked in the description.

